PR #13, opened 2020-06-27
Hide peer detail view if multiple are selected

promag:
Currently if multiple peers are selected the peer detail view shows the first new selected peer.

With this PR the peer detail view is hidden when multiple peers are selected. It is also a slight refactor to simplify and remove duplicate code.

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -492,9 +492,7 @@ RPCConsole::RPCConsole(interfaces::Node& node, const PlatformStyle *_platformSty
     m_node.rpcSetTimerInterfaceIfUnset(rpcTimerInterface);
 
     setTrafficGraphRange(INITIAL_TRAFFIC_GRAPH_MINS);
-
-    ui->detailWidget->hide();
-    ui->peerHeading->setText(tr("Select a peer to view detailed information."));
+    updateDetailWidget();
 
     consoleFontSize = settings.value(fontSizeSettingsKey, QFont().pointSize()).toInt();
     clear();
@@ -625,7 +623,7 @@ void RPCConsole::setClientModel(ClientModel *model)
         connect(disconnectAction, &QAction::triggered, this, &RPCConsole::disconnectSelectedNode);
 
         // peer table signal handling - update peer details when selecting new node
-        connect(ui->peerWidget->selectionModel(), &QItemSelectionModel::selectionChanged, this, &RPCConsole::peerSelected);
+        connect(ui->peerWidget->selectionModel(), &QItemSelectionModel::selectionChanged, this, &RPCConsole::updateDetailWidget);
         // peer table signal handling - update peer details when new nodes are added to the model
         connect(model->getPeerTableModel(), &PeerTableModel::layoutChanged, this, &RPCConsole::peerLayoutChanged);
         // peer table signal handling - cache selected node ids
@@ -1017,18 +1015,6 @@ void RPCConsole::updateTrafficStats(quint64 totalBytesIn, quint64 totalBytesOut)
     ui->lblBytesOut->setText(GUIUtil::formatBytes(totalBytesOut));
 }
 
-void RPCConsole::peerSelected(const QItemSelection &selected, const QItemSelection &deselected)
-{
-    Q_UNUSED(deselected);
-
-    if (!clientModel || !clientModel->getPeerTableModel() || selected.indexes().isEmpty())
-        return;
-
-    const CNodeCombinedStats *stats = clientModel->getPeerTableModel()->getNodeStats(selected.indexes().first().row());
-    if (stats)
-        updateNodeDetail(stats);
-}
-
 void RPCConsole::peerLayoutAboutToChange()
 {
     QModelIndexList selected = ui->peerWidget->selectionModel()->selectedIndexes();
@@ -1045,7 +1031,6 @@ void RPCConsole::peerLayoutChanged()
     if (!clientModel || !clientModel->getPeerTableModel())
         return;
 
-    const CNodeCombinedStats *stats = nullptr;
     bool fUnselect = false;
     bool fReselect = false;
 
@@ -1076,9 +1061,6 @@ void RPCConsole::peerLayoutChanged()
             fUnselect = true;
             fReselect = true;
         }
-
-        // get fresh stats on the detail node.
-        stats = clientModel->getPeerTableModel()->getNodeStats(detailNodeRow);
     }
 
     if (fUnselect && selectedRow >= 0) {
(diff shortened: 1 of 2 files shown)


promag:
This was done while reviewing #6, @hebasto maybe you could take a look too.

hebasto:
Concept ACK.

src/qt/rpcconsole.cpp
```diff
@@ -1092,13 +1076,17 @@ void RPCConsole::peerLayoutChanged()
...
-    if (stats)
-        updateNodeDetail(stats);
 }
 
-void RPCConsole::updateNodeDetail(const CNodeCombinedStats *stats)
+void RPCConsole::updateDetailWidget()
 {
+    auto selected_rows = ui->peerWidget->selectionModel()->selectedRows();
+    if (!clientModel || !clientModel->getPeerTableModel() || selected_rows.size() != 1) {
+        ui->detailWidget->hide();
```
Saibato, line 1085: Nit check for access to clientModel ?
``` code 
if (!clientModel) return;
    auto selected_rows = ui->peerWidget->selectionModel()->selectedRows();

**Comment by promag** [comment]
Strange, in macos that didn't happen. Was that on startup? On opening peers tab?

**Comment by hebasto** [comment]
> Strange, in macos that didn't happen. Was that on startup? On opening peers tab?

On startup.

**Comment by promag** [comment]
@hebasto should be fixed now, `selectionModel()` can return nullptr.

**Comment by hebasto** [comment]
> @hebasto should be fixed now...

Testing right now.

> `selectionModel()` can return nullptr.

The [Qt docs](https://doc.qt.io/qt-5/qabstractitemview.html#selectionModel) keeps silence about `nullptr` :)

**Comment by promag** [comment]
Right, I saw qabstractitemview.cpp where it also has the check.

**Comment by hebasto** [comment]
@promag 
How about the following patch which fixes the problem described in the OP:
```diff
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -593,7 +593,7 @@ void RPCConsole::setClientModel(ClientModel *model)
         ui->peerWidget->verticalHeader()->hide();
         ui->peerWidget->setEditTriggers(QAbstractItemView::NoEditTriggers);
         ui->peerWidget->setSelectionBehavior(QAbstractItemView::SelectRows);
-        ui->peerWidget->setSelectionMode(QAbstractItemView::ExtendedSelection);
+        ui->peerWidget->setSelectionMode(QAbstractItemView::SingleSelection);
         ui->peerWidget->setContextMenuPolicy(Qt::CustomContextMenu);
         ui->peerWidget->setColumnWidth(PeerTableModel::Address, ADDRESS_COLUMN_WIDTH);
         ui->peerWidget->setColumnWidth(PeerTableModel::Subversion, SUBVERSION_COLUMN_WIDTH);
```
?

promag:
Multiple selection is enabled and used when you use the context menu.

src/qt/rpcconsole.cpp
```diff
@@ -1078,7 +1063,6 @@ void RPCConsole::peerLayoutChanged()
         }
 
         // get fresh stats on the detail node.
-        stats = clientModel->getPeerTableModel()->getNodeStats(detailNodeRow);
```
hebasto, line 1081: nit: not sure about the two lines above?

promag: Removed those.

src/qt/rpcconsole.cpp
```diff
@@ -1093,12 +1075,20 @@ void RPCConsole::peerLayoutChanged()
...
 {
+    QModelIndexList selected_rows;
+    auto selection_model = ui->peerWidget->selectionModel();
+    if (selection_model) selected_rows = selection_model->selectedRows();
+    if (!clientModel || !clientModel->getPeerTableModel() || selected_rows.size() != 1) {
+        ui->detailWidget->hide();
+        ui->peerHeading->setText(tr("Select a peer to view detailed information."));
+        return;
+    }
+    const CNodeCombinedStats *stats = clientModel->getPeerTableModel()->getNodeStats(selected_rows.first().row());
```
hebasto, line 1091: nanonit:
```suggestion

    const CNodeCombinedStats* stats = clientModel->getPeerTableModel()->getNodeStats(selected_rows.first().row());
```

or even use `const auto` (that seems preferable to me).

src/qt/rpcconsole.cpp
```diff
@@ -1093,12 +1075,20 @@ void RPCConsole::peerLayoutChanged()
...
-    if (stats)
-        updateNodeDetail(stats);
+    updateDetailWidget();
 }
 
-void RPCConsole::updateNodeDetail(const CNodeCombinedStats *stats)
+void RPCConsole::updateDetailWidget()
 {
+    QModelIndexList selected_rows;
+    auto selection_model = ui->peerWidget->selectionModel();
```
hebasto, line 1084: nit:
```suggestion
    const auto selection_model = ui->peerWidget->selectionModel();
```

Saibato:
tested  https://github.com/bitcoin-core/gui/commit/76277cc77dea39b53e09ee1c440cd37270826201 on debian and lubuntu VM's on top of master and other changes so far LGTM

hebasto:
@promag 
Due to this connection
https://github.com/bitcoin-core/gui/blob/76277cc77dea39b53e09ee1c440cd37270826201/src/qt/rpcconsole.cpp#L626

not sure if added in [the latest push](https://github.com/bitcoin-core/gui/compare/df2987a0bd07bfbabbffbafeca53e5b36fd4fdc8..76277cc77dea39b53e09ee1c440cd37270826201) `updateDetailWidget();` is required at all.

promag:
@hebasto right, I've added it just to be sure view is updated. I also thought of blocking signals to prevent unnecessary updates because of that signal. Another thing we could do is to add `const CNodeCombinedStats* m_current_node_stats`, this way we can skip unnecessary work in `updateDetailWidget`.

luke-jr:
Weak Concept NACK

My mail program also shows the most-recently-touched message when selecting multiple, so the current behaviour isn't too weird. I'm not sure there's a benefit from hiding it?

hebasto:
@luke-jr 
> Weak Concept NACK
> 
> My mail program also shows the most-recently-touched message when selecting multiple, so the current behaviour isn't too weird. I'm not sure there's a benefit from hiding it?

But the GUI shows not the most-recently-touched peer details rather the first-touched one's. Therefore, this pull is an improvement, I think.

promag:
Either hide or show last selected looks good to me.

jonasschnelli:
Tested ACK 76277cc77dea39b53e09ee1c440cd37270826201.
Should this be considered a bugfix?

promag:
Thanks.

PR #15, opened 2020-06-30
Update pruning tooltip

BitcoinErrorLog:
This tooltip was a bit confusing and a deterrent more than informative. Can we adjust to something that fits a user looking for a way to save disk space?

### DIFF
--- a/src/qt/forms/optionsdialog.ui
+++ b/src/qt/forms/optionsdialog.ui
@@ -55,7 +55,7 @@
         <item>
          <widget class="QCheckBox" name="prune">
           <property name="toolTip">
-           <string>>Enabling pruning significantly reduces the disk space required to verify transactions by removing some advanced features. Reverting this setting requires downloading the entire blockchain again.</string>
+           <string>Enabling pruning significantly reduces the disk space required to store transactions. All blocks are still fully validated. Reverting this setting requires re-downloading the entire blockchain.</string>
           </property>
           <property name="text">
            <string>Prune &amp;block storage to</string>
--- a/src/qt/forms/optionsdialog.ui
+++ b/src/qt/forms/optionsdialog.ui
@@ -55,7 +55,7 @@
         <item>
          <widget class="QCheckBox" name="prune">
           <property name="toolTip">
-           <string>Disables some advanced features but all blocks will still be fully validated. Reverting this setting requires re-downloading the entire blockchain. Actual disk usage may be somewhat higher.</string>
+           <string>>Enabling pruning significantly reduces the disk space required to verify transactions by removing some advanced features. Reverting this setting requires downloading the entire blockchain again.</string>
           </property>
           <property name="text">
            <string>Prune &amp;block storage to</string>


src/qt/forms/optionsdialog.ui
```diff
@@ -55,7 +55,7 @@
         <item>
          <widget class="QCheckBox" name="prune">
           <property name="toolTip">
-           <string>Disables some advanced features but all blocks will still be fully validated. Reverting this setting requires re-downloading the entire blockchain. Actual disk usage may be somewhat higher.</string>
+           <string>>Enabling pruning significantly reduces the disk space required to verify transactions by removing some advanced features. Reverting this setting requires downloading the entire blockchain again.</string>
```
hebasto, line 58: Please remove redundant `>`.

hebasto:
@BitcoinErrorLog 
To make reviewing more constructive mind elaborating what exactly is confusing in the current tooltip for you?

BitcoinErrorLog:
I'm totally okay with nitpicking for accuracy as long as it results in a useful tooltip. If I read it as is released, it sounds risky, resource-intensive and does not communicate that I could run Bitcoin with much smaller footprint. Pruning is a huge thing for node adoption, it should be clear to a typical user there is a huge benefit to using it.

Sjors:
Concept ACK on making the option more appealing.

Original message:

> Disables some advanced features but all blocks will still be fully validated. Reverting this setting requires re-downloading the entire blockchain. Actual disk usage may be somewhat higher.

Current proposal in this PR:

>  Enabling pruning significantly reduces the disk space required to verify transactions by removing some advanced features. Reverting this setting requires downloading the entire blockchain again.

This suggest a cause and effect that isn't really correct. We never needed the disk space to verify transactions, and removing the advanced features is just a side-effect; they're not the reason pruning isn't a default.

Here's my attempt (with @hebasto's points included):

> Enabling pruning significantly reduces the disk space required to verify transactions. All blocks are still fully validated, but the are removed after some time. Pruning disables some advanced features and reverting this setting requires re-downloading the entire blockchain.

> actually "significantly" depends on the -prune setting value

Which by default is extremely small, so that's not worth mentioning imo.

BitcoinErrorLog:
**Edited for typo and brevity:**

> Enabling pruning significantly reduces the disk space required to store transactions. All blocks are still fully validated. Pruning disables some advanced features. Reverting this setting requires re-downloading the entire blockchain.

**Also, please consider:**
> Enabling pruning significantly reduces the disk space required to store transactions. All blocks are still fully validated. Reverting this setting requires re-downloading the entire blockchain.

This is even shorter, as I doubt any advanced feature user cares about a tooltip UI detail like that. Are such advanced features even in the UI?

Sjors:
Either looks good to me. It could be annoying for someone, once they decide they need an index, to then realize they shouldn't have pruned. But I'm fine with leaving the sentence about advanced features out.

BitcoinErrorLog:
File has been updated with shorter version.

BitcoinErrorLog:
Now where do I go to lobby/edit for prompting the user to choose whether to turn on pruning on first run at install? :)

Sjors:
Pruning is automatically suggested for users with insufficient disk space, see https://github.com/bitcoin/bitcoin/pull/16714.

Currently it suggests pruning if your free disk space less than the expected blockchain size plus 10 GB. The easiest change you could propose is to increase that margin, e.g. to 100 GB. People with giant disks are likely to care less about pruning.
  
```
        if (ui->prune->isEnabled()) {
            ui->prune->setChecked(m_bytes_available < (m_blockchain_size_gb + m_chain_state_size_gb + 10) * GB_BYTES);
```

ACK 1c82a82

hebasto:
Let me share my concerns regarding this PR.

1. The GUI has two places where the `-prune` option is available for an user to change:
- the intro window, and
- the "Main" tab in the "Options" window

It seems reasonable to synchronize tooltips in both places.

2. As pruning consequences (e.g., time-consuming block chain re-downloading) could be frustrating for an inexperienced user, I'd move important details from tooltips to window labels to let the user to focus on them. Also see [Human Interface Guidelines - Help Tags (Tooltips)](https://developer.apple.com/design/human-interface-guidelines/macos/user-interaction/help/).

3. We could start to align the Bitcoin Core GUI with [Optech Style Guide](https://github.com/bitcoinops/bitcoinops.github.io/blob/master/STYLE.md) (cc @harding @jnewbery):
- s/"blockchain"/"block chain"/

harding:
Block chain as two words is historically consistent with Bitcoin Core naming, e.g. https://github.com/bitcoin/bitcoin/pull/6666 , and it's what nearly everyone used when I started writing about Bitcoin.  I've continued using two words for consistency with my previous writing and that's why it's that way in the Optech style guide.  However, I have no opposition to the one-word term.

Beyond the one-word/two-word debate, the phrasing in this PR LGTM.  @BitcoinErrorLog FYI: the maintainers will probably ask for this to be squashed before they merge.

BitcoinErrorLog:
1. Could you link me to the other tooltip? I'm not exactly expert in this environment, but happy to review it and try to keep things consistent. I'm not sure this will be important since it is two different situations for a user. Or maybe you are proposing only having the tooltip text appear once in the code to remove redundancy?
2. I'm not personally interested in excessive hand-holding. I'm skeptical there is/will be problematic frustration about reverting pruning consequences, and I don't think window labels, as I understand them, are notably user-friendly. 
3. Blockchain is one word. There is no debate on this anymore.

ryanofsky:
> 3. Blockchain is one word. There is no debate on this anymore.

Blockchain is a word, but it has some negative connotations and associations that "block chain" avoids

hebasto:
@BitcoinErrorLog 
> 1. Could you link me to the other tooltip?

https://github.com/bitcoin-core/gui/blob/d6fe5b28dff4c512bad7c9576232d6def10608d4/src/qt/forms/intro.ui#L216

BitcoinErrorLog:
> Blockchain is a word, but it has some negative connotations and associations that "block chain" avoids

With respect, this is software, not a political forum. There is no useful statement being made by making it two words. I'll just end up trolling such topics to meaninglessness by requesting "timechain" or "time chain" is a better argument, etc. 

Would prefer productivity, personally.

Sjors:
Let's not touch the spelling of "blockchain" in this PR. That can be done for the whole codebase in a separate PR, which I'll probably NACK and mute :-)

Please squash. Regarding the other tooltip: no strong preference if you tweak it. By the time a user visits those settings, as opposed to the intro screen, they probably have a bit more knowledge.

> block chain re-downloading) could be frustrating for an inexperienced user

But I don't expect inexperienced users to want this, especially GUI users. E.g. a txindex is more likely useful for a server-type application.

harding:
> I'll just end up trolling such topics to meaninglessness by requesting "timechain" or "time chain" is a better argument, etc.

I think this threat should be taken seriously. @BitcoinErrorLog has a [history](https://twitter.com/hrdng/status/792393416093102081) of such [trolling](https://twitter.com/hrdng/status/878226283162992640).  I suggest that this message serve as a warning that such behavior here will result in a permanent ban from this project's hosted repositories.

MarcoFalke:
Please squash your commits according to https://github.com/bitcoin-core/gui/blob/master/CONTRIBUTING.md#squashing-commits

PR #21, opened 2020-07-02
Update pruning tooltip, original author BitcoinErrorLog

fluffypony:
Squashed commits from BitcoinErrorLog at his request, per the original discussion on #15: this tooltip has been adjusted to be more user-friendly and reflect what the net effect of pruning is for the user.

### DIFF
--- a/src/qt/forms/optionsdialog.ui
+++ b/src/qt/forms/optionsdialog.ui
@@ -55,7 +55,7 @@
         <item>
          <widget class="QCheckBox" name="prune">
           <property name="toolTip">
-           <string>Disables some advanced features but all blocks will still be fully validated. Reverting this setting requires re-downloading the entire blockchain. Actual disk usage may be somewhat higher.</string>
+           <string>Enabling pruning significantly reduces the disk space required to store transactions. All blocks are still fully validated. Reverting this setting requires re-downloading the entire blockchain.</string>
           </property>
           <property name="text">
            <string>Prune &amp;block storage to</string>


jnewbery:
Thanks for the PR @fluffypony ! Can you change your PR description to describe what this change does? The maintainers' merge tool includes the PR description in the merge commit, so unless BitcoinErrorLog wants to be bombarded with notifications from github for this, you should remove his handle.

fluffypony:
@jnewbery done

fluffypony:
Also let me know if you want me to flip it round and put what it does at the beginning of the description, and attribution notes at the end.

harding:
Untested ACK 2fc5efc55c886f1b874ce6cd02c9082b5bb6435a

Sjors:
utACK 2fc5efc55c886f1b874ce6cd02c9082b5bb6435a and welcome to the dark side!

fluffypony:
@Sjors Thanks! I was going to, but I GPG-sign my commits, so I thought it would be a little odd to have BEL as the author but with my GPG sig;)

Sjors:
Another approach is to put `Co-Authored-By: ` in the commit message (to be clear: no need to).

Rspigler:
I think it is very important that users know pruning disables some advanced features. Otherwise, why not always enable pruning? Rerunning IBD is a hassle - currently, users are only warned that it will happen if reverted, but not why they may have to revert.

fluffypony:
@Rspigler there's been enough bike-shedding on the previous PR; I would suggest this get merged so that feedback can be elicited from actual users, and then the tooltip can be iterated on.

Rspigler:
Feedback already has been brought by users, this has been a requested change:
https://twitter.com/mikeinspace/status/1278453115835035653

fluffypony:
Ehhhh I'm skeptical that users running the Bitcoin Core GUI are hooking a block explorer up to it. I agree that such warnings are warranted in the CLI and documentation. Is there any functionality *within the GUI* that is affected by enabling pruning? That should be noted. Some externality is (generally speaking) not the concern of this piece of software.

fluffypony:
To add: any additional software you're trying to hook into Bitcoin will come with it's own instructions, which includes a note on disabling pruning (see Electrum Personal Server, for instance, which actually works with pruning and notes when you'd want to leave it disabled).

harding:
AFAIK, the only thing pruning disables in the GUI is the ability to close a wallet, keep it closed until the last block it saw has dropped beyond the prune horizon (plus one more block), and then open up that wallet again.  We already have a warning for that case:



(I'm not counting RPCs affected by pruning that are technically available in the GUI via its RPC console.)

I personally think it's not particularly useful to say "disables some advanced features" without also providing the user with a list of those features---and that's getting beyond the amount of information a tooltip should probably contain.  Additionally, although re-downloading blocks is certainly very annoying, anyone who did it once can probably do it again without too much issue.  I don't think we need to worry too much about it, although anyone writing extended documentation about pruning should certainly consider describing the tradeoffs.

fluffypony:
@harding thanks for the extra colour - fully agree that extended docs should go into detail about what may be affected by pruning, and that it's out-of-scope for the tooltip.

Rspigler:
I think that's a very good point you brought up, pruning doesn't affect functionality /within the GUI/. Warnings in the CLI and documentation should cover it.

luke-jr:
No mention of lost functionality at all?

This impacts the ability to restore backups...

harding:
@luke-jr if you lost your original wallet file, you probably also lost your chainstate, so you probably need to IBD again anyway.

jonasschnelli:
ACK 2fc5efc55c886f1b874ce6cd02c9082b5bb6435a
Since translation freeze for 0.21.0 happened already, we have to postpone this for 0.22

PR #27, opened 2020-07-03
Move the UI window style to a top to bottom layout

Saibato:
This is part of {wip} side project i had to make the ui more
user friendly when run on mobile i.e in termux or other linux mobile environtments
that can show our qt ui.

### DIFF
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>740</width>
-    <height>430</height>
+    <width>440</width>
+    <height>440</height>
    </rect>
   </property>
   <property name="windowTitle">
@@ -15,28 +15,9 @@
   </property>
   <layout class="QVBoxLayout" name="verticalLayout_2">
    <item>
-    <widget class="QLabel" name="label_alerts">
-     <property name="visible">
-      <bool>false</bool>
-     </property>
-     <property name="styleSheet">
-      <string notr="true">QLabel { background-color: qlineargradient(x1: 0, y1: 0, x2: 1, y2: 0, stop:0 #F0D0A0, stop:1 #F8D488); color:#000000; }</string>
-     </property>
-     <property name="wordWrap">
-      <bool>true</bool>
-     </property>
-     <property name="margin">
-      <number>3</number>
-     </property>
-     <property name="textInteractionFlags">
-      <set>Qt::TextSelectableByMouse</set>
-     </property>
-    </widget>
-   </item>
-   <item>
     <widget class="QTabWidget" name="tabWidget">
      <property name="currentIndex">
-      <number>0</number>
+      <number>2</number>
      </property>
      <widget class="QWidget" name="tab_info">
       <attribute name="title">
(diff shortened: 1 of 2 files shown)


Saibato:
GUI-1

Saibato:
GUI-2

luke-jr:
This seems very undesirable on desktops... Shouldn't it be an option, or at least based on screen ratio?

Saibato:
I fiddled around with long v3 Tor and i2p addresses and they where truncated way to much and even on desktop, i guess that could make sense to have this layout for ppl, to cherrypick on top of there compile of core if testing v3 or other things when you want multiple nodes side by side. 

And since it make no sense to use the pre-builds, your are right, its an option, for sovereign multiplatform coiners.      

> be an option, or at least based on screen ratio?

i.e.` ./configure --gui-top-bottom-design ` ?

luke-jr:
I was thinking a checkbox in the GUI settings at runtime.

Saibato:
> I was thinking a checkbox in the GUI settings at runtime.

even better, but since the ui is fixed and the headers are created at compile time.
I am not quite sure how to do that? Two layouts ? ping @hebasto

hebasto:
> > I was thinking a checkbox in the GUI settings at runtime.
> 
> even better, but since the ui is fixed and the headers are created at compile time.
> I am not quite sure how to do that? Two layouts ? ping @hebasto

Probably building layouts dynamically rather using `*.ui` forms is a way, no?

luke-jr:
Or modifying the default layout in code should also be possible.

RandyMcMillan:
I believe that the over all interface should maintain a "Landscape" layout - This approach seems to be forcing a "Portrait" layout...which isn't optimal.

Either way - optimizing the use of the empty space may guide layout decisions...

For example:

Screen Shot 2020-09-14 at 6 01 11 PM


Screen Shot 2020-09-14 at 6 02 27 PM

Notice how much empty space can be utilized in other ways...

By formatting the layout of the text - this extra space can be used by the peer detail widget...

Some simple changes could be in the NodeID column - this could be simplified to ID
I don't think sorting by NodeID is particularly useful and could be disabled for better display of the label.

Also removing the slashes from the User Agent display would look better..

jarolrod (commented):
Concept ACK on reordering widgets on `overviewpage.ui`. NACK on reordering widgets under `debugwindow.ui`. This leads to a lot of wasted space under the peers tab (shown below).
peer-view

jonasschnelli:
The layout should depend on the screensize ratio (or on the window-size ratio). I agree that Core should also have an option to run on rather portrait mode screens and devices.

Ideally, the design-implementation can handle run-time window size changes and switches from landscape to top-down mode. Similar to how webpages have a floating design.

If too complicated, I agree with @luke-jr about the checkbox-in-settings approach.
A compile option seems not ideal to me.

rebroad:
Agree with @luke-jr - this should not be hardcoded but instead based on screen ratio - or better still, user-overridable.

hebasto:
The flowing GUI could be implemented in https://github.com/bitcoin-core/gui-qml/blob/main/src/qml/README.md

Closing this for now.

PR #30, opened 2020-07-11
Disable the main window toolbar when the modal overlay is shown

### LINKED ISSUE #22: The GUI appears non-responsive when the 'still syncing' overlay is visible
After the initial install there is an overlay that provides a description and progress of syncing. If a user clicks on the main UI buttons (Overview, Send, Receive, Transactions) while that overlay is visible, nothing seems to happen. The buttons respond as if clicked (background color changes) but other than that nothing seems to respond. if you look REALLY closely, you can see that actually the UI is responding and changing, but it's masked out behind the overlay screen.

Two suggestion, in my personal order of preference:

1. Clicking anywhere outside of the overlay just closes the overlay.
2. The overlay covers everything and you're essentially forced into clicking the 'Hide' button to reveal the UI behind it.

image

There's a short video of the behavior posted here:
https://twitter.com/statusquont/status/1278465322278948864

hebasto:
Keeping the main window toolbar activated while the modal overlay is shown could create the appearance of the non-responsive GUI.

Fixes #22.

---

On master (ca055885c631de8ac0ffe24be6b02835dbcc039d):

Screenshot from 2020-07-11 13-07-00

With this PR:

Screenshot from 2020-07-11 13-07-39

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -112,6 +112,8 @@ BitcoinGUI::BitcoinGUI(interfaces::Node& node, const PlatformStyle *_platformSty
         Q_EMIT consoleShown(rpcConsole);
     }
 
+    modalOverlay = new ModalOverlay(enableWallet, this->centralWidget());
+
     // Accept D&D of URIs
     setAcceptDrops(true);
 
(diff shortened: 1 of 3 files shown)


harding:
Tested ACK d0cc1f6df740e03ca0213a3754c3277b01ae2c05.  Tested on Linux/X11 as much as I could given it's a pretty small change; seems like a nice improvement.  I'm not experienced in Qt, but I don't see anything obviously problematic about the code.

leonardo-sj:
Tested ACK. Tested on Ubuntu 20.04 LTS. It can improve user experience.

polylunar:
Not sure whether it's still helpful but I tested it on MacOS and confirm it worked as well, nicely done @hebasto

jonatack:
@polylunar thanks! It is very helpful to review before or after a merge. Great to see and looking forward to more tests and review from you :rocket:

PR #34, opened 2020-07-15
Show permissions instead of whitelisted

laanwj:
Show detailed permissions instead of legacy "whitelisted" flag in the peer list details.
These are formatted with `&` in between just like services flags. It reuses the "N/A" translation message if there are no special permissions.
This removes the one-but-last use of `legacyWhitelisted`.

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -1120,7 +1120,15 @@ void RPCConsole::updateNodeDetail(const CNodeCombinedStats *stats)
     ui->peerSubversion->setText(QString::fromStdString(stats->nodeStats.cleanSubVer));
     ui->peerDirection->setText(stats->nodeStats.fInbound ? tr("Inbound") : tr("Outbound"));
     ui->peerHeight->setText(QString::number(stats->nodeStats.nStartingHeight));
-    ui->peerWhitelisted->setText(stats->nodeStats.m_legacyWhitelisted ? tr("Yes") : tr("No"));
+    if (stats->nodeStats.m_permissionFlags == PF_NONE) {
+        ui->peerPermissions->setText(tr("N/A"));
+    } else {
+        QStringList permissions;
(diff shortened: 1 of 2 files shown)


hebasto:
Concept ACK.

MarcoFalke:
Screenshot from 2020-07-16 08-03-03

Tested aCK

src/qt/rpcconsole.cpp
```diff
@@ -1120,7 +1120,15 @@ void RPCConsole::updateNodeDetail(const CNodeCombinedStats *stats)
     ui->peerSubversion->setText(QString::fromStdString(stats->nodeStats.cleanSubVer));
     ui->peerDirection->setText(stats->nodeStats.fInbound ? tr("Inbound") : tr("Outbound"));
     ui->peerHeight->setText(QString::number(stats->nodeStats.nStartingHeight));
-    ui->peerWhitelisted->setText(stats->nodeStats.m_legacyWhitelisted ? tr("Yes") : tr("No"));
```
promag, line 1123: Why wasn't `m_legacyWhitelisted` removed?

src/qt/rpcconsole.cpp
```diff
@@ -1120,7 +1120,15 @@ void RPCConsole::updateNodeDetail(const CNodeCombinedStats *stats)
...
+    if (stats->nodeStats.m_permissionFlags == PF_NONE) {
```
promag, line 1123: nit, could `#include <net_permissions.h>`.

Sjors:
Post merge concept ACK, nice!

src/qt/rpcconsole.cpp
MarcoFalke: See OP

>  This removes the one-but-last use of legacyWhitelisted.

promag: Oh right, miss interpreted "one-but-last", sorry.

amitiuttarwar: done in https://github.com/bitcoin/bitcoin/pull/20755 :)

PR #35, opened 2020-07-16
Parse params directly instead of through node (partial revert #10244)

ryanofsky:
This PR is part of the [process separation project](https://github.com/bitcoin/bitcoin/projects/10).

---

This is a partial revert of https://github.com/bitcoin/bitcoin/pull/10244. It changes gui code to go back to using gArgs and Params() functions directly instead of using interfaces::Node to handle arguments.

These changes were originally pushed as part of https://github.com/bitcoin/bitcoin/pull/19461. Motivation is to support a new GUI process connecting to an already running node process. Details are explained in commit messages, but in addition to spawning a new bitcoin-node process, we want bitcoin-gui to connect to an existing bitcoin-node process. So for that reason it should be able to parse its own parameters, rather than rely on the node.

### DIFF
--- a/src/interfaces/node.h
+++ b/src/interfaces/node.h
@@ -55,41 +55,6 @@ class Node
 public:
     virtual ~Node() {}
 
-    //! Send init error.
-    virtual void initError(const bilingual_str& message) = 0;
-
-    //! Set command line arguments.
-    virtual bool parseParameters(int argc, const char* const argv[], std::string& error) = 0;
-
-    //! Set a command line argument
-    virtual void forceSetArg(const std::string& arg, const std::string& value) = 0;
-
-    //! Set a command line argument if it doesn't already have a value
-    virtual bool softSetArg(const std::string& arg, const std::string& value) = 0;
-
-    //! Set a command line boolean argument if it doesn't already have a value
-    virtual bool softSetBoolArg(const std::string& arg, bool value) = 0;
-
-    //! Load settings from configuration file.
-    virtual bool readConfigFiles(std::string& error) = 0;
-
-    //! Choose network parameters.
-    virtual void selectParams(const std::string& network) = 0;
-
-    //! Read and update <datadir>/settings.json file with saved settings. This
-    //! needs to be called after selectParams() because the settings file
-    //! location is network-specific.
-    virtual bool initSettings(std::string& error) = 0;
-
-    //! Get the (assumed) blockchain size.
-    virtual uint64_t getAssumedBlockchainSize() = 0;
-
-    //! Get the (assumed) chain state size.
-    virtual uint64_t getAssumedChainStateSize() = 0;
-
-    //! Get network name.
-    virtual std::string getNetwork() = 0;
-
     //! Init logging.
     virtual void initLogging() = 0;
 
@@ -116,9 +81,6 @@ public:
 
     //! Return whether shutdown was requested.
     virtual bool shutdownRequested() = 0;
-
-    //! Setup arguments
-    virtual void setupServerArgs() = 0;
 
     //! Map port.
     virtual void mapPort(bool use_upnp) = 0;
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -196,10 +196,9 @@ void BitcoinCore::shutdown()
 static int qt_argc = 1;
 static const char* qt_argv = "bitcoin-qt";
 
-BitcoinApplication::BitcoinApplication(interfaces::Node& node):
+BitcoinApplication::BitcoinApplication():
     QApplication(qt_argc, const_cast<char **>(&qt_argv)),
     coreThread(nullptr),
-    m_node(node),
     optionsModel(nullptr),
     clientModel(nullptr),
     window(nullptr),
@@ -249,12 +248,13 @@ void BitcoinApplication::createPaymentServer()
 
 void BitcoinApplication::createOptionsModel(bool resetSettings)
 {
-    optionsModel = new OptionsModel(m_node, this, resetSettings);
+    optionsModel = new OptionsModel(this, resetSettings);
+    optionsModel->setNode(node());
 }
 
 void BitcoinApplication::createWindow(const NetworkStyle *networkStyle)
 {
-    window = new BitcoinGUI(m_node, platformStyle, networkStyle, nullptr);
+    window = new BitcoinGUI(node(), platformStyle, networkStyle, nullptr);
 
     pollShutdownTimer = new QTimer(window);
     connect(pollShutdownTimer, &QTimer::timeout, window, &BitcoinGUI::detectShutdown);
@@ -262,17 +262,25 @@ void BitcoinApplication::createWindow(const NetworkStyle *networkStyle)
 
 void BitcoinApplication::createSplashScreen(const NetworkStyle *networkStyle)
 {
-    SplashScreen *splash = new SplashScreen(m_node, nullptr, networkStyle);
+    assert(!m_splash);
+    m_splash = new SplashScreen(nullptr, networkStyle);
+    m_splash->setNode(node());
     // We don't hold a direct pointer to the splash screen after creation, but the splash
     // screen will take care of deleting itself when finish() happens.
-    splash->show();
-    connect(this, &BitcoinApplication::splashFinished, splash, &SplashScreen::finish);
-    connect(this, &BitcoinApplication::requestedShutdown, splash, &QWidget::close);
+    m_splash->show();
+    connect(this, &BitcoinApplication::splashFinished, m_splash, &SplashScreen::finish);
+    connect(this, &BitcoinApplication::requestedShutdown, m_splash, &QWidget::close);
+}
+
+void BitcoinApplication::setNode(interfaces::Node& node)
+{
+    assert(!m_node);
+    m_node = &node;
 }
 
 bool BitcoinApplication::baseInitialize()
 {
-    return m_node.baseInitialize();
+    return node().baseInitialize();
 }
 
 void BitcoinApplication::startThread()
@@ -280,7 +288,7 @@ void BitcoinApplication::startThread()
     if(coreThread)
         return;
     coreThread = new QThread(this);
-    BitcoinCore *executor = new BitcoinCore(m_node);
+    BitcoinCore *executor = new BitcoinCore(node());
     executor->moveToThread(coreThread);
 
     /*  communication to and from thread */
@@ -334,7 +342,7 @@ void BitcoinApplication::requestShutdown()
     window->unsubscribeFromCoreSignals();
     // Request node shutdown, which can interrupt long operations, like
     // rescanning a wallet.
-    m_node.startShutdown();
+    node().startShutdown();
     // Unsetting the client model can cause the current thread to wait for node
     // to complete an operation, like wait for a RPC execution to complete.
     window->setClientModel(nullptr);
@@ -356,7 +364,7 @@ void BitcoinApplication::initializeResult(bool success, interfaces::BlockAndHead
     {
         // Log this only after AppInitMain finishes, as then logging setup is guaranteed complete
         qInfo() << "Platform customization:" << platformStyle->getName();
-        clientModel = new ClientModel(m_node, optionsModel);
+        clientModel = new ClientModel(node(), optionsModel);
         window->setClientModel(clientModel, &tip_info);
 #ifdef ENABLE_WALLET
         if (WalletModel::isWalletEnabled()) {
@@ -456,7 +464,8 @@ int GuiMain(int argc, char* argv[])
     QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
 #endif
 
-    BitcoinApplication app(*node);
+    BitcoinApplication app;
+    app.setNode(*node);
 
     /// 2. Parse command-line options. We do this after qt in order to show an error if there are problems parsing these
     // Command-line options take precedence:
@@ -612,10 +621,10 @@ int GuiMain(int argc, char* argv[])
         }
     } catch (const std::exception& e) {
         PrintExceptionContinue(&e, "Runaway exception");
-        app.handleRunawayException(QString::fromStdString(node->getWarnings().translated));
+        app.handleRunawayException(QString::fromStdString(app.node().getWarnings().translated));
     } catch (...) {
         PrintExceptionContinue(nullptr, "Runaway exception");
-        app.handleRunawayException(QString::fromStdString(node->getWarnings().translated));
+        app.handleRunawayException(QString::fromStdString(app.node().getWarnings().translated));
     }
     return rv;
 }
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -490,7 +490,7 @@ int GuiMain(int argc, char* argv[])
     // Show help message immediately after parsing command-line options (for "-lang") and setting locale,
     // but before showing splash screen.
     if (HelpRequested(gArgs) || gArgs.IsArgSet("-version")) {
-        HelpMessageDialog help(*node, nullptr, gArgs.IsArgSet("-version"));
+        HelpMessageDialog help(nullptr, gArgs.IsArgSet("-version"));
         help.showOrPrint();
         return EXIT_SUCCESS;
     }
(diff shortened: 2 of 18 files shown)


Sjors:
Concept ACK. The reasoning as explained in the commit message makes sense to me. Maybe a better PR title would be "Parse params directly instead of through node (Partially revert #10244)". And then in the PR description: "In addition to spawning a new bitcoin-node process, we want bitcoin-gui to connect to an existing bitcoin-node proces. So for that reason it should be able to parse its own parameters, rather than rely on the node."

src/qt/test/test_main.cpp
```diff
@@ -70,7 +70,7 @@ int main(int argc, char* argv[])
     BitcoinApplication app(*node);
     app.setApplicationName("Bitcoin-Qt-test");
 
-    node->setupServerArgs();            // Make gArgs available in the NodeContext
+    node->context()->args = &gArgs;     // Make gArgs available in the NodeContext
     node->context()->args->ClearArgs(); // Clear added args again
```
MarcoFalke, line 74: This line won't be needed anymore

src/qt/optionsmodel.cpp
```diff
@@ -244,10 +244,10 @@ void OptionsModel::SetPruneEnabled(bool prune, bool force)
     const int64_t prune_target_mib = PruneGBtoMiB(settings.value("nPruneSize").toInt());
     std::string prune_val = prune ? ToString(prune_target_mib) : "0";
     if (force) {
-        m_node.forceSetArg("-prune", prune_val);
+        gArgs.ForceSetArg("-prune", prune_val);
         return;
     }
-    if (!m_node.softSetArg("-prune", prune_val)) {
+    if (!gArgs.SoftSetArg("-prune", prune_val)) {
```
MarcoFalke, line 250: Can't this use `this->node().context()->args` instead of the global, which should be gone hopefully :soon: ?

hebasto:
I'll try to describe the desired behavior as I understand it.

When a `bitcoin-gui` process starts there are two possibilities: (a) to spawn a new `bitcoin-node` process and connect to it, or (b) to connect to an existing `bitcoin-node` process.

In variant (a) the `bitcoin-gui` process parses arguments (from the command line and from a config file) and passes them to the `bitcoin-node` process, which in turn does not even look into the config file.

In variant (b) the existing `bitcoin-node` process already has parsed its arguments (from the command line and from a config file), and  the `bitcoin-gui` process parses its own command-line arguments and (the same?) config file.

Did I understand the concept correct?

How the possible conflicts between different process parsed arguments should be resolved?

src/qt/optionsmodel.cpp
ryanofsky: re: https://github.com/bitcoin-core/gui/pull/35#discussion_r456261621

> Can't this use `this->node().context()->args` instead of the global, which should be gone hopefully 🔜 ?

After https://github.com/bitcoin/bitcoin/pull/10102, there is no NodeContext in the gui process. An ArgsManager reference could be passed to OptionsModel if desired, though.

src/qt/test/test_main.cpp
ryanofsky: re: https://github.com/bitcoin-core/gui/pull/35#discussion_r456258986

> This line won't be needed anymore

Thanks, removed

hebasto:
Concept ACK.

hebasto:
Not passing an `interfaces::Node` argument to the `BitcoinApplication` and `OptionsModel` constructors implies that the internal states of the constructed objects with `m_node` initialized to `nullptr` are consistent and legitimate. Therefore, their `node()` getters shouldn't include `assert(m_node)`. I think the responsibility to check the pointer should be moved to callers.

ryanofsky:
Thanks for reviewing! I looked into many things but wound up not making changes for reasons described below.

re: https://github.com/bitcoin-core/gui/pull/35#issuecomment-660603032
> Not passing an interfaces::Node argument to the BitcoinApplication and OptionsModel constructors implies that the internal states of the constructed objects with m_node initialized to nullptr are consistent and legitimate. Therefore, their node() getters shouldn't include assert(m_node).

I don't think this follows. It's ok for functions to have preconditions, and this is a pretty straightforward precondition to have for an accessor returning a reference.

> I think the responsibility to check the pointer should be moved to callers.

We might just have a style disagreement here, but if there are specific changes you are looking for, maybe you could post them or describe them in more detail. I don't think we need to have pointers and null checks in places where the pointers can never be null. There's only one place in the PR that actually needs to check for a null node: the splash screen code handling the Q keypress needs to check if the node is connected before sending it a shutdown request. Everywhere else, references and not pointers are more appropriate and have been retained.

re: https://github.com/bitcoin-core/gui/pull/35#pullrequestreview-451120856

> I think the `m_shutdown` member should be dropped due to the following reasons:

m_shutdown is needed so the splash screen can be shown instantly without waiting for anything and the Q key can function requesting a shutdown as soon as possible. Maybe this change could be moved to https://github.com/bitcoin/bitcoin/pull/19461, but you could also say the same about all other changes in this PR. I hope the cleanups in this PR make sense by themselves, but the actual motivation for this PR is to enable https://github.com/bitcoin/bitcoin/pull/19461.

> the m_shutdown member could require be protected by a mutex

A mutex would only be required if Qt signals are used incorrectly. GUI state should be accessed by one thread.

> nano-nit: why not apply [`clang-format-diff.py`](https://github.com/bitcoin/bitcoin/tree/master/contrib/devtools#clang-format-diffpy) for each commit?

I read diffs to understand changes and having unrelated formatting changes mixed into commits makes this harder.

hebasto:
> > I think the `m_shutdown` member should be dropped due to the following reasons:
> 
> m_shutdown is needed so the splash screen can be shown instantly without waiting for anything and the Q key can function requesting a shutdown as soon as possible.

I understand this motivation. But the splash screen is not derived from `QDialog` with own event loop, it is derived from `QWidget` and it _will wait_ for the main event loop which starts in https://github.com/bitcoin-core/gui/blob/49b993ddb2cea18b7ba8691ea9b572c9e9bbcb9c/src/qt/bitcoin.cpp#L604

ryanofsky:
> > > I think the `m_shutdown` member should be dropped due to the following reasons:
> > 
> > 
> > m_shutdown is needed so the splash screen can be shown instantly without waiting for anything and the Q key can function requesting a shutdown as soon as possible.
> 
> I understand this motivation. But the splash screen is not derived from `QDialog` with its own event loop, it is derived from `QWidget` and it _will wait_ for the main event loop which starts in
> 
> https://github.com/bitcoin-core/gui/blob/49b993ddb2cea18b7ba8691ea9b572c9e9bbcb9c/src/qt/bitcoin.cpp#L604

What is this comment referring to and what is it asking for? I think we both agree handling a null pointer in the splashscreen is not strictly needed here. It's needed to handle the delayed gui->node connection implemented in https://github.com/bitcoin/bitcoin/pull/19461. The point of all the changes in this PR is to reduce size and complexity of https://github.com/bitcoin/bitcoin/pull/19461. I think it makes sense to handle null node pointer in the splash screen here in this PR which is first introducing the pointer variable, even though the variable won't have a null value until https://github.com/bitcoin/bitcoin/pull/19461.

promag:
Concept ACK.

luke-jr:
Concept ACK

hebasto:
@ryanofsky Mind rebasing? (no DrahtBot's notifications in this repo for now)

ryanofsky:
Rebased 49b993ddb2cea18b7ba8691ea9b572c9e9bbcb9c -> 1a0875f151be1827a2f880cc63b1022455665c0d ([`pr/ipc-guiarg.2`](https://github.com/ryanofsky/gui/commits/pr/ipc-guiarg.2) -> [`pr/ipc-guiarg.3`](https://github.com/ryanofsky/gui/commits/pr/ipc-guiarg.3), [compare](https://github.com/ryanofsky/gui/compare/pr/ipc-guiarg.2-rebase..pr/ipc-guiarg.3)) due to conflicts with https://github.com/bitcoin/bitcoin/pull/19561 and https://github.com/bitcoin/bitcoin/pull/15935

ryanofsky:
Rebased 1a0875f151be1827a2f880cc63b1022455665c0d -> 37b7eba5c4d3a77aa3fbd40fa7f066dad6d393c1 ([`pr/ipc-guiarg.3`](https://github.com/ryanofsky/gui/commits/pr/ipc-guiarg.3) -> [`pr/ipc-guiarg.4`](https://github.com/ryanofsky/gui/commits/pr/ipc-guiarg.4), [compare](https://github.com/ryanofsky/gui/compare/pr/ipc-guiarg.3-rebase..pr/ipc-guiarg.4)) due to conflict with https://github.com/bitcoin/bitcoin/pull/19098

ryanofsky:
Rebased 37b7eba5c4d3a77aa3fbd40fa7f066dad6d393c1 -> 333507a1f3e1941367756dc7974c7c9720d03b4a ([`pr/ipc-guiarg.4`](https://github.com/ryanofsky/gui/commits/pr/ipc-guiarg.4) -> [`pr/ipc-guiarg.5`](https://github.com/ryanofsky/gui/commits/pr/ipc-guiarg.5), [compare](https://github.com/ryanofsky/gui/compare/pr/ipc-guiarg.4-rebase..pr/ipc-guiarg.5)) due to conflict with https://github.com/bitcoin/bitcoin/pull/19011

src/interfaces/node.h
```diff
@@ -55,32 +55,6 @@ class Node
...
-    //! Choose network parameters.
-    virtual void selectParams(const std::string& network) = 0;
-
-    //! Read and update <datadir>/settings.json file with saved settings. This
-    //! needs to be called after selectParams() because the settings file
-    //! location is network-specific.
-    virtual bool initSettings(std::string& error) = 0;
-
     //! Get the (assumed) blockchain size.
     virtual uint64_t getAssumedBlockchainSize() = 0;
```
MarcoFalke, line 59: in commit 427c8da711

why is this not removed?

src/qt/bitcoin.cpp
```diff
@@ -457,11 +460,11 @@ int GuiMain(int argc, char* argv[])
 
     /// 2. Parse command-line options. We do this after qt in order to show an error if there are problems parsing these
     // Command-line options take precedence:
-    node->setupServerArgs();
+    SetupServerArgs(*node->context());
```
MarcoFalke, line 471: commit 427c8da7113527cf3809cb568cd3fe529e56b259: The documentation for context() says "Useful for testing". Did you mean to write `node_context` instead?
```suggestion
    SetupServerArgs(node_context);
```

src/interfaces/node.h
ryanofsky: re: https://github.com/bitcoin-core/gui/pull/35#discussion_r471384999

> in commit [427c8da](https://github.com/bitcoin-core/gui/commit/427c8da7113527cf3809cb568cd3fe529e56b259)
> 
> why is this not removed?

Good catch, removed now

src/qt/bitcoin.cpp
ryanofsky: re: https://github.com/bitcoin-core/gui/pull/35/files#r471397426

> commit [427c8da](https://github.com/bitcoin-core/gui/commit/427c8da7113527cf3809cb568cd3fe529e56b259): The documentation for context() says "Useful for testing". Did you mean to write `node_context` instead?

Changed, agree this is a little better. SetupServerArgs should really not be using NodeContext at all (argument is removed in https://github.com/bitcoin/bitcoin/pull/10102), but for this PR which is not in the GUI repository, I wanted to avoid changing non-gui code.

MarcoFalke:
re-run ci

src/qt/bitcoin.cpp
```diff
@@ -453,15 +464,15 @@ int GuiMain(int argc, char* argv[])
     QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
 #endif
 
-    BitcoinApplication app(*node);
+    BitcoinApplication app;
 
     /// 2. Parse command-line options. We do this after qt in order to show an error if there are problems parsing these
     // Command-line options take precedence:
-    node->setupServerArgs();
+    SetupServerArgs(node_context);
```
MarcoFalke, line 471: nit: Could add an unused reference to the argsmanager that is supposed to be used by gui code (instead of the global).

Would either be `const ArgsManager& args = *node_context.args` or `= gArgs`?

Feel free to ingore.

src/qt/bitcoin.cpp
```diff
@@ -453,15 +464,15 @@ int GuiMain(int argc, char* argv[])
...
     SetupUIArgs(gArgs);
     std::string error;
-    if (!node->parseParameters(argc, argv, error)) {
-        node->initError(strprintf(Untranslated("Error parsing command line arguments: %s\n"), error));
+    if (!gArgs.ParseParameters(argc, argv, error)) {
```
MarcoFalke, line 474: nit: (if feedback is addressed)

Could use 
```suggestion
    if (!args.ParseParameters(argc, argv, error)) {
```

here?

MarcoFalke (commented):
will re-review after rebase

src/qt/bitcoin.cpp
ryanofsky: re: https://github.com/bitcoin-core/gui/pull/35#discussion_r477205453

> Could use

(same feedback)

MarcoFalke:
re-ACK 519cae8fd6, only change is rebase and addressed nits of my previous review last week 🌄

<details><summary>Show signature and timestamp</summary>

Signature:

```
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA512

re-ACK 519cae8fd6, only change is rebase and addressed nits of my previous review last week 🌄
-----BEGIN PGP SIGNATURE-----

iQGzBAEBCgAdFiEE+rVPoUahrI9sLGYTzit1aX5ppUgFAlwqrYAACgkQzit1aX5p
pUj6Ggv+LwP3f/BQc39c/CE6YAwsMHH0yuGCrTpPSZ2Mg4LFVAMNs5Frf7X/WTK3
knd0Wqn8uXL1ohllgYw/RpANRre871edpLXP9DANpAF1mkFM8oQIwLesi2B26TWK
eQWmTbXHeXzmx3TDeyRlNbnd7YQjYWv46qaLaqgtstSzTM5KQG24D2qLL85YgdvK
iGzfDQqNsfVlt8KQa1edWtMi/tbnSKIERKdC1oGjBdWbg/hliwabsDpjaFtgACSS
CwWyz4yVrKcN3fd9KRkkA+InJIdCGxNx2Tkqvof1f7d3P9RZ3OSlVhz6b7/Z2V6a
kucVgTo/bFOdza6AQ+7K6yT3fuYOyhRvwUY2U9CD1heLUFVdaDouCVwqj1AsSTU7
/eqMwOmldW/hrxxVM+Pnm767VryXYMSkOzjBUlEKjrBgSaQSR9PPe4gZ2spAzDYX
R4CIzdXuxLnmYqgY6ynhaQI//XlBg+Kj68oNYcT/ccKuEkMreYjmnta3T7vtiB3b
SONP0kHM
=zDRb
-----END PGP SIGNATURE-----
```

[Timestamp of file with hash `ecc68c0f8f95794d4615f84e2642dbc3ee12d9db23d521fe606e763336bf3551  -`](https://opentimestamps.org/info/?004f70656e54696d657374616d7073000050726f6f6600bf89e2e884e892940108ecc68c0f8f95794d4615f84e2642dbc3ee12d9db23d521fe606e763336bf3551f010239724073576afdee67524de0753111608fff010014a38fef4c8faa94d2755b2974310ad08f1045f467155f008c13ae609bd9290e60083dfe30d2ef90c8e2c2b68747470733a2f2f626f622e6274632e63616c656e6461722e6f70656e74696d657374616d70732e6f7267fff01015083eb288009cae624da76a15c6bc5308f1045f467155f00838429a2b8eb508c00083dfe30d2ef90c8e2e2d68747470733a2f2f616c6963652e6274632e63616c656e6461722e6f70656e74696d657374616d70732e6f7267fff0105edde3c0e13206806381a82b913b6dd008f1045f467155f0086514de7d9f2f22400083dfe30d2ef90c8e292868747470733a2f2f66696e6e65792e63616c656e6461722e657465726e69747977616c6c2e636f6df010c11d147da8f5304861126df53c84118d08f1045f467155f008f9255ab0e04412aa0083dfe30d2ef90c8e232268747470733a2f2f6274632e63616c656e6461722e636174616c6c6178792e636f6d)

</details>

Issue #1, opened 2020-06-16
gui: Remove unused `Decrypt` wallet


MarcoFalke:
The Decrypt case is unused and should be removed. See

```
src/qt/askpassphrasedialog.h:        Decrypt     /**< Ask passphrase and decrypt wallet */
```

#### Want to work on this issue?

For guidance on contributing, please read [CONTRIBUTING.md](https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md) before opening your pull request.

sehyunc:
Hey MarcoFalke, I'm interested in working on this issue.

hebasto:
@sehyunc 
> Hey MarcoFalke, I'm interested in working on this issue.

Thanks! Will review your PR :)

michaelfolkson:
Have you made any progress on this @sehyunc?

sehyunc:
> Have you made any progress on this @sehyunc?

Hello, yes, I've spent some time familiarizing myself with the code base and working on this issue.

hebasto:
@MarcoFalke Mind looking into #109 ?

PR #37, opened 2020-07-19
indicate explicit to the user that the wallet balances shown is watch only.

Saibato:
This little change got me back to earth when i realized that the
balance hint *Your current spendable balance* was just a label and
not the result of of some random pubkey addresses i imported with
importaddress warping into keys for those addresses,

At least for some short moment i felt mainnet btc rich, but this change
in hint wording should be done.

edit@saibato Todo: before merge run translation Makefile after we agree in wording.

### DIFF
--- a/src/qt/overviewpage.cpp
+++ b/src/qt/overviewpage.cpp
@@ -201,7 +201,10 @@ void OverviewPage::setBalance(const interfaces::WalletBalances& balances)
     m_balances = balances;
     if (walletModel->wallet().isLegacy()) {
         if (walletModel->wallet().privateKeysDisabled()) {
+            ui->labelSpendable->setText(tr("OverviewPage", "Watch-only:"));
+            ui->labelSpendable->setVisible(true);
             ui->labelBalance->setText(BitcoinUnits::formatWithPrivacy(unit, balances.watch_only_balance, BitcoinUnits::SeparatorStyle::ALWAYS, m_privacy));
+            ui->labelBalance->setToolTip(tr("OverviewPage", "Your current balance in watch-only addresses"));
             ui->labelUnconfirmed->setText(BitcoinUnits::formatWithPrivacy(unit, balances.unconfirmed_watch_only_balance, BitcoinUnits::SeparatorStyle::ALWAYS, m_privacy));
             ui->labelImmature->setText(BitcoinUnits::formatWithPrivacy(unit, balances.immature_watch_only_balance, BitcoinUnits::SeparatorStyle::ALWAYS, m_privacy));
             ui->labelTotal->setText(BitcoinUnits::formatWithPrivacy(unit, balances.watch_only_balance + balances.unconfirmed_watch_only_balance + balances.immature_watch_only_balance, BitcoinUnits::SeparatorStyle::ALWAYS, m_privacy));


Saibato:
Untitled

michaelfolkson:
My understanding is that the language "spendable balance" is deliberate as it does not include unconfirmed transactions which are considered not yet "[spendable](https://github.com/bitcoin/bitcoin/blob/44307449f758b239fca0ad6b722bc36795ac6727/src/qt/forms/overviewpage.ui#L156)". If my understanding is correct it is a Concept NACK from me.

Saibato:
> My understanding is that the language "spendable balance" is deliberate as it does not include unconfirmed transactions 

@michaelfolkson  hmm,,. this is not about unconfirmed transactions, they are confirmed and the balance is in sum correct.
The wallet has no private keys to this balance, do you think they are "*spendable*" by this wallet and me?

michaelfolkson:
Spendable as in whoever has the keys can spend them. Not yet spendable would be funds in unconfirmed transactions. The language isn't referring to who has the keys, it is referring to whether the transactions are confirmed or not. In the example image you posted all your funds are spendable.

Saibato:
Your ... is a synonym for whoever hmm...

MarcoFalke:
Unrelated, but the wallet generally considers unconfirmed change as spendable as well

Saibato:
> Unrelated, but the wallet generally considers unconfirmed change as spendable as well

Yup, when you begin to see what we do with user eye;s, there is some work ahead.

And this is probably anyway an edge case and this wrong hint is somehow a minor bug, but since we allow watch only wallets, i still would like to have the ui state in any from that those balances are not spendable at all by the wallet that is currently visible to the user.

In general we clearly show that  watch-only utxos are unspendable and have split row to diff     , but not in this case.  An unaware user might think since in this case the ui show's up almost like the normal full spendable wallet. And i must admit when testing back and forth with wallets, for a short moment, even i thought wow they are spendbable, only after inspecting deeper i saw that this was only watched utxos,

Saibato:
@MarcoFalke  Btw. since rona has gave me lots of free timeslots, if you know others thing to be addressed, that you have in your backlog and think, would be nice to have, pls ping.

MarcoFalke:
There'd be #1 for example

Saibato:
> There'd be #1 for example

thx, for hint,
I only have looked in the codebase for two weeks or so i am not so deep in bitcoin and c++ voodoo, aside from net, tor or ip things or a bit gui. But I have a quite remarkable learning by doing pace, if i must  or have fun. But i also want not do push others, that can not keep up with such a pace hide away or think, why can she this in minutes? I am quite different from a healthy human.
So ping @sehyunc have you begun fix #1?

Saibato:
fixup-Update;
Since we now have different wallet types and one ui can show different wallets, it;s still my view that we should indicate this to the user.

I considered @michaelfolkson argument and he is right that those confirmed balances are spendable, if you have the keys, what we know from the wallet, if the keys are there, but even
for a watch only wallet, the user might have the keys,

So this fixup changes in addition also the` balancetext` label
to indicated that the wallet displayed has no keys and that in general we could only say that those shown wallet balances are current confirmed.

This is the current way we display watch only wallets, they look the same as a normal spendable full key wallet, although the wallet has no keys and we might not had them also never.


After change;

new_balance

Saibato:
> We already have the watch only icon on the status bar, maybe it should be more visible?

thx, whatever helps, my point is that up till now this design is not optimal and serve;s not  perfect our high standards .

An unaware user could even have there keys stripped from there wallet by i.e. some rpc remote or local  bytecode insert exploit or simple fileops filesytem wallet.dat exchange.
And be visual somewhat unaware of that his keys are gone, until he tries to spend in the other window,

And even long term experts could in my view, be somewhat astound, if they switch to newer node software. Where keyless and multiple wallets are possible.

luke-jr:
I can't imagine a case where it would make sense for malware to turn a read/write wallet into watch-only. Typically, they would just spend the coins to a key they control exclusively. Even if they wanted to pull a trick like this, it would be comparable to just corrupt the private keys...

Also, aren't watch-only funds spendable with stuff like hardware wallets?

Saibato:
@luke-jr I know, its not easy to follow or see the links of sense i would like to have changed in core and some of my PR are only hints and I never cared honestly, if any is merged, i done my footprint early on, but be assured, if you at some point will see the 3 stage picture that i see water clear, you will think omg, the heist of the century. imho

Saibato:
> Also, aren't watch-only funds spendable with stuff like hardware wallets?

Sure but i would say, that's technically an other wallet?

And btw, I will not push script hint pr's , but honestly why even doubt, that what i outlined
here makes obviously sense, if you compare the overall watch-only wallet layout and the standard HD layout?

And have you hovered about the little disabled hd icon on the bottom, if you tell me
great choice of words, turing-test?

hebasto:
@achow101 @meshcollider Mind looking into this PR?

meshcollider:
I'm ~0 on this, as promag said, there is already an indicator. And spendable isn't the same as confirmed so I'm not convinced on that change either,

michaelfolkson:
Spendable takes into account confirmed/unconfirmed, whether wallet contains all the key(s) needed to spend, whether sufficient time has passed for coinbase output to be spent etc. So as @meshcollider says spendable meets more conditions than just confirmed. As a result it is a Concept NACK on this change. I do wonder though whether it should be made clearer to the GUI user how "spendable" is defined either in the UI or in documentation.

kristapsk (approved):
ACK 990c2cd6b8d1f0f39fb8804c065ed1c7243318be. I had similar idea when switching between spendable and watch-only wallets on a testnet few days ago and got confused for a brief moment.

kristapsk:
> We already have the watch only icon on the status bar, maybe it should be more visible?

Yes, I didn't notice it before this comment actually. Also, with wallets having both spendable and watch-only addresses you have "Watch-only" / "Spendable" above amounts, so one might expect the same behaviour when having loaded wallet with only watch-only addresses.

Saibato:
@kristapsk 
> Yes, I didn't notice it before this comment actually. Also, with wallets having both spendable and watch-only addresses you have "Watch-only" / "Spendable" above amounts, so one might expect the same behaviour when having loaded wallet with only watch-only addresses.

exactly, if I interpret correctly the case u describe?
That's the reason the change is in its last version only for the pure watch only wallets
Since as @michaelfolkson pointed out, to change the hint in general is a bit overdone. 

Mixed stay untouched since they already display in two columns with a hinting header and this seams to be fine AFAICS.

src/qt/overviewpage.cpp
```diff
@@ -180,7 +180,10 @@ void OverviewPage::setBalance(const interfaces::WalletBalances& balances)
     m_balances = balances;
     if (walletModel->wallet().isLegacy()) {
         if (walletModel->wallet().privateKeysDisabled()) {
+            ui->labelSpendable->setText(QApplication::translate("OverviewPage", "Watch-only:", Q_NULLPTR));
```
hebasto, line 183:
```suggestion
            ui->labelSpendable->setText(tr("Watch-only:"));
```

src/qt/overviewpage.cpp
```diff
@@ -180,7 +180,10 @@ void OverviewPage::setBalance(const interfaces::WalletBalances& balances)
...
+            ui->labelSpendable->setVisible(true);
             ui->labelBalance->setText(BitcoinUnits::formatWithPrivacy(unit, balances.watch_only_balance, BitcoinUnits::SeparatorStyle::ALWAYS, m_privacy));
+            ui->labelBalance->setToolTip(QApplication::translate("OverviewPage", "Your current balance in watch-only addresses", Q_NULLPTR));
```
hebasto, line 186:
```suggestion
            ui->labelBalance->setToolTip(tr("Your current balance in watch-only addresses"));
```

hebasto (dismissed):

jarolrod:
@Saibato are you going to revisit this?

ShaMan239 (commented):
tACK 990c2cd6b8d1f0f39fb8804c065ed1c7243318be

This PR is trying to update the interface around a “Watch-Only” wallet to make it easier for a user to identify the type of wallet, by clearly displaying the wallet type over the balance of that wallet, and also by mentioning it in the ToolTip text for the balance. It works exactly as the op mentioned in the PR.

I think this PR is an improvement, and I agree with adding a label of “Watch-Only” over the balance, as it would suggest to the user that the balance is Watch-Only. But I think editing the ToolTip message is quite over-the-top and is not needed, because the message is properly conveyed to the user that the wallet is “Watch-Only” by the wallet heading. And I think the tooltip message should remain unchanged, to maintain uniformity over all types of wallets. 

**Testing on PR Branch:**
I have tested the original PR over the testnet network. It works correctly and as intended. I am attaching a screenshot here of a “Watch-Only” Wallet and the message displayed when the cursor hovers on the balance of the wallet.
Screenshot from 2021-08-02 18-29-08


**Testing Rebased on Master:**
I have rebased this branch over the latest state of the master branch of bitcoin-GUI. Here is the link to the rebased branch. The reason that I tested it rebased is that the PR branch is very old and I wanted to ensure that functionality remains the same if it is merged over the master. Additionally, rebasing on master lets me test the PR on the signet network. The PR seems to work fine on the signet network on the rebased branch. Here is a screenshot of it.
Screenshot from 2021-08-02 18-24-38

src/qt/overviewpage.cpp
Saibato: effect is the the same, but to translate her by tr makes sense since we have that in the object, and fever code is better code ;-)

Saibato:
thx, for testing this.

I merged the suggestion from @hebasto to use the Qt` r()` function here for auto translate, since we have tr in the object and so we have shorter code. ;-)
The effect is the same, and you might want to revisit and confirm that this has still the desired effect

While revisiting this PR I found a minor bug when displaying icons if the wallets are reloaded, https://github.com/bitcoin-core/gui/pull/407 you might want to test that too?

Rspigler:
With descriptor wallets, users aren't going to have mixed wallets anymore.  So I don't know if this is really necessary.  The balance should either be all watch only, or all single key, or all multi-signature, etc.

There probably could be some GUI improvements, but I don't think this is complete/or going in the right direction.

Saibato:
> With descriptor wallets, users aren't going to have mixed wallets anymore. So I don't know if this is really necessary. The balance should either be all watch only, or all single key, or all multi-signature, etc.
> 
> There probably could be some GUI improvements, but I don't think this is complete/or going in the right direction.

this is for sure an edge case of bare keyless and not mixed wallets. afaics mixed wallets are handled already correct.
I would guess, core will for ever have wallets without keys, and some bold obvious distinction in UI to the user that those spendable coins are in a watch-only wallets without keys, is in my view good UX. 
Afaics that is what this PR intends and archives.

Rspigler:
How would it label a multisignature wallet where less than `m` privkeys are on the device? It's neither "watch only" or "spendable"?

Rspigler:
After further thought, I like the current change (the "Watch-Only" label).  I just don't know what the solution is for multisignature wallets.

There's lots of different ways for wallets to be setup and I think that complicates this.  Don't know if @Bosch-0 has thought about this.  For example:

You could have a cold storage situation with a full node being in watch only mode.  Then `n` wallets each with a privkey.  The GUI for the full node vs for the wallets should probably represent the balance in different ways? (The wallets have one "piece of the puzzle" to spend, the node has none).

Bosch-0:
I think Specter and Nunchuk both go about solving this issue quite eloquently. I think these type of products are what the GUI should strive to be, hopefully a lot easier with the work being done with QML GUI.

You can actually go through this UX flow using `n` local core wallets in Specter.

hebasto:
@Saibato 

Please squash your commits, and rebase the branch to avoid CI failures.

shaavan (commented):
reACK 84fa30fccfe9d88ec8db019251ea0bae4852dbeb

The only change since my last review is that `QApplication::translate` has been replaced with `tr().`
Adding new screenshots to indicate no change in perceived behavior

**PR:**
pr

**PR Rebased on latest Master branch:**
rebased

Just squash the commits, and it would be perfect!

achow101:
Concept ACK

I don't think this should be limited to just legacy wallets - descriptor wallets can have private keys disabled as well. This should probably apply to all wallets that have private keys disabled, as the the watch-only icon does.

hebasto:
@Saibato 

Are you still working on this? If so, mind addressing @achow101's https://github.com/bitcoin-core/gui/pull/37#issuecomment-1575628144 please?

pablomartin4btc (commented):
tACK 8f9fd44146c3a0512dd9da657aeedb4339e050d1

Current change doesn't display the desirable goal:

image

I agree with @achow101, the change should be moved outside the legacy logic.

The commit title/ message could start with "`gui: `", it shouldn't finish with a dot I think, and perhaps the commit body could be rephrased a bit to explain more direct the purpose of this change, eg (please check the commit [guidelines](https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#committing-patches)):
```
gui: Show spendable balance for keyless watch-only

Make labelSpendable visible for watch only wallets
with their private keys disabled.
Also set the labelBalance accordingly for the same
condition.

```

hebasto:
Closing due to lack of support from the author.

PR #38, opened 2020-07-20
Add ability to integrate the client with desktop environment on Linux

### LINKED ISSUE #31: Generate .desktop file for Gnome and other desktop environments.
I got this idea from Android Studio/IntelliJ where they have a menu option to "Create Desktop entry..." and it generates a .desktop file for Gnome for example.

I created an entry for myself, but I didn't know where to get the [icons](https://github.com/bitcoin/bitcoin/tree/master/share/pixmaps) at first. And I still don't know where the green icon for Testnet comes from. It would be convenient to have this option on bitcoin-qt as well.

hebasto:
On Linux a new menu item "Settings" --> "Integrate with desktop environment" is available now (non-regtest only).

Here are some screenshots:

Screenshot from 2020-07-20 18-36-52

DeepinScreenshot_select-area_20200720190707

Screenshot from 2020-07-20 18-41-08

Closes #31

---

Localization of the  `*.desktop` files is out of this PR scope.

### DIFF
--- a/build_msvc/libbitcoin_qt/libbitcoin_qt.vcxproj
+++ b/build_msvc/libbitcoin_qt/libbitcoin_qt.vcxproj
@@ -23,6 +23,7 @@
     <ClCompile Include="..\..\src\qt\createwalletdialog.cpp" />
     <ClCompile Include="..\..\src\qt\csvmodelwriter.cpp" />
     <ClCompile Include="..\..\src\qt\editaddressdialog.cpp" />
+    <ClCompile Include="..\..\src\qt\guifileutil.cpp" />
     <ClCompile Include="..\..\src\qt\guiutil.cpp" />
     <ClCompile Include="..\..\src\qt\intro.cpp" />
     <ClCompile Include="..\..\src\qt\modaloverlay.cpp" />
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -119,6 +119,7 @@ BITCOIN_QT_H = \
   qt/csvmodelwriter.h \
   qt/editaddressdialog.h \
   qt/guiconstants.h \
+  qt/guifileutil.h \
   qt/guiutil.h \
   qt/intro.h \
   qt/macdockiconhandler.h \
@@ -219,6 +220,7 @@ BITCOIN_QT_BASE_CPP = \
   qt/bitcoinunits.cpp \
   qt/clientmodel.cpp \
   qt/csvmodelwriter.cpp \
+  qt/guifileutil.cpp \
   qt/guiutil.cpp \
   qt/intro.cpp \
   qt/modaloverlay.cpp \
--- a/src/qt/addressbookpage.cpp
+++ b/src/qt/addressbookpage.cpp
@@ -12,6 +12,7 @@
 #include <qt/addresstablemodel.h>
 #include <qt/csvmodelwriter.h>
 #include <qt/editaddressdialog.h>
+#include <qt/guifileutil.h>
 #include <qt/guiutil.h>
 #include <qt/platformstyle.h>
 
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -8,6 +8,7 @@
 #include <qt/clientmodel.h>
 #include <qt/createwalletdialog.h>
 #include <qt/guiconstants.h>
+#include <qt/guifileutil.h>
 #include <qt/guiutil.h>
 #include <qt/modaloverlay.h>
 #include <qt/networkstyle.h>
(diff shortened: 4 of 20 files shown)


jonatack:
Concept ACK

hebasto:
Updated 2bd1394dce8cbe50a82c1e11aeb1edba6feabebb -> 5dc09b5526a28c92f6b8640fc5234ff3a93a0ee3 ([pr38.01](https://github.com/hebasto/gui/commits/pr38.01) -> [pr38.02](https://github.com/hebasto/gui/commits/pr38.02), [diff](https://github.com/hebasto/gui/compare/pr38.01..pr38.02)):

- improved user-faced "Comment" field
- screenshot in the OP has been updated

luke-jr:
Approach NACK. Just restore the `.desktop` file removed in fa0e1e2f630e6ca6cdfca64d8d92a7f0f85baa4d and add the missing bits to `configure.ac`. There's no reason to go through the application to generate this...

caiofaustino:
Tested, works well for me. 
File created at `~/.local/share/applications/org.bitcoincore.BitcoinQt.desktop`
Screenshot from 2020-07-21 00-44-20

caiofaustino:
> Approach NACK. Just restore the `.desktop` file removed in [fa0e1e2](https://github.com/bitcoin-core/gui/commit/fa0e1e2f630e6ca6cdfca64d8d92a7f0f85baa4d) and add the missing bits to `configure.ac`. There's no reason to go through the application to generate this...
Can you explain what behaviour that approach would have for a user downloading the binary from the website?
I'm new to most of this, my understanding was that `configure.ac.` only affected  compilation.

luke-jr:
>Can you explain what behaviour that approach would have for a user downloading the binary from the website?

There'd presumably be a .desktop file in the tarball they could install to the right place.

caiofaustino:
> There'd presumably be a .desktop file in the tarball they could install to the right place.

Thanks for the explanation, yeah that could be an alternative, although I think this one is more user friendly, I'm not sure what level of technical expertise is expected from the users.

hebasto:
> Approach NACK. Just restore the `.desktop` file removed in [fa0e1e2](https://github.com/bitcoin-core/gui/commit/fa0e1e2f630e6ca6cdfca64d8d92a7f0f85baa4d) and add the missing bits to `configure.ac`. There's no reason to go through the application to generate this...

> > Can you explain what behaviour that approach would have for a user downloading the binary from the website?
> 
> There'd presumably be a .desktop file in the tarball they could install to the right place.

Placing a `*.desktop` file into the tarball requires placing icon files into it as well.

Manually installing a `*.desktop` and icon files seems not user-friendly.

As a `*.desktop` and icon files usually are parts of packages (e.g., https://snapcraft.io/bitcoin-core, https://packages.debian.org/sid/main/bitcoin-qt), it is possible that a `configure` option is required to disable this feature for package maintainers.
@MarcoFalke What do you think about it?

luke-jr:
Packagers would simply switch to the bundled icon/desktop files...

hebasto:
> Packagers would simply switch to the bundled icon/desktop files...

Yes, that is what I mean: if icon/desktop files are bundled into the package there is no need to compile the feature introduced in this PR; therefor it could be disabled by the `configure` script.

hebasto:
> Approach NACK. Just restore the `.desktop` file removed in [fa0e1e2](https://github.com/bitcoin-core/gui/commit/fa0e1e2f630e6ca6cdfca64d8d92a7f0f85baa4d) and add the missing bits to `configure.ac`. There's no reason to go through the application to generate this...

The alternative approach is https://github.com/bitcoin/bitcoin/pull/15068

hebasto:
Rebased 5dc09b5526a28c92f6b8640fc5234ff3a93a0ee3 -> 95c0cff67f98aa1798cc1b0e9fd39fa33bd4dbe5 ([pr38.02](https://github.com/hebasto/gui/commits/pr38.02) -> [pr38.03](https://github.com/hebasto/gui/commits/pr38.03)) due to the conflict with https://github.com/bitcoin/bitcoin/pull/15704.

jarolrod (approved):
Tested ACK on Arch Linux and KDE Plasma. Works smoothly.
Also tested to make sure this option doesn't show up on macOS. 

One question: the option does not show up when testing on FreeBSD, how do we add support for the BSD's?

hebasto:
@jarolrod 
> Tested ACK on Arch Linux and KDE Plasma. Works smoothly.
> Also tested to make sure this option doesn't show up on macOS.

Thank you for testing.

> One question: the option does not show up when testing on FreeBSD, how do we add support for the BSD's?

I'm not familiar with BSD world, except of macOS :)

hebasto:
Closing and leaving it up for grabs.

caiofaustino:
What does leaving for grabs mean? Is this feature still considered or does it need further discussion?

hebasto:
> What does leaving for grabs mean? Is this feature still considered or does it need further discussion?

Anyone could continue this work. Although, a new pr will be a subject for discussion.

PR #39, opened 2020-07-21
Add visual accenting for the 'Create new receiving address' button

### LINKED ISSUE #24: Give primary action items more visual weight
The Receive panel indicates that "all fields are optional" and if that is indeed the case, there is no clear visual cues what the correct action is -- it's to click the 'Create New Receiving Address' button.

Suggestion: give primary action buttons an orange background.

image

Here's an example from GitHub settings where a UI has multiple actions, but they visually guide the user to the action they should be taking with a colored button.

image

hebasto:
Fix #24 

The first commit:
- visual improvement with no behavior change

The second commit:
- removes a bunch of LOCs
- slightly change behavior and makes it standard

With this PR:
DeepinScreenshot_select-area_20200721213040

### DIFF
--- a/src/qt/forms/receivecoinsdialog.ui
+++ b/src/qt/forms/receivecoinsdialog.ui
@@ -114,6 +114,12 @@
              <iconset resource="../bitcoin.qrc">
               <normaloff>:/icons/receiving_addresses</normaloff>:/icons/receiving_addresses</iconset>
             </property>
+            <property name="autoDefault">
+             <bool>false</bool>
+            </property>
+            <property name="default">
+             <bool>true</bool>
+            </property>
            </widget>
           </item>
           <item>
--- a/src/qt/receivecoinsdialog.cpp
+++ b/src/qt/receivecoinsdialog.cpp
@@ -242,22 +242,6 @@ void ReceiveCoinsDialog::resizeEvent(QResizeEvent *event)
     columnResizingFixer->stretchColumnWidth(RecentRequestsTableModel::Message);
 }
 
-void ReceiveCoinsDialog::keyPressEvent(QKeyEvent *event)
-{
-    if (event->key() == Qt::Key_Return)
-    {
-        // press return -> submit form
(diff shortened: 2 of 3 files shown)


Saibato:
Concept tACK https://github.com/hebasto/gui/tree/200721-prim

While testing this, at first my intuition was to click the bottom down menu, but there we have no create address.
When i locked closer, i wondered if we could change the bottom down text for show addresses.

from  From *Receiving addresses* to *Show all receiving addresses* ?
To indicate that we do not create there new addresses

new_receive

promag:
Tested ACK 4ec49f8d1e25b330e6a0f79ae897d98d29ff32f9 on macos.

src/qt/forms/receivecoinsdialog.ui
```diff
@@ -114,6 +114,12 @@
              <iconset resource="../bitcoin.qrc">
               <normaloff>:/icons/receiving_addresses</normaloff>:/icons/receiving_addresses</iconset>
             </property>
+            <property name="autoDefault">
+             <bool>false</bool>
+            </property>
```
luke-jr, line 119: I don't think we need to mess with this?

hebasto: This change _is_ required.

From Qt [docs](https://doc.qt.io/qt-5/qpushbutton.html#autoDefault-prop):
> **autoDefault : bool**
> ...
> This property's default is true for buttons that have a QDialog parent; otherwise it defaults to false.

 As `ReceiveCoinsDialog` is derived from `QDialog` its buttons have `autoDefault == true` by default.

Having `autoDefault == true` breaks rendering of the default button when tabbing through it (when it is focused out). At least on Linux Mint 20 with system Qt 5.12.8.

luke-jr: Weird, sounds like a Qt bug?

hebasto: Maybe. Didn't track it down though.

Sjors:
utACK

PR #40, opened 2020-07-25
Clarify block height label

hebasto:
Prefer "block height" instead of "number of blocks".

This was done while testing https://github.com/bitcoin/bitcoin/pull/16981.

### DIFF
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -290,7 +290,7 @@
        <item row="11" column="0">
         <widget class="QLabel" name="label_3">
          <property name="text">
-          <string>Current number of blocks</string>
+          <string>Current block height</string>
          </property>
         </widget>
        </item>


src/qt/forms/debugwindow.ui
```diff
@@ -290,7 +290,7 @@
        <item row="11" column="0">
         <widget class="QLabel" name="label_3">
          <property name="text">
-          <string>Current number of blocks</string>
+          <string>Current block height</string>
```
promag, line 293: Funny that "height" is not in the translations.

michaelfolkson:
ACK b6dcc6d74186eee15eda2cb6e8a7ab5b5b4a05f8. I don't think there are any other obvious examples in the GUI where "block height" should replace "number of blocks" except for translations.

MarcoFalke:
```
$ git grep --ignore-case 'block height' | wc
     65     793    7793
$ git grep --ignore-case 'number of blocks' | wc
    141    1125   12941

**Comment by hebasto** [comment]
```
$ git grep --ignore-case 'block height' -- :^doc :^src/qt/locale | wc -l
59
$ git grep --ignore-case 'number of blocks' -- :^doc :^src/qt/locale | wc -l
36
```

MarcoFalke:
Fair enough. I wanted to show that both wordings are used, so it seems arbitrary to change only one?

MarcoFalke:
cr ACK b6dcc6d74186eee15eda2cb6e8a7ab5b5b4a05f8

probably doesn't matter much, but shouldn't hurt either

hebasto:
> Fair enough. I wanted to show that both wordings are used, so it seems arbitrary to change only one?

In the GUI this is one, I guess.

MarcoFalke:
block height is probably better because it implies linearity (chain) instead of number of blocks, which could also mean the number of blocks in the block tree.

hebasto:
@MarcoFalke 
> block height is probably better because it implies linearity (chain) instead of number of blocks, which could also mean the number of blocks in the block tree.

Exactly.

PR #42, opened 2020-07-31
qt: remove unused 'Decrypt' wallet

sehyunc:
Fix #1
remove 'Decrypt' case from enum in AskPassPhraseDialog
remove 'Decrypt' case from switch statements
show only Encrypt option in dialog
would appreciate feedback, specifically on src/qt/walletview.cpp:265

PR #43, opened 2020-08-03
bugfix: Call setWalletActionsEnabled(true) only for the first wallet

hebasto:
On master (a78742830aa35bf57bcb0a4730977a1e5a1876bc) there is a bug:
- open an encrypted wallet; please note that the "Encrypt Wallet..." menu item is disabled that is expected:
Screenshot from 2020-08-03 12-38-37
- then open any other wallet; note that the "Encrypt Wallet..." menu item gets enabled that is wrong:
Screenshot from 2020-08-03 12-42-36

This PR fixes this bug.

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -664,14 +664,15 @@ void BitcoinGUI::addWallet(WalletModel* walletModel)
 {
     if (!walletFrame) return;
     if (!walletFrame->addWallet(walletModel)) return;
-    const QString display_name = walletModel->getDisplayName();
-    setWalletActionsEnabled(true);
     rpcConsole->addWallet(walletModel);
-    m_wallet_selector->addItem(display_name, QVariant::fromValue(walletModel));
-    if (m_wallet_selector->count() == 2) {
+    if (m_wallet_selector->count() == 0) {
+        setWalletActionsEnabled(true);
+    } else if (m_wallet_selector->count() == 1) {
         m_wallet_selector_label_action->setVisible(true);
         m_wallet_selector_action->setVisible(true);
     }
+    const QString display_name = walletModel->getDisplayName();
+    m_wallet_selector->addItem(display_name, QVariant::fromValue(walletModel));
 }
 
 void BitcoinGUI::removeWallet(WalletModel* walletModel)


hebasto:
The bug initially was reported by @Saibato in #42:
- https://github.com/bitcoin-core/gui/pull/42#issuecomment-667697341
- https://github.com/bitcoin-core/gui/pull/42#issuecomment-667827606

hebasto:
friendly ping @achow101 @meshcollider @promag @Sjors

src/qt/bitcoingui.cpp
```diff
@@ -664,14 +664,15 @@ void BitcoinGUI::addWallet(WalletModel* walletModel)
...
     if (!walletFrame) return;
     if (!walletFrame->addWallet(walletModel)) return;
-    const QString display_name = walletModel->getDisplayName();
-    setWalletActionsEnabled(true);
     rpcConsole->addWallet(walletModel);
-    m_wallet_selector->addItem(display_name, QVariant::fromValue(walletModel));
-    if (m_wallet_selector->count() == 2) {
+    if (m_wallet_selector->count() == 0) {
+        setWalletActionsEnabled(true);
+    } else if (m_wallet_selector->count() == 1) {
```
achow101, line 670: Why does this change from 2 to 1?

hebasto: Because `m_wallet_selector->addItem()` is called after this code now.

hebasto:
@achow101 
> It's not clear to me why this fixes the problem. Wouldn't a better fix be to enable/disable that menu item depending on the wallet being selected?

This is exactly how it works with this PR :)
When a new wallet added in `m_wallet_selector->addItem()`, a signal calls `BitcoinGUI::setCurrentWalletBySelectorIndex()` slot that in turn calls `BitcoinGUI::setCurrentWallet()`.

Things are broken in master due to the unconditional call `setWalletActionsEnabled(true)`.

achow101:
> When a new wallet added in `m_wallet_selector->addItem()`, a signal calls `BitcoinGUI::setCurrentWalletBySelectorIndex()` slot that in turn calls `BitcoinGUI::setCurrentWallet()`.
> 
> Things are broken in master due to the _unconditional_ call `setWalletActionsEnabled(true)`.

`BitcoinGUI::setCurrentWallet()` should be occurring after `setWalletActionsEnabled(true)`, so in theory, it should be fine to do it unconditionally. So why doesn't it work?

hebasto:
@achow101 
I was a bit wrong describing adding a new wallet ((

When no wallets are open, and a new first wallet is added, the `currentIndexChanged(0)` signal is emitted with parameter `index=0`. This signal calls `setCurrentWalletBySelectorIndex(0)` slot. Result: wallet menu items are set correctly.

When any wallet is open already, and a non-first  wallet is added, the `currentIndexChanged()` signal is _not_ emitted. And `setCurrentWalletBySelectorIndex()` is not called. Result: wallet menu items are ignored.

In the second case the `setWalletActionsEnabled(true)` call enables all wallet menu items unconditionally. That is wrong.

src/qt/bitcoingui.cpp
promag: Why not just add the `count == 1` case in the old code?

hebasto: > Why not just add the `count == 1` case in the old code?

It would change wallet selector behavior, and it won't fix a bug due to the unconditional `setWalletActionsEnabled(true)` call.

> Edit: if it's due to some other signal/slot then I'd say to refactor to remove that dependency.

Not sure what do you mean. FWIW, moving this logic into `setCurrentWalletBySelectorIndex()` or `setCurrentWallet()` won't work as these methods do not know if a wallet has been added or removed.

jonasschnelli:
Tested ACK 20c9e035543892e322c7134e89eb33115678bb30 - I could reproduce the issue on master and have verify that this PR fixes it.

achow101:
ACK 20c9e035543892e322c7134e89eb33115678bb30

PR #59, opened 2020-08-14
Do not block GUI thread in RPCConsole

### LINKED ISSUE #53: Shutdown is blocked if called before gettxoutsetinfo is finished
Ported from https://github.com/bitcoin/bitcoin/issues/13217.

Steps to reproduce:
1) Open QT console.
2) Run gettxoutsetinfo.
3) Exit the debug window (no blocking) -> exit the program (will block).

This is what is shown on Windows (10) for several minutes:
[https://i.imgur.com/AyexgkC.png](https://i.imgur.com/AyexgkC.png)

hebasto:
Ported from https://github.com/bitcoin/bitcoin/pull/17659.

On master (b4d0366b47dd9b8fe29cc9a100dcdf6ca1d3cabf) the GUI thread is blocked with [`QThread::wait()`](https://doc.qt.io/qt-5/qthread.html#wait) during `bitcoin-qt` shutdown routine. This causes unresponsive GUI if some commands are passed to the RPC console (#53) before shutdown initiating.

This PR:
- removes blocking call and uses additional signal-to-slot connections.
- makes `bitcoin-qt` shutdown routine more streamlined: the only [`QApplication::exec()`](https://doc.qt.io/qt-5/qapplication.html#exec) is used in `bitcoin-qt`. Therefore, the main event loop never interrupts until shutdown.

Fix #53

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -98,6 +98,8 @@ BitcoinGUI::BitcoinGUI(interfaces::Node& node, const PlatformStyle *_platformSty
     updateWindowTitle();
 
     rpcConsole = new RPCConsole(node, _platformStyle, nullptr);
+    connect(rpcConsole, &RPCConsole::executorThreadFinished, this, &BitcoinGUI::rpcExecutorThreadFinished);
+
     helpMessageDialog = new HelpMessageDialog(this, false);
 #ifdef ENABLE_WALLET
     if(enableWallet)
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -228,8 +228,6 @@ BitcoinGUI::~BitcoinGUI()
 
     QSettings settings;
     settings.setValue("MainWindowGeometry", saveGeometry());
-    if(trayIcon) // Hide tray icon, as deleting will let it linger until quit (on Ubuntu)
-        trayIcon->hide();
 #ifdef Q_OS_MAC
     delete m_app_nap_inhibitor;
     delete appMenuBar;
@@ -376,8 +374,6 @@ void BitcoinGUI::createActions()
     connect(toggleHideAction, &QAction::triggered, this, &BitcoinGUI::toggleHidden);
     connect(showHelpMessageAction, &QAction::triggered, this, &BitcoinGUI::showHelpMessageClicked);
     connect(openRPCConsoleAction, &QAction::triggered, this, &BitcoinGUI::showDebugWindow);
-    // prevents an open debug window from becoming stuck/unusable on client shutdown
-    connect(quitAction, &QAction::triggered, rpcConsole, &QWidget::hide);
 
 #ifdef ENABLE_WALLET
     if(walletFrame)
@@ -625,11 +621,6 @@ void BitcoinGUI::setClientModel(ClientModel *_clientModel, interfaces::BlockAndH
     } else {
         // Disable possibility to show main window via action
         toggleHideAction->setEnabled(false);
-        if(trayIconMenu)
-        {
-            // Disable context menu on tray icon
-            trayIconMenu->clear();
-        }
         // Propagate cleared model to child objects
         rpcConsole->setClientModel(nullptr);
 #ifdef ENABLE_WALLET
@@ -1167,9 +1158,6 @@ void BitcoinGUI::closeEvent(QCloseEvent *event)
     {
         if(!clientModel->getOptionsModel()->getMinimizeOnClose())
         {
-            // close rpcConsole in case it was open to make some space for the shutdown window
-            rpcConsole->close();
-
             QApplication::quit();
         }
         else
@@ -1363,8 +1351,6 @@ void BitcoinGUI::detectShutdown()
 {
     if (m_node.shutdownRequested())
     {
-        if(rpcConsole)
-            rpcConsole->hide();
         qApp->quit();
     }
 }
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -1395,6 +1395,14 @@ void BitcoinGUI::showModalOverlay()
         modalOverlay->toggleVisibility();
 }
 
+void BitcoinGUI::hideAll()
+{
+    trayIconMenu->clear();
+    trayIcon->hide();
+    rpcConsole->hide();
+    hide();
+}
+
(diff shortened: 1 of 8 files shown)


hebasto:
Rebased 47912c505eebc79c849cf14fcf9d2654caa0945a -> 5549c453b1817084dd9b5f463ecd4d8277b973e8 ([pr59.01](https://github.com/hebasto/gui/commits/pr59.01) -> [pr59.02](https://github.com/hebasto/gui/commits/pr59.02)) due to the conflict with #35.

hebasto:
@MarcoFalke Maybe add "Bug" label?

src/qt/bitcoingui.cpp
```diff
@@ -1395,6 +1395,14 @@ void BitcoinGUI::showModalOverlay()
         modalOverlay->toggleVisibility();
 }
 
+void BitcoinGUI::clearGUI()
+{
+    trayIconMenu->clear();
```
promag, line 1389: Why clear here?

hebasto: The `trayIconMenu` contains some `QAction` instances. I think we do not need any chance to activate any of them during shutdown.

hebasto:
Updated 5549c453b1817084dd9b5f463ecd4d8277b973e8 -> 38efc5a11294ac56a9a098139813b94faf3979e9 ([pr59.02](https://github.com/hebasto/gui/commits/pr59.02) -> [pr59.03](https://github.com/hebasto/gui/commits/pr59.03), [diff](https://github.com/hebasto/gui/compare/pr59.02..pr59.03)):
- addressed @promag's [comment](https://github.com/bitcoin-core/gui/pull/59#discussion_r494936991):
> `hideAll`?

src/qt/bitcoingui.h
```diff
@@ -320,6 +320,9 @@ public Q_SLOTS:
     void setTrayIconVisible(bool);
 
     void showModalOverlay();
+
+    /** Hide all windows and tray icon. */
+    void clearGUI();
```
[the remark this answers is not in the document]
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui/pull/59#issuecomment-699646853).

jonasschnelli:
Tested this a bit.
* I can't shutdown with this PR when I call `gettxoutsetinfo` in the RPC console
* Calling `generate(1000, getnewaddress())` still blocks the GUI thread

What I'd like to see is:
* Keep the RPC console synchronous "single channeled". If I call `gettxoutsetinfo`, regardless of the non-blocking GUI mode, it should **not** allow to enter a next command before the old has completed.
* Things like "`generate(1000, getnewaddress())`" should not block the GUI and there should be an idle indicator (maybe just the text "executing...") in the RPC console.

luke-jr:
Seems like it would be even better to allow multiple RPC executions, and insert the return values in the applicable place :)

promag:
> Seems like it would be even better to allow multiple RPC executions, and insert the return values in the applicable place :)

Why? Seems a source of confusion.

hebasto:
@jonasschnelli 
>  - Calling `generate(1000, getnewaddress())` still blocks the GUI thread

I've made some debugging, and now I believe that this is [another](https://github.com/bitcoin-core/gui/issues/122) issue, which is uncorrelated to #53, and it has nothing to do with the `RPCExecutor` thread.

Not sure if #122 should be fixed at all, as it is a pretty edge case, and a possible fix could touch the code at least in three important places.

hebasto:
@jonasschnelli 
> What I'd like to see is:
> 
>  - Keep the RPC console synchronous "single channeled". If I call `gettxoutsetinfo`, regardless of the non-blocking GUI mode, it should **not** allow to enter a next command before the old has completed.
> 
>  - Things like "`generate(1000, getnewaddress())`" should not block the GUI and there should be an idle indicator (maybe just the text "executing...") in the RPC console.

Mind looking into #123 ?

jarolrod:
ACK 38efc5a11294ac56a9a098139813b94faf3979e9, Tested on macOS 11.1 with Qt 5.15.2

Confirming that I can replicate the bug described in #53 on master.

Confirming that this PR allows me to shutdown the GUI right after calling `gettxoutsetinfo` both in the `Console Window` and through `bitcoin-cli gettxoutsetinfo` after running the GUI with the `-server` option.

Confirming that I can replicate the issue laid out by @jonasschnelli [comment](https://github.com/bitcoin-core/gui/pull/59#issuecomment-715275185) about `generate`. There is no longer a generate command available in the `Console Window`, instead I replicated with `bitcoin-cli -generate 10000` after running bitcoin-qt with `-server`. But, this bug is meant to be fixed by #123.

jarolrod (commented):
ACK 38efc5a11294ac56a9a098139813b94faf3979e9, Tested on macOS 11.1 with Qt 5.15.2

Confirming that I can replicate the bug described in #53 on master.

Confirming that this PR allows me to shutdown the GUI right after calling `gettxoutsetinfo` both in the `Console Window` and through `bitcoin-cli gettxoutsetinfo` after running the GUI with the `-server` option.

Confirming that I can replicate the issue laid out by @jonasschnelli [comment](https://github.com/bitcoin-core/gui/pull/59#issuecomment-715275185) about `generate`. There is no longer a generate command available in the `Console Window`, instead I replicated with `bitcoin-cli -generate 10000` after running bitcoin-qt with `-server`. But, this bug is meant to be fixed by #123.

hebasto:
Rebased 38efc5a11294ac56a9a098139813b94faf3979e9 -> 84a9a7ce4ddf6627afca42e6ed6fd8ead55503f9 ([pr59.03](https://github.com/hebasto/gui/commits/pr59.03) -> [pr59.04](https://github.com/hebasto/gui/commits/pr59.04)) due to the conflicts with #115 and https://github.com/bitcoin/bitcoin/pull/21328.

src/qt/bitcoingui.cpp
promag: How about `trayIconMenu->setEnabled(false)` instead?

hebasto:
https://github.com/bitcoin-core/gui/issues/53#issuecomment-931322978:

> On the current master ([bd40cd8](https://github.com/bitcoin-core/gui/commit/bd40cd81083a5bde15b90e66855255457c8fb58c)) this bug is no longer reproducible. However, the blocking call `QThread::wait()` is still present in the `RPCConsole::setClientModel` function.
> 
> Tested on Linux Mint 20.2 (Qt 5.12.8) and on Windows 10 Pro 20H2 (MSVC build).
> 
> Closing for now.

Closing.

PR #61, opened 2020-08-14
refactor: Drop unneeded Q_DECLARE_METATYPE

hebasto:
Ported from https://github.com/bitcoin/bitcoin/pull/19108.

For `typedef`s, which is `CAmount`, [`qRegisterMetaType(const char*)`](https://doc.qt.io/qt-5/qmetatype.html#qRegisterMetaType) only is required.

No behavior change.

### DIFF
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -44,6 +44,7 @@
 #include <QLibraryInfo>
 #include <QLocale>
 #include <QMessageBox>
+#include <QMetaType>
 #include <QSettings>
 #include <QThread>
 #include <QTimer>
@@ -60,9 +61,11 @@ Q_IMPORT_PLUGIN(QCocoaIntegrationPlugin);
 #endif
 #endif
 
-// Declare meta types used for QMetaObject::invokeMethod
+// Make types (not aliases) available to QVariant and other template-based Qt functions.
+// Additionally qRegisterMetaType() calls are required to make them available to
+// non-template based Qt functions, including those that use Qt::QueuedConnection.
+// Run `QT_FATAL_WARNINGS=1 bitcoin-qt -debug=qt` to identify missing type registrations.
 Q_DECLARE_METATYPE(bool*)
-Q_DECLARE_METATYPE(CAmount)
 Q_DECLARE_METATYPE(SynchronizationState)
 Q_DECLARE_METATYPE(uint256)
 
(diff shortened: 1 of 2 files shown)


hebasto:
bitcoin/bitcoin#19108 has ACK from @promag.

src/qt/bitcoin.cpp
```diff
@@ -60,9 +61,10 @@ Q_IMPORT_PLUGIN(QCocoaIntegrationPlugin);
 #endif
 #endif
 
-// Declare meta types used for QMetaObject::invokeMethod
+// Make types (not aliases) available to QVariant and other template-based Qt functions.
+// Additionally qRegisterMetaType() calls are required to make them available to
+// non-template based Qt functions, including those that use Qt::QueuedConnection.
```
promag, line 66: nit, could mention env `QT_FATAL_WARNINGS=1` as a way to identify missing registrations.

hebasto:
Updated 9657a8661be3e79a579b38893d58fc0a06f56bd9 -> 7c944086b3c62f9833fa9fbdbda2d5d4b34b37e9 ([pr61.01](https://github.com/hebasto/gui/commits/pr61.01) -> [pr61.02](https://github.com/hebasto/gui/commits/pr61.02), [diff](https://github.com/hebasto/gui/compare/pr61.01..pr61.02)):
- addressed @promag's [comment](https://github.com/bitcoin-core/gui/pull/61#discussion_r500080463):
> nit, could mention env `QT_FATAL_WARNINGS=1` as a way to identify missing registrations.

src/qt/bitcoin.cpp
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui/pull/61#issuecomment-704127075).

hebasto:
There was unanswered [comment](https://github.com/bitcoin/bitcoin/pull/19108#issuecomment-639627488) from @luke-jr:
> Is not needing it for typedefs considered defined behaviour?

I think this excerpt from [Qt docs](https://doc.qt.io/qt-5/qmetatype.html#qRegisterMetaType) is an answer:
> This [`qRegisterMetaType`] function is useful only for registering an alias (typedef)...

luke-jr:
If there's no behaviour change, I'd figure just leave it alone so things don't subtly break if the type of `CAmount` were to ever be changed. Kinda the point of the typedef anyway is that you don't need to care so much what the underlying type is...

hebasto:
> If there's no behaviour change, I'd figure just leave it alone so things don't subtly break if the type of `CAmount` were to ever be changed. Kinda the point of the typedef anyway is that you don't need to care so much what the underlying type is...

Closing.

promag:
> if the type of `CAmount` were to ever be changed

Ok 😅

Issue #58, opened 2020-08-14
Tor status indicator icon in bottom right


fanquake:
Moved from: https://github.com/bitcoin/bitcoin/issues/7734.

The current QT GUI shows some network status indicators in the bottom right.

I'd like someone to add one for Tor. It might be the Tor onion logo that is either green, yellow (if going through some not-quite-ready stage) or with a circle and line striking through it when not connected via Tor.

This would complement the Tor-related improvements in 0.12.0 and make it easier for users to determine whether they are connected via Tor without doing additional investigatory steps, such as using the  `-debug=tor` switch.

jonatack:
Concept ACK

Bosch-0:
Here is a suggestion for the three icon states.

At some stage it would be nice to have some colored icons, but whilst the others are still in a black and white I think its important to stick with this style for now to keep consistency. I followed the same style that the connect icons use also shown below.

Not connected -> connecting / not-quite-ready stage -> connected.

![image](https://user-images.githubusercontent.com/55287964/90750022-cef9ae80-e306-11ea-863e-1e062cea5d0f.png)

jonatack:
Nice!

michaelfolkson:
I like the minimalism, I think that is important for Core. The "Not connected icon" and "connected" icons are clear. Maybe the "connecting" icon could be fainter. Similar to the "Not connected icon" but without the cross.

MarcoFalke:
Why not pick the icons that look good even when scaled down (non high-DPI displays)?

Tor made an effort and recently changed those: https://trac.torproject.org/projects/tor/ticket/25702

![Screenshot from 2020-08-21 07-16-37](https://user-images.githubusercontent.com/6399679/90855005-c2b13800-e37e-11ea-8002-266adcb71952.png)
![Screenshot from 2020-08-21 07-15-19](https://user-images.githubusercontent.com/6399679/90855013-c5139200-e37e-11ea-8978-772e7533df81.png)

Bosch-0:
> Why not pick the icons that look good even when scaled down (non high-DPI displays)?

SVG will look fine when scaled down or up

MarcoFalke:
When rendering an SVG, there will always be a limit until it can no longer be scaled down and still be distinguished from a blurry blob. Generally, the less detail an SVG has, the less pixels are needed to still represent the details.

Bosch-0:
Here is another, much more minimal, icon iteration. 

[...]

PR #67, opened 2020-08-20
Added Tor icons to repo

Bosch-0:
Added the below icons to the repo which can be used for Tor network connectivity indicators #58. 

From previous discussion on issue.

> At some stage it would be nice to have some colored icons, but whilst the others are still in a black and white I think its important to stick with this style for now to keep consistency. I followed the same style that the connect icons use...

Link to the figma file where I did the icon designs > https://www.figma.com/file/FJ02rY3m8V9ZCDvoXjW39W/Bitcoin-Core?node-id=486%3A312

**In order:** Not connected -> connecting -> connected. 

image

Here is some examples of how they could be shown

image

image

image


hebasto:
Concept ACK

MarcoFalke:
What is the point of adding the icons if they are not shown in the gui?

hebasto:
> What is the point of adding the icons if they are not shown in the gui?

I'll take care about it :)

MarcoFalke:
Please add source/license to new files (contrib/debian/copyright) and optimize the png versions: https://github.com/bitcoin-core/bitcoin-maintainer-tools/blob/master/optimize-pngs.py

jonatack:
Concept ACK

(make sure it renders to look more like an onion than, say, a bomb ;)

Bosch-0:
> What is the point of adding the icons if they are not shown in the gui?

They will be soon as hebasto said. We have been discussing the implementation over at the bitcoin design slack. 

Though yes I guess I could have just passed the icons onto hebasto and he do a PR implementing them into the GUI but I figured some feedback on the design was important.

Bosch-0:
> Concept ACK
> 
> 
> 
> (make sure it renders to look more like an onion than, say, a bomb ;)

There is both png and svg versions to play around with so the implementation can be a little flexible (using the svg's over pngs).

jonasschnelli:
Concept ACK though I would prefer a simpler, less fine-grained icon.

fanquake:
> What is the point of adding the icons if they are not shown in the gui?

I agree. These should be merged when they are being used; I don't think there's any need for us to start pulling in unused assets/imagery. If people are looking for feedback on designs, opening new issues in this repo or commenting in the existing threads is the way to go (I can see that was done in https://github.com/bitcoin-core/gui/issues/58).

PR #71, opened 2020-08-23
Fix visual quality of text in QR image

### LINKED ISSUE #54: QRCode address in image is not readable
Moved from: https://github.com/bitcoin/bitcoin/issues/19103.

Somehow the resolution of the QRCode image is not high enough.
Native wpkh addresses are non readable and legacy/p2sh addresses are not good readable.
Haven't tested on other OSes than macOS.

Bildschirmfoto 2020-05-29 um 09 52 07

Bildschirmfoto 2020-05-29 um 09 53 21

hebasto:
Master (197450f80868fe752c6107955e5da80704212b34):
DeepinScreenshot_select-area_20200824001800

This PR (6954156b4091bc1e561502f0eef0cece56c76eec):
- macOS 10.15.6
Screenshot from 2020-09-07 15-40-30

- Linux Mint 20
Screenshot from 2020-09-07 15-48-13

Fix #54
Fix https://github.com/bitcoin/bitcoin/issues/19103

---

The first commit is easy to review with [`git diff --word-diff`](https://github.com/bitcoin-core/gui/pull/71/commits/8071c75d45e12c2bca04b170c687bebd30ad19ac?w=1).

### DIFF
--- a/src/qt/qrimagewidget.cpp
+++ b/src/qt/qrimagewidget.cpp
@@ -66,24 +66,25 @@ bool QRImageWidget::setQR(const QString& data, const QString& text)
 
     QImage qrAddrImage = QImage(QR_IMAGE_SIZE, QR_IMAGE_SIZE + (text.isEmpty() ? 0 : 20), QImage::Format_RGB32);
     qrAddrImage.fill(0xffffff);
-    QPainter painter(&qrAddrImage);
-    painter.drawImage(0, 0, qrImage.scaled(QR_IMAGE_SIZE, QR_IMAGE_SIZE));
+    {
+        QPainter painter(&qrAddrImage);
+        painter.drawImage(0, 0, qrImage.scaled(QR_IMAGE_SIZE, QR_IMAGE_SIZE));
 
-    if (!text.isEmpty()) {
-        QFont font = GUIUtil::fixedPitchFont();
-        font.setStyleStrategy(QFont::NoAntialias);
-        QRect paddedRect = qrAddrImage.rect();
+        if (!text.isEmpty()) {
+            QFont font = GUIUtil::fixedPitchFont();
+            font.setStyleStrategy(QFont::NoAntialias);
+            QRect paddedRect = qrAddrImage.rect();
 
-        // calculate ideal font size
-        qreal font_size = GUIUtil::calculateIdealFontSize(paddedRect.width() - 20, text, font);
-        font.setPointSizeF(font_size);
+            // calculate ideal font size
+            qreal font_size = GUIUtil::calculateIdealFontSize(paddedRect.width() - 20, text, font);
+            font.setPointSizeF(font_size);
 
-        painter.setFont(font);
-        paddedRect.setHeight(QR_IMAGE_SIZE+12);
-        painter.drawText(paddedRect, Qt::AlignBottom|Qt::AlignCenter, text);
+            painter.setFont(font);
+            paddedRect.setHeight(QR_IMAGE_SIZE+12);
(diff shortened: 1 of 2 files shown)


MarcoFalke:
Concept ACK

Sjors:
f4016dd looks way better on macOS. Can you explain what fixed it?

hebasto:
@Sjors 
> [f4016dd](https://github.com/bitcoin-core/gui/commit/f4016dd0190cd60bfc86f095d41ad41afbc6c891) looks way better on macOS. Can you explain what fixed it?

- `QFontDatabase::GeneralFont` instead of `QFontDatabase::FixedFont`
- dropped `QFont::NoAntialias`
- added `QFont::AbsoluteSpacing` 1 px

michaelfolkson:
Looking for Concept (N)ACKs (plus comments if you have them) from designers cc @GBKS @Bosch-0 @johnsBeharry

michaelfolkson:
Concept ACK from me on fixing visual quality. I'll leave the font discussion to the designers.

Bosch-0:
Concept ACK

If https://github.com/bitcoin-core/gui/pull/79 gets merged and we use an embedded mono font this should be used here.

Bosch-0:
Both

luke-jr:
863-412-9652

luke-jr:
Would prefer to keep the font monospaced

GBKS:
Concept ACK on fixing the visual quality.

A monospace font would be nice, and also a narrow space every 4 characters (like in credit card numbers, some exploration [here](https://www.figma.com/file/iUhCkFNFwn29j5w4YFDsQq/Bitcoin-address-and-amount-display?node-id=0%3A1)) for easier comparison, but all of that seems like a different issue.

hebasto:
Updated f4016dd0190cd60bfc86f095d41ad41afbc6c891 -> 6954156b4091bc1e561502f0eef0cece56c76eec ([pr71.01](https://github.com/hebasto/gui/commits/pr71.01) -> [pr71.02](https://github.com/hebasto/gui/commits/pr71.02), [diff](https://github.com/hebasto/gui/compare/pr71.01..pr71.02)):
- reverted back to a system monospaced font as requested by @Bosch-0, @luke-jr and @GBKS

luke-jr:
Before:
before
After:
after

I don't see a noteworthy improvement...

hebasto:
@luke-jr 
> I don't see a noteworthy improvement...

Mind sharing your system details?

On your screenshots the address string does not even fit in the allocated space...
Going to reproduce such behavior.

luke-jr:
Qt GUI 5.14.2

Preferred font https://luke.dashjr.org/education/tonal/glyphs/fonts/Console/

(Note that it is a bitmap font and only supports one size)

hebasto:
> (Note that it is a bitmap font and only supports one size)

This explains the string size issue.

RandyMcMillan:
It may be best to not add the address to the qrcode image.
This would free up space to display a larger qrcode image in the imageView.
Maybe the save image dialogue should auto populate the name field with the address.
This would allow the user to check the address as they are saving the image.
Maybe the image name field should not be editable by default - and an edit check box option added 
to ensure that the user only changes the name of the image intentionally.

jonasschnelli:
Tested ACK 6954156b4091bc1e561502f0eef0cece56c76eec - tested on macOS 10.15. Fixes the problem.

The font still remains non-HiDPI.
This is the image saved through the dialogs save function: 
test

The image seems to use a HiDPI resolution but the font isn't rendered in HiDPI.

PR #75, opened 2020-08-26
Pass argsmanager reference to intro instead of relying on global

MarcoFalke:
Motivation: (copied from https://github.com/bitcoin/bitcoin/pull/19779)

The gArgs global has several issues:

* gArgs is used by each process (bitcoind, bitcoin-qt, bitcoin-wallet, bitcoin-cli, bitcoin-tx, ...), but it is hard to determine which arguments are actually used by each process. For example arguments that have never been registered, but are still used, will always return the fallback value.
* Tests may run several sub-tests, which need different settings. So globals will have to be overwritten, but that is fragile on its own: e.g. https://github.com/bitcoin/bitcoin/pull/19704#issuecomment-678259092 or #19511

The goal is to remove gArgs, but as a first step in that direction this pull will change gArgs to use a passed-in reference instead.

### DIFF
--- a/src/qt/intro.h
+++ b/src/qt/intro.h
@@ -11,6 +11,7 @@
 
 static const bool DEFAULT_CHOOSE_DATADIR = false;
 
+class ArgsManager;
 class FreespaceChecker;
 
 namespace interfaces {
@@ -30,8 +31,7 @@ class Intro : public QDialog
     Q_OBJECT
 
 public:
-    explicit Intro(QWidget *parent = nullptr,
-                   int64_t blockchain_size_gb = 0, int64_t chain_state_size_gb = 0);
+    explicit Intro(const ArgsManager& args, QWidget* parent = nullptr, int64_t blockchain_size_gb = 0, int64_t chain_state_size_gb = 0);
     ~Intro();
 
     QString getDataDirectory();
@@ -47,7 +47,7 @@ public:
      * @note do NOT call global GetDataDir() before calling this function, this
      * will cause the wrong path to be cached.
      */
-    static bool showIfNeeded(bool& did_show_intro, bool& prune);
+    static bool showIfNeeded(ArgsManager& args, bool& did_show_intro, bool& prune);
 
 Q_SIGNALS:
     void requestCheck();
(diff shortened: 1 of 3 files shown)


MarcoFalke:
requested by @ryanofsky in https://github.com/bitcoin-core/gui/pull/35#discussion_r477265883

hebasto:
Concept ACK.

src/qt/intro.h
```diff
@@ -47,7 +47,7 @@ class Intro : public QDialog
      * @note do NOT call global GetDataDir() before calling this function, this
      * will cause the wrong path to be cached.
      */
-    static bool showIfNeeded(bool& did_show_intro, bool& prune);
+    static bool showIfNeeded(ArgsManager& args, bool& did_show_intro, bool& prune);
```
hebasto, line 50: As `arg` is an in-_out_ parameter, maybe place it at the end of the parameter list?

hebasto (approved):
ACK fa1b478ddb7eccea2ae33246212a36ec503cdbca, tested on Linux Mint 20 (x86_64) with the `-choosedatadir` command-line option.

src/qt/intro.h
MarcoFalke: every param in this list is an in-out param

hebasto: Indeed.

MarcoFalke:
@ryanofsky Would be good to have your eyes over this as a sanity check :eyes:

ryanofsky:
I don't think this change is beneficial in it's current form. Before this change there are 46 references to the `gArgs` global in `src/qt/`, and after this change there are 40 references to `gArgs`. I think if you really want to get rid of the `gArgs` global in the GUI you could do it in a single PR instead of having `gArgs` and `args` variables used different places unpredictably.

Also I don't think there is a necessarily a reason to eliminate the `gArgs` in the GUI. It's understandable to want to get rid of global variables in node and wallet code to allow more code reuse and help testing, but the GUI already relies on a global QApplication instance, a global event loop, global settings state and probably other global state, so getting rid of `gArgs` there doesn't seem to accomplish much. I think removing `gArgs` from node and wallet code and then making `gArgs` a GUI-only variable would satisfy goals stated in the PR description while keeping GUI code less verbose, and requiring fewer changes.

MarcoFalke:
Ok, thanks for looking.

PR #79, opened 2020-08-28
gui: Add IBM Plex Mono font

hebasto:
Qt does not guarantee that the actual applied font matches to the requested one.
It was noted (https://github.com/bitcoin/bitcoin/pull/16432#issuecomment-514486077):
> the monospace font looks a bit weird no macOS

... because it is _not_ monospaced.
Also some discrepancies I've noted on Windows while testing Qt 5.15 ([#19716](https://github.com/bitcoin/bitcoin/pull/19716)).

Of course, we could check the actual font with `QFontInfo`, and try to choose another font.

But this PR suggests to just embed a monospaced font, and get the GUI look (partially) independent from a platform.

I've [asked](https://bitcoindesign.slack.com/archives/C018PM159JS/p1598567309121600) Bitcoin Design community for a font to embed, and was pointed to [IBM Plex Mono](https://fonts.google.com/specimen/IBM+Plex+Mono?preview.text=%E2%82%BF21,000,000&preview.text_type=custom&category=Monospace).

Also a minor GUI change: `#` is replaced with `*` as a mask symbol in the privacy mode to improve look with IBM Plex Mono font.


hebasto:
Many thanks to @GBKS !

Sjors:
Before and after screenshots on various OS's would be nice.

I'm generally not a fan of messing with the default font choices of operating systems / desktop environments. However, if upstream QT doesn't use a proper monospaced font for macOS, then it makes sense for at least that platform.

hebasto:
Just for record: another PR that embeds a font is https://github.com/bitcoin/bitcoin/pull/16883.

Bosch-0:
Another suggestion https://fonts.google.com/specimen/Roboto+Mono

image

RandyMcMillan:
There have been concerns about OFL Licensing in the past...

https://github.com/bitcoin/bitcoin/pull/17311#issuecomment-550453875

https://github.com/bitcoin/bitcoin/pull/17311#issuecomment-550591089

But if OFL isn't an issue - then fonts-liberation is a good candidate as well.

https://en.wikipedia.org/wiki/Liberation_fonts

michaelfolkson:
Looking for Concept (N)ACKs (plus comments if you have them) from designers cc @GBKS @Bosch-0 @johnsBeharry

Bosch-0:
Concept (N)ACK

Using [Roboto Mono](https://fonts.google.com/specimen/Roboto+Mono) would be a better alternative:

- [IBM Plex Mono](https://fonts.google.com/specimen/IBM+Plex+Mono#license) uses a OFL license which as pointed out in https://github.com/bitcoin/bitcoin/pull/17311#issuecomment-550591089 has some concerns about its 'free' status.  Roboto mono uses [Apache Licence, Version 2.0](http://www.apache.org/licenses/LICENSE-2.0) which as far as I'm concerned isn't an issue. 

- Roboto Mono is a [variable font](https://web.dev/variable-fonts/) which allows it to have its properties changed (weight, width etc.) without needing a separate font file. This will keep the overall font file size down. 

- This may be a subjective take but I think Roboto Mono is also slightly more readable. 

IBM Plex Mono
image

Roboto Mono
image

image

hebasto:
About font sizes:
```
$ ls -go
total 380
-rw-rw-r-- 1 111532 Mar 11  2018 IBMPlexMono-Medium.ttf
-rw-rw-r-- 1  86820 May 13  2015 RobotoMono-Medium.ttf
-rw-rw-r-- 1 182172 May 13  2015 RobotoMono-VariableFont_wght.ttf
```

Bosch-0:
Hebasto asked in [slack](https://bitcoindesign.slack.com/archives/C018PM159JS/p1599122925039100) whether choosing # or * would be more applicable for hidden/masked amounts. I'd go for the star * over the hash #, it looks cleaner (see below). Each item being concealed should also used a fixed number of strings as @achow101 [pointed out here](https://github.com/bitcoin/bitcoin/pull/16432#issuecomment-513789511) - I used 9 here so it fills the same amount of space as the unconcealed items. Having a fixed number of strings also gives consistency resulting in a cleaner UI.

image

I also did a design for the suggestion that @promag made regarding [having a toggle](https://github.com/bitcoin/bitcoin/pull/16432#issuecomment-513814481) icon on the overview page that users can click on to easily turn privacy mode on or off. Does privacy mode conceal any information in the send, receive or transactions sections? I think this mode could be extended to those areas also if not. If so this toggle could be moved to the top menu, possibly to the left of the wallet selection drop down.

hebasto:
It appears that Qt (at least 5.12.8) cannot work with variable fonts.
At least [`QFont::setWeight()`](https://doc.qt.io/qt-5/qfont.html#setWeight) is noop for `RobotoMono-VariableFont_wght.ttf`.

Bosch-0:
Ahh that's a shame, not a huge issue though, as the mono font is only going to be used in a few places the benefits of using a variable font is negligible. The static medium Roboto Mono font style should be all we need to use. 

I think the licensing issue with OFL still makes Roboto Mono a more suitable option than IBM Plex Mono.

hebasto:
Updated 15f37db53124f31d5499ce0604f907b9564aa506 -> 3fdd5b6bd17a679d6e3876682266092159c52d59 ([pr79.01](https://github.com/hebasto/gui/commits/pr79.01) -> [pr79.02](https://github.com/hebasto/gui/commits/pr79.02), [diff](https://github.com/hebasto/gui/compare/pr79.01..pr79.02)):
- used Roboto Mono instead of IBM Plex Mono as requested by @Bosch-0 
- changes are scoped only the Overview page as it is a clear visual bug fix

Going to update OP soon.

hebasto:
@Sjors 
> Before and after screenshots on various OS's would be nice.

Done.

> I'm generally not a fan of messing with the default font choices of operating systems / desktop environments. However, if upstream QT doesn't use a proper monospaced font for macOS, then it makes sense for at least that platform.

This is under discussion now in Bitcoin Design Community (https://github.com/BitcoinDesign, https://bitcoindesign.slack.com). Please join the discussion :)

luke-jr:
Concept NACK to overriding the platform. If users don't like their platform's font choices, they should change it there.

Also concept NACK to moving discussion to a private proprietary platform.

moneyball:
I agree with @luke-jr that the bitcoin design community should not require Core contributors to join Slack in order to debate changes to the Core GUI (although developers are more than welcome to join the conversation there!). I think it is fine if there is discussion on Slack or video calls or any medium, but for any kind of formal proposal or discussion or debate, it should occur on GitHub as that is the existing process for the project.

flack:
Just my two cents for the screenshot in the OP: Using monospace for the balance numbers seems like a weird workaround (weird because you have two different fonts for the same line of information, the label in the normal UI font and then the number & unit in the monospace one). It would be much nicer to use a non-monospace font throughout that supports tabular figures (see https://www.fonts.com/content/learning/fontology/level-3/numbers/proportional-vs-tabular-figures). Or doesn't that work in Qt?

hebasto:
@flack 
> Just my two cents for the screenshot in the OP: Using monospace for the balance numbers seems like a weird workaround (weird because you have two different fonts for the same line of information, the label in the normal UI font and then the number & unit in the monospace one). It would be much nicer to use a non-monospace font throughout that supports tabular figures (see https://www.fonts.com/content/learning/fontology/level-3/numbers/proportional-vs-tabular-figures). Or doesn't that work in Qt?

Thanks! TIL. I didn't know about "tabular figures".

RandyMcMillan:
@hebasto - 

The Roboto Family is an excellent choice as an embedded font (imo).
I believe we should be proactive and embed the entire family.
It would eliminate this unknown factor in Gui design moving forward.

Keep in mind -  fonts do pose a security issue.
https://www.zdnet.com/article/google-open-sources-internal-tool-for-finding-font-related-security-bugs/
https://security.stackexchange.com/questions/91347/how-can-a-font-be-used-for-privilege-escalation

In essence - embedding fonts explicitly enhances determinism.

hebasto:
@luke-jr 
> Concept NACK to overriding the platform. If users don't like their platform's font choices, they should change it there.

Being a _financial_ application the Bitcoin Core requires a monospaced/fixed-pitch font (or, at least, tabular figures font).

The Qt's font matching algorithm does not guarantee that:
- the monospaced font will be chosen at all (see https://github.com/bitcoin/bitcoin/pull/16432#issuecomment-514486077)
- the chosen monospaced font has expected style and weight (see https://github.com/bitcoin-core/gui/pull/87#issuecomment-768920736)
- an upgrade of the system or Qt won't change the Bitcoin Core look unexpectedly

A agree that user's ability to customize fonts is good. The question is _how users could customize the monospaced font for Qt application now?_

I'm not aware of such possibilities on Windows 10 and macOS 11.1. On Linux I used the `qt5ct` package, but it fails to customize the _monospaced_ font.

If there are no ways to customize the monospaced font for Qt application, are you going to insist on your point?

laanwj:
> The Roboto Family is an excellent choice as an embedded font (imo). I believe we should be proactive and embed the entire family.

Despite really disliking it on a conceptual level, I think 'ship your own fonts' is the modern way. The less assumptions we have to make about the operating system, platform, and what is already installed, the better.

michaelfolkson:
> Concept NACK to overriding the platform. If users don't like their platform's font choices, they should change it there.

If this purely a design choice (there are no negative side effects to this change) I'm happy to leave to the designers. If the designers are a Concept ACK I'm a Concept ACK.

laanwj:
Code review and license review ACK 3fdd5b6bd17a679d6e3876682266092159c52d59
Please do update `contrib/debian/copyright` to explitily list the apache license for the embedded font files.

jonasschnelli:
Concept ACK 3fdd5b6bd17a679d6e3876682266092159c52d59 - please add the license attribution

luke-jr:
> The Qt's font matching algorithm does not guarantee that:

> the monospaced font will be chosen at all (see bitcoin/bitcoin#16432 (comment))
> the chosen monospaced font has expected style and weight (see #87 (comment))

We already have functional workarounds for these.

> an upgrade of the system or Qt won't change the Bitcoin Core look unexpectedly

That's desirable and expected behaviour.

>I'm not aware of such possibilities on Windows 10 and macOS 11.1.

Crazy. It looks like Windows 10 removed the ability for users to configure their fonts. But that's a Windows problem, not necessarily ours...

I'm not surprised Apple denies users choice.

>On Linux I used the qt5ct package, but it fails to customize the monospaced font.

It's right there in System Settings for me, as a "Fixed width" font choice.

hebasto:
Updated 3fdd5b6bd17a679d6e3876682266092159c52d59 -> 67f26319a0ca7e34e3db17d9330133622a40de09 ([pr79.02](https://github.com/hebasto/gui/commits/pr79.02) -> [pr79.04](https://github.com/hebasto/gui/commits/pr79.04)):
- rebased due to tyhe conflict with #115
- added the license attribution as requested by @laanwj and @jonasschnelli 
- added an option to opt-out embedded monospaced font in Menu -> Options-> Display as requested by @luke-jr (as I understand his idea)

luke-jr:
> added an option to opt-out embedded monospaced font in Menu -> Options-> Display as requested by @luke-jr (as I understand his idea)

That seems like a reasonable compromise.

Can you also not add the font to the repo, make it optional at build time (perhaps configurable?), and bundle it via depends?

hebasto:
> Can you also not add the font to the repo, make it optional at build time (perhaps configurable?), and bundle it via depends?

What reasons do you see for that?

luke-jr:
So that people who don't want this can build without it...

laanwj:
Tested ACK 67f26319a0ca7e34e3db17d9330133622a40de09

A configure option to build without internal fonts can be added later. IMO adding the font to the repository itself is fine. It's only 87kB, not worth the overhead of an external depends download.

rebroad:
image
I'm a little confused, what version is the "before" picture from? I'm running 0.21 and it's already a fixed font.

hebasto:
@rebroad 
> I'm a little confused, what version is the "before" picture from? I'm running 0.21 and it's already a fixed font.

I suppose your system is not macOS, right?

PR #84, opened 2020-09-07
Changed confusing "BLOCKLIST" to "DENYLIST".

Hey:
A while ago "BLACKLIST" was changed to "BLOCKLIST".
Considering this is a blockchain, the word "block" may be somewhat confusing.

I understand that the world "blacklist" may be considered racism, so I believe the world "DENYLIST" makes more sense.


fanquake:
This has already been changed in the main repo. It just hasn't been pushed here yet.

PR #85, opened 2020-09-07
Remove unused "What's This" button in dialogs on Windows OS

### LINKED ISSUE #74: Help '?' button on create wallet and encrypt wallet modals does nothing
The help button '?' on the create wallet modal (and the following encrypt wallet modal) does not do anything on WindowsOS. This does not exist on the Linux version, have not tested Mac.

image

hebasto:
Fix #74.

From [Qt docs](https://doc.qt.io/qt-5/qdialog.html#QDialog):
> The widget flags _f_ are passed on to the `QWidget` constructor. If, for example, you don't want a **What's This** button in the title bar of the dialog, pass `Qt::WindowTitleHint | Qt::WindowSystemMenuHint` in _f_.

Screenshot on Windows 10 (2004):
- master (3ba25e3bdde3464eed5d2743d68546e48b005544)
Screenshot from 2020-09-07 16-55-42

- this PR (e322fe7e19ac504272d14b9b4f9b28b13df888ed)
Screenshot from 2020-09-07 18-31-16

### DIFF
--- a/src/qt/receivecoinsdialog.cpp
+++ b/src/qt/receivecoinsdialog.cpp
@@ -8,6 +8,7 @@
 #include <qt/forms/ui_receivecoinsdialog.h>
 
 #include <qt/addresstablemodel.h>
+#include <qt/guiutil.h>
 #include <qt/optionsmodel.h>
 #include <qt/platformstyle.h>
 #include <qt/receiverequestdialog.h>
--- a/src/qt/receivecoinsdialog.cpp
+++ b/src/qt/receivecoinsdialog.cpp
@@ -22,7 +22,7 @@
 #include <QTextDocument>
 
 ReceiveCoinsDialog::ReceiveCoinsDialog(const PlatformStyle *_platformStyle, QWidget *parent) :
-    QDialog(parent),
+    QDialog(parent, GUIUtil::dialog_flags),
     ui(new Ui::ReceiveCoinsDialog),
     columnResizingFixer(nullptr),
     model(nullptr),
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -53,7 +53,7 @@ int getIndexForConfTarget(int target) {
 }
 
 SendCoinsDialog::SendCoinsDialog(const PlatformStyle *_platformStyle, QWidget *parent) :
-    QDialog(parent),
+    QDialog(parent, GUIUtil::dialog_flags),
     ui(new Ui::SendCoinsDialog),
     clientModel(nullptr),
     model(nullptr),
--- a/src/qt/addressbookpage.cpp
+++ b/src/qt/addressbookpage.cpp
@@ -56,7 +56,7 @@ protected:
 };
 
 AddressBookPage::AddressBookPage(const PlatformStyle *platformStyle, Mode _mode, Tabs _tab, QWidget *parent) :
-    QDialog(parent),
+    QDialog(parent, GUIUtil::dialog_flags),
     ui(new Ui::AddressBookPage),
     model(nullptr),
     mode(_mode),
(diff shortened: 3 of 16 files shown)


MarcoFalke:
Please split this into two commits, otherwise literally every commit could be a scripted diff by including the diff as a patch in the commit body

hebasto:
Updated e322fe7e19ac504272d14b9b4f9b28b13df888ed -> ac7ccd67d7f2b09e36dd57405f899e4698dd3d78 ([pr85.01](https://github.com/hebasto/gui/commits/pr85.01) -> [pr85.02](https://github.com/hebasto/gui/commits/pr85.02), [diff](https://github.com/hebasto/gui/compare/pr85.01..pr85.02)):
- addressed @MarcoFalke's [comment](https://github.com/bitcoin-core/gui/pull/85#issuecomment-688407464):
> Please split this into two commits, otherwise literally every commit could be a scripted diff by including the diff as a patch in the commit body

Bosch-0:
Concept ACK

Bosch-0:
Testing ACK

Tested on Windows 10.0.18363 Build 18363. 

Master ([`56a461f`](https://github.com/bitcoin-core/gui/commit/9af7c1993b3512a7230f723b1923abd496989c59)) vs this PR ([`ac7ccd6`](https://github.com/bitcoin-core/gui/commit/ac7ccd67d7f2b09e36dd57405f899e4698dd3d78))

Group 109

This PR also removes this unused dialog button from the following windows on Windows 10 compared to master. 

- About Bitcoin Core
- Command-line options
- Request payment modal when generating a new receive address
- Create Wallet
- Encrypt wallet when creating a new wallet
- Signature signing / verifying 
- Open Bitcoin URI 
- Receive addresses 
- Sending addresses

luke-jr:
Is there a way we can exclude the "What's This?" button rather than explicitly list included buttons?

hebasto:
> Is there a way we can exclude the "What's This?" button rather than explicitly list included buttons?

Qt docs do not provide that way.

src/qt/sendcoinsdialog.cpp
```diff
@@ -53,7 +53,7 @@ int getIndexForConfTarget(int target) {
 }
 
 SendCoinsDialog::SendCoinsDialog(const PlatformStyle *_platformStyle, QWidget *parent) :
-    QDialog(parent),
+    QDialog(parent, GUIUtil::dialog_flags),
```
promag, line 56: It extends dialog but it's not used as such. I think you could drop this change?

src/qt/receivecoinsdialog.cpp
```diff
@@ -21,7 +22,7 @@
 #include <QTextDocument>
 
 ReceiveCoinsDialog::ReceiveCoinsDialog(const PlatformStyle *_platformStyle, QWidget *parent) :
-    QDialog(parent),
+    QDialog(parent, GUIUtil::dialog_flags),
```
promag, line 25: Same as `SendCoinsDialog`.

promag (commented):
Code review ACK ac7ccd67d7f2b09e36dd57405f899e4698dd3d78 but with some suggestions.

Does it make sense to have a `QDialog` subclass? There's also the call `handleCloseWindowShortcut`.

src/qt/sendcoinsdialog.cpp
hebasto: > I think you could drop this change?

Adding widget flags to remove "What's This" button is the goal of this PR. Or did I understand you in a wrong way?

jonasschnelli (approved):
utACK ac7ccd67d7f2b09e36dd57405f899e4698dd3d78

hebasto:
@luke-jr 
> Is there a way we can exclude the "What's This?" button rather than explicitly list included buttons?

I was [wrong](https://github.com/bitcoin-core/gui/pull/85#issuecomment-716048754).

The `Qt::AA_DisableWindowContextHelpButton` application attribute value was added in Qt 5.10.

Unfortunately, we have minimum Qt version 5.5.1.

PR #87, opened 2020-09-10
Choose monospaced font in C++ code rather in `*.ui` file

hebasto:
This change makes macOS choose the correct monospaced font.

**fanquake**  noted (https://github.com/bitcoin/bitcoin/pull/16432#issuecomment-514486077):
> the monospace font looks a bit weird on macOS

... because it is _not_ monospaced.

This is an alternative to #79 as that PR has a Concept NACK from [**luke-jr**](https://github.com/bitcoin-core/gui/pull/79#issuecomment-686510806).

I'm personally still lean to embedding font as it makes the GUI more stable.

### DIFF
--- a/src/qt/overviewpage.cpp
+++ b/src/qt/overviewpage.cpp
@@ -141,6 +141,17 @@ OverviewPage::OverviewPage(const PlatformStyle *platformStyle, QWidget *parent) 
     showOutOfSyncWarning(true);
     connect(ui->labelWalletStatus, &QPushButton::clicked, this, &OverviewPage::handleOutOfSyncWarningClicks);
     connect(ui->labelTransactionsStatus, &QPushButton::clicked, this, &OverviewPage::handleOutOfSyncWarningClicks);
+
+    QFont f = GUIUtil::fixedPitchFont();
+    f.setWeight(QFont::Bold);
+    ui->labelBalance->setFont(f);
+    ui->labelUnconfirmed->setFont(f);
+    ui->labelImmature->setFont(f);
+    ui->labelTotal->setFont(f);
+    ui->labelWatchAvailable->setFont(f);
+    ui->labelWatchPending->setFont(f);
+    ui->labelWatchImmature->setFont(f);
+    ui->labelWatchTotal->setFont(f);
 }
 
 void OverviewPage::handleTransactionClicked(const QModelIndex &index)
--- a/src/qt/forms/overviewpage.ui
+++ b/src/qt/forms/overviewpage.ui
@@ -116,13 +116,6 @@
             </property>
             <item row="2" column="2">
              <widget class="QLabel" name="labelWatchPending">
-              <property name="font">
-               <font>
-                <family>Monospace</family>
-                <weight>75</weight>
-                <bold>true</bold>
-               </font>
-              </property>
               <property name="cursor">
                <cursorShape>IBeamCursor</cursorShape>
               </property>
@@ -142,13 +135,6 @@
             </item>
             <item row="2" column="1">
              <widget class="QLabel" name="labelUnconfirmed">
-              <property name="font">
-               <font>
-                <family>Monospace</family>
-                <weight>75</weight>
-                <bold>true</bold>
-               </font>
-              </property>
               <property name="cursor">
                <cursorShape>IBeamCursor</cursorShape>
               </property>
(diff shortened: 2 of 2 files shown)


hebasto:
cc @fanquake @luke-jr

hebasto:
@jonasschnelli Should this be considered a bugfix?

jonasschnelli:
Unsure about this.
IMO this is a MVC violation (which we have plenty but unsure to add more).

hebasto:
> IMO this is a MVC violation (which we have plenty but unsure to add more).

How is it so? I think both `qt/overviewpage.cpp` and `qt/forms/overviewpage.ui` are parts of "View" in MVC, no?

jonasschnelli:
> How is it so? I think both qt/overviewpage.cpp and qt/forms/overviewpage.ui belong to "View" in MVC, no?

I'd say *.ui is view, *.cpp is controller (the non-automatic generated ones).

hebasto:
> > How is it so? I think both qt/overviewpage.cpp and qt/forms/overviewpage.ui belong to "View" in MVC, no?
> 
> I'd say *.ui is view, *.cpp is controller (the non-automatic generated ones).

Right. But Qt programming usually assumes "the view and the controller objects are combined" (https://doc.qt.io/qt-5/model-view-programming.html).

src/qt/overviewpage.cpp
```diff
@@ -141,6 +141,18 @@ OverviewPage::OverviewPage(const PlatformStyle *platformStyle, QWidget *parent)
     showOutOfSyncWarning(true);
     connect(ui->labelWalletStatus, &QPushButton::clicked, this, &OverviewPage::handleOutOfSyncWarningClicks);
     connect(ui->labelTransactionsStatus, &QPushButton::clicked, this, &OverviewPage::handleOutOfSyncWarningClicks);
+
+    QFont f = GUIUtil::fixedPitchFont();
+    f.setWeight(QFont::Bold);
+    f.setPointSize(10);
```
luke-jr, line 147: Let's not override this for no reason?

luke-jr (changes requested):
Looks good to me, one change requested

src/qt/overviewpage.cpp
hebasto: The only reason is to make things explicitly rather than implicitly. Does `setPointSize` call has any downsides?

luke-jr: Yes, the downside is that it is explicit and overrides the user's preference (which may not be 10pt).

hebasto:
Updated a44ba74ceccf6f61c65972e2ae37327de0db4c79 -> 2e386cd3dd3fea5a22dc8ae871b80a80f8cdb3f7 ([pr87.01](https://github.com/hebasto/gui/commits/pr87.01) -> [pr87.02](https://github.com/hebasto/gui/commits/pr87.02), [diff](https://github.com/hebasto/gui/compare/pr87.01..pr87.02)):

- addressed @luke-jr's [comment](https://github.com/bitcoin-core/gui/pull/87#discussion_r511600141):
> Yes, the downside is that it is explicit and overrides the user's preference (which may not be 10pt).

src/qt/overviewpage.cpp
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui/pull/87#issuecomment-716163108).

Now I'm able to change the font size with the `qt5ct` utility on Linux Mint 20 :)

luke-jr (approved):
utACK

jarolrod (approved):
Tested ACK on macOS 10.15.7

Master
master

PR
pr

jonasschnelli:
Master:
Bildschirmfoto 2021-01-28 um 10 26 02

This PR:
Bildschirmfoto 2021-01-28 um 10 23 57


I think the correct monospace font should be bigger and eventually bold.

jonasschnelli:
Adding `f.setPointSize(12);` makes it look much better (on macOS, unsure about windows/linux):

hebasto:
> Adding `f.setPointSize(12);` makes it look much better (on macOS, unsure about windows/linux):

Right. But @luke-jr's opinion is [opposite](https://github.com/bitcoin-core/gui/pull/87#discussion_r511600141).

RandyMcMillan:
It would be great to add a font size selector - similar to the console interface. 

Screen Shot 2021-01-28 at 10 07 51 PM

hebasto:
Closed in favor of #207.

PR #88, opened 2020-09-11
Support block hash alias in console

promag:
Add support to use @best or @height instead of block hash in the
RPC console. The approach used is to simply replace those expressions
with the relevant RPC method.

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -32,6 +32,7 @@
 #include <QKeyEvent>
 #include <QMenu>
 #include <QMessageBox>
+#include <QRegExp>
 #include <QScreen>
 #include <QScrollBar>
 #include <QSettings>
@@ -81,7 +82,7 @@ public:
     explicit RPCExecutor(interfaces::Node& node) : m_node(node) {}
 
 public Q_SLOTS:
-    void request(const QString &command, const WalletModel* wallet_model);
+    void request(QString command, const WalletModel* wallet_model);
 
 Q_SIGNALS:
     void reply(int category, const QString &command);
@@ -383,10 +384,13 @@ bool RPCConsole::RPCParseCommandLine(interfaces::Node* node, std::string &strRes
     }
 }
 
-void RPCExecutor::request(const QString &command, const WalletModel* wallet_model)
+void RPCExecutor::request(QString command, const WalletModel* wallet_model)
 {
     try
     {
+        command.replace(QRegExp("@best"), "getbestblockhash()");
+        command.replace(QRegExp("@([\\d]+)"), "getblockhash(\\1)");
+
         std::string result;
         std::string executableCommand = command.toStdString() + "\n";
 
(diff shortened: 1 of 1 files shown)


promag:
Close and reopen, travis didn't pick latest commit.

promag:
Now using `QRegExp`, not sure why it fails to build with `QRegularExpression` since it's (supposedly) available since Qt 5.0.

hebasto:
> Now using `QRegExp`, not sure why it fails to build with `QRegularExpression` since it's (supposedly) available since Qt 5.0.

https://github.com/bitcoin/bitcoin/blob/9366a73d695136fec8de12d3942254543903c0cd/depends/packages/qt.mk#L99:
```
$(package)_config_opts += -no-feature-regularexpression
```

promag:
This is an alternative approach for the GUI change in https://github.com/bitcoin/bitcoin/pull/16439

src/qt/rpcconsole.cpp
```diff
@@ -383,10 +384,13 @@ bool RPCConsole::RPCParseCommandLine(interfaces::Node* node, std::string &strRes
     }
 }
 
-void RPCExecutor::request(const QString &command, const WalletModel* wallet_model)
+void RPCExecutor::request(QString command, const WalletModel* wallet_model)
 {
     try
     {
+        command.replace(QRegExp("@best"), "getbestblockhash()");
+        command.replace(QRegExp("@([\\d]+)"), "getblockhash(\\1)");
```
ajtowns, line 392: Would it be better to have "\b@best\b" (and likewise for digits) so you don't replace "bob@bestdomainever.com" in a label or similar?

hebasto, line 392: The opening word boundary doesn't work with a "word" that starts with the `@` character. But the closing word boundary indeed improves behavior and error handling:
```suggestion
        command.replace(QRegExp("@best\\b"), "getbestblockhash()");
        command.replace(QRegExp("@([\\d]+)\\b"), "getblockhash(\\1)");
```

hebasto (changes requested):
Approach ACK d2379e5c8f831abc244616069b6382626d6a6382.

Only `help-console` message provides a user with info about new aliases. The help messages of the affected RPC's still lack these updates, but I think it is OK for the GUI console.

This PR seems preferable than the [2f9f2d66cbada9134238f25aa12691876bdeff6b](https://github.com/bitcoin/bitcoin/pull/16439/commits/2f9f2d66cbada9134238f25aa12691876bdeff6b) commit from https://github.com/bitcoin/bitcoin/pull/16439.

src/qt/rpcconsole.cpp
```diff
@@ -32,6 +32,7 @@
 #include <QKeyEvent>
 #include <QMenu>
 #include <QMessageBox>
+#include <QRegExp>
```
luke-jr, line 35: I don't think this warrants adding `QRegExp` to the project

src/qt/rpcconsole.cpp
luke-jr, line 392: I think `RPCParseCommandLine` would be the proper place to do this.

luke-jr (changes requested):

src/qt/rpcconsole.cpp
promag: We already have it. Maybe you are thinking of `QRegularExpression`?

hebasto:
@promag Any update on this? :)

promag:
@hebasto I'm not convinced this is a good change. Closing for now and leaving it up for grabs.

Issue #77, opened 2020-08-27
[Design] New 'Create Wallet' user flow. 


Bosch-0:
**Overview of current and new create wallet user flow**
![image](https://user-images.githubusercontent.com/55287964/91424131-fc58d600-e88b-11ea-9e23-bed5aa18126f.png)

**1.0 Who is this being designed for?** 

I am under the assumption that currently the Bitcoin core GUI primarily caters to Bitcoin power users and or developers who are okay with usability trade offs for improved security and more robustly audited code. With the GUI being one of the most secure, simple ways to run a full node, and if a goal of the Bitcoin community is to have as many full nodes as possible (which I presume it is), then we need the GUI to cater to a wider, less technical audience. 

The aim of my designs is to remove friction for beginner and power users without having any security and or privacy trade offs, as well as have as minimal technical hurdles. Primarily, however, my goal is to make the GUI more beginner friendly. The Bitcoin core GUI in my view should be a benchmark / reference implementation and should not aim to be a flashy node software like [umbrel](https://getumbrel.com/) or [mynode](https://mynodebtc.com/) – but this does not mean usability improvements cant be made. A well designed GUI will also further improve the security and privacy of its users - for this reason core should put more emphasis on design implementations. With good design, both the beginners and power users will be able to send and receive bitcoin backed by their own full node using the GUI without trade offs. 

**2.0 Problems being solved**

The current user flow for creating a new wallet does not explain to users the types of wallets you can create or how each wallet type is different. Furthermore, the types of wallets available to create use terminology that is likely unfamiliar to many users (e.g. ‘Disable private keys’ could just be named ‘Watch Only Wallet’ which is more common and describes the same kind of wallet) - Luke Jr brang up this concern in https://github.com/bitcoin/bitcoin/pull/17879.

[...]

PR #96, opened 2020-09-18
Slight improve create wallet dialog

Sjors:
Previously only users who needed a second wallet had to use to the create wallet dialog. With the merge of https://github.com/bitcoin/bitcoin/pull/15454 now all new users have to. I don't think it was user-friendly enough for that.

Schermafbeelding 2020-09-18 om 09 41 44

This PR makes a few simple improvements so that new users don't have to think too much:

Schermafbeelding 2020-10-15 om 16 45 22

It's lightly inspired by #77. It would be better if those changes made it into the upcoming release, but this PR is a good start imo.

* wallet encryption is no longer checked by default, because such a change in the default needs a separate discussion (fwiw, I suspect it increases the number of users losing access to coins)
* watch-only and descriptor wallet stuff is moved to advanced, so new users know they can safely ignore these check boxes
* bonus: when you click on "disable private keys" it disables encrypt wallet and checks blank wallet
* label changes: see screenshot
* tooltip changes: see code diff

Note that a blank wallet name isn't allowed in the dialog; I haven't addressed that.

_Update 2020-10-30_, dropped the new strings for now:
Schermafbeelding 2020-10-30 om 11 26 55

### DIFF
--- a/src/qt/createwalletdialog.cpp
+++ b/src/qt/createwalletdialog.cpp
@@ -35,11 +35,28 @@ CreateWalletDialog::CreateWalletDialog(QWidget* parent) :
         }
     });
 
-#ifndef USE_SQLITE
-    ui->descriptor_checkbox->setToolTip(tr("Compiled without sqlite support (required for descriptor wallets)"));
-    ui->descriptor_checkbox->setEnabled(false);
-    ui->descriptor_checkbox->setChecked(false);
-#endif
+    connect(ui->disable_privkeys_checkbox, &QCheckBox::toggled, [this](bool checked) {
+        // Disable the encrypt_wallet_checkbox when isDisablePrivateKeysChecked is
+        // set to true, enable it when isDisablePrivateKeysChecked is false.
+        ui->encrypt_wallet_checkbox->setEnabled(!checked);
+
+        // Wallets without private keys start out blank
+        if (checked) {
+            ui->blank_wallet_checkbox->setChecked(true);
+        }
+
+        // When the encrypt_wallet_checkbox is disabled, uncheck it.
+        if (!ui->encrypt_wallet_checkbox->isEnabled()) {
+            ui->encrypt_wallet_checkbox->setChecked(false);
+        }
+    });
+
+    #ifndef USE_SQLITE
+        ui->descriptor_checkbox->setToolTip(tr("Compiled without sqlite support (required for descriptor wallets)"));
+        ui->descriptor_checkbox->setEnabled(false);
+        ui->descriptor_checkbox->setChecked(false);
+    #endif
+
 }
 
 CreateWalletDialog::~CreateWalletDialog()
--- a/src/qt/forms/createwalletdialog.ui
+++ b/src/qt/forms/createwalletdialog.ui
@@ -68,12 +68,12 @@
     <string>Encrypt Wallet</string>
    </property>
    <property name="checked">
-    <bool>true</bool>
+    <bool>false</bool>
    </property>
   </widget>
   <widget class="QCheckBox" name="disable_privkeys_checkbox">
    <property name="enabled">
-    <bool>false</bool>
+    <bool>true</bool>
    </property>
    <property name="geometry">
     <rect>
--- a/src/qt/forms/createwalletdialog.ui
+++ b/src/qt/forms/createwalletdialog.ui
@@ -74,6 +74,22 @@
     <bool>false</bool>
    </property>
   </widget>
+  <widget class="QLabel" name="advanced_options_label">
+   <property name="geometry">
+    <rect>
+     <x>20</x>
+     <y>90</y>
+     <width>130</width>
+     <height>21</height>
+    </rect>
+   </property>
+   <property name="styleSheet">
+    <string notr="true">font-weight:bold;</string>
+   </property>
+   <property name="text">
+    <string>Advanced options</string>
+   </property>
+  </widget>
   <widget class="QCheckBox" name="disable_privkeys_checkbox">
    <property name="enabled">
     <bool>true</bool>
@@ -81,7 +97,7 @@
    <property name="geometry">
     <rect>
      <x>20</x>
-     <y>80</y>
+     <y>115</y>
      <width>171</width>
      <height>22</height>
     </rect>
@@ -97,8 +113,8 @@
    <property name="geometry">
     <rect>
      <x>20</x>
-     <y>110</y>
-     <width>171</width>
+     <y>135</y>
+     <width>220</width>
      <height>22</height>
     </rect>
    </property>
@@ -113,7 +129,7 @@
    <property name="geometry">
     <rect>
      <x>20</x>
-     <y>140</y>
+     <y>155</y>
      <width>171</width>
      <height>22</height>
     </rect>
@@ -131,6 +147,7 @@
   <tabstop>encrypt_wallet_checkbox</tabstop>
   <tabstop>disable_privkeys_checkbox</tabstop>
   <tabstop>blank_wallet_checkbox</tabstop>
+  <tabstop>descriptor_checkbox</tabstop>
  </tabstops>
  <resources/>
  <connections>
--- a/src/qt/forms/createwalletdialog.ui
+++ b/src/qt/forms/createwalletdialog.ui
@@ -38,6 +38,9 @@
      <height>24</height>
     </rect>
    </property>
+   <property name="placeholderText">
+     <string>Wallet</string>
+   </property>
   </widget>
   <widget class="QLabel" name="label">
    <property name="geometry">


jonatack:
Concept ACK, seems much better.

I'm probably missing something, but why "blank wallet" instead of "watchonly"?

Sjors:
A blank wallet has no keys at all. A watch-only wallet has no private keys. A watch-only wallet starts out blank. It's confusing, and probably deserves an even more obscure place in the UI :-)

jonatack:
> A blank wallet has no keys at all, but you can add private keys to it. A watch-only wallet has no private keys. A watch-only wallet starts out blank, and then you can't add private keys to it. It's confusing, and probably deserves an even more obscure place in the UI :-)

Thanks. Will test.

GBKS:
Good changes. Could "Encrypt wallet" also be "Protect wallet with a password"? Would it be possible to add links to resources that explain the advanced options to more easily read up on what they do?

Sjors:
@GBKS I think I'll leave that for whoever implements #77

Bosch-0:
Concept ACK, could you not call 'disable private keys' watch-only wallet? It's both technically correct and I'd assume is a more familiar term to users. 

> Good changes. Could "Encrypt wallet" also be "Protect wallet with a password"?

Wouldn't Encrypt private keys be more accurate? The wallet isn't being encrypted only the private keys from my understanding - though I think wallet contents a long with meta data like tx history should also be encrypted at some point (again, this is also a familiar process to many users). If this change is made Encrypt wallet or protect wallet with a password is more suitable.

Relevant discussion https://github.com/bitcoin/bitcoin/issues/18085

Sjors:
I updated same labels and tooltips.

harding:
Concept ACK.  I like the idea of separating out the confusing and less-often-needed features into an advanced section.

MarcoFalke:
Concept ACK

src/qt/createwalletdialog.cpp
```diff
@@ -34,6 +34,22 @@ CreateWalletDialog::CreateWalletDialog(QWidget* parent) :
...
     });
+
+    connect(ui->disable_privkeys_checkbox, &QCheckBox::toggled, [this](bool checked) {
+        // Disable the encrypt_wallet_checkbox when isDisablePrivateKeysChecked is
+        // set to true, enable it when isDisablePrivateKeysChecked is false.
+        ui->encrypt_wallet_checkbox->setEnabled(!checked);
+
+        // Wallets without private keys start out blank
+        if (checked) {
+            ui->blank_wallet_checkbox->setChecked(true);
```
promag, line 45: Should also disable `blank_wallet_checkbox`?

Sjors: Neh, it gets too tedious. It doesn't matter what you do with this checkbox, it's ignored for watch-only.

src/qt/forms/createwalletdialog.ui
```diff
@@ -74,14 +74,27 @@
     <bool>false</bool>
    </property>
   </widget>
+  <widget class="QLabel" name="label">
```
hebasto, line 77: The "label" name is already in use: see line 45.

hebasto (commented):
Concept ACK e84e07c810f3e187a8536d947254fab2e29107d0, tested on Linux Mint 20 (x86_64, Qt 5.12.8).

- master (875e1ccc9fe01e026e564dfd39a64d9a4b332a89):
Screenshot from 2020-10-05 15-47-27

- this PR (e84e07c810f3e187a8536d947254fab2e29107d0):
Screenshot from 2020-10-05 15-52-44

I'd prefer to see more designers (@Bosch-0, @GBKS) ACKs or comments after the [recent](https://github.com/bitcoin-core/gui/pull/96#issuecomment-694786668) update.

joernroeder:
I am only following the discussion here, therefore I might miss something. First of all, I like the separation into advanced features, but I don't like the idea to disable the encryption by default and would prefer a "fail safe" strategy here. 
As users might not know the consequences, add a description explaining them and potential loss of funds next to the password input fields in the next step or – when users check it off, describe the potential risk next to/below the checkbox.

Both decisions have their pro and cons, I would at least explain why encryption was the default in the first place.

Another minor question: Why `PascalCase` in the placeholder?

GBKS:
I'm wondering if "Manually import seed or keys" would be a more helpful label since that is what the user wants to do (Skipping the seed generation is what allows them to import a seed or keys, but it's not their actual goal)? What do you think?

Otherwise, looks great.

Bosch-0:
Concept ACK, looks good on Windows 10. I agree with with Christoph on renaming **'Skip seed generation'** to **'Manually import seed or keys'**, it's more action driven.

image

harding:
@joernroeder 

> I don't like the idea to disable the encryption by default and would prefer a "fail safe" strategy here.

FYI, no released version of Bitcoin Core has ever created encrypted wallets by default; this PR just preserves that legacy.

As background, there's an open question between experts about whether or not the use of wallet encryption in typical user wallets saves more money than it loses.  It saves money if someone gets a hold of just your wallet file (if they get direct access to your computer, they can observe your passphrase and steal your funds any way).  It loses money if you forget your passphrase.  Some experts believe the number of occasions where a wallet file has fallen into a attacker's hands without that attacker getting direct access to the user's computer is small compared to the large number of occasions where a user has forgotten their passphrase, so it's on average safer not to use encryption.

(I don't hold a strong opinion here myself.  I assume that any bitcoins stored in my hot wallet will be stolen someday and prefer leaving my wallet unencrypted so I'm likely to learn about the theft as early as possible.)

---

I also like @GBKS's suggested label rename to `Manually import seed or keys`.

laanwj:
> Some experts believe the number of occasions where a wallet file has fallen into a attacker's hands without that attacker getting direct access to the user's computer is small compared to the large number of occasions where a user has forgotten their passphrase, so it's on average safer not to use encryption.

If it's anything to go by I've received way more sob stories about people losing their wallet passphrase than about stolen funds, regarding bitcoin core. I'm not sure that's a reliable measure, but because of this I'm partial to not encrypting by default.

Bosch-0:
I agree with not encrypting by default. Hot wallets should only be used for small amounts and should not warrant the extra security complexity. An external signer (coming soon to the GUI https://github.com/bitcoin-core/gui/pull/4) should be used for larger amounts at a minimum. Though what consists of a larger amount will vary from person to person. The various security levels and trade offs of each wallet type should be communicated to users clearly (Getting there https://github.com/bitcoin-core/gui/issues/77).

joernroeder:
Thank you @harding for all the details. It seems like I misinterpreted the first screenshot as the current default which has encryption checked. Also thanks @Bosch-0 for pointing me to #77 which is what I had in mind.

Sjors:
@joernroeder the default for the (only recently introduced) wallet creation dialog was to enable encryption. But the default wallet that Bitcoin Core creates for new users has encryption disabled. Until recently the wallet creation dialog was only used by people creating a _second_ wallet, which presumable is a more experienced subset of users.

hebasto:
@Sjors 
> ... but prefer to conserve ACKs.

Which one? :smiley:

Sjors:
Right... rebased and changed the label.

hebasto (changes requested):
ACK ad0874e0f53f7c9a731c701451cfa50da412304c, tested on Linux Mint 20 (x86_64, Qt 5.12.8).

For the nice commit history mind squashing commits "gui: create wallet: add advanced section" and "gui: rename duplicate label in createwalletdialog.ui"?

jonatack:
Approach ACK, on my test shortlist.

src/qt/forms/createwalletdialog.ui
```diff
@@ -62,64 +65,77 @@
...
      <height>22</height>
     </rect>
    </property>
    <property name="toolTip">
-    <string>Use descriptors for scriptPubKey management</string>
+    <string>Use descriptors for scriptPubKey management.</string>
    </property>
    <property name="text">
-    <string>Descriptor Wallet</string>
+    <string>Use descriptors</string>
```
jonatack, line 138: "Descriptor wallet" seems better?

src/qt/forms/createwalletdialog.ui
```diff
@@ -62,64 +65,77 @@
...
      <height>22</height>
     </rect>
    </property>
    <property name="toolTip">
-    <string>Make a blank wallet. Blank wallets do not initially have private keys or scripts. Private keys and addresses can be imported, or an HD seed can be set, at a later time.</string>
+    <string>Blank wallets do not initially have private keys or scripts. Private keys and addresses can be imported, or an HD seed can be set, at a later time.</string>
    </property>
    <property name="text">
-    <string>Make Blank Wallet</string>
+    <string>Manually import seed or keys</string>
```
jonatack, line 122: Unless the manual import takes place immediately here on creation, this change seems confusing; "Blank wallet", along with the previous tooltip, seems more clear.

jonatack (commented):
Reading the code on GitHub, because I'm building another branch locally.

ISTM these 3 "advanced" options would read coherently:

- [ ] Watch-only wallet
- [ ] Blank wallet
- [ ] Descriptor wallet

src/qt/forms/createwalletdialog.ui
Sjors: The word "wallet" seems redundant. Also, you can mix and match, e.g. a watch-only descriptor wallet (not a watch-only wallet descriptor wallet)

src/qt/forms/createwalletdialog.ui
Sjors: Anyone using this feature will probably be following a tutorial, so I don't think we should bikeshed this to death. I can see a followup design improvement adding a wizard where you can copy-paste descriptors at this step.

src/qt/forms/createwalletdialog.ui
jonatack: Hm, good points. WDYT about:

- Watch-only
- Blank
- Descriptors

or

- Watch-only
- Skip seed generation
- Use descriptors

jonatack: I guess we should mention what can be mixed with what.

src/qt/forms/createwalletdialog.ui
jonatack: P.S. apart from this ^, I'll be very reactive in feedback/review here to try to move it forward.

src/qt/forms/createwalletdialog.ui
Sjors: Watch-only implies blank (checkbox enforces that), other than that there's no restriction. See also the `createwallet` RPC documentation, which this is dumb mimic of.

The main point of this PR is to make v0.21 slightly more usable than it is now, because we no longer create a blank wallet. I'd rather not fine tune this too much, that's what #77 is for.

src/qt/forms/createwalletdialog.ui
jonatack: Fair enough, but this change seems strictly worse, as it's really misleading (and wasn't in your original proposal), nothing is being imported at this step.

jonatack: Maybe either:

Make blank wallet / tootip: Make a blank wallet. Blank wallets do not initially have private keys or scripts. Private keys and addresses can be imported, or an HD seed can be set, at a later time.

or

Skip seed generation / tooltip: Skip seed generation and make a blank wallet without private keys or scripts. Private keys and addresses can be imported, or an HD seed can be set, at a later time.

jonatack: I'll test the rest.

jonatack:
"I'd rather not finetune this" / "I don't think we should bikeshed this to death."

Ok sorry. I wanted to be help move forward but one of the changes doesn't make sense to me. I've removed my review feedback except for the one comment.

hebasto:
> Anyone using this feature will probably be following a tutorial, so I don't think we should bikeshed this to death. I can see a followup design improvement adding a wizard where you can copy-paste descriptors at this step.

We have a nice opportunity to avoid bikeshedding in this repo completely. All that we need is to realize that the GUI repo is a new home for collaboration of devs and [designers](https://github.com/BitcoinDesign). Code is devs' responsibility. Look-and-feel is designers' responsibility. Let's rely on opinion of designers (@Bosch-0 and @GBKS in this particular PR).

Let bikeshedding be designers' burden :)

Sjors:
That only works until there's more designers :-)

hebasto:
> That only works until there's more designers :-)

The [extrapolation](https://xkcd.com/605/) is optimistic :)

jonatack:
> Code is devs' responsibility. Look-and-feel is designers' responsibility. Let's rely on opinion of designers (@Bosch-0 and @GBKS in this particular PR).

Well, this is about text and I'm somewhat competent at communicating by writing, so I disagree. I also don't see why these comments were fine earlier on and are now called bikeshedding. But whatever, when I see the word "bikeshedding" I'm outtie.

Sjors:
@jonatack I've been resisting text changes to this PR since the start: https://github.com/bitcoin-core/gui/pull/96#issuecomment-694768086

I occasionally make them anyway if I need to fix a bug or rebase. But some of these comments are going in circles, where I change it and then someone asks why not use the original.

jonatack:
Well, apparently I'm that someone who is saying to not make confusing changes. Anyway, if I'm seen as bikeshedding then I apologize and will let you all move forward without me.

hebasto:
@jonatack 
> Well, this is about text and I'm somewhat competent at communicating by writing...

Being a non-native English speaker, I always rely on that your competence :)

MarcoFalke:
Concept ACK

Sjors:
Both @GBKS and @harding suggested the change to `Manually import seed or keys`, provided reasoning for that, nobody objected for a week, so I made the change. I don't really care either way, because these options are incomprehensible and useless without a manual. What I want is for new users to ignore these options, which is why I put them under "advanced". What I don't want to do is change the text once, and then change it back, and then change it again and miss the feature freeze deadline. The net result of that is keeping the unsafe version (with encryption enabled by default). Everyone is welcome to leave suggestions about text, I just stated again (and again) that I don't necessary plan to honour those suggestions.

ryanofsky (approved):
Code review ACK e84e07c810f3e187a8536d947254fab2e29107d0. Ignore my bike shed color list!

@jonatack, in the future could you maybe edit comments in ~strikethrough~ to update them instead of deleting them entirely? Deleted comments make github threads even harder to follow than normal. (Also, I happened to agree with the comments)

Rspigler:
Checking `Encrypt private keys` disables `Watch-only` :heavy_check_mark:

Checking `Watch-only` disables `Encrypt private keys` and enables `Manually import seed or keys`.  That makes sense to me when `Manually import...` was originally labelled `Blank Wallet` - but you can't import seed or keys to a watch-only wallet, so this seems like an error.  Because of this, I think naming it back to `Blank Wallet` makes more sense?

Unchecking `Watch-only` re-enables `Encrypt private keys`, and keeps `Manually import seed or keys enabled`. :heavy_check_mark:

What is the difference between a wallet that has `Watch-only` enabled, and one that has `Watch-only` and `Manually import seed or keys` enabled? In other words, should unchecking `Manually import seed or keys` also uncheck `Watch-only`? I believe so, and this is another reason why remaining `Manually import seed or keys` to `Blank Wallet` makes more sense IMO. (Unchecking `Blank Wallet` should uncheck `Watch-only`).

I also like @ryanofsky 's suggestion of `Skip seed generation`.

But I also understand I am coming in late, and there has been frustrations over bikeshedding.

Sjors:
Given that we missed the [translations freeze](https://github.com/bitcoin-core/gui/pull/21#issuecomment-715219566), it seems best to revert all text to their originals. @jonasschnelli?

Sjors:
I had to rebase. I dropped ad0874e with the new strings, because I'd still like to get this into 0.21. For a followup, here's the last list of strings I used:

> Encrypt private keys:
> Private keys in the wallet will be encrypted with a passphrase of your choice. Addresses, transaction history and notes are not encrypted.

> Watch-only
> Disable private keys for this wallet. Watch-only wallets have no private keys and cannot have an HD seed or imported private keys.

> Manually import seed or keys (_but see discussion above_):
> Blank wallets do not initially have private keys or scripts. Private keys and addresses can be imported, or an HD seed can be set, at a later time.

> Use descriptors:
> Use descriptors for scriptPubKey management.

ryanofsky (approved):
Code review ACK d69a436565c74484cac8b2814d2bee1bfeb59e72. Since last review, this was rebased and all changes to existing strings were reverted

src/qt/forms/createwalletdialog.ui
```diff
@@ -68,17 +71,30 @@
...
+   <property name="geometry">
+    <rect>
+     <x>20</x>
+     <y>90</y>
+     <width>130</width>
+     <height>21</height>
+    </rect>
+   </property>
+   <property name="text">
+    <string>Advanced options</string>
```
ysangkok, line 87: you say that there are no new strings, but this looks like a new string. why is it not a new string?

src/qt/createwalletdialog.cpp
```diff
@@ -35,11 +35,28 @@ CreateWalletDialog::CreateWalletDialog(QWidget* parent) :
...
+        }
+
+        // When the encrypt_wallet_checkbox is disabled, uncheck it.
+        if (!ui->encrypt_wallet_checkbox->isEnabled()) {
+            ui->encrypt_wallet_checkbox->setChecked(false);
+        }
+    });
+
+    #ifndef USE_SQLITE
+        ui->descriptor_checkbox->setToolTip(tr("Compiled without sqlite support (required for descriptor wallets)"));
```
ysangkok, line 55: this also looks like a new string ?

src/qt/forms/createwalletdialog.ui
ysangkok: why is it not a heading like h1-6? seems like it is supposed to cover multiple items below?

src/qt/forms/createwalletdialog.ui
```diff
@@ -68,17 +71,30 @@
     <string>Encrypt Wallet</string>
    </property>
    <property name="checked">
-    <bool>true</bool>
+    <bool>false</bool>
+   </property>
+  </widget>
+  <widget class="QLabel" name="encrypt_wallet_label">
```
ysangkok, line 77: why is the property with text "Advanced options" under a widget named 'encrypt_wallet_label' in the XML hierarchy? encrypting the wallet is just one option.

ysangkok (changes requested):

ysangkok:
@Sjors did you know about [QGroupBox](https://doc.qt.io/qt-5/qgroupbox.html)? was it a conscious choice not to use it? it could be better for accessibility, since Qt would know the "Advanced options" label is a heading and could tell a screen reader that.

src/qt/forms/createwalletdialog.ui
Sjors: There's no modified strings, but indeed this is added. It will show in English for all users.

src/qt/createwalletdialog.cpp
Sjors: No, it's moved. Try viewing with ignored whitespace (add `?w=1` to github URL)

src/qt/forms/createwalletdialog.ui
Sjors: That should have been `advanced_options_label`. I'll update that if I need to make another change.

Sjors:
QGroupBox looks like a better approach, but I'm not handy enough with manipulation UI files and or using the QT view edit, dealing with size constraints and such. So would prefer to leave that to a followup.

src/qt/forms/createwalletdialog.ui
Sjors: I changed it to bold font, same as with Coin Selection in the send screen.

jonatack:
Will review/test tomorrow for 0.21

src/qt/forms/createwalletdialog.ui
jonasschnelli: Hmm... thats unfortunate. Would using "Options" (already translated) be an option?

src/qt/forms/createwalletdialog.ui
```diff
@@ -38,6 +38,9 @@
      <height>24</height>
     </rect>
    </property>
+   <property name="placeholderText">
+     <string>MyWallet</string>
```
jonasschnelli, line 42: this is also a new string,... though probably acceptable. Could use "Wallet Name" though,... but we already have this as label for the textfield.

Sjors: Ah yes, one alternative could be to allow empty names.

jonasschnelli:
Tested ACK d393708c9bf1d2ea2a953a236816dabc48a072f1 module the missing string for "Advanced options". If we are going to merge this for 0.21 (I don't see a pressing reason to do so but fine with it), we could switch to just "Option" (already translated) and change that post 0.21 branch-off.

Sjors:
I pushed an additional commit that allows the wallet name to be empty, which creates the default wallet (or throws an error if you already have it). That gets rid of the "MyWallet" string.

I prefer to stick to "Advanced options" rather than just "Options" in order to discourage their blindly using them. My guess is that if someone doesn't understand the English term, they'll be discouraged :-)

ryanofsky (changes requested):
Code review ACK d393708c9bf1d2ea2a953a236816dabc48a072f1 which just tweaks createdialog label id, tabstop & stylesheet property since last review.

I don't think newly added commit 03ebe1b5938a73b41e34a1251520a2db29fafd95 is a good idea, though. Main motivation seems be avoiding adding a translation string (https://github.com/bitcoin-core/gui/pull/96#discussion_r522738216), but the new commit increases the scope of the PR beyond "slightly improving the create wallet dialog", to adding a GUI feature: creating and encouraging creation of a top level unnamed wallet which contains other wallets. I don't think the change is great, but even if it were a good idea, I think it would make more sense in its own feature PR, not a PR that's just consisted of dialog box tweaks up to this point and is still described that way.

Sjors:
>  creation of a top level unnamed wallet which contains other wallets

I'm not sure what you mean. This is basically the original behaviour where the first wallet is `wallet.dat`. The UI already handles it.

ryanofsky:
> I'm not sure what you mean. This is basically the original behaviour where the first wallet is `wallet.dat`. The UI already handles it.

Right, and it was intentionally disabled in 78863e290006e61060622dbdbecc5b58c0fefa05 from #15450. The new commit 03ebe1b5938a73b41e34a1251520a2db29fafd95 removes the code which disables it. I don't think the change is a good idea and wrote some reasons why. Would be happy to discuss more, but the point is this change goes beyond tweaking the dialog and shouldn't be part of this PR

Sjors:
bitcoin/bitcoin#15450 introduced the ability to add _additional_ wallets in the GUI, basically an advanced feature. It wasn't designed to be part of the onboarding flow. It makes sense that it didn't allow a blank wallet name, but it would never be used to create the initial wallet.

The PR description of  bitcoin/bitcoin#15454 says (emphasis mine):
> Instead of **automatically** creating and loading a default wallet, users should instead explicitly create their wallet or load it on start.

It wasn't clear to me that intention of that PR was to get rid of the unnamed wallet entirely. It seemed more like an oversight to me.

That said, I don't object to ditching the unnamed wallet. But it does create the annoying problem of the user having to pick a name.

I can drop the last commit (or the maintainer can just omit it from the merge) if needed.

ryanofsky:
PR title "Slight improve create wallet dialog" title is currently misleading. Right now this PR is also adding back ability to create a top-level unnamed wallet containing other wallets from the GUI that was removed in #15454 (definitely intentional for me, though I can't speak for other participants).

I don't think commit 03ebe1b5938a73b41e34a1251520a2db29fafd95 is a good idea. I think it is confusing for a wallet to contain other wallets. I think wallets that are unnamed are more likely to get lost and not backed up properly than wallets that have recognizable names. I think if the last commit 03ebe1b5938a73b41e34a1251520a2db29fafd95 is desirable at all, it should be a standalone PR, which should have some rationale stating why it is desirable. It shouldn't be tacked on to a PR doing other cleanup as an extra commit with no rationale or description. Even from discussion in this PR, I haven't seen any justification of the commit, except as workaround to avoid adding a translation string.

It seems fine to just drop the placeholder string if we can't have a translation now and don't want an untranslated string. I think commit 03ebe1b5938a73b41e34a1251520a2db29fafd95 is beyond scope of this PR.

jonatack:
Tested ACK d393708c9bf1d2ea2a953a236816dabc48a072f1 "gui: create wallet: add advanced section" with the following reservations:

- I agree with @ryanofsky about dropping the last commit.
- Clicking on the watch-only option does indeed auto-check the blank wallet option but also violates the principle of least surprise and may look to many users like a bug in the code, which may reduce user confidence in the wallet. Perhaps remove this auto-checking, or maybe append ` (auto-selects "Make Blank Wallet")` to the watch-only text but the dialog may need to be larger to accomodate that.
  
> For a followup, here's the last list of strings I used:
> 
> > Encrypt private keys:
> > Private keys in the wallet will be encrypted with a passphrase of your choice. Addresses, transaction history and notes are not encrypted.
> 
> > Watch-only
> > Disable private keys for this wallet. Watch-only wallets have no private keys and cannot have an HD seed or imported private keys.

For now or later, I like these two changes.

> Another minor question: Why `PascalCase` in the placeholder?

Had the same thought. Maybe "My Wallet".

src/qt/forms/createwalletdialog.ui
```diff
@@ -68,17 +68,33 @@
...
+   <property name="geometry">
+    <rect>
+     <x>20</x>
+     <y>90</y>
+     <width>130</width>
+     <height>21</height>
+    </rect>
+   </property>
+   <property name="styleSheet">
+    <string notr="true">font-weight:bold;</string>
```
jonatack, line 84: (QT newb question: what is `notr`?)

jonatack: Sorry, should have looked it up. Per https://doc.qt.io/qt-5/designer-creating-custom-widgets.html, if the `notr` optional attribute is true, the value is not meant to be translated.

Sjors: I just copy-pasted that from another ui file :-)

Sjors:
@ryanofsky I still don't understand what you mean with "wallet to contain other wallets". It's just one wallet.

I dropped the last commit. To the degree it wasn't clear whether disallowing the `""` wallet was intentional, it is now :-)

I changed "MyWallet" to "Wallet" to avoid CamelCase and at the same avoid spaces in directory names, which tend to trip up badly written backup commands. For future versions we can translate it, but I suggest hinting to the translators to avoid spaces. Unicode characters (e.g. Angličtina)  are probably fine, because afaik we have tests for those. But Arab seems to have a space:  محفظة نقود

I don't think dropping placeholder completely, yet not allowing a blank name, is a good idea: users won't know why Create is disabled and they'll probably start selecting random options.

Of course it's just a placeholder text so the user might type something completely different. I'm still not a fan of forcing users to pick a name, but agree it should be out of scope for this PR.

> Clicking on the watch-only option does indeed auto-check the blank wallet option but also violates the principle of least surprise

It's similar to the pre-existing behaviour of clicking on Encrypt wallet. Explaining the behaviour better makes sense, but since I can't change the strings, that would be for a followup (I also don't want to risk whack-a-mole by touching window size, though that's probably safe).

src/qt/forms/createwalletdialog.ui
hebasto: string is **no**t **tr**anslated :)

hebasto (approved):
re-ACK ac64cec4ce35cb7e6fbf3678c1ffeac0137791ae

Two new strings are added after translation freeze:
- https://github.com/bitcoin-core/gui/blob/ac64cec4ce35cb7e6fbf3678c1ffeac0137791ae/src/qt/forms/createwalletdialog.ui#L42
- https://github.com/bitcoin-core/gui/blob/ac64cec4ce35cb7e6fbf3678c1ffeac0137791ae/src/qt/forms/createwalletdialog.ui#L90

I'm ok with them though, as the main goal--to disable wallet encryption by default--could be achieved in 0.21.

@harding 
> FYI, no released version of Bitcoin Core has ever created encrypted wallets by default; this PR just preserves that legacy.

A screenshot from 0.20.1:
Screenshot from 2020-11-14 18-10-17

I agree that implementation details could be improved (https://github.com/bitcoin-core/gui/pull/96#issuecomment-726337165) in a followup.

@Sjors 
I apologize for the delay in the review.

Rspigler:
`Encrypt Wallet` is still incorrectly described.

Checking `Disable Private Keys` checks `Make Blank Wallet` (and disables `Encrypt Wallet`).  (This is correct).
However, unchecking `Make Blank Wallet` should uncheck `Disable Private Keys` (and re-enable for checking `Encrypt Wallet`).  Currently, it does nothing.

Currently, you are left with settings: `Disable Private Keys` checked, `Make Blank Wallet` unchecked, and `Encrypt Wallet` disabled.  This seems like a bug when compared to the second sentence in my post.

If we are trying to fit this in 0.21 I guess these can be fixed in another PR

fjahr:
Tested ACK ac64cec4ce35cb7e6fbf3678c1ffeac0137791ae

Tested on MacOS 10.15.7. Seems like a great improvement!

ryanofsky (approved):
Code review ACK ac64cec4ce35cb7e6fbf3678c1ffeac0137791ae. Only changes since last review are tweaking placeholder text and dropping "allow nameless" commit

> @ryanofsky I still don't understand what you mean with "wallet to contain other wallets". It's just one wallet.

Sorry, didn't see earlier question. This is just referring to one wallet directory containing another wallet directory, e.g:

```
wallets/wallet.dat
wallets/db.log
wallets/wallet2/wallet.dat
wallets/wallet2/db.log
```

instead of 

```
wallets/wallet1/wallet.dat
wallets/wallet1/db.log
wallets/wallet2/wallet.dat
wallets/wallet2/db.log
```

jonatack:
> Tested ACK [d393708](https://github.com/bitcoin-core/gui/commit/d393708c9bf1d2ea2a953a236816dabc48a072f1) "gui: create wallet: add advanced section" with the following reservations:
> 
>I agree with ryanofsky about dropping the last commit.
> 
>Clicking on the watch-only option does indeed auto-check the blank wallet option but also violates the principle of least surprise and may look to many users like a bug in the code, which may reduce user confidence in the wallet. Perhaps remove this auto-checking, or maybe append ` (auto-selects "Make Blank Wallet")` to the watch-only text to make clear to users what is going on, but the dialog may need to be larger to accomodate that.
> 
> 
> > For a followup, here's the last list of strings I used:
> > > Encrypt private keys:
> > > Private keys in the wallet will be encrypted with a passphrase of your choice. Addresses, transaction history and notes are not encrypted.
> > 
> > 
> > > Watch-only
> > > Disable private keys for this wallet. Watch-only wallets have no private keys and cannot have an HD seed or imported private keys.
>
> For now or later, I like these two changes.

re-ACK ac64cec4ce35cb7e6fbf3678c1ffeac0137791ae, per `git diff d393708 ac64cec` only change since my last review is improving the placeholder from "MyWallet" to "Wallet" and dropping the last commit. Tested creating a dozen wallets in signet with different combinations of options and then verifying/comparing their characteristics in the console with getwalletinfo. My remaining caveats are the need for less user surprise by improving the user info or with less auto-(un)selecting as mentioned in https://github.com/bitcoin-core/gui/pull/96#issuecomment-727017409 and I prefer the above wording and descriptions over the current ones; hopefully these can be addressed in a follow-up.

jonatack:
> Sorry, didn't see earlier question. This is just referring to one wallet directory containing another wallet directory, e.g:
> 
> ```
> wallets/wallet.dat
> wallets/db.log
> wallets/wallet2/wallet.dat
> wallets/wallet2/db.log
> ```
> 
> instead of
> 
> ```
> wallets/wallet1/wallet.dat
> wallets/wallet1/db.log
> wallets/wallet2/wallet.dat
> wallets/wallet2/db.log
> ```

Agree, I'm seeing the same. And also, in signet the wallets do not even have their own `wallets` subdirectory.

Sjors:
> in signet the wallets do not even have their own wallets subdirectory

That would be very odd, maybe worth a seperate github issue. Did you start with pre-existing `/signet` directory?

jonatack:
> > in signet the wallets do not even have their own wallets subdirectory
> 
> That would be very odd, maybe worth a seperate github issue. Did you start with pre-existing `/signet` directory?

Thanks for circling back on that. I tested several different versions of the signet implementation PR before it was merged, but also rm -rf'ed that dir as many times. I guess I should start once again with a fresh signet dir to check.

Sjors:
Thanks for the merge!

I'm leaving the text changes followup up for grabs.

PR #97, opened 2020-09-18
Relax GUI freezes during IBD (when using wallets)

jonasschnelli:
Calling `updateSmartFeeLabel` and therefore `estimateSmartFee` is pointless during IBD.

GUI freezes appear because `estimateSmartFee` competes with `processBlock` for the `m_cs_fee_estimator` lock leading to multiple seconds of blocking the GUI thread in `updateSmartFeeLabel`.

### DIFF
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -28,6 +28,8 @@
 #include <wallet/fees.h>
 #include <wallet/wallet.h>
 
+#include <validation.h>
+
 #include <QFontMetrics>
 #include <QScrollBar>
 #include <QSettings>
@@ -134,7 +136,7 @@ void SendCoinsDialog::setClientModel(ClientModel *_clientModel)
     this->clientModel = _clientModel;
 
     if (_clientModel) {
-        connect(_clientModel, &ClientModel::numBlocksChanged, this, &SendCoinsDialog::updateSmartFeeLabel);
+        connect(_clientModel, &ClientModel::numBlocksChanged, this, &SendCoinsDialog::updateNumberOfBlocks);
     }
 }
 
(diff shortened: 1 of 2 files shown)


Sjors:
I can't reproduce GUI freezes in the Send dialog during IBD, but the change looks reasonable to me.

jonasschnelli:
@Sjors: I think you need to be at a height where blocks require more CPU cycles to verify.
I can 100% reproduce the freezes when syncing up a mainnet node after – say – a week of offlinetime and thus falling out of sync.

Sjors:
I see, in that case `invalidateblock` and `reconsiderblock` should be handy for testing.

ryanofsky (approved):
Code review ACK 0d9d2a1f7c26dc9c7b233ea8c3182fe1f8936bca. Clever fix. Didn't test but I remember I could reproduce the startup issue easily before by putting a sleep in estimateSmartFee.

src/qt/sendcoinsdialog.cpp
```diff
@@ -134,7 +136,7 @@ void SendCoinsDialog::setClientModel(ClientModel *_clientModel)
     this->clientModel = _clientModel;
 
     if (_clientModel) {
-        connect(_clientModel, &ClientModel::numBlocksChanged, this, &SendCoinsDialog::updateSmartFeeLabel);
+        connect(_clientModel, &ClientModel::numBlocksChanged, this, &SendCoinsDialog::updateNumberOfBlocks);
```
promag, line 139: nit, could connect to lambda to avoid changes in header file:
```cpp
connect(_clientModel, &ClientModel::numBlocksChanged, [this](int, const QDateTime&, double, bool, SynchronizationState sync_state) {
    if (sync_state == SynchronizationState::POST_INIT) {
        updateSmartFeeLabel();
    }
});

**Review by promag — COMMENTED** [review]
Code review ACK 0d9d2a1f7c26dc9c7b233ea8c3182fe1f8936bca.

**Comment by hebasto** [comment]
Concept ACK.

File: src/qt/sendcoinsdialog.cpp [hunk 2]
```diff
@@ -28,6 +28,8 @@
 #include <wallet/fees.h>
 #include <wallet/wallet.h>
 
+#include <validation.h>
```
hebasto, line 31: nit: No reasons to create a separate group of headers.

src/qt/sendcoinsdialog.h
```diff
@@ -98,6 +99,7 @@ private Q_SLOTS:
     void coinControlClipboardLowOutput();
     void coinControlClipboardChange();
     void updateFeeSectionControls();
+    void updateNumberOfBlocks(int count, const QDateTime& blockDate, double nVerificationProgress, bool headers, SynchronizationState sync_state);
```
hebasto, line 102: nit: `updateNumberOfBlocks` slot name for an object that knows nothing about "number of blocks" looks a bit weird. But I cannot suggest a better name (

hebasto (approved):
ACK 0d9d2a1f7c26dc9c7b233ea8c3182fe1f8936bca, tested on Linux Mint 20 (x86_64) with `QT_FATAL_WARNINGS=1` and `-debug=qt`.

PR #107, opened 2020-10-23
Fix main window geometry save/restore

### LINKED ISSUE #104: Default window size in Ubuntu 200% (HiDPI) mode is wrong
Starting Bitcoin-Qt on Ubuntu 20.04.01 with display scale of 200% results in a extremely tiny main window.

Bildschirmfoto 2020-10-23 um 12 50 45

hebasto:
Fix #104 ?

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -82,10 +82,10 @@ BitcoinGUI::BitcoinGUI(interfaces::Node& node, const PlatformStyle *_platformSty
     m_network_style(networkStyle)
 {
     QSettings settings;
-    if (!restoreGeometry(settings.value("MainWindowGeometry").toByteArray())) {
-        // Restore failed (perhaps missing setting), center the window
-        move(QGuiApplication::primaryScreen()->availableGeometry().center() - frameGeometry().center());
-    }
+    settings.beginGroup("MainWindow");
+    resize(settings.value("size", QSize(800, 540)).toSize());
+    move(settings.value("pos", QGuiApplication::primaryScreen()->availableGeometry().center() - frameGeometry().center()).toPoint());
+    settings.endGroup();
 
 #ifdef ENABLE_WALLET
     enableWallet = WalletModel::isWalletEnabled();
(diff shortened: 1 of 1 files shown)


src/qt/bitcoingui.cpp
```diff
@@ -82,10 +82,10 @@ BitcoinGUI::BitcoinGUI(interfaces::Node& node, const PlatformStyle *_platformSty
     m_network_style(networkStyle)
 {
     QSettings settings;
-    if (!restoreGeometry(settings.value("MainWindowGeometry").toByteArray())) {
-        // Restore failed (perhaps missing setting), center the window
-        move(QGuiApplication::primaryScreen()->availableGeometry().center() - frameGeometry().center());
-    }
+    settings.beginGroup("MainWindow");
+    resize(settings.value("size", QSize(800, 540)).toSize());
```
luke-jr, line 86: Not sure hard-coding a size in pixels is a good idea. It should auto-size based on what Qt determines is needed when a wallet is loaded (ie, all widgets visible)

luke-jr (changes requested):

src/qt/bitcoingui.cpp
hebasto: This is a default size that is used only at first run or after `-resetguisettings`.

luke-jr: I'm aware. But the ideal default size isn't measured in pixels.

hebasto: > But the ideal default size isn't measured in pixels.

Correct. The ideal size is set by users and it will be saved in settings. I don't think the default size could be ideal for more than one person :)

luke-jr: Until we removed the default wallet, the default window size was automatically sane for any platform. We just need to go back to that.

hebasto: @luke-jr Do you suggest to revert https://github.com/bitcoin/bitcoin/pull/15454 back or to calculate the default window size as if a wallet is loaded?

luke-jr: The latter of course

hebasto: Reworked in #116.

PR #90, opened 2020-09-14
Enlarge Network Traffic Graph

RandyMcMillan:
This minor GUI change allows for a larger display of the Network Graph.

Before/master:

Screen Shot 2020-09-20 at 1 28 32 AM

After/PR:

Screen Shot 2020-09-20 at 1 36 11 AM

Before/master:

Screen Shot 2020-09-20 at 1 32 07 AM

After/PR:

Screen Shot 2020-09-20 at 1 35 10 AM

I capitalized "Window" in the title per standard "Human Interface Guidelines" and common sense.

I added some accessibility settings as part of an on going personal effort to make the GUI more user friendly for impaired users.

PR #18, opened 2020-07-01
Add peertablesortproxy module

hebasto:
The "Peers" table in the "Node" window does not hold multiple selection after sorting.

This PR introduces a `QSortFilterProxyModel` subclass, that is a standard Qt [practice](https://doc.qt.io/qt-5/model-view-programming.html#custom-sorting-models) for such cases.

Now the sorting code is encapsulated into the dedicated Qt class, and we do not need to maintain it.

Fixes #283 (additionally).

---

On **master** (7ae86b3c6845873ca96650fc69beb4ae5285c801):
- rows are sorted by "Ping", and a selection is made
Screenshot from 2020-11-28 22-53-11

- rows are sorted by "NodeId", and the previous selection is _lost_
Screenshot from 2020-11-28 22-53-21

With **this PR**:
- rows are sorted by "Ping", and a selection is made
Screenshot from 2020-11-28 22-39-41

- rows are sorted by "NodeId", and the row are still selected
Screenshot from 2020-11-28 22-39-53

Issue #16, opened 2020-06-30
[Open Wallet] click freeze system, when in sync node


Saibato:
**Steps to reproduce:**
disconnect a node that has no "wallets" dir, in -datadir.
wait until you missed say 20 new blocks.
start the node with ui.
click [Open Wallet] ....

If your whole system ui hangs.
**Don;t panic!!!.** whole System should leave dead lock after < 5 min.
Close ui., hmm... :woman_shrugging: 

fixed by https://github.com/bitcoin/bitcoin/pull/19419

hebasto:
> **Steps to reproduce:**
> disconnect a node that has no "wallets" dir, in -datadir.
> wait until you missed say 20 new blocks.
> start the node with ui.
> click [Open Wallet] ....
> 
> If your whole system ui hangs.

Couldn't reproduce it. Using Linux Mint 20, Qt 5.12.8, master (49464004701e5ac256fe1f3e969464012af7c598).

Saibato:
> > **Steps to reproduce:**
> > disconnect a node that has no "wallets" dir, in -datadir.
> > wait until you missed say 20 new blocks.
> > start the node with ui.
> > click [Open Wallet] ....
> > If your whole system ui hangs.
> 
> Couldn't reproduce it. Using Linux Mint 20, Qt 5.12.8, master ([4946400](https://github.com/bitcoin-core/gui/commit/49464004701e5ac256fe1f3e969464012af7c598)).

Could obviously be system dependent, And maybe you have to stay only a bit shorter online again to start sync and get hight and then close the connection and restart with a proxy that yield to nowhere.
I guess it also can depend on the files driver and format as well
 
BTW i use a full indexed node. When we look at #18095 your system where also not able to reproduce,
BTW. i tested  #18095 that also, but had no effect on my system, ec/eci was not even triggered.

Saibato:
@hebasto
tyi, a node of mine, compiled with no wallet, was free of this issue and synced fine in GUI with 21rc1
Wallets in core ( if hot ) are anyway not that smart to use with hot keys, but as long as we have that build in by default, maybe we should just warn newbe users of this ugly annoying behavior
Because I see not much what we can do, if the wallet support is compiled in?

I still can reproduce this issue with any version above 19 up to 21rc2. lower I had not tried.

promag:
@Saibato what about `bitcoin-cli listwalletdir`, does it take too long?

Saibato:
@promag 
[...]

PR #108, opened 2020-10-23
WIP: mempool stats chart

jonasschnelli:
This PR adds a mempool statistics to the "debug" window.
The diagram is heavy inspired by [Jochen Hoenicke's webversion](https://jochen-hoenicke.de/queue/#0,24h) (Thanks @jhoenicke for that).

Bildschirmfoto 2020-10-23 um 16 30 51

The approach is it to collect the fees also when the mempool chart or the node window is not open (collect it from the start of the application).
Clicking on the fee-range color square highlights the fee group over time.

To keep the scope minimal:
* flexible timeframe is out of scope (can be added later)

See also https://github.com/bitcoin/bitcoin/pull/8550 (an initial attempt, different concept).

TODO:

- [ ] Make it non HiDPI compatible (linux/win/etc.).
- [ ] Analyze memory consumption (should not be too bad)
- [ ] Optimize performance
- [ ] Try to make fee ranges dynamic


jonatack:
Concept ACK -- nice!

RandyMcMillan:
Should the background color of the Network Traffic display be changed to white as well? - to create a consistent UX within the different tabs? If so - should they share a common theme - so they can be toggled together?

RandyMcMillan:
Concept ACK :) 

It seems that the parent window should trigger a repaint/resize of the view simular to the peer table data model...

https://github.com/bitcoin-core/gui/blob/d67883d01e507dd22d1281f4a4860e79d6a46a47/src/qt/peertablemodel.cpp#L219

Screen Shot 2020-10-25 at 6 33 17 PM


---

- A similar implementation may be useful to actually hide the tab until the node is up and running and some initial values have been populated
- Should flushing the mempool prior to shut down  - as an option - be added to the tab as a check box? default no? This would allow the window to check cached data to speed up the initial paint.


https://github.com/bitcoin-core/gui/blob/d67883d01e507dd22d1281f4a4860e79d6a46a47/src/qt/bitcoingui.cpp#L840

We can attempt to use similar settings to make a more consistent UX between the tabs by using the same dimensions for the data presentation view and the margin underneath...

PR https://github.com/bitcoin-core/gui/pull/90

promag:
Concept ACK.

Sjors:
Very cool, concept ACK.

jonasschnelli:
@RandyMcMillan 
Thanks for the feedback.
I'd like to avoid visual alignment of both of the graphs in this PR. Off course this should happen but my takeaway from bitcoin/bitcoin#8550 is that scope creep must be avoided if we want to get this in master.

I recommend to do the visual alignment later on.

src/qt/mempoolstats.cpp
```diff
@@ -0,0 +1,291 @@
...
+        c_y-=c_margin;
+        int i = 0;
+        for (const interfaces::mempool_feeinfo& list_entry : m_clientmodel->m_mempool_feehist[0].second) {
+            if (i > display_up_to_range) {
+                continue;
+            }
+            ClickableRectItem *fee_rect = new ClickableRectItem();
+            fee_rect->setRect(4, c_y, c_w, c_h);
+
+            QColor brush_color = colors[(i < static_cast<int>(colors.size()) ? i : static_cast<int>(colors.size())-1)];
```
Sjors, line 176:
```
comparison of integers of different signs: 'int' and 'std::__1::vector<QColor, std::__1::allocator<QColor> >::size_type' (aka 'unsigned long') [-Werror,-Wsign-compare]
```
(a bunch of other warnings too)

Sjors (commented):
The peers tab is dropped from the menu, try expanding `std::vector<TabTypes> tabs()` in `rpcconsole.h`

I managed to get a segfault by clicking on the chart (macOS), though only once. Also when clicking on an item in the legend.

You could split this into two PR's, one where the node collects all the histogram data and exposes it via RPC (with a functional test). And a GUI PR on top.

hebasto:
Concept ACK.

src/Makefile.qt.include
```diff
@@ -22,6 +22,7 @@ QT_FORMS_UI = \
   qt/forms/helpmessagedialog.ui \
   qt/forms/intro.ui \
```
thantkyaw1, line 23: Need

src/Makefile.qt.include
```diff
@@ -22,6 +22,7 @@ QT_FORMS_UI = \
...
   qt/forms/modaloverlay.ui \
+  qt/forms/mempoolstats.ui \
   qt/forms/openuridialog.ui \
```
thantkyaw1, line 26: Need

jarolrod (commented):
Concept ACK

One Note:
On macOS 11.1, Qt 5.15.2 - I can reliably cause a segmentation fault by clicking on any of the text on the left side. I can click on the corresponding box, but clicking on the text will cause a segmentation fault. Here is a pastebin link to the error report: https://pastebin.com/E7my5X8V

prayank23:
IMO the bar graph used by https://btc.bitaps.com is better than https://jochen-hoenicke.de/queue/#BTC%20(default%20mempool),24h

Reason: Its easy to look at it and use the information while broadcasting a bitcoin transaction

image

rebroad:
concept ACK. IMHO the window needs to be resizeable though. I also preferred the earlier graph that included a graph of the current feefilter.

rebroad:
image
Would also be good if it could use the colors available rather than try to use colours that aren't.

rebroad:
I also think it would make more sense to make the vertical axis memory usage, not tx count, as this way it's easier to see when blocks are smaller than they could be, and it also makes it clearer when your mempool is full (or close to full).

rebroad:
Thread 1 "bitcoin-qt.21" received signal SIGILL, Illegal instruction.
0x00000000083ae50b in MempoolStats::drawChart (this=0xa021060) at qt/mempoolstats.cpp:134
134                 max_txcount_graph = step*amount_of_h_lines;
(gdb) bt
#0  0x00000000083ae50b in MempoolStats::drawChart (this=0xa021060) at qt/mempoolstats.cpp:134
#1  0x00000000083b241b in QtPrivate::FunctorCall<QtPrivate::IndexesList<>, QtPrivate::List<>, void, void (MempoolStats::*)()>::call(void (MempoolStats::*)(), MempoolStats*, void**) (
    f=(void (MempoolStats::*)(MempoolStats * const)) 0x83adf00 <MempoolStats::drawChart()>, o=0xa021060, arg=0xa12ac60)
    at /usr/include/x86_64-linux-gnu/qt5/QtCore/qobjectdefs_impl.h:136
#2  0x00000000083b2365 in QtPrivate::FunctionPointer<void (MempoolStats::*)()>::call<QtPrivate::List<>, void>(void (MempoolStats::*)(), MempoolStats*, void**) (
    f=(void (MempoolStats::*)(MempoolStats * const)) 0x83adf00 <MempoolStats::drawChart()>, o=0xa021060, arg=0xa12ac60)
    at /usr/include/x86_64-linux-gnu/qt5/QtCore/qobjectdefs_impl.h:169
#3  0x00000000083b2233 in QtPrivate::QSlotObject<void (MempoolStats::*)(), QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*) (which=1, this_=0xa16fc70, r=0xa021060, a=0xa12ac60, ret=0x0)
    at /usr/include/x86_64-linux-gnu/qt5/QtCore/qobject_impl.h:120
#4  0x00007ffffe6ea0c2 in QObject::event(QEvent*) () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#5  0x00007ffffd60375b in QWidget::event(QEvent*) () from /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#6  0x00007ffffd5c483c in QApplicationPrivate::notify_helper(QObject*, QEvent*) ()
   from /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#7  0x00007ffffd5cc104 in QApplication::notify(QObject*, QEvent*) () from /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#8  0x00007ffffe6ba8d8 in QCoreApplication::notifyInternal2(QObject*, QEvent*) ()
   from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#9  0x00007ffffe6bd04d in QCoreApplicationPrivate::sendPostedEvents(QObject*, int, QThreadData*) ()
   from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#10 0x00007ffffe714263 in ?? () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#11 0x00007ffffa23c417 in g_main_context_dispatch () from /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#12 0x00007ffffa23c650 in ?? () from /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#13 0x00007ffffa23c6dc in g_main_context_iteration () from /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#14 0x00007ffffe71388f in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) ()
   from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#15 0x00007ffffe6b890a in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) ()
   from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#16 0x00007ffffe6c19b4 in QCoreApplication::exec() () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#17 0x00000000082458e9 in GuiMain (argc=1, argv=0x7ffffffee218) at qt/bitcoin.cpp:618
#18 0x000000000823eadf in main (argc=1, argv=0x7ffffffee218) at qt/main.cpp:21

hebasto:
> IMO the bar graph used by https://btc.bitaps.com is better than https://jochen-hoenicke.de/queue/#BTC%20(default%20mempool),24h
> 
> Reason: Its easy to look at it and use the information while broadcasting a bitcoin transaction

I like it. It uses the widget area more effective.

rebroad:
image
It would be idea if the functionality of the previous graph and the new could be combined. e.g. allow different durations to be shown (from 3 hours, to 2 weeks), have one line for tx count, but show the fee rates within the memory usage graph. The scale (tx count and MB usage) could both me shown on the right, with the fee-rates shown on the left. Change the colours slightly so that the minfeerate and tx count are still clearly visible if plotted over the memusage/feerate.

At least, this would be perfection IMHO :)

rebroad:
image
The functionality of clicking on a fee range and it displaying the number of transactions seems a bit odd. In this example it's saying there are over 1,000,000 transactions in the 30-40 satoshi range, yet the vertical axis only goes up to 20,000 tx.

rebroad:
Rebasing this today to core/master

RandyMcMillan:
#320 is a rebase of this PR.

hebasto:
The "22.0" milestone seems unrealistic here. Removing it.

rebroad:
image
Seems to get the font sizes wrong on some setups (using VcXsrv on Windows 10).

rebroad:
image
Thinking it might be nice if the vertical axis scaled like it did on the old graph so that more detail can be seen...

Rspigler:
@jonasschnelli Do you still have plans to work on this? Thanks!

laanwj:
Concept ACK, this looks really neat. I think it's good for people running the GUI to have some introspection into what is happening on their node, and visualizations like this are useful for that.

hebasto:
Closing due to inactivity. Feel free to reopen.

Rspigler:
Mark up for grabs?

hebasto:
> Mark up for grabs?

Done.

PR #114, opened 2020-10-24
macOS: Post user notifications using QSystemTrayIcon

### LINKED ISSUE #112: RFC: Migration from NSUserNotificationCenter to UNUserNotificationCenter
Apple suggests to use [`UNUserNotificationCenter`](https://developer.apple.com/documentation/usernotifications/unusernotificationcenter?language=objc) instead of deprecated [`NSUserNotificationCenter`](https://developer.apple.com/documentation/foundation/nsusernotificationcenter?language=objc).

I've played a bit with `UNUserNotificationCenter` and always getting the error "Notifications are not allowed for this application".

Not sure, but it seems the root of this error is unsigned/[unnotarized](https://github.com/bitcoin/bitcoin/issues/15774) application.

hebasto:
Pros:
- drop deprecated [`NSUserNotificationCenter`](https://developer.apple.com/documentation/foundation/nsusernotificationcenter?language=objc)
- get rid of some macOS-specific, i.e., Objective-C, code

Cons:
- this approach does not support launching a just compiled binary from the Terminal (this affects developers rather end users)

Close  #112

Note: no Growl is required :)


hebasto:
cc @jonasschnelli

jonasschnelli:
Will this still use macOS'es Notification Center?
How does this interact with non code signed and non "notarized" apps?

hebasto:
> Will this still use macOS'es Notification Center?

Yes, it will.

> How does this interact with non code signed and non "notarized" apps?

Tested on non code signed apps: `Bitcoin-Core.dmg` and gitian `bitcoin-717f9a25f414-osx-unsigned.dmg`.

jonasschnelli:
Concept ACK

Sjors:
How do I produce a notification on macOS this way?

```
2020-10-26T11:43:32Z [main] System tray properties:
2020-10-26T11:43:32Z [main]  is available: yes
2020-10-26T11:43:32Z [main]  supports balloon messages: yes
```

But sending transactions back and forth doesn't trigger a notification. I tried making a DMG and installing from that locally.

I don't mind making things a bit more annoying for self-compilation on macOS if it allows dropping a whole bunch of deprecated code.

hebasto:
@Sjors 
> How do I produce a notification on macOS this way?
>
> But sending transactions back and forth doesn't trigger a notification. I tried making a DMG and installing from that locally.

Tested on macOS 10.15:
- installed `Bitcoin-Core.dmg` which was cross compiled on Linux Mint 20
- installed gitian built `bitcoin-717f9a25f414-osx-unsigned.dmg`

While compiling locally on macOS 10.15 I get an error:
```
% make deploy
...
+ Finalizing .dmg disk image +
hdiutil: convert failed - Resource temporarily unavailable
make: *** [Makefile:1279: Bitcoin-Core.dmg] Error 1
```

Application has actually been installed into `Applications` folder, but, indeed, has no interactions with the Notification Center.

Is `make deploy` broken?

fanquake:
> Is make deploy broken?

No.

> + Finalizing .dmg disk image +
hdiutil: convert failed - Resource temporarily unavailable
make: *** [Makefile:1279: Bitcoin-Core.dmg] Error 1

This error happens when you run `make deploy`, but there is already a `Bitcoin-Core.dmg` mounted.

I tested this branch and notifications seem to work fine, although I don't remember seeing the cloud with the exclamation point before:

popup

notif_center

hebasto:
@Sjors @fanquake 

What are your environments (macOS version, Qt version)?

fanquake:
> What are your environments (macOS version, Qt version)?

macOS 10.15.7. Qt 5.15.1.

hebasto:
I've re-tested one more time.

> > Is make deploy broken?
> 
> No.

Correct. `make deploy` works fine. It was wrong usage by me :)

To make this PR work on macOS one should run `Bitcoin Core.app` (installation of `Bitcoin-Qt.dmg` is not required).

Some screenshots:

- **Catalina 10.15.7 (19H2)**
Screenshot from 2020-10-27 10-54-15
Screenshot from 2020-10-27 10-54-30

- **Big Sur 11.0 Beta (20A5395g)**
Screenshot from 2020-10-27 10-53-45
Screenshot from 2020-10-27 10-52-38
Screenshot from 2020-10-27 10-53-11

jonasschnelli:
Tested ACK 717f9a25f4144fc34cb926a8decf0ac6c9b9dfd8

jonasschnelli:
Is the current notification support on macOS in master broken? If so, this qualifies as a bugfix I'd say.

hebasto:
> Is the current notification support on macOS in master broken?

No, it isn't.

Sjors:
macOS 10.15.7 with QT 5.15.1 via homebrew. I tried using `Bitcoin-Qt.app` rather than the DMG route and it works fine.

The tray icon is a bit weird, because nothing happens when you click on it:
Schermafbeelding 2020-10-30 om 15 38 46

That's because `trayIconActivated` still has an exception for macOS. It's a bit annoying that QT requires this tray icon to be visible in order for notifications to work, but anyway. 

The exclamation mark in notifications could be avoided, by tweaking `void Notificator::notifySystray`:

```cpp
{
    QSystemTrayIcon::MessageIcon sicon = QSystemTrayIcon::NoIcon;
    switch(cls) // Set icon based on class
    {
#ifdef Q_OS_MAC
    case Information: sicon = QSystemTrayIcon::NoIcon; break;
#else
    case Information: sicon = QSystemTrayIcon::Information; break;
#endif
```

Schermafbeelding 2020-10-30 om 15 44 43

hebasto:
@Sjors 

Thank you for reviewing!

> The tray icon is a bit weird, because nothing happens when you click on it:

I think some behavior could be assigned to the tray icon, but it would be unusual for macOS UX, no?

> The exclamation mark in notifications could be avoided, by tweaking `void Notificator::notifySystray`:

Do we need to avoid it?

Sjors:
Given that we can't avoid the tray icon (afaik), it would be good if it shows / hides the window. That way at least it doesn't feel broken.

Typically on macOS these icons provide a menu with shortcuts to various functions. That might be better, if it's not too much work:
Schermafbeelding 2020-10-31 om 09 23 05

As for the explanation mark, I would prefer to avoid it, because it's very non standard. It could be interpreted as an error. That said, it's not the end of the world.

hebasto:
> Given that we can't avoid the tray icon (afaik), it would be good if it shows / hides the window. That way at least it doesn't feel broken.
>
> Typically on macOS these icons provide a menu with shortcuts to various functions. That might be better, if it's not too much work:

I'd leave adding tray icon functionality for a follow up as it will probably raise another discussion.

> As for the explanation mark, I would prefer to avoid it, because it's very non standard. It could be interpreted as an error. That said, it's not the end of the world.

Ok, Going to patch it.

johnsBeharry:
### Notification Title
Nitpick for the notification title. Since the logo is already included its not necessary to repeat "Bitcoin Core".

Frame 9

---
### Context Menu

WRT the menu bar context menu options. Personally I'd like to be able to quickly start/stop syncing cause I'm not always on a consistent connection.

```
Network Activity: Enabled
Disable Network Activity
---
Receive
Send
---
Open Bitcoin Core
```

### Menu Bar App: Concept
@gbks took conceptual look at what the design for a menu bar app could look like. Perhaps he can give some input.

EdiZBWuXsAAAaGb
-- [source](https://twitter.com/GBKS/status/1285945949739790339/photo/1)

Sjors:
@johnsBeharry macOS notifications always have the application name in the title. It's also good for accessibility, think voice readers.

jonasschnelli:
I missed the new tray icon in my tests and only focused on the actual notification.
The visible global menu icon is a NACK element to me.
Can we use the system notification without a visible global menu icon? Under the hood, they are two separated frameworks. But maybe Qt combines them as one.

jonasschnelli:
This is the mac Qt code for further analysis: https://github.com/qt/qtbase/blob/81e09ae404b632a92e1e4c27f5875bdf027c5401/src/plugins/platforms/cocoa/qcocoasystemtrayicon.mm

hebasto:
> Can we use the system notification without a visible global menu icon?

No.

> Under the hood, they are two separated frameworks. But maybe Qt combines them as one.

If a tray icon is hidden, `m_statusItem == nil` and `QCocoaSystemTrayIcon::showMessage` is noop.

hebasto:
> This is the mac Qt code for further analysis: https://github.com/qt/qtbase/blob/81e09ae404b632a92e1e4c27f5875bdf027c5401/src/plugins/platforms/cocoa/qcocoasystemtrayicon.mm

It uses the deprecated `NSUserNotificationCenter`, and in Qt community I've found no signs of moving from `NSUserNotificationCenter` to `UNUserNotificationCenter`.

hebasto:
@luke-jr 
> While the menu app concept looks nice, I do think users should be able to disable it if they want, and shouldn't lose notifications by doing so...

Only visible icon allows notifications via `QSystemTrayIcon`. Adding menu to that icon is not a goal of this PR.

Another approach is to use the new `UNUserNotificationCenter`, but I have [no success](https://github.com/bitcoin-core/gui/issues/112) to authorize `bitcoin-qt` for notifications.

I'll readily close this PR in favor of any alternative one.

RandyMcMillan:
MacOS 10.14.6 (18G103)
Qt Creator 4.13.3
Based on Qt 5.15.2 (Clang 11.0 (Apple), 64 bit)
Built on Nov 13 2020 12:39:07
From revision 524cad144a

---

I agree with @jonasschnelli on the menu icon.
Until more functionality is added to the icon - it should be omitted (imo).

Screen Shot 2020-12-06 at 8 35 35 PM

Screen Shot 2020-12-06 at 7 46 16 PM

Note:

The date field seems to be broken. Tx values are correct but the date field isn't incrementing with the record.

I didn't rescan to test. I will try to get to it.

---
build command used...
---
rm -f src/bitcoind && rm -rf Bitcoin-Qt.app && rm -rf ~/Library/Saved\ Application\ State/org.bitcoinfoundation.Bitcoin-Qt.savedState && rm -rf ~/Library/Preferences/org.bitcoin.Bitcoin-Qt.plist && make appbundle && wait && ./Bitcoin-Qt.app/Contents/MacOS/Bitcoin-Qt

---

hebasto:
Closing as I cannot see the way to provide notifications in Qt way without the visible global menu icon.

hebasto:
@jonasschnelli Could be labeled "Up for grabs" :)

PR #115, opened 2020-10-24
Replace "Hide tray icon" option with positive "Show tray icon" one 

hebasto:
This change makes easier both (1) using this option, and (2) reasoning about the code.

### DIFF
--- a/src/qt/optionsmodel.cpp
+++ b/src/qt/optionsmodel.cpp
@@ -54,14 +54,15 @@ void OptionsModel::Init(bool resetSettings)
     // These are Qt-only settings:
 
     // Window
-    if (!settings.contains("fHideTrayIcon"))
+    if (!settings.contains("fHideTrayIcon")) {
         settings.setValue("fHideTrayIcon", false);
-    fHideTrayIcon = settings.value("fHideTrayIcon").toBool();
-    Q_EMIT hideTrayIconChanged(fHideTrayIcon);
+    }
+    m_show_tray_icon = !settings.value("fHideTrayIcon").toBool();
+    Q_EMIT showTrayIconChanged(m_show_tray_icon);
 
     if (!settings.contains("fMinimizeToTray"))
         settings.setValue("fMinimizeToTray", false);
-    fMinimizeToTray = settings.value("fMinimizeToTray").toBool() && !fHideTrayIcon;
+    fMinimizeToTray = settings.value("fMinimizeToTray").toBool() && m_show_tray_icon;
 
     if (!settings.contains("fMinimizeOnClose"))
         settings.setValue("fMinimizeOnClose", false);
@@ -272,8 +273,8 @@ QVariant OptionsModel::data(const QModelIndex & index, int role) const
         {
         case StartAtStartup:
             return GUIUtil::GetStartOnSystemStartup();
-        case HideTrayIcon:
-            return fHideTrayIcon;
+        case ShowTrayIcon:
+            return m_show_tray_icon;
         case MinimizeToTray:
             return fMinimizeToTray;
         case MapPortUPnP:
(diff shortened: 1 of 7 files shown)


src/qt/optionsmodel.cpp
```diff
@@ -54,14 +54,15 @@ void OptionsModel::Init(bool resetSettings)
     // These are Qt-only settings:
 
     // Window
-    if (!settings.contains("fHideTrayIcon"))
-        settings.setValue("fHideTrayIcon", false);
-    fHideTrayIcon = settings.value("fHideTrayIcon").toBool();
-    Q_EMIT hideTrayIconChanged(fHideTrayIcon);
+    if (!settings.contains("bShowTrayIcon")) {
```
promag, line 57: nit, is there a reason to change setting key? No big deal in this case, but ideally "old" key would be migrated - if you wish to rename the key.

promag (commented):
Code review ACK a124ee763e294a75871823c642f55f46633cdb31.

src/qt/optionsmodel.cpp
hebasto: It is not just renaming here. `fHideTrayIcon` and `bShowTrayIcon` are different semantically. As the `Bitcoin-Qt.conf` is written in the human readable format, I'd prefer to introduce a new setting key to avoid any possible confusing.

An old key migration sounds good to me. But is it worth to add more code for such a non-critical key?

promag: > It is not just renaming here. `fHideTrayIcon` and `bShowTrayIcon` are different semantically.

Of course, you would have to take the !fHideTrayIcon.

> An old key migration sounds good to me. But is it worth to add more code for such a non-critical key?

I've started with "nit" 😅

luke-jr: >It is not just renaming here. fHideTrayIcon and bShowTrayIcon are different semantically.

How?

>As the Bitcoin-Qt.conf is written in the human readable format, I'd prefer to introduce a new setting key to avoid any possible confusing.

That's not a good reason IMO

luke-jr (changes requested):

src/qt/optionsmodel.cpp
hebasto: > > It is not just renaming here. fHideTrayIcon and bShowTrayIcon are different semantically.
> 
> How?

The meaning of `fHideTrayIcon` is opposite to `bShowTrayIcon`, no?

> 
> > As the Bitcoin-Qt.conf is written in the human readable format, I'd prefer to introduce a new setting key to avoid any possible confusing.
> 
> That's not a good reason IMO

What do you suggest?

luke-jr: > The meaning of fHideTrayIcon is opposite to bShowTrayIcon, no?

It's a strict opposite.

>What do you suggest?

Leave `fHideTrayIcon` alone and just change the UI presentation.

hebasto:
Updated a124ee763e294a75871823c642f55f46633cdb31 -> 03edb52eee5a87af16161c23bdc6cde91a2e5b8b ([pr115.01](https://github.com/hebasto/gui/commits/pr115.01) -> [pr115.02](https://github.com/hebasto/gui/commits/pr115.02), [diff](https://github.com/hebasto/gui/compare/pr115.01..pr115.02)):

- addressed @luke-jr's [comment](https://github.com/bitcoin-core/gui/pull/115#discussion_r523783728):
> Leave `fHideTrayIcon` alone and just change the UI presentation.

src/qt/optionsmodel.cpp
hebasto: [Updated](https://github.com/bitcoin-core/gui/pull/115#issuecomment-727605555).

luke-jr (approved):
utACK

nit: I would split it into more commits. (eg, one changing GUI, one changing internals)

jonasschnelli (approved):
utACK 03edb52eee5a87af16161c23bdc6cde91a2e5b8b

PR #120, opened 2020-10-28
Fix multiwallet transaction notifications

promag:
Currently `vQueueNotifications` holds transactions of any wallet, but the queue is dispatched on a given wallet and it assumes notifications are of that wallet.

This means that some transactions can be missed if multiple wallets are loaded.

Fix this by having a queue for each wallet.

### DIFF
--- a/src/qt/transactiontablemodel.cpp
+++ b/src/qt/transactiontablemodel.cpp
@@ -54,6 +54,33 @@ struct TxLessThan
     }
 };
 
+// queue notifications to show a non freezing progress dialog e.g. for rescan
+struct TransactionNotification
+{
+public:
+    TransactionNotification() {}
+    TransactionNotification(uint256 _hash, ChangeType _status, bool _showTransaction):
+        hash(_hash), status(_status), showTransaction(_showTransaction) {}
+
+    void invoke(QObject *ttm)
+    {
+        QString strHash = QString::fromStdString(hash.GetHex());
+        qDebug() << "NotifyTransactionChanged: " + strHash + " status= " + QString::number(status);
+        bool invoked = QMetaObject::invokeMethod(ttm, "updateTransaction", Qt::QueuedConnection,
+                                  Q_ARG(QString, strHash),
+                                  Q_ARG(int, status),
+                                  Q_ARG(bool, showTransaction));
+        assert(invoked);
+    }
+private:
+    uint256 hash;
+    ChangeType status;
+    bool showTransaction;
+};
+
+static bool fQueueNotifications = false;
+static std::vector< TransactionNotification > vQueueNotifications;
+
 // Private implementation
 class TransactionTablePriv
 {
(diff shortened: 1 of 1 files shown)


src/qt/transactiontablemodel.cpp
```diff
@@ -734,7 +734,7 @@ void TransactionTablePriv::ShowProgress(const std::string &title, int nProgress)
         }
         for (unsigned int i = 0; i < vQueueNotifications.size(); ++i)
         {
-            if (vQueueNotifications.size() - i <= 10) {
+            if (vQueueNotifications.size() == 10 + 1 + i) {
```
ryanofsky, line 737: In commit "qt: Enqueue setProcessingQueuedTransactions(false) only once" (5ff42b72e93508dfae41e082bbbc7c7ee4aacf6b)

This is kind of hard to follow, and not 100% sure I understand it. IIUC, calling `setProcessingQueuedTransactions` with `false` argument turns on balloons, calling it with `true` argument turns off balloons.

Previously if queue size more than `10`, balloons were turned off before looping through transactions. Then for every transaction starting at `i >= size-10` (i.e. for the most recent 10 transactions) balloons were turned on before updating each transactions.

Now balloons are turned on at `i == size-11` so an extra balloon will be shown. (For example if size is `11`, balloons are now turned on at `i=0` instead of `i=1`).

If this is right, I'd make a few suggestions:

1. Update commit description to say what motivation for the change is: e.g. that this is a optimization not supposed to change behavior. Not a bugfix or something else.
2. Drop `+ 1` term which does not seem correct
3. Add comment before if statement like `// Re-enable balloons for the most recent 10 transactions`
4. Maybe write expression as `i == size-10` instead of `size == 10 + i`, which seems more obscure

ryanofsky (commented):
Code review almost-ACK 278ad5f47549cf88cce3e8684ab2e341c2e3c8c0. I think there's a minor off by one error in the third commit, and it doesn't seem as clear as it could be, but otherwise this looks very good

src/qt/transactiontablemodel.cpp
promag: Ok, I'll just drop this commit, it's unrelated anyway to the PR goal. This ballon filtering should be done elsewhere anyway.

promag: Removed 5ff42b7,

promag:
@jonasschnelli @fanquake this is a bugfix.

hebasto (approved):
ACK 241434200ec2067673d8522fee4f1228abfd8247, I have reviewed the code and it looks OK, I agree it can be merged.

nit: Could follow our style guide (at least apply `clang-format-diff.py`) for non-move-only commits?

ryanofsky (approved):
Code review ACK 241434200ec2067673d8522fee4f1228abfd8247. Only change is dropping one commit

jonasschnelli:
utACK 241434200ec2067673d8522fee4f1228abfd8247

PR #121, opened 2020-10-28
Early subscribe core signals in transaction table model

promag:
This fixes the case where transaction notifications arrive between `getWalletTxs` and `subscribeToCoreSignals`. Basically notifications are queued until `getWalletTxs` and wallet rescan complete.

This is also a requirement to call `getWalletTxs` in a background thread.

Motivated by https://github.com/bitcoin/bitcoin/issues/20241.

### DIFF
--- a/src/qt/transactiontablemodel.cpp
+++ b/src/qt/transactiontablemodel.cpp
@@ -735,22 +735,20 @@ void TransactionTablePriv::DispatchNotifications()
 {
     if (!m_loaded || m_loading) return;
 
+    if (vQueueNotifications.size() > 10) { // prevent balloon spam, show maximum 10 balloons
+        bool invoked = QMetaObject::invokeMethod(parent, "setProcessingQueuedTransactions", Qt::QueuedConnection, Q_ARG(bool, true));
+        assert(invoked);
+    }
+    for (unsigned int i = 0; i < vQueueNotifications.size(); ++i)
     {
-        if (vQueueNotifications.size() > 10) { // prevent balloon spam, show maximum 10 balloons
-            bool invoked = QMetaObject::invokeMethod(parent, "setProcessingQueuedTransactions", Qt::QueuedConnection, Q_ARG(bool, true));
+        if (vQueueNotifications.size() - i <= 10) {
+            bool invoked = QMetaObject::invokeMethod(parent, "setProcessingQueuedTransactions", Qt::QueuedConnection, Q_ARG(bool, false));
             assert(invoked);
         }
-        for (unsigned int i = 0; i < vQueueNotifications.size(); ++i)
-        {
-            if (vQueueNotifications.size() - i <= 10) {
-                bool invoked = QMetaObject::invokeMethod(parent, "setProcessingQueuedTransactions", Qt::QueuedConnection, Q_ARG(bool, false));
-                assert(invoked);
-            }
 
-            vQueueNotifications[i].invoke(parent);
-        }
-        vQueueNotifications.clear();
+        vQueueNotifications[i].invoke(parent);
     }
+    vQueueNotifications.clear();
 }
 
 void TransactionTableModel::subscribeToCoreSignals()
--- a/src/qt/transactiontablemodel.cpp
+++ b/src/qt/transactiontablemodel.cpp
@@ -96,13 +96,14 @@ public:
      */
     QList<TransactionRecord> cachedWallet;
 
-    bool fQueueNotifications = false;
     /** True when model finishes loading all wallet transactions on start */
     bool m_loaded = false;
+    /** True when transactions are being notified, for instance when scanning */
+    bool m_loading = false;
     std::vector< TransactionNotification > vQueueNotifications;
 
     void NotifyTransactionChanged(const uint256 &hash, ChangeType status);
-    void ShowProgress(const std::string &title, int nProgress);
+    void DispatchNotifications();
 
     /* Query entire wallet anew from core.
      */
@@ -721,7 +722,7 @@ void TransactionTablePriv::NotifyTransactionChanged(const uint256 &hash, ChangeT
 
     TransactionNotification notification(hash, status, showTransaction);
 
-    if (fQueueNotifications)
+    if (m_loading)
     {
         vQueueNotifications.push_back(notification);
         return;
@@ -729,14 +730,11 @@ void TransactionTablePriv::NotifyTransactionChanged(const uint256 &hash, ChangeT
     notification.invoke(parent);
 }
 
-void TransactionTablePriv::ShowProgress(const std::string &title, int nProgress)
+void TransactionTablePriv::DispatchNotifications()
 {
-    if (nProgress == 0)
-        fQueueNotifications = true;
+    if (m_loading) return;
 
-    if (nProgress == 100)
     {
-        fQueueNotifications = false;
         if (vQueueNotifications.size() > 10) { // prevent balloon spam, show maximum 10 balloons
             bool invoked = QMetaObject::invokeMethod(parent, "setProcessingQueuedTransactions", Qt::QueuedConnection, Q_ARG(bool, true));
             assert(invoked);
@@ -758,7 +756,10 @@ void TransactionTableModel::subscribeToCoreSignals()
 {
     // Connect signals to wallet
     m_handler_transaction_changed = walletModel->wallet().handleTransactionChanged(std::bind(&TransactionTablePriv::NotifyTransactionChanged, priv, std::placeholders::_1, std::placeholders::_2));
-    m_handler_show_progress = walletModel->wallet().handleShowProgress(std::bind(&TransactionTablePriv::ShowProgress, priv, std::placeholders::_1, std::placeholders::_2));
+    m_handler_show_progress = walletModel->wallet().handleShowProgress([this](const std::string&, int progress) {
+        priv->m_loading = progress < 100;
+        priv->DispatchNotifications();
+    });
 }
 
 void TransactionTableModel::unsubscribeFromCoreSignals()
--- a/src/qt/transactiontablemodel.cpp
+++ b/src/qt/transactiontablemodel.cpp
@@ -118,6 +118,7 @@ public:
             }
         }
         m_loaded = true;
+        DispatchNotifications();
     }
 
     /* Update our model of the wallet incrementally, to synchronize our model of the wallet
@@ -252,12 +253,12 @@ TransactionTableModel::TransactionTableModel(const PlatformStyle *_platformStyle
         fProcessingQueuedTransactions(false),
         platformStyle(_platformStyle)
 {
+    subscribeToCoreSignals();
+
     columns << QString() << QString() << tr("Date") << tr("Type") << tr("Label") << BitcoinUnits::getAmountColumnTitle(walletModel->getOptionsModel()->getDisplayUnit());
     priv->refreshWallet(walletModel->wallet());
 
     connect(walletModel->getOptionsModel(), &OptionsModel::displayUnitChanged, this, &TransactionTableModel::updateDisplayUnit);
-
-    subscribeToCoreSignals();
 }
 
 TransactionTableModel::~TransactionTableModel()
@@ -722,7 +723,7 @@ void TransactionTablePriv::NotifyTransactionChanged(const uint256 &hash, ChangeT
 
     TransactionNotification notification(hash, status, showTransaction);
 
-    if (m_loading)
+    if (!m_loaded || m_loading)
     {
         vQueueNotifications.push_back(notification);
         return;
@@ -732,7 +733,7 @@ void TransactionTablePriv::NotifyTransactionChanged(const uint256 &hash, ChangeT
 
 void TransactionTablePriv::DispatchNotifications()
 {
-    if (m_loading) return;
+    if (!m_loaded || m_loading) return;
 
     {
         if (vQueueNotifications.size() > 10) { // prevent balloon spam, show maximum 10 balloons
--- a/src/qt/transactiontablemodel.cpp
+++ b/src/qt/transactiontablemodel.cpp
@@ -97,6 +97,8 @@ public:
     QList<TransactionRecord> cachedWallet;
 
     bool fQueueNotifications = false;
+    /** True when model finishes loading all wallet transactions on start */
+    bool m_loaded = false;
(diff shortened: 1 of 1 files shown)


promag:
Please ignore whitespaces when reviewing.

promag:
Rebased due to conflict with updated #120.

hebasto:
Concept ACK.

hebasto:
> This fixes the case where transaction notifications arrive between `getWalletTxs` and `subscribeToCoreSignals`.

Yes, the current behavior is indeed buggy. It is easy to verify by patching `TransactionTableModel` ctor with a delay before `subscribeToCoreSignals` call.

src/qt/transactiontablemodel.cpp
```diff
@@ -95,25 +95,27 @@ class TransactionTablePriv
      */
     QList<TransactionRecord> cachedWallet;
 
-    bool fQueueNotifications = false;
+    bool m_loaded = false;
+    int m_progress = 100;
```
hebasto, line 99: It would be nice to add doxygen comments for new data members.

hebasto, line 99: Why have an `int` when we are only interested in `bool` if progress < 100?

src/qt/transactiontablemodel.cpp
```diff
@@ -95,25 +95,27 @@ class TransactionTablePriv
...
-    void ShowProgress(const std::string &title, int nProgress);
+    void DispatchNotifications();
 
     /* Query entire wallet anew from core.
      */
     void refreshWallet(interfaces::Wallet& wallet)
     {
-        qDebug() << "TransactionTablePriv::refreshWallet";
-        cachedWallet.clear();
+        assert(cachedWallet.empty());
```
hebasto, line 109: With new logic I'd prefer to `assert(!m_loaded);` or 
```suggestion
        assert(cachedWallet.empty() && !m_loaded);
```

src/qt/transactiontablemodel.cpp
```diff
@@ -95,25 +95,27 @@ class TransactionTablePriv
...
 
     void NotifyTransactionChanged(const uint256 &hash, ChangeType status);
-    void ShowProgress(const std::string &title, int nProgress);
+    void DispatchNotifications();
 
     /* Query entire wallet anew from core.
      */
     void refreshWallet(interfaces::Wallet& wallet)
     {
-        qDebug() << "TransactionTablePriv::refreshWallet";
```
hebasto, line 108: Why dropped?

hebasto (changes requested):
Approach ACK 5d7d36344c1241fb940766905265747492570446.

ryanofsky (approved):
Code review ACK 5d7d36344c1241fb940766905265747492570446. This looks good but is a pretty confusing change. I think it could be clearer split up into two commits: one pure refactoring commit replacing `ShowProgress` with `DispatchNotifications`, and another commit changing initialization order moving the `subscribeToCoreSignals` call before the `refreshWallet` call and adding the new `DispatchNotifications` call at the end of the `refreshWallet` implementation.

I will say I don't fully understand the motivation for the two `refreshMethod` related changes. They seem safe but I don't know if they are actually changing behavior now or just setting something up for the future. If they were moved to standalone commit, purpose of the changes could be made more clear either in the commit description or code comments.

In case it helps other reviewers, I found the refactoring changes replacing `ShowProgress` with `DispatchNotifications` easier to understand after ignoring whitespace and rereading a previous explanation of the code https://github.com/bitcoin-core/gui/pull/120#discussion_r513658333.

Hebasto's suggestions also all seem reasonable and good. Basically a lot could be made clearer here, but this PR does seems fine to merge in its current form.

This PR might be interesting to people dealing with large wallets that have a lot of transactions, since it seems like it could help move to loading transactions asynchronously without locking up the GUI.

src/qt/transactiontablemodel.cpp
promag: What is the goal of this log? To know when `refreshWallet` is called? Seems one of those logs you add when developing, but now it's useless. I'll revert this change if you think otherwise. It's unrelated to this PR but committed since I was touching nearby code.

promag (commented):
Thanks for reviewing, I'll improve readability.

src/qt/transactiontablemodel.cpp
hebasto: > To know when `refreshWallet` is called?

Rather to know that `refreshWallet` _is_ called, for example for old wallets that were not synced for a long time, no?

promag: But it's called unconditionally in `TransactionTableModel` constructor.

hebasto: > But it's called unconditionally in `TransactionTableModel` constructor.

Right. But it is an implementation detail that (as we assume) is hidden from users.

Anyway, feel free to ignore this concern and mark it as resolved :)

laanwj:
> This is also a requirement to call getWalletTxs in a background thread.

Concept ACK, thanks for working on that.

hebasto (approved):
ACK 5d7d36344c1241fb940766905265747492570446, agree with https://github.com/bitcoin-core/gui/pull/121#pullrequestreview-575666547
> this PR does seems fine to merge in its current form.

src/qt/transactiontablemodel.cpp
promag: > But it is an implementation detail that (as we assume) is hidden from users.

Right, so why log it? Following the same idea there are many things we could also log.

src/qt/transactiontablemodel.cpp
promag: Added.

promag: Changed.

promag (commented):
Rebased and addressed @hebasto review.

src/qt/transactiontablemodel.cpp
```diff
@@ -96,25 +96,29 @@ class TransactionTablePriv
      */
     QList<TransactionRecord> cachedWallet;
 
-    bool fQueueNotifications = false;
+    /** True when model finishes loading all wallet transaction on start */
```
jonatack, line 99: nit, s/transaction/transactions/

src/qt/transactiontablemodel.cpp
```diff
@@ -96,25 +96,29 @@ class TransactionTablePriv
...
+    bool m_loaded = false;
+    /** True when transactions are being notified, for instance when scanning */
+    int m_loading = false;
```
jonatack, line 102: (maybe a leftover from an initial version?) this can work but unless I'm misreading `m_loading` will only be set to true or false (1 or 0)
```suggestion
    bool m_loading = false;
```

jonatack (commented):
Approach ACK modulo that it's a fairly confusing change and separating into more commits may help here (as mentioned above).

src/qt/transactiontablemodel.cpp
promag: Fixed.

src/qt/transactiontablemodel.cpp
promag: Fixed.

promag:
Ok, I've tried to split in decent commits.

jonatack:
Per `git range-diff 16d0051c 8f16516 cafef08`, the code is indeed the same. Having a look.
```
1:  8f1651678 < -:  --------- qt: Early subscribe core signals in transaction table model
-:  --------- > 1:  3bccd50ad qt: Set flag after inital load on transaction table model
-:  --------- > 2:  c6cbdf1a9 qt: Refactor ShowProgress to DispatchNotifications
-:  --------- > 3:  57785fb7f qt: Early subscribe core signals in transaction table model
-:  --------- > 4:  cafef080a qt: Refactor to remove unnecessary block in DispatchNotifications
```

src/qt/transactiontablemodel.cpp
```diff
@@ -96,25 +96,29 @@ class TransactionTablePriv
...
-        cachedWallet.clear();
+        assert(!m_loaded);
         {
             for (const auto& wtx : wallet.getWalletTxs()) {
                 if (TransactionRecord::showTransaction()) {
                     cachedWallet.append(TransactionRecord::decomposeTransaction(wtx));
                 }
             }
         }
+        m_loaded = true;
```
jonatack, line 120: 3bccd50ad2f38 nit, can drop extra brackets
```diff
diff --git a/src/qt/transactiontablemodel.cpp b/src/qt/transactiontablemodel.cpp
index 0175c88e7..8dc6d4c40 100644
--- a/src/qt/transactiontablemodel.cpp
+++ b/src/qt/transactiontablemodel.cpp
@@ -110,11 +110,9 @@ public:
     void refreshWallet(interfaces::Wallet& wallet)
     {
         assert(!m_loaded);
-        {
-            for (const auto& wtx : wallet.getWalletTxs()) {
-                if (TransactionRecord::showTransaction()) {
-                    cachedWallet.append(TransactionRecord::decomposeTransaction(wtx));
-                }
+        for (const auto& wtx : wallet.getWalletTxs()) {
+            if (TransactionRecord::showTransaction()) {
+                cachedWallet.append(TransactionRecord::decomposeTransaction(wtx));
             }
         }
         m_loaded = true;
```

jonatack: (similar to the change in the last commit, ` git show cafef080a2e --ignore-all-space`)

src/qt/transactiontablemodel.cpp
```diff
@@ -719,44 +723,42 @@ void TransactionTablePriv::NotifyTransactionChanged(const uint256 &hash, ChangeT
...
-        fQueueNotifications = true;
+    if (!m_loaded || m_loading) return;
 
-    if (nProgress == 100)
+    if (vQueueNotifications.size() > 10) { // prevent balloon spam, show maximum 10 balloons
+        bool invoked = QMetaObject::invokeMethod(parent, "setProcessingQueuedTransactions", Qt::QueuedConnection, Q_ARG(bool, true));
+        assert(invoked);
+    }
+    for (unsigned int i = 0; i < vQueueNotifications.size(); ++i)
     {
```
jonatack, line 743: cafef080a2e59c2 nit, while removing the brackets in this commit could also
```diff
diff --git a/src/qt/transactiontablemodel.cpp b/src/qt/transactiontablemodel.cpp
index 0175c88e7..734605a47 100644
--- a/src/qt/transactiontablemodel.cpp
+++ b/src/qt/transactiontablemodel.cpp
@@ -739,8 +739,7 @@ void TransactionTablePriv::DispatchNotifications()
         bool invoked = QMetaObject::invokeMethod(parent, "setProcessingQueuedTransactions", Qt::QueuedConnection, Q_ARG(bool, true));
         assert(invoked);
     }
-    for (unsigned int i = 0; i < vQueueNotifications.size(); ++i)
-    {
+    for (unsigned int i = 0; i < vQueueNotifications.size(); ++i) {
```

jonatack (commented):
tACK cafef080a2e59c2bcae6baeee3c3c7e187e931ae

happy to re-ACK for the above

src/qt/transactiontablemodel.cpp
```diff
@@ -96,25 +96,29 @@ class TransactionTablePriv
      */
     QList<TransactionRecord> cachedWallet;
 
-    bool fQueueNotifications = false;
+    /** True when model finishes loading all wallet transactions on start */
+    bool m_loaded = false;
+    /** True when transactions are being notified, for instance when scanning */
```
hebasto, line 101: It could be just my bad English, but it seems should be "... are _not_ being notified", no?

If I'm correct, maybe re-word the comment in positive one?

Or maybe we think about different notifications: (a) from the node vs (b) to a user

src/qt/transactiontablemodel.cpp
```diff
@@ -96,25 +96,29 @@ class TransactionTablePriv
...
+    bool m_loading = false;
     std::vector< TransactionNotification > vQueueNotifications;
 
     void NotifyTransactionChanged(const uint256 &hash, ChangeType status);
-    void ShowProgress(const std::string &title, int nProgress);
+    void DispatchNotifications();
```
hebasto, line 106: A doxygen comment?

src/qt/transactiontablemodel.cpp
```diff
@@ -719,44 +723,42 @@ void TransactionTablePriv::NotifyTransactionChanged(const uint256 &hash, ChangeT
 
     TransactionNotification notification(hash, status, showTransaction);
 
-    if (fQueueNotifications)
+    if (!m_loaded || m_loading)
     {
```
hebasto, line 727: style nit:
```suggestion
    if (!m_loaded || m_loading) {
```

hebasto (commented):
Thanks for splitting commits! Much easier to review.

src/qt/transactiontablemodel.cpp
promag: Right, I'll rephrase.

ryanofsky: In commit "qt: Refactor ShowProgress to DispatchNotifications" (c6cbdf1a90a253fef0259b365a782bf88cd437f2)

re: https://github.com/bitcoin-core/gui/pull/121#discussion_r623855013

> Right, I'll rephrase.

If you're still planning to update this, my attempt at a comment would be: "This will be true when blocks in the chain are being scanned for new transactions, and transaction notification popups should be suppressed until the scan is done. It can be true even after m_loaded is true if a rescan has been triggered." I think I would also call this `m_delay_notifications` instead of `m_loading` because it can be true when `m_loaded` is also true, but feel free to stick with whatever name you prefer.

ryanofsky (approved):
Code review ACK cafef080a2e59c2bcae6baeee3c3c7e187e931ae. Only change since last review is splitting commits and replacing m_progress with m_loading.

Thanks for the updates! I think this change is much easier to understand now.

It would be good to clarify in the PR description what exactly is meant by "This fixes the case where transaction notifications arrive between `getWalletTxs` and `subscribeToCoreSignals`." It's not clear from the description what the old behavior was and what the new behavior is.

IIUC, the problem was that notifications about new transactions discovered while the wallet was being loaded were dropped previously, and now they are shown?

meshcollider (commented):
Code review ACK cafef080a2e59c2bcae6baeee3c3c7e187e931ae

PR #123, opened 2020-10-28
rpc: Do not accept command while executing another one

hebasto:
On master (3f512f3d563954547061ee743648b57a900cbe04) it is possible to enter another command while the current command is still being executed. That makes a mess in the output.

With this PR:
Screenshot from 2020-10-29 20-48-55

Some previous context: https://github.com/bitcoin-core/gui/pull/59#issuecomment-715275185

---

It is still possible to enter and execute the `stop` command any time.

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -958,6 +958,8 @@ void RPCConsole::on_lineEdit_returnPressed()
 #endif
 
         message(CMD_REQUEST, QString::fromStdString(strFilteredCmd));
+        //: A console message indicating an entered command is currently being executed.
+        message(CMD_REPLY, tr("Executing…"));
         Q_EMIT cmdRequest(cmd, m_last_wallet_model);
 
         cmd = QString::fromStdString(strFilteredCmd);
@@ -1004,7 +1006,12 @@ void RPCConsole::startExecutor()
     executor->moveToThread(&thread);
 
     // Replies from executor object must go to this object
-    connect(executor, &RPCExecutor::reply, this, qOverload<int, const QString&>(&RPCConsole::message));
+    connect(executor, &RPCExecutor::reply, this, [this](int category, const QString& command) {
+        // Remove "Executing…" message.
+        ui->messagesWidget->undo();
+        message(category, command);
+        scrollToEnd();
+    });
 
     // Requests from this object must go to executor
     connect(this, &RPCConsole::cmdRequest, executor, &RPCExecutor::request);
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -926,57 +926,57 @@ void RPCConsole::on_lineEdit_returnPressed()
 {
     QString cmd = ui->lineEdit->text();
 
-    if(!cmd.isEmpty())
-    {
-        std::string strFilteredCmd;
-        try {
-            std::string dummy;
-            if (!RPCParseCommandLine(nullptr, dummy, cmd.toStdString(), false, &strFilteredCmd)) {
-                // Failed to parse command, so we cannot even filter it for the history
-                throw std::runtime_error("Invalid command line");
-            }
-        } catch (const std::exception& e) {
-            QMessageBox::critical(this, "Error", QString("Error: ") + QString::fromStdString(e.what()));
-            return;
+    if (cmd.isEmpty()) {
+        return;
+    }
+
+    std::string strFilteredCmd;
+    try {
+        std::string dummy;
+        if (!RPCParseCommandLine(nullptr, dummy, cmd.toStdString(), false, &strFilteredCmd)) {
+            // Failed to parse command, so we cannot even filter it for the history
+            throw std::runtime_error("Invalid command line");
         }
+    } catch (const std::exception& e) {
+        QMessageBox::critical(this, "Error", QString("Error: ") + QString::fromStdString(e.what()));
+        return;
+    }
 
-        ui->lineEdit->clear();
-
-        cmdBeforeBrowsing = QString();
+    ui->lineEdit->clear();
 
 #ifdef ENABLE_WALLET
-        WalletModel* wallet_model = ui->WalletSelector->currentData().value<WalletModel*>();
+    WalletModel* wallet_model = ui->WalletSelector->currentData().value<WalletModel*>();
 
-        if (m_last_wallet_model != wallet_model) {
-            if (wallet_model) {
-                message(CMD_REQUEST, tr("Executing command using \"%1\" wallet").arg(wallet_model->getWalletName()));
-            } else {
-                message(CMD_REQUEST, tr("Executing command without any wallet"));
-            }
-            m_last_wallet_model = wallet_model;
+    if (m_last_wallet_model != wallet_model) {
+        if (wallet_model) {
+            message(CMD_REQUEST, tr("Executing command using \"%1\" wallet").arg(wallet_model->getWalletName()));
+        } else {
+            message(CMD_REQUEST, tr("Executing command without any wallet"));
         }
-#endif
-
-        message(CMD_REQUEST, QString::fromStdString(strFilteredCmd));
-        //: A console message indicating an entered command is currently being executed.
-        message(CMD_REPLY, tr("Executing…"));
-        Q_EMIT cmdRequest(cmd, m_last_wallet_model);
-
-        cmd = QString::fromStdString(strFilteredCmd);
-
-        // Remove command, if already in history
-        history.removeOne(cmd);
-        // Append command to history
-        history.append(cmd);
-        // Enforce maximum history size
-        while(history.size() > CONSOLE_HISTORY)
-            history.removeFirst();
-        // Set pointer to end of history
-        historyPtr = history.size();
-
-        // Scroll console view to end
-        scrollToEnd();
+        m_last_wallet_model = wallet_model;
     }
+#endif // ENABLE_WALLET
+
+    message(CMD_REQUEST, QString::fromStdString(strFilteredCmd));
+    //: A console message indicating an entered command is currently being executed.
+    message(CMD_REPLY, tr("Executing…"));
+    Q_EMIT cmdRequest(cmd, m_last_wallet_model);
+
+    cmd = QString::fromStdString(strFilteredCmd);
+
+    // Remove command, if already in history
+    history.removeOne(cmd);
+    // Append command to history
+    history.append(cmd);
+    // Enforce maximum history size
+    while (history.size() > CONSOLE_HISTORY) {
+        history.removeFirst();
+    }
+    // Set pointer to end of history
+    historyPtr = history.size();
+
+    // Scroll console view to end
+    scrollToEnd();
 }
 
 void RPCConsole::browseHistory(int offset)
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -924,7 +924,7 @@ void RPCConsole::setMempoolSize(long numberOfTxs, size_t dynUsage)
 
 void RPCConsole::on_lineEdit_returnPressed()
 {
-    QString cmd = ui->lineEdit->text();
+    QString cmd = ui->lineEdit->text().trimmed();
 
     if (cmd.isEmpty()) {
         return;
@@ -939,6 +939,13 @@ void RPCConsole::on_lineEdit_returnPressed()
         }
     } catch (const std::exception& e) {
         QMessageBox::critical(this, "Error", QString("Error: ") + QString::fromStdString(e.what()));
+        return;
+    }
+
+    // A special case allows to request shutdown even a long-running command is executed.
+    if (cmd == QLatin1String("stop")) {
+        std::string dummy;
+        RPCExecuteCommandLine(m_node, dummy, cmd.toStdString());
         return;
     }
 
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -949,6 +949,10 @@ void RPCConsole::on_lineEdit_returnPressed()
         return;
     }
 
+    if (m_is_executing) {
+        return;
+    }
+
     ui->lineEdit->clear();
 
 #ifdef ENABLE_WALLET
@@ -967,6 +971,7 @@ void RPCConsole::on_lineEdit_returnPressed()
     message(CMD_REQUEST, QString::fromStdString(strFilteredCmd));
     //: A console message indicating an entered command is currently being executed.
     message(CMD_REPLY, tr("Executing…"));
+    m_is_executing = true;
     Q_EMIT cmdRequest(cmd, m_last_wallet_model);
 
     cmd = QString::fromStdString(strFilteredCmd);
@@ -1018,6 +1023,7 @@ void RPCConsole::startExecutor()
         ui->messagesWidget->undo();
         message(category, command);
         scrollToEnd();
+        m_is_executing = false;
     });
 
     // Requests from this object must go to executor
--- a/src/qt/rpcconsole.h
+++ b/src/qt/rpcconsole.h
@@ -166,6 +166,7 @@ private:
     QCompleter *autoCompleter = nullptr;
     QThread thread;
     WalletModel* m_last_wallet_model{nullptr};
+    bool m_is_executing{false};
 
     /** Update UI with latest network info from model. */
     void updateNetworkState();
--- a/src/qt/test/apptests.cpp
+++ b/src/qt/test/apptests.cpp
@@ -40,7 +40,7 @@ void TestRpcCommand(RPCConsole* console)
     QTest::keyClicks(lineEdit, "getblockchaininfo");
     QTest::keyClick(lineEdit, Qt::Key_Return);
     QVERIFY(mw_spy.wait(1000));
-    QCOMPARE(mw_spy.count(), 2);
+    QCOMPARE(mw_spy.count(), 4);
     QString output = messagesWidget->toPlainText();
     UniValue value;
     value.read(output.right(output.size() - output.lastIndexOf(QChar::ObjectReplacementCharacter) - 1).toStdString());


jonasschnelli:
Tested ACK 0b8b65392a590dc0ac04818b4933676e3f2c5a7c - works as intended.

A follow up would be to add a cancel/abort button. Especially for things like "waitforblock 99999999" (would even be useful for `gettxoutsetinfo`).

hebasto:
> A follow up would be to add a cancel/abort button. Especially for things like "waitforblock 99999999" (would even be useful for `gettxoutsetinfo`).

Combine cancel/abort functionality into already present Clear button?

src/qt/rpcconsole.cpp
```diff
@@ -890,8 +888,10 @@ void RPCConsole::on_lineEdit_returnPressed()
 {
     QString cmd = ui->lineEdit->text();
 
-    if(!cmd.isEmpty())
-    {
+    if (!cmd.isEmpty()) {
+        ui->lineEdit->setEnabled(false);
```
promag, line 892: I think it's nice to be able to write the next command.

promag (commented):
One way to abort running command is to run ~~`quit`~~`stop`. With this change this is not possible.

src/qt/rpcconsole.cpp
jonasschnelli: Good point. What about allowing input but disallow executing (`enter` won't clear the line, etc.). The `executing...` text would need to be somewhere else though...

promag: @jonasschnelli and this https://github.com/bitcoin-core/gui/pull/123#pullrequestreview-519587704?

hebasto:
> One way to abort running command is to run `quit`.

When the `quit` command became a thing in the GUI console?

promag:
I just mean it won't work anymore. I happen to use it, but can quit from the GUI.

hebasto:
> I just mean it won't work anymore.

I'm going to address your concerns.

> I happen to use it, but can quit from the GUI.

I cannot find the `quit` command now.

promag:
s/quit/stop 🤦

jonasschnelli:
@promag 
> One way to abort running command is to run `stop`. With this change this is not possible.

This would shutdown bitcoin-qt. Right?
I don't think this is an adequate action for stopping a single RPC call. It's already possible with Ctrl-Q.

Why not just emulate the shell (bitcoin-cli usage) as good as possible?
* No parallel commands are possible in a single shell window
* Ctrl-C abort the call
* Clear visual sign that something is "working.." (no prompt visible)
* Allow entering a next command (but impossible to execute before old has been finished)

This would translate that into our Qt world with:
* Once a command has been started, forbid executing a new one until the old has "returned" (done in this PR)
* Have a specific "abort" button or support something like Ctrl-C if it looks silly (**should be added**)
* * If the "abort" button looks bad (quick show/hide) for commands taking only a few milliseconds, maybe only show the abort button after 1s of execution time
* Show that a command is executed (done in this PR, but maybe should be done differently to allow entering a next command).
* Allow entering a followup command but can't execute (**should be added**)

hebasto:
@jonasschnelli Spotted your comment too late, so, probably, the last commit in the latest push should be dropped, right?

src/qt/rpcconsole.cpp
hebasto: Updated.

src/qt/rpcconsole.cpp
```diff
@@ -892,53 +892,53 @@ void RPCConsole::on_lineEdit_returnPressed()
 {
     QString cmd = ui->lineEdit->text();
```
promag, line 893: nit, could trim.

promag (commented):
Another sensible approach would be to disable the line edit while keep the 2nd command there and execute it when the 1st ends.

hebasto:
> Another sensible approach would be to disable the line edit while keep the 2nd command there and execute it when the 1st ends.

This deprives the user of the ability to change his mind when the 1st command output is arrived.

src/qt/rpcconsole.cpp
```diff
@@ -834,12 +834,14 @@ void RPCConsole::message(int category, const QString &message, bool html)
...
-    if(html)
+    if (html) {
         out += message;
-    else
+    } else {
         out += GUIUtil::HtmlEscape(message, false);
+    }
     out += "</td></tr></table>";
     ui->messagesWidget->append(out);
+    scrollToEnd();
```
luke-jr, line 844: If the user has scrolled up, we shouldn't ignore that and reset scroll to the bottom...

luke-jr (changes requested):

hebasto:
Updated 980b8e968459f6fa7e16e5c79eaad782da9e89c0 -> e8753bf07bc2cb9c4023e771f7e10e65ba7571dd ([pr123.02](https://github.com/hebasto/gui/commits/pr123.02) -> [pr123.03](https://github.com/hebasto/gui/commits/pr123.03), [diff](https://github.com/hebasto/gui/compare/pr123.02..pr123.03)):

- addressed @promag's [comment](https://github.com/bitcoin-core/gui/pull/123#discussion_r517197531):
> nit, could trim.

- addressed @luke-jr's [comment](https://github.com/bitcoin-core/gui/pull/123#discussion_r523164157):
> If the user has scrolled up, we shouldn't ignore that and reset scroll to the bottom...

src/qt/rpcconsole.cpp
hebasto: [Updated](https://github.com/bitcoin-core/gui/pull/123#issuecomment-727557347).

src/qt/rpcconsole.cpp
hebasto: [Updated](https://github.com/bitcoin-core/gui/pull/123#issuecomment-727557347).

jarolrod (commented):
ACK 08d599e4f3b2e3a1143863d3cf1e64a8215a59b1, Tested on macOS 11.1 with Qt 5.15.2

**Master:** Allowed to input several commands at once which will lead to messy output
Screen Shot 2021-01-19 at 3 57 30 PM

**PR:** Can only execute one command at a time
Screen Shot 2021-01-19 at 4 22 56 PM

**Additional Thoughts (for a followup pr)**
As others have stated, it would be nice to have a cancel button on the command. This should be done in a follow up PR. An example of when this is useful:
- I want to enter the `gettxout` command
- I mistakingly autocomplete to the `gettxoutsetinfo` which takes some time to run
- It would be nice to be able to cancel the `gettxoutsetinfo` command ran by accident, and not have to wait until it is finished to accomplish what I set out to accomplish -> run `gettxout`

Perhaps something like this: (rough mockup)
cancel-mockup

src/qt/rpcconsole.cpp
```diff
@@ -888,57 +888,69 @@ void RPCConsole::setMempoolSize(long numberOfTxs, size_t dynUsage)
...
+        if (!RPCParseCommandLine(nullptr, dummy, cmd.toStdString(), false, &strFilteredCmd)) {
+            // Failed to parse command, so we cannot even filter it for the history
+            throw std::runtime_error("Invalid command line");
         }
+    } catch (const std::exception& e) {
+        QMessageBox::critical(this, "Error", QString("Error: ") + QString::fromStdString(e.what()));
+        return;
+    }
+
+    if (cmd.startsWith("stop")) {
```
Talkless, line 909: nit: `QString::startsWith()` has overload taking ` QLatin1String` which is recommended to use if exists. It will not allocate new `QString`, and comparing to latin1 instead of utf-16 is cheaper memory-bandwith-wise.

Also, maybe `startsWith` is not needed at all as the `cmd` string was trimmed beforehand? i.e. `cmd == QLatin1String("stop")` would suffice?

src/qt/rpcconsole.cpp
```diff
@@ -888,57 +888,69 @@ void RPCConsole::setMempoolSize(long numberOfTxs, size_t dynUsage)
...
-
-        message(CMD_REQUEST, QString::fromStdString(strFilteredCmd));
-        Q_EMIT cmdRequest(cmd, m_last_wallet_model);
+        m_last_wallet_model = wallet_model;
+    }
+#endif // ENABLE_WALLET
 
-        cmd = QString::fromStdString(strFilteredCmd);
+    message(CMD_REQUEST, QString::fromStdString(strFilteredCmd));
+    message(CMD_REPLY, "Executing...");
```
Talkless, line 935: nit: `QStringLiteral("Executing...")`

Talkless (approved):
tACK 08d599e4f, tested on Debian Sid with Qt 5.15.2.

I've generated bunch of transactions on regtest using `generatetoaddress` and later used `listunspent` as long-running command to reproduce issue on master, invoking multiple commands while first one is still running, and I see that PR code works as expecting afterwards.

hebasto:
Updated 08d599e4f3b2e3a1143863d3cf1e64a8215a59b1 -> b87a4cd6bb58a6458dea7527a19433ead0ea848d ([pr123.04](https://github.com/hebasto/gui/commits/pr123.04) -> [pr123.05](https://github.com/hebasto/gui/commits/pr123.05), [diff](https://github.com/hebasto/gui/compare/pr123.04..pr123.05)):

- addressed @Talkless' [comment](https://github.com/bitcoin-core/gui/pull/123#discussion_r597168009)

- translated user-faced string

src/qt/rpcconsole.cpp
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui/pull/123#issuecomment-803235476).

src/qt/rpcconsole.cpp
hebasto: This string is translated [now](https://github.com/bitcoin-core/gui/pull/123#issuecomment-803235476).

Talkless (approved):
re-ACK b87a4cd6bb58a6458dea7527a19433ead0ea848d after tweaks.

leonardojobim (commented):
Tested https://github.com/bitcoin-core/gui/pull/123/commits/b87a4cd6bb58a6458dea7527a19433ead0ea848d on Ubuntu 20.04 Qt 5.12.8.

When executing commands enough to fill the entire result dialog, instead of scrolling down, it scrolls up to the beginning ("Welcome to the Bitcoin Core RPC console." line).
On the master branch, the scrolling apparently works OK.

src/qt/rpcconsole.cpp
```diff
@@ -906,6 +906,16 @@ void RPCConsole::on_lineEdit_returnPressed()
         return;
     }
 
+    if (cmd == QLatin1String("stop")) {
```
promag, line 909: b87a4cd6bb58a6458dea7527a19433ead0ea848d

Could add a comment explaining this case.

promag (commented):
Tested ACK 328efdcb69eedbd3bce6d6de865cca0e3ee38374.

if new output is appended and its not visible (user scrolled) then it could show some feedback/alert so user knows command finished and can scroll down.

nit, could make "qt, rpc: Accept stop RPC even another command is executing" before "qt, rpc: Do not accept command while executing another one" to avoid adding `|| m_is_executing` and then removing.

hebasto:
@leonardojobim @Talkless @jarolrod @luke-jr @promag @jonasschnelli 

Thanks you for your reviewing and testing!

Updated b87a4cd6bb58a6458dea7527a19433ead0ea848d -> 716139584ddf7ef554962e0188ae8b8a74e310e4 ([pr123.05](https://github.com/hebasto/gui/commits/pr123.05) -> [pr123.06](https://github.com/hebasto/gui/commits/pr123.06), [diff](https://github.com/hebasto/gui/compare/pr123.05..pr123.06)):

- reordered commits (@promag)
- added comment for special case for `stop` command (@promag)
- fixed scrolling (@leonardojobim)
- added translator comment

src/qt/rpcconsole.cpp
hebasto: [Updated](https://github.com/bitcoin-core/gui/pull/123#issuecomment-831538996).

Talkless:
re-tACK 716139584ddf7ef554962e0188ae8b8a74e310e4, tested on same setup as previously.

Talkless:
re-utACK 716139584ddf7ef554962e0188ae8b8a74e310e4, range-diff does not show anything to actually review or test.

Talkless:
Wait @hebasto, a bit off topic, should I have had to make actual GitHub review in this interface, or comment is enough? Maybe sticking to _always_ checking all files and producing GitHub review would be "better"?

hebasto:
@Talkless 
> Wait @hebasto, a bit off topic, should I have had to make actual GitHub review in this interface, or comment is enough? Maybe sticking to _always_ checking all files and producing GitHub review would be "better"?

Not sure if I understand your question in part "producing GitHub review"...

Anyway, your https://github.com/bitcoin-core/gui/pull/123#issuecomment-840720795 is straightforward. While reviewing pulls after rebasing I do the same locally `git range-diff master pr-before-rebase pr-after-rebase`. If rebasing is not trivial it could be valuable re-check all changes.

Talkless:
> Not sure if I understand your question in part "producing GitHub review"...

In Github web interface, we can go to "Files changed" tab, check "Viewed" checkboxes on the files, and click "Review changes" -> "Submit review", which produces sort of "special" kind of "comment". Maybe it's "better" for maintainers to check for these kind of "reviews", instead of "just" a comment?

hebasto:
> > Not sure if I understand your question in part "producing GitHub review"...
> 
> In Github web interface, we can go to "Files changed" tab, check "Viewed" checkboxes on the files, and click "Review changes" -> "Submit review", which produces sort of "special" kind of "comment". Maybe it's "better" for maintainers to check for these kind of "reviews", instead of "just" a comment?

In this project we strive to rely on GitHub as little as possible :)
Ofc, a reviewer could use any GitHub feature he like. Features you mentioned make no difference for maintainers.

hebasto:
@Talkless 

In https://github.com/bitcoin-core/gui/pull/123#issuecomment-840720795
> re-utACK [7161395](https://github.com/bitcoin-core/gui/commit/716139584ddf7ef554962e0188ae8b8a74e310e4), range-diff does not show anything to actually review or test.

Did you mean the old top 716139584ddf7ef554962e0188ae8b8a74e310e4 or the new cb48c5964c901bccf5d2ffcd9f3f3e4552ae2fb1?

Talkless:
> @Talkless
> 
> In [#123 (comment)](https://github.com/bitcoin-core/gui/pull/123#issuecomment-840720795)
> 
> > re-utACK [7161395](https://github.com/bitcoin-core/gui/commit/716139584ddf7ef554962e0188ae8b8a74e310e4), range-diff does not show anything to actually review or test.
> 
> Did you mean the old top [7161395](https://github.com/bitcoin-core/gui/commit/716139584ddf7ef554962e0188ae8b8a74e310e4) or the new [cb48c59](https://github.com/bitcoin-core/gui/commit/cb48c5964c901bccf5d2ffcd9f3f3e4552ae2fb1)?

Ugh, not again... I did range-diff for review, but forgot to actually swtich to new tip before copying latest commit hash..

re-ACK cb48c5964c901bccf5d2ffcd9f3f3e4552ae2fb1

hebasto:
@jarolrod @promag @leonardojobim 

Do you mind having another (final?) look into at this PR?

promag (commented):
Tested ACK cb48c5964c901bccf5d2ffcd9f3f3e4552ae2fb1.

src/qt/rpcconsole.cpp
```diff
@@ -897,57 +897,71 @@ void RPCConsole::setMempoolSize(long numberOfTxs, size_t dynUsage)
...
-        // Set pointer to end of history
-        historyPtr = history.size();
+        m_last_wallet_model = wallet_model;
+    }
+#endif // ENABLE_WALLET
 
-        // Scroll console view to end
-        scrollToEnd();
+    message(CMD_REQUEST, QString::fromStdString(strFilteredCmd));
+    //: A message in the GUI console while an entered command being executed.
```
jarolrod, line 945: nit
```suggestion
    //: A console message indicating an entered command is currently being executed.
```

src/qt/rpcconsole.cpp
```diff
@@ -897,57 +897,71 @@ void RPCConsole::setMempoolSize(long numberOfTxs, size_t dynUsage)
 
 void RPCConsole::on_lineEdit_returnPressed()
 {
-    QString cmd = ui->lineEdit->text();
+    QString cmd = ui->lineEdit->text().trimmed();
```
jarolrod, line 900: nit

we should document that we are using this function in the commit message for proper documentation

jarolrod (commented):
ACK cb48c5964c901bccf5d2ffcd9f3f3e4552ae2fb1

Tested functionality on `Arch Linux Qt 5.15.2` and `macOS 11.3 Qt 5.15.2`. Tested each commit individually, cherry-picking each on top of master. This is a nice patch. A couple of nits, no need to address unless you *absolutely* have to retouch. 🥃

**Notes on commits:**
- [x] 9d9f7f1158323e77d6c7ec59f1ca375d6d87a9ea
  - Tested that "Executing..." message is displayed and removed upon response
- [x] 614cc38fa94567c90b567514cc5db785be658841
  - Concept ACK on refactoring the logic here to return early if the command is empty
- [x] d4e7fcc2a8b36e7bf1450edeb67bb468aac8d602
  - Tested that we can call `stop`, although true testing of this commit can only be done with the next commit applied. I think we should have kept this commit after `qt, rpc: Do not accept command while executing another one`
  - ACK on using the [trimmed](https://doc.qt.io/qt-5/qstring.html#trimmed) Qstring function % [nit](https://github.com/bitcoin-core/gui/pull/123#discussion_r642592794)
- [x] cb48c5964c901bccf5d2ffcd9f3f3e4552ae2fb1
  - Tested that we *cannot* input another command while another is being executed
  - Tested that we *can* still call the stop command
    - It's annoying that this command just quits the application. It would be nice to change this in a followup.

**Screenshots:**
Below are screenshots showing the behavior of master vs pr. Note that one can continue to input commands on master, but cannot with this pr:

| Master    | Description |
| ----------- | --------------- |
| master | pr |

src/qt/rpcconsole.cpp
hebasto: This could be addressed in a follow up.

src/qt/rpcconsole.cpp
hebasto: This could be addressed in a follow up.

hebasto:
Updated cb48c5964c901bccf5d2ffcd9f3f3e4552ae2fb1 -> 38eb37c0bd29b4cb825de905e8eec87636a5221b ([pr123.07](https://github.com/hebasto/gui/commits/pr123.07) -> [pr123.08](https://github.com/hebasto/gui/commits/pr123.08)):

- rebased to fix silent merge conflict with #335 
- addressed the recent @jarolrod's comments

src/qt/rpcconsole.cpp
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui/pull/123#issuecomment-851674465).

src/qt/rpcconsole.cpp
hebasto: [Updated](https://github.com/bitcoin-core/gui/pull/123#issuecomment-851674465).

jarolrod:
ACK  38eb37c

re-tested functionality. Ran tests to confirm this fixes the silent merge conflict with #335.

Rebased `pr123.07` on top of the commit `pr123.08` is based on, 933c6466c2bda3a06d3c5da0de0d8f05111c9f4c,  and got the following diff which shows the only changes from my last review are; changes to [translator comment](https://github.com/bitcoin-core/gui/pull/123/files#diff-cec3a0a10196329610d46fbd0e01ce6496e5ae9e23c057d8012d48c2f0d3b5c9R972) to address my [review comment](https://github.com/bitcoin-core/gui/pull/123#discussion_r642588665) and [update](https://github.com/bitcoin-core/gui/pull/123/files#diff-41e328d9ba5879d2bd026a700743339b5b300b9bee50844795ef9e27f60bdfb8R43) to apptests to reflect the expected signal count:

``` diff
$ git diff pr123.07 pr123.08
index 803946dcf..c762f2b9c 100644
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -969,7 +969,7 @@ void RPCConsole::on_lineEdit_returnPressed()
 #endif // ENABLE_WALLET
 
     message(CMD_REQUEST, QString::fromStdString(strFilteredCmd));
-    //: A message in the GUI console while an entered command being executed.
+    //: A console message indicating an entered command is currently being executed.
     message(CMD_REPLY, tr("Executing…"));
     m_is_executing = true;
     Q_EMIT cmdRequest(cmd, m_last_wallet_model);
diff --git a/src/qt/test/apptests.cpp b/src/qt/test/apptests.cpp
index c1d5f84be..cb3dbd226 100644
--- a/src/qt/test/apptests.cpp
+++ b/src/qt/test/apptests.cpp
@@ -40,7 +40,7 @@ void TestRpcCommand(RPCConsole* console)
     QTest::keyClicks(lineEdit, "getblockchaininfo");
     QVERIFY(mw_spy.wait(1000));
-    QCOMPARE(mw_spy.count(), 2);
+    QCOMPARE(mw_spy.count(), 4);
     QString output = messagesWidget->toPlainText();
     UniValue value;
     value.read(output.right(output.size() - output.lastIndexOf(QChar::ObjectReplacementCharacter) - 1).toStdString());

```

hebasto:
@Talkless @promag 

Hoping on your re-reviewing.

promag (commented):
Tested ACK 38eb37c0bd29b4cb825de905e8eec87636a5221b.

Added a dummy sleep in some call just to simplify testing.

PR #125, opened 2020-10-29
Enable changing the autoprune block space size in intro dialog

luke-jr:
Screenshot_20200911_095102

Moved from https://github.com/bitcoin/bitcoin/pull/18728

### DIFF
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -315,11 +315,9 @@ void BitcoinApplication::parameterSetup()
     InitParameterInteraction(gArgs);
 }
 
-void BitcoinApplication::InitializePruneSetting(bool prune)
+void BitcoinApplication::InitPruneSetting(int64_t prune_MiB)
 {
-    // If prune is set, intentionally override existing prune size with
-    // the default size since this is called when choosing a new datadir.
-    optionsModel->SetPruneTargetGB(prune ? DEFAULT_PRUNE_TARGET_GB : 0, true);
+    optionsModel->SetPruneTargetGB(PruneMiBtoGB(prune_MiB), true);
 }
 
 void BitcoinApplication::requestInitialize()
@@ -508,9 +506,9 @@ int GuiMain(int argc, char* argv[])
     /// 5. Now that settings and translations are available, ask user for data directory
     // User language is set up: pick a data directory
     bool did_show_intro = false;
-    bool prune = false; // Intro dialog prune check box
+    int64_t prune_MiB = 0;  // Intro dialog prune configuration
     // Gracefully exit if the user cancels
-    if (!Intro::showIfNeeded(did_show_intro, prune)) return EXIT_SUCCESS;
+    if (!Intro::showIfNeeded(did_show_intro, prune_MiB)) return EXIT_SUCCESS;
 
     /// 6. Determine availability of data directory and parse bitcoin.conf
     /// - Do not call GetDataDir(true) before this step finishes
@@ -594,7 +592,7 @@ int GuiMain(int argc, char* argv[])
 
     if (did_show_intro) {
         // Store intro dialog settings other than datadir (network specific)
-        app.InitializePruneSetting(prune);
+        app.InitPruneSetting(prune_MiB);
     }
 
     if (gArgs.GetBoolArg("-splash", DEFAULT_SPLASHSCREEN) && !gArgs.GetBoolArg("-min", false))
--- a/src/qt/forms/intro.ui
+++ b/src/qt/forms/intro.ui
@@ -220,6 +220,13 @@
       </widget>
      </item>
      <item>
+      <widget class="QLabel" name="lblPruneSuffix">
+       <property name="buddy">
+        <cstring>pruneGB</cstring>
+       </property>
+      </widget>
+     </item>
+     <item>
       <spacer name="horizontalSpacer_2">
        <property name="orientation">
         <enum>Qt::Horizontal</enum>
--- a/src/qt/forms/intro.ui
+++ b/src/qt/forms/intro.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>674</width>
-    <height>415</height>
+    <height>447</height>
    </rect>
   </property>
   <property name="windowTitle">
@@ -201,22 +201,46 @@
     </layout>
    </item>
    <item>
+    <layout class="QHBoxLayout" name="pruneOptLayout">
+     <item>
+      <widget class="QCheckBox" name="prune">
+       <property name="text">
+        <string>Limit block chain storage to</string>
+       </property>
+       <property name="toolTip">
+        <string>Reverting this setting requires re-downloading the entire blockchain. It is faster to download the full chain first and prune it later. Disables some advanced features.</string>
+       </property>
+      </widget>
+     </item>
+     <item>
+      <widget class="QSpinBox" name="pruneGB">
+       <property name="suffix">
+        <string> GB</string>
+       </property>
+      </widget>
+     </item>
+     <item>
+      <spacer name="horizontalSpacer_2">
+       <property name="orientation">
+        <enum>Qt::Horizontal</enum>
+       </property>
+       <property name="sizeHint" stdset="0">
+        <size>
+         <width>40</width>
+         <height>20</height>
+        </size>
+       </property>
+      </spacer>
+     </item>
+    </layout>
+   </item>
+   <item>
     <widget class="QLabel" name="lblExplanation1">
      <property name="text">
       <string>When you click OK, %1 will begin to download and process the full %4 block chain (%2GB) starting with the earliest transactions in %3 when %4 initially launched.</string>
      </property>
      <property name="wordWrap">
       <bool>true</bool>
-     </property>
-    </widget>
-   </item>
-   <item>
-    <widget class="QCheckBox" name="prune">
-     <property name="toolTip">
-      <string>Reverting this setting requires re-downloading the entire blockchain. It is faster to download the full chain first and prune it later. Disables some advanced features.</string>
-     </property>
-     <property name="text">
-      <string></string>
      </property>
     </widget>
    </item>
--- a/src/qt/forms/intro.ui
+++ b/src/qt/forms/intro.ui
@@ -201,6 +201,36 @@
     </layout>
    </item>
    <item>
+    <widget class="QLabel" name="lblExplanation1">
+     <property name="text">
+      <string>When you click OK, %1 will begin to download and process the full %4 block chain (%2GB) starting with the earliest transactions in %3 when %4 initially launched.</string>
+     </property>
+     <property name="wordWrap">
+      <bool>true</bool>
+     </property>
+    </widget>
+   </item>
+   <item>
+    <widget class="QLabel" name="lblExplanation2">
+     <property name="text">
+      <string>This initial synchronisation is very demanding, and may expose hardware problems with your computer that had previously gone unnoticed. Each time you run %1, it will continue downloading where it left off.</string>
+     </property>
+     <property name="wordWrap">
+      <bool>true</bool>
+     </property>
+    </widget>
+   </item>
+   <item>
+    <widget class="QLabel" name="lblExplanation3">
+     <property name="text">
+      <string>If you have chosen to limit block chain storage (pruning), the historical data must still be downloaded and processed, but will be deleted afterward to keep your disk usage low.</string>
+     </property>
+     <property name="wordWrap">
+      <bool>true</bool>
+     </property>
+    </widget>
+   </item>
+   <item>
     <layout class="QHBoxLayout" name="pruneOptLayout">
      <item>
       <widget class="QCheckBox" name="prune">
@@ -240,36 +270,6 @@
       </spacer>
      </item>
     </layout>
-   </item>
-   <item>
-    <widget class="QLabel" name="lblExplanation1">
-     <property name="text">
-      <string>When you click OK, %1 will begin to download and process the full %4 block chain (%2GB) starting with the earliest transactions in %3 when %4 initially launched.</string>
-     </property>
-     <property name="wordWrap">
-      <bool>true</bool>
-     </property>
-    </widget>
-   </item>
-   <item>
-    <widget class="QLabel" name="lblExplanation2">
-     <property name="text">
-      <string>This initial synchronisation is very demanding, and may expose hardware problems with your computer that had previously gone unnoticed. Each time you run %1, it will continue downloading where it left off.</string>
-     </property>
-     <property name="wordWrap">
-      <bool>true</bool>
-     </property>
-    </widget>
-   </item>
-   <item>
-    <widget class="QLabel" name="lblExplanation3">
-     <property name="text">
-      <string>If you have chosen to limit block chain storage (pruning), the historical data must still be downloaded and processed, but will be deleted afterward to keep your disk usage low.</string>
-     </property>
-     <property name="wordWrap">
-      <bool>true</bool>
-     </property>
-    </widget>
    </item>
    <item>
     <spacer name="verticalSpacer">
--- a/src/qt/intro.cpp
+++ b/src/qt/intro.cpp
@@ -182,6 +182,16 @@ void Intro::setDataDirectory(const QString &dataDir)
     }
 }
 
+int64_t Intro::getPruneMiB() const
+{
+    switch (ui->prune->checkState()) {
+    case Qt::Checked:
+        return PruneGBtoMiB(m_prune_target_gb);
+    case Qt::Unchecked: default:
+        return 0;
+    }
+}
+
 bool Intro::showIfNeeded(bool& did_show_intro, int64_t& prune_MiB)
 {
     did_show_intro = false;
@@ -233,7 +243,7 @@ bool Intro::showIfNeeded(bool& did_show_intro, int64_t& prune_MiB)
         }
 
         // Additional preferences:
-        prune_MiB = intro.ui->prune->isChecked() ? PruneGBtoMiB(intro.m_prune_target_gb) : int64_t(0);
+        prune_MiB = intro.getPruneMiB();
 
         settings.setValue("strDataDir", dataDir);
         settings.setValue("fReset", false);
--- a/src/qt/intro.cpp
+++ b/src/qt/intro.cpp
@@ -17,6 +17,7 @@
 
 #include <interfaces/node.h>
 #include <util/system.h>
+#include <validation.h>
 
 #include <QFileDialog>
 #include <QSettings>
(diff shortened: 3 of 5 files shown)


jonasschnelli:
Concept ACK

src/qt/bitcoin.cpp
```diff
@@ -315,11 +315,9 @@ void BitcoinApplication::parameterSetup()
     InitParameterInteraction(gArgs);
 }
 
-void BitcoinApplication::InitializePruneSetting(bool prune)
+void BitcoinApplication::InitPruneSetting(int64_t prune_opt)
```
promag, line 318: 273e57bd4cd91e8399f3567c79241bbc416ea690

Why the rename? To ensure calls are refactored?

src/qt/intro.cpp
```diff
@@ -139,17 +140,25 @@ Intro::Intro(QWidget *parent, int64_t blockchain_size_gb, int64_t chain_state_si
...
-    ui->prune->setText(tr("Discard blocks after verification, except most recent %1 GB (prune)").arg(m_prune_target_gb));
+    ui->pruneGB->setValue(m_prune_target_gb);
+    ui->pruneGB->setToolTip(ui->prune->toolTip());
     UpdatePruneLabels(ui->prune->isChecked());
 
     connect(ui->prune, &QCheckBox::toggled, [this](bool prune_checked) {
         UpdatePruneLabels(prune_checked);
         UpdateFreeSpaceLabel();
     });
+    connect(ui->pruneGB, static_cast<void (QSpinBox::*)(int)>(&QSpinBox::valueChanged), [this](int prune_GB) {
```
promag, line 158: 77f3f2dda6918b0986f785b3061cfb8c97ed2b5c

just noting that this cast is only required because of `valueChanged` signal is overloaded:
```cpp
#if QT_DEPRECATED_SINCE(5, 14)
    QT_DEPRECATED_X("Use textChanged(QString) instead")
    void valueChanged(const QString &);
#endif
```

src/qt/intro.cpp
```diff
@@ -17,6 +17,7 @@
 
 #include <interfaces/node.h>
 #include <util/system.h>
+#include <validation.h>
```
promag, line 20: 77f3f2dda6918b0986f785b3061cfb8c97ed2b5c

It's unfortunate to include this just because of `MIN_DISK_SPACE_FOR_BLOCK_FILES`.

src/qt/intro.cpp
```diff
@@ -381,6 +382,9 @@ void Intro::UpdatePruneLabels(bool prune_checked)
     }
     ui->lblExplanation3->setVisible(prune_checked);
     ui->pruneGB->setEnabled(prune_checked);
+    static const uint64_t nPowTargetSpacing = 10 * 60;  // from chainparams, which we don't have at this stage
+    static const uint32_t expected_block_data_size = 2250000;  // includes undo data
+    ui->lblPruneSuffix->setText(tr("(sufficient to restore backups %n day(s) old)", "block chain pruning", m_prune_target_gb * 1e9 / (uint64_t(expected_block_data_size) * 86400 / nPowTargetSpacing)));
```
promag, line 387: 7d40a4805229618a364dc53a84cd5af2ac273b7d

nit, suggest to declare a local variable.

src/qt/intro.cpp
```diff
@@ -139,17 +140,25 @@ Intro::Intro(QWidget *parent, int64_t blockchain_size_gb, int64_t chain_state_si
     );
     ui->lblExplanation2->setText(ui->lblExplanation2->text().arg(PACKAGE_NAME));
 
+    const int min_prune_target = std::ceil(MIN_DISK_SPACE_FOR_BLOCK_FILES / 1e9);
+    ui->pruneGB->setRange(min_prune_target, std::numeric_limits<int>::max());
```
promag, line 144: 77f3f2dda6918b0986f785b3061cfb8c97ed2b5c

Just `setMinimum`?

nit, `min_prune_target_gb`?

promag (commented):
Concept ACK.

src/qt/bitcoin.cpp
luke-jr: Yes, otherwise a bool might get silently treated as a 1

src/qt/intro.cpp
luke-jr: Need to set the maximum too... Default is IIRC 100

Bosch-0:
Testing ACK 7e37329 on Windows 10.0.18363 Build 18363. Articulating how exactly the pruned storage works is a great UX move. Many users seem to think pruning only stores data relevant to them, not that its only storing a recent history of blocks. Can be quite a shock importing an old wallet that you believe has coins only to get a zero balance displayed. 

Screenshot 2020-11-09 144538

Just some minor nits that could be fixed along with this PR:

"The wallet will also be stored in this directory" should be changed to "Wallets will also be stored in this directory" in the top dialogue. Users can / may have multiple wallets.

Blockchain size is around ~350GB now, may want to update the 320GB text in the second last paragraph.

ryanofsky (approved):
Code review ACK 7d40a4805229618a364dc53a84cd5af2ac273b7d

promag (commented):
Tested ACK 7d40a4805229618a364dc53a84cd5af2ac273b7d.

Could fix the following when setting an invalid prune target (big number or empty string):
Screenshot 2020-11-12 at 09 54 10

Sjors:
Concept ACK

sipa:
@Bosch-0 Does rescanning a wallet with a pruned node that misses relevant blocks result in an incorrect balance? That sounds like a bug - it should just abort with missing blocks or so.

hebasto:
@Bosch-0 
> Just some minor nits that could be fixed along with this PR:
> 
> "The wallet will also be stored in this directory" should be changed to "Wallets will also be stored in this directory" in the top dialogue. Users can / may have multiple wallets.

Usually, we are trying to keep PRs focused :)
A small pull with a label correction could be easy to review.

> Blockchain size is around ~350GB now, may want to update the 320GB text in the second last paragraph.

This number update is a part of the [release process](https://github.com/bitcoin/bitcoin/blob/master/doc/release-process.md#before-branch-off), e.g., https://github.com/bitcoin/bitcoin/pull/20263.

src/qt/intro.cpp
hebasto, line 158: 77f3f2dda6918b0986f785b3061cfb8c97ed2b5c
```suggestion
    connect(ui->pruneGB, QOverload<int>::of(&QSpinBox::valueChanged), [this](int prune_GB) {
```
[looks](https://doc.qt.io/qt-5/signalsandslots-syntaxes.html#selecting-overloaded-signals-and-slots) more neat.

src/qt/forms/intro.ui
```diff
@@ -219,6 +219,13 @@
        </property>
       </widget>
      </item>
+     <item>
+      <widget class="QLabel" name="lblPruneSuffix">
+       <property name="buddy">
+        <cstring>pruneGB</cstring>
+       </property>
```
hebasto, line 226: 7d40a4805229618a364dc53a84cd5af2ac273b7d
Where is this used?

src/qt/intro.cpp
```diff
@@ -381,6 +382,9 @@ void Intro::UpdatePruneLabels(bool prune_checked)
     }
     ui->lblExplanation3->setVisible(prune_checked);
     ui->pruneGB->setEnabled(prune_checked);
+    static const uint64_t nPowTargetSpacing = 10 * 60;  // from chainparams, which we don't have at this stage
+    static const uint32_t expected_block_data_size = 2250000;  // includes undo data
```
hebasto, line 386: 7d40a4805229618a364dc53a84cd5af2ac273b7d, nit:
```suggestion
    static constexpr uint64_t nPowTargetSpacing = 10 * 60;  // from chainparams, which we don't have at this stage
    static constexpr uint32_t expected_block_data_size = 2250000;  // includes undo data
```

hebasto (changes requested):
Concept ACK 7d40a4805229618a364dc53a84cd5af2ac273b7d.

Master (fb7726e56d0f0617076c89cbc279547cc9d07341):
DeepinScreenshot_select-area_20201116112204

This PR:
DeepinScreenshot_select-area_20201116112251

The visual distance between two closely related items has been increased (see the screeshots above). This change makes harder for novices to grasp the pruning details. I'd suggest the opposite: to place those items side by side.

src/qt/intro.cpp
```diff
@@ -182,6 +182,16 @@ void Intro::setDataDirectory(const QString &dataDir)
...
 }
 
+int64_t Intro::getPrune() const
+{
+    switch (ui->prune->checkState()) {
+    case Qt::Checked:
+         return PruneGBtoMiB(m_prune_target_gb);
+    case Qt::Unchecked: default:
+        return 0;
+    }
```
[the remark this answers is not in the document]
luke-jr: It becomes tri-state in a followup PR.

src/qt/forms/intro.ui
```diff
@@ -200,6 +200,40 @@
...
    </item>
+   <item>
+    <layout class="QHBoxLayout" name="pruneOptLayout">
+     <item>
+      <widget class="QCheckBox" name="prune">
+       <property name="text">
+        <string>Limit block chain storage to</string>
+       </property>
+       <property name="toolTip">
+        <string>Reverting this setting requires re-downloading the entire blockchain. It is faster to download the full chain first and prune it later. Disables some advanced features.</string>
```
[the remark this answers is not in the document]
luke-jr: "blockchain" and "block chain" are tied in existing strings. I think the single word is better.

src/qt/forms/intro.ui
luke-jr: It's here for correctness. I don't know if there's a platform that acts on it.

luke-jr:
Addressed @hebasto's comments.

Ping @promag @ryanofsky @Bosch-0 for re-ACKing

Bosch-0:
> @Bosch-0 Does rescanning a wallet with a pruned node that misses relevant blocks result in an incorrect balance? That sounds like a bug - it should just abort with missing blocks or so.

I haven't actually tested this but I don't think it does - will test this today. The only time an incorrect balance may be shown is when importing a wallet post-prune that has tx history older than the block history stored as far as I'm concerned.

Bosch-0:
ACK ea7926527ce36c05cacb99602b2b59579ff646e8, looks good.

hebasto:
> ACK [ea79265](https://github.com/bitcoin-core/gui/commit/ea7926527ce36c05cacb99602b2b59579ff646e8), looks good.

Which commit did you mean?

src/qt/intro.cpp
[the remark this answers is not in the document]
hebasto: Have you a branch to look into upcoming changes?

src/qt/forms/intro.ui
hebasto: > It's here for correctness. I don't know if there's a platform that acts on it.

I cannot parse this message. Correctness of what?

hebasto:
@luke-jr 
> Addressed @hebasto's comments.

Sorry, but I don't think just that marking a comment as resolved means that it is addressed. If the current state of code is preferred  that could be stated explicitly, right?

src/qt/forms/intro.ui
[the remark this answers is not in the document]
hebasto: Screenshot from 2020-11-19 07-54-29

https://github.com/bitcoinops/bitcoinops.github.io/blob/master/STYLE.md

Bosch-0:
> Which commit did you mean?

I just re-tested the changes made since my first ACK, I did not review the code

hebasto:
> > Which commit did you mean?
> 
> I just re-tested the changes made since my first ACK, I did not review the code

I mean that ACK should be followed by the top commit of the pr branch, i.e. 7d40a4805229618a364dc53a84cd5af2ac273b7d for now, not the master branch (that ea7926527ce36c05cacb99602b2b59579ff646e8 is).

ryanofsky:
Is an update supposed to be pushed? My review was requested but there haven't been any changes since my last review https://github.com/bitcoin-core/gui/pull/125#pullrequestreview-528515367

src/qt/intro.cpp
```diff
@@ -361,6 +381,11 @@ void Intro::UpdatePruneLabels(bool prune_checked)
         storageRequiresMsg = tr("Approximately %1 GB of data will be stored in this directory.");
     }
     ui->lblExplanation3->setVisible(prune_checked);
+    ui->pruneGB->setEnabled(prune_checked);
+    static constexpr uint64_t nPowTargetSpacing = 10 * 60;  // from chainparams, which we don't have at this stage
+    static constexpr uint32_t expected_block_data_size = 2250000;  // includes undo data
+    static const uint64_t expected_backup_days = m_prune_target_gb * 1e9 / (uint64_t(expected_block_data_size) * 86400 / nPowTargetSpacing);
```
jarolrod, line 387: setting `expected_backup_days` as a `static const` means that the number of days the prune target is good for will never update as the prune target is changed.
```suggestion
    uint64_t expected_backup_days = m_prune_target_gb * 1e9 / (uint64_t(expected_block_data_size) * 86400 / nPowTargetSpacing);
```

jarolrod (changes requested):

src/qt/intro.cpp
luke-jr: Fixed (const can remain)

src/qt/intro.cpp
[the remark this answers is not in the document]
jarolrod: @luke-jr is this still going to become tri-state? otherwise hebasto's suggestion would be preferred

jarolrod (approved):

jarolrod (approved):
Tested ACK 415fb2e.
Stress-tested on Ubuntu 20.10 and Arch Linux. Synced a node on each machine with no issues. Ubuntu machine set to 6gb prune target, Arch machine set to 9gb prune target.

luke-jr:
QOverload doesn't work in Qt 5.5 ... reverted that change

luke-jr:
master is bumped to Qt 5.9, so went back to QOverload...

Talkless:
@lucash-dev feeling dejavu - I see `Intro::getPruneMiB()` added here and int #149 :)

Talkless (approved):
tACK 415fb2e1abac189fcbe8eccf2ea065724d17460f, tested on Debian Sid with Qt 5.15.2.

src/qt/bitcoin.cpp
```diff
@@ -508,9 +506,9 @@ int GuiMain(int argc, char* argv[])
     /// 5. Now that settings and translations are available, ask user for data directory
     // User language is set up: pick a data directory
     bool did_show_intro = false;
-    bool prune = false; // Intro dialog prune check box
+    int64_t prune_MiB = 0;  // Intro dialog prune configuration
```
hebasto, line 509: style piconit: [`clang-format-diff.py`](https://github.com/bitcoin/bitcoin/tree/master/contrib/devtools#clang-format-diffpy) suggests
```suggestion
    int64_t prune_MiB = 0; // Intro dialog prune configuration
```

src/qt/forms/intro.ui
```diff
@@ -240,6 +230,47 @@
...
+       </property>
```
hebasto, line 242: Qt Designer insists on changing properties order:
```suggestion
       <property name="toolTip">
        <string>Reverting this setting requires re-downloading the entire blockchain. It is faster to download the full chain first and prune it later. Disables some advanced features.</string>
       </property>
       <property name="text">
        <string>Limit block chain storage to</string>
       </property>
```

src/qt/intro.cpp
```diff
@@ -139,17 +140,26 @@ Intro::Intro(QWidget *parent, int64_t blockchain_size_gb, int64_t chain_state_si
     );
     ui->lblExplanation2->setText(ui->lblExplanation2->text().arg(PACKAGE_NAME));
 
+    const int min_prune_target_GB = std::ceil(MIN_DISK_SPACE_FOR_BLOCK_FILES / 1e9);
```
hebasto, line 143: nit: why here and after not use https://github.com/bitcoin-core/gui/blob/fa2a5b8f3aeb978b64189203cbbe95a00ac92a6a/src/qt/guiconstants.h#L52-L53
?

src/qt/intro.cpp
```diff
@@ -139,17 +140,26 @@ Intro::Intro(QWidget *parent, int64_t blockchain_size_gb, int64_t chain_state_si
...
+    ui->pruneGB->setValue(m_prune_target_gb);
+    ui->pruneGB->setToolTip(ui->prune->toolTip());
+    ui->lblPruneSuffix->setToolTip(ui->prune->toolTip());
     UpdatePruneLabels(ui->prune->isChecked());
 
     connect(ui->prune, &QCheckBox::toggled, [this](bool prune_checked) {
         UpdatePruneLabels(prune_checked);
         UpdateFreeSpaceLabel();
     });
+    connect(ui->pruneGB, QOverload<int>::of(&QSpinBox::valueChanged), [this](int prune_GB) {
```
hebasto, line 158: The `QOverload` helper class is required for C++11 code. Having C++17, the `qOverload` should be used.

src/qt/intro.cpp
```diff
@@ -182,7 +192,17 @@ void Intro::setDataDirectory(const QString &dataDir)
...
 }
 
-bool Intro::showIfNeeded(bool& did_show_intro, bool& prune)
+int64_t Intro::getPruneMiB() const
+{
+    switch (ui->prune->checkState()) {
+    case Qt::Checked:
+        return PruneGBtoMiB(m_prune_target_gb);
+    case Qt::Unchecked: default:
+        return 0;
```
hebasto, line 201:
```suggestion
    case Qt::Unchecked:
    [[fallthrough]];
    default:
        return 0;
```

src/qt/intro.cpp
```diff
@@ -361,6 +381,11 @@ void Intro::UpdatePruneLabels(bool prune_checked)
         storageRequiresMsg = tr("Approximately %1 GB of data will be stored in this directory.");
     }
     ui->lblExplanation3->setVisible(prune_checked);
+    ui->pruneGB->setEnabled(prune_checked);
+    static constexpr uint64_t nPowTargetSpacing = 10 * 60;  // from chainparams, which we don't have at this stage
+    static constexpr uint32_t expected_block_data_size = 2250000;  // includes undo data
```
hebasto, line 386: style piconit: [`clang-format-diff.py`](https://github.com/bitcoin/bitcoin/tree/master/contrib/devtools#clang-format-diffpy) suggests
```suggestion
    static constexpr uint64_t nPowTargetSpacing = 10 * 60;        // from chainparams, which we don't have at this stage
    static constexpr uint32_t expected_block_data_size = 2250000; // includes undo data
```

src/qt/intro.h
```diff
@@ -36,6 +36,7 @@ class Intro : public QDialog
 
     QString getDataDirectory();
     void setDataDirectory(const QString &dataDir);
+    int64_t getPruneMiB() const;
```
hebasto, line 39: Could be private?

src/qt/forms/intro.ui
hebasto: @luke-jr 

From [`QLabel::setBuddy`](https://doc.qt.io/qt-5/qlabel.html#setBuddy):
> The buddy mechanism is only available for `QLabel`s that contain text in which one character is prefixed with an ampersand, '&'. This character is set as the shortcut key.

You do not set a shortcut key, therefore linking `lblPruneSuffix` to `pruneGB` buddy is noop, and should be dropped.

hebasto (commented):
Tested 415fb2e1abac189fcbe8eccf2ea065724d17460f on Linux Mint 20.1 (Qt 5.12.8).
Works great!

I believe the commit "GUI/Intro: Move prune setting below explanation" (415fb2e1abac189fcbe8eccf2ea065724d17460f) should be reordered and squashed into the "GUI/Intro: Rework UI flow to let the user set prune size in GBs" (e2dcd957fa206a28404ff824fb764b8889705fb2).

src/qt/intro.cpp
hebasto: > IIRC qOverload function template does not work with Visual Studio 2017 due to implementation limitations

Oh, I was not aware of it. I'd like to know more about this pitfall. Any proof/link/test is appreciated :)

hebasto (approved):
ACK 415fb2e1abac189fcbe8eccf2ea065724d17460f, my unresolved comments are not blockers, and they could be resolved in follow ups.

Rspigler:
Post-merge tACK.  Great improvement!

src/qt/intro.cpp
Talkless: @hebasto see https://bugreports.qt.io/browse/QTBUG-61667?focusedCommentId=462231&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-462231

hebasto: @Talkless

Thanks! Since https://github.com/bitcoin/bitcoin/pull/21811 is merged, it isn't a problem, is it?

Btw, could you look at #257?

Talkless: It seem so, yes. Maybe I should try building on some older GCC to make sure other compiler does not have similar limitations (though I doubt it).

I will take a look at 257 at some time.

hebasto:
@luke-jr 

Could you add release notes into https://github.com/bitcoin-core/bitcoin-devwiki/wiki/22.0-Release-Notes-draft#gui-changes?

PR #127, opened 2020-11-02
Replace QMetaObject::invokeMethod with atomic operations

hebasto:
No need to read/write a data member in one thread if it is atomic.

### DIFF
--- a/src/qt/transactiontablemodel.cpp
+++ b/src/qt/transactiontablemodel.cpp
@@ -215,7 +215,6 @@ TransactionTableModel::TransactionTableModel(const PlatformStyle *_platformStyle
         QAbstractTableModel(parent),
         walletModel(parent),
         priv(new TransactionTablePriv(this)),
-        fProcessingQueuedTransactions(false),
         platformStyle(_platformStyle)
 {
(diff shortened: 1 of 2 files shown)


promag (commented):
An atomic bool would be enough?

Didn't want to mess with this in #120 because I think ballon filtering doesn't belong in transaction model. IMO the right place should be aware of what wallets are open.

hebasto:
@promag 
> An atomic bool would be enough?

You're right. Reworked.

promag (commented):
This doesn't work: `setProcessingQueuedTransactions` needs to be enqueued and proceed in between transaction notifications notifications.

hebasto:
> This doesn't work: `setProcessingQueuedTransactions` needs to be enqueued and proceed in between transaction notifications.

Why?

promag:
Transaction notifications are enqueued in `TransactionNotification::invoke` and by the time they are processed `setProcessingQueuedTransactions(false)` was already called.

hebasto:
You mean that `setProcessingQueuedTransactions` must be placed into the same event queue as `updateTransaction`, i.e., coordinated?

promag:
@hebasto yes, `processingQueuedTransactions()` value is worthless.

PR #133, opened 2020-11-20
Qt Creator specific .gitignore files 

RandyMcMillan:
Ignore some files created by Qt

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,8 @@ src/univalue/gen
 src/qt/*.moc
 src/qt/moc_*.cpp
 src/qt/forms/ui_*.h
+src/qt/forms/*.ui.autosave
+src/qt/*.o.tmp
 
 src/qt/test/moc*.cpp
 
@@ -56,6 +58,7 @@ src/qt/bitcoin-qt.creator
 src/qt/bitcoin-qt.creator.user
 src/qt/bitcoin-qt.files
 src/qt/bitcoin-qt.includes
+contrib/bitcoin-qt.pro.user
 
 .deps
 .dirstamp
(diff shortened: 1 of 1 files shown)


Awramienko23 (approved):

.gitignore
```diff
@@ -98,6 +101,7 @@ background.tiff*
 
 # Qt Creator
 Makefile.am.user
+build-bitcoin-qt*
```
jonasschnelli, line 104: What or why would this file be created?

.gitignore
```diff
@@ -56,6 +58,7 @@ src/qt/bitcoin-qt.creator
 src/qt/bitcoin-qt.creator.user
 src/qt/bitcoin-qt.files
 src/qt/bitcoin-qt.includes
+contrib/bitcoin-qt.pro.user
```
jonasschnelli, line 61: I guess thats the Qt Creator project file? Why would that be stored in contrib? (Asking because mine is in the root folder AFAIK).

.gitignore
RandyMcMillan: Folders with this prefix are created by Qt creator.

.gitignore
RandyMcMillan: Good question. I will try to determine why. Thank.

hebasto:
If all changes are related to Qt Creator:
- s/Qt/Qt Creator/ in the OP
- all files/dirs could be gathered here: https://github.com/bitcoin-core/gui/blob/762cbd287f63035dc8f814b998c0b1af2e497820/.gitignore#L100

Out of curiosity, do you use Qt Creator as a build environment?

PR #46, opened 2020-08-09
refactor: Fix deprecation warnings when building against Qt 5.15

hebasto:
[What's New in Qt 5.15](https://doc.qt.io/qt-5/whatsnew515.html#deprecated-modules):
> To help preparing for the transition to Qt 6, numerous classes and member functions that will be removed from Qt 6.0 have been marked as deprecated in the Qt 5.15 release.

Fixes #36

PR #137, opened 2020-11-27
refactor: Replace deprecated Qt::SystemLocale{Short,Long}Date

hebasto:
As all deprecated warning in Qt 5.15.0 were eliminated in #46, Qt 5.15.1 introduced another one that is fixed in this PR.

Required for https://github.com/bitcoin/bitcoin/pull/20182.

Details in Qt docs:
- https://doc.qt.io/qt-5/qdatetime.html#toString-1
- https://doc.qt.io/qt-5/qdate.html#toString-1

### DIFF
--- a/src/qt/bantablemodel.cpp
+++ b/src/qt/bantablemodel.cpp
@@ -11,6 +11,7 @@
 
 #include <QDateTime>
 #include <QList>
+#include <QLocale>
 #include <QModelIndex>
 #include <QVariant>
 
@@ -122,7 +123,7 @@ QVariant BanTableModel::data(const QModelIndex &index, int role) const
         case Bantime:
             QDateTime date = QDateTime::fromMSecsSinceEpoch(0);
             date = date.addSecs(rec->banEntry.nBanUntil);
-            return date.toString(Qt::SystemLocaleLongDate);
+            return QLocale::system().toString(date, QLocale::LongFormat);
         }
     }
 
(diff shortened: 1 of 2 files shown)


promag:
Was this identified with the latest 5.15.2?

hebasto:
> Was this identified with the latest 5.15.2?

5.15.2 was released on 2020-11-20, but the recent Qt the Homebrew provides still 5.15.1.
Warnings are fired on 5.15.1. Did not test on 5.15.2.

promag:
Ok, tmight make sense to a

> UPDATE: Looking into [code](https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/global/qnamespace.h?h=5.15.0#n1274) it seems I just missed those warnings in #46 🤷‍♂️

Yeah, easy to miss.

jarolrod (approved):
Tested ACK 86b1ab6 on MacOS 10.15.7 and Arch Linux both with Qt 5.15.1
- Compiled both Master and PR with `make -j{n} |& tee compile.txt`, then grepped this file with `grep -r "deprecated"`
- Can confirm that master raises these deprecation warnings and the PR fixes them

I also compiled with Qt 5.15.2 on Linux and can confirm that master still raises, and PR fixes. Furthermore, there are no new deprecated warnings.

jonasschnelli (approved):
Tested ACK 86b1ab64b1a5b56518787ef16ea54ddbbc97d83e

Issue #136, opened 2020-11-25
macOS 11 (Big Sur) visual bugs


jonasschnelli:
Nighy build on Big Sur "dark mode":
![Screenshot 2020-11-24 at 20 13 04](https://user-images.githubusercontent.com/178464/100140847-cda1ca00-2e91-11eb-9cc7-f0afbf542df8.png)

Self compiled on Big Sur "night mode" with depends:
![Screenshot 2020-11-24 at 20 10 38](https://user-images.githubusercontent.com/178464/100141076-1e192780-2e92-11eb-88fc-fb0c8975d2bc.png)

Two major issues:
* [ ] disabling the night mode color palette or adopt it correctly (probably a QT upstream issue)
* [ ] get rid of the white glitch in the debug-window tab-bar-menu 

Qt upstream issue: https://bugreports.qt.io/browse/QTBUG-85546

mjdietzx:
I also noticed what seems to be a bug on Big Sur. When I go to send a transaction from my encrypted wallet, I get a popup that says "Unlock wallet". I enter the correct passphrase, but no matter what the "OK" button stays disabled and can't be clicked. Usually I'd expect once you enter at least 1 char into the passphrase input, that button would become enabled/clickable

goums:
I've tried to update the `QPalette` and the `styleSheet` of the widgets to fix the display glitches, but I didn't manage to fix them as the bugs are directly in the `QMacStyle` widgets implementations.

The QT team is working on it,  but the underlying QT issues will probably take some time to be resolved and is tagged for QT6:
https://bugreports.qt.io/browse/QTBUG-86513

So one solution instead of waiting for QT6 release (and hoping for QMacStyle to be fixed) could be to force `QApplication` to use the default `"Fusion"` style on macos: 
https://doc.qt.io/qt-5/gallery.html

This is a one line change, no added dependencies, and it supports Light /  Dark modes out of the box.

- Some screenshots using Fusion on DarkMode : 
![image](https://user-images.githubusercontent.com/5577626/102623653-22f49280-4143-11eb-81b5-277ee90eb433.png)
![image](https://user-images.githubusercontent.com/5577626/102623667-2ab43700-4143-11eb-86ea-b76f59a3bf10.png)
![image](https://user-images.githubusercontent.com/5577626/102623689-343d9f00-4143-11eb-8622-aceb0da2576f.png)
[...]

PR #138, opened 2020-11-27
unlock encrypted wallet "OK" button bugfix

mjdietzx:
When trying to send a transaction from an encrypted wallet, the ask
passphrase dialog would not allow the user to click the "OK" button
and proceed. Therefore it was impossible to send a transaction
through the gui. It was not enabling the "OK" button after the
passphrase was entered by the user, because it was using the same
form validation logic as the "Change passphrase" flow.

I reported this in a comment in https://github.com/bitcoin-core/gui/issues/136. But then I realized this seems to be a flat out bug.

### DIFF
--- a/src/qt/askpassphrasedialog.cpp
+++ b/src/qt/askpassphrasedialog.cpp
@@ -199,6 +199,8 @@ void AskPassphraseDialog::textChanged()
         acceptable = !ui->passEdit2->text().isEmpty() && !ui->passEdit3->text().isEmpty();
         break;
     case Unlock: // Old passphrase x1
+        acceptable = !ui->passEdit1->text().isEmpty();
+        break;
     case ChangePass: // Old passphrase x1, new passphrase x2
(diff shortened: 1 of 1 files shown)


mjdietzx:
This bug was introduced in this commit https://github.com/bitcoin-core/gui/commit/6e950118a31fd6a85026d934fc6adb6255e47e23#diff-6fb84ee9edb2f6effebfefeea02e3548508d2043ea997132a17d9f44505f28d4L223, and it's bc the `Decrypt` case was removed from the switch statement (which the `Unlock` case had been falling back to / relying on)

MarcoFalke:
review ACK 8008ef770f3d0b14d03e22371314500373732143

nice catch!

hebasto (approved):
ACK 8008ef770f3d0b14d03e22371314500373732143, I have reviewed the code and it looks OK, I agree it can be merged.

Sorry for introducing this bug :man_facepalming:

Sjors:
It would be good to push the git tags from the main repo to this repo as well, so it's more clear which releases the regression is present in (looks like this one isn't in v0.21).

PR #135, opened 2020-11-24
peers-tab: cleaner presentation - more functionality

RandyMcMillan:
Initial Presentation:

Screen Shot 2020-11-24 at 6 27 28 PM

peer table populated:

Screen Shot 2020-11-24 at 6 27 36 PM

Detail View of selected peer:

Screen Shot 2020-11-24 at 6 27 50 PM

Window enlarged:
Peertable and detail view:
Screen Shot 2020-11-24 at 6 30 57 PM

Additional Notes:

The widgets were renamed to offer a "self documenting" naming convention during the development.

Screen Shot 2020-11-24 at 6 39 18 PM

Issue #100, opened 2020-10-07
Use black icon for bitcoin-core account


Bosch-0:
Currently the account that hosts this repo and several others uses the orange bitcoin icon which (as far as I'm concerned) is the Bitcoin brand icon not the Bitcoin Core brand icon. Would it not be more on-brand to use the black icon for this account and orange Bitcoin icon for the bitcoin account? A discussions was had several weeks ago around branding here https://github.com/bitcoin-core/gui/issues/89

[Here](https://www.figma.com/file/4dO1LqPY0WAgp5d1Bz3U8j/Bitcoin-Core-Designs?node-id=16%3A15) is the black icon on a Figma that can be exported and used.

Visualized below before / after:

![Group 113](https://user-images.githubusercontent.com/55287964/95337707-b65c4c80-08e4-11eb-8bde-6892f5a12ff2.png)

This icon also has the 'B' filled white (the current icon is not filled) which looks slightly better when, for e.g. content is shared on twitter.  

![image](https://user-images.githubusercontent.com/55287964/95329863-36c98000-08da-11eb-8e47-a25320bd84cc.png)

Posted an issue over at https://github.com/bitcoin/bitcoin/issues/20098 regarding changing the black icon to orange for the bitcoin account.

GBKS:
ACK

MarcoFalke:
Is figma better than an SVG https://github.com/bitcoin/bitcoin/pull/16571/files#diff-ae926a4501336e7914755a32cc3af751 ?

Bosch-0:
Figma is just an editor, you can export the logo I linked to as an SVG, PNG, JPG etc. apologies I missed this pull.

johnsBeharry:
ACK

---

_"Consistency over correctness"_ as I've been told.

Bosch-0:
Both of these repos should use this icon based on discussions here https://github.com/bitcoin-core/gui/issues/89

![Bitcoin Core Logo](https://user-images.githubusercontent.com/55287964/104463344-9e375000-55ec-11eb-8057-e6e3a3616ec7.png)

[Source](https://www.figma.com/file/0oqnohjahRtprjRyaetDOL/Bitcoin-Core-Design-System?node-id=1851%3A135)

Bosch-0:
Closing this issue based on my comment above.

Issue #89, opened 2020-09-14
[Discussion] Bitcoin Core logo


Bosch-0:
Wanted to open a discussion around the Bitcoin Core logo. Currently there is some inconsistencies in logo style / presentation as well as concerns around the currently used font. Having a more consistent logo assists with branding of Bitcoin Core as well as being important in preventing concerns over core 'controlling Bitcoin' as it will help it be seen as distinct from Bitcoin itself. I've also been doing some work around creating a Bitcoin Core design system of which branding will be a section covered, however I'll leave that for a separate discussion. 

# Icon + colorgraphy 

In https://github.com/bitcoin-core/bitcoincore.org/tree/master/assets/images there exists a orange and black version, what would be considered the primary logo to to use to represent core? 

I'd personally lean towards using the black version with some white (for dark backgrounds) and minimalist (no circle) iterations, both are shown below. The idea here is to have some branding distinction between Bitcoin Core and Bitcoin which the logo is a good way to achieve this. Why? Using the orange logo may confuse some people and think that Bitcoin Core IS Bitcoin - which is not the case. As mentioned above it is important for Bitcoin Core and Bitcoin to be distinct to prevent them being seen as the same entity. The original Bitcoin logo, in my view, should be used to symbolized the network itself not a product (such as the core GUI and related website bitcoincore.org) built on top of that network. Bitcoin core is currently the only product that uses the original logo as part of its branding (though some exceptions may apply such as bitcoin.org but this quite a different kind of entity to Bitcoin Core which ships tools users use). 

One concern brought up by the Bitcoin Design community was this approach could make Bitcoin Core be seen as as a bitcoin fork, though I think this is unlikely. Many Bitcoin wallets exist with branding that could be confused as a fork, e.g. Blockstreams GreenWallet, but has not caused confusion from my understanding. 

# Font

The current logo uses Neue Helvetica font, which is a [licensed font](https://www.myfonts.com/fonts/linotype/neue-helvetica/) - It may be worth switching to a free open source font to avoid issues. I've included one example, inter, below but I will be exploring some more. 

![image](https://user-images.githubusercontent.com/55287964/93116857-71336900-f6f0-11ea-89fd-3f9748af10ee.png)

Here are some quick examples of a window in the GUI that could be changed for more consistent branding.

## Launching GUI screen
![Group 124](https://user-images.githubusercontent.com/55287964/93116355-b99e5700-f6ef-11ea-81dc-95e36c085614.png)

[...]

PR #140, opened 2020-11-29
about:version:help clean up add Bitcoin branding

RandyMcMillan:
Before:

Screen Shot 2020-11-29 at 3 26 26 AM

After:

> Updated default presentations:
> 
> Screen Shot 2020-12-04 at 3 19 07 PM
> 
> Screen Shot 2020-12-04 at 3 18 56 PM

### DIFF
--- a/src/bitcoind.cpp
+++ b/src/bitcoind.cpp
@@ -55,13 +55,12 @@ static bool AppInit(int argc, char* argv[])
 
     // Process help and version before taking care about datadir
     if (HelpRequested(args) || args.IsArgSet("-version")) {
-        std::string strUsage = PACKAGE_NAME " version " + FormatFullVersion() + "\n";
+        std::string strUsage = PACKAGE_NAME " " + FormatFullVersion() + "\n" + LicenseInfo()  +"\n" + CopyrightInfo() + "\n";
 
-        if (args.IsArgSet("-version")) {
-            strUsage += FormatParagraph(LicenseInfo()) + "\n";
-        } else {
-            strUsage += "\nUsage:  bitcoind [options]                     Start " PACKAGE_NAME "\n";
-            strUsage += "\n" + args.GetHelpMessage();
+        if (!args.IsArgSet("-version")) {
+            strUsage += "\nUsage:  bitcoind [options]                     Start " PACKAGE_NAME "\n"
+                "\n";
+            strUsage += args.GetHelpMessage();
         }
 
         tfm::format(std::cout, "%s", strUsage);
--- a/src/clientversion.cpp
+++ b/src/clientversion.cpp
@@ -47,6 +47,11 @@ static std::string FormatVersion(int nVersion)
     return strprintf("%d.%d.%d", nVersion / 10000, (nVersion / 100) % 100, nVersion % 100);
 }
 
+std::string FormatVersion()
+{
+    return BUILD_DESC "-" BUILD_GIT_COMMIT;
+}
+
 std::string FormatFullVersion()
 {
     return CLIENT_BUILD;
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -591,23 +591,28 @@ void SetupServerArgs(NodeContext& node)
     argsman.AddHiddenArgs(hidden_args);
 }
 
+std::string VersionInfo()
+{
+    return FormatVersion();//from clientversion.cpp
+}
+
+std::string CopyrightInfo()
+{
+    return CopyrightHolders(strprintf(_("Copyright (C) %i-%i ").translated, 2009, COPYRIGHT_YEAR));
+}
+
+std::string AboutMessageInfo()
+{
+    const std::string SOURCE_CODE_URL = "<https://github.com/bitcoin/bitcoin>";
+    return strprintf(_("Please contribute if you find %s useful." "\n"
+                       "Visit %s for further information.").translated, PACKAGE_NAME, "<" PACKAGE_URL ">") + "\n" +
+           strprintf(_("The source code is available at: %s").translated, SOURCE_CODE_URL);
+}
+
 std::string LicenseInfo()
 {
-    const std::string URL_SOURCE_CODE = "<https://github.com/bitcoin/bitcoin>";
-
-    return CopyrightHolders(strprintf(_("Copyright (C) %i-%i").translated, 2009, COPYRIGHT_YEAR) + " ") + "\n" +
-           "\n" +
-           strprintf(_("Please contribute if you find %s useful. "
-                       "Visit %s for further information about the software.").translated,
-               PACKAGE_NAME, "<" PACKAGE_URL ">") +
-           "\n" +
-           strprintf(_("The source code is available from %s.").translated,
-               URL_SOURCE_CODE) +
-           "\n" +
-           "\n" +
-           _("This is experimental software.").translated + "\n" +
-           strprintf(_("Distributed under the MIT software license, see the accompanying file %s or %s").translated, "COPYING", "<https://opensource.org/licenses/MIT>") +
-           "\n";
+    const std::string LICENSE_URL = "<https://github.com/bitcoin/bitcoin/blob/master/COPYING>";
+    return strprintf(_("\nThis is experimental software.\nDistributed under the MIT software license:\n%s").translated, LICENSE_URL);
 }
 
 static bool fHaveGenesis = false;
@@ -881,7 +886,7 @@ void InitLogging(const ArgsManager& args)
 
     fLogIPs = args.GetBoolArg("-logips", DEFAULT_LOGIPS);
 
-    std::string version_string = FormatFullVersion();
+    std::string version_string = FormatVersion();
 #ifdef DEBUG
     version_string += " (debug build)";
 #else
--- a/src/init.h
+++ b/src/init.h
@@ -68,7 +68,10 @@ bool AppInitMain(const util::Ref& context, NodeContext& node, interfaces::BlockA
  */
 void SetupServerArgs(NodeContext& node);
 
-/** Returns licensing information (for -version) */
+/** Returns information (for -version) */
+std::string VersionInfo();
+std::string CopyrightInfo();
+std::string AboutMessageInfo();
 std::string LicenseInfo();
 
 #endif // BITCOIN_INIT_H
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -846,6 +846,7 @@ void BitcoinGUI::showDebugWindowActivateConsole()
 void BitcoinGUI::showHelpMessageClicked()
 {
     helpMessageDialog->show();
+    helpMessageDialog->raise();//bring to top if obscured
 }
 
 #ifdef ENABLE_WALLET
--- a/src/qt/forms/helpmessagedialog.ui
+++ b/src/qt/forms/helpmessagedialog.ui
@@ -6,9 +6,18 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>780</width>
-    <height>400</height>
+    <width>500</width>
+    <height>545</height>
    </rect>
+  </property>
+  <property name="sizePolicy">
+   <sizepolicy hsizetype="Preferred" vsizetype="Minimum">
+    <horstretch>0</horstretch>
+    <verstretch>0</verstretch>
+   </sizepolicy>
+  </property>
+  <property name="styleSheet">
+   <string notr="true"/>
   </property>
   <layout class="QHBoxLayout" name="horizontalLayout_2">
    <property name="spacing">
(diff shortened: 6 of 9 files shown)


src/init.cpp
```diff
@@ -591,23 +591,28 @@ void SetupServerArgs(NodeContext& node)
...
+{
+    return FormatVersion();
+}
+
+std::string CopyrightInfo()
+{
+    return CopyrightHolders(strprintf(_("Copyright (C) %i-%i").translated, 2009, COPYRIGHT_YEAR) + " ");
+}
+
+std::string AboutMessageInfo()
```
RandyMcMillan, line 604: The AboutMessageInfo() is simple and can be modified more easily in the future.

src/init.h
```diff
@@ -68,7 +68,16 @@ bool AppInitMain(const util::Ref& context, NodeContext& node, interfaces::BlockA
  */
 void SetupServerArgs(NodeContext& node);
 
-/** Returns licensing information (for -version) */
+/** Returns versionInfo text for utilitydialogue  */
```
RandyMcMillan, line 71: utilitydialogue is mentoned in the comments for better documentation.

src/qt/bitcoin.qrc
```diff
@@ -44,6 +44,7 @@
         <file alias="hd_disabled">res/icons/hd_disabled.png</file>
         <file alias="network_disabled">res/icons/network_disabled.png</file>
         <file alias="proxy">res/icons/proxy.png</file>
+        <file alias="bitcoin_core">res/icons/bitcoin_core.png</file>
```
RandyMcMillan, line 47: Added the png art that is used in the header of the about view.

src/qt/forms/helpmessagedialog.ui
```diff
@@ -6,10 +6,13 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>780</width>
-    <height>400</height>
+    <width>602</width>
```
RandyMcMillan, line 9: The new configuration makes a better use of space - therefore the window doesn't need to open as big.

src/qt/utilitydialog.cpp
```diff
@@ -41,21 +41,37 @@ HelpMessageDialog::HelpMessageDialog(QWidget *parent, bool about) :
         setWindowTitle(tr("About %1").arg(PACKAGE_NAME));
 
         std::string licenseInfo = LicenseInfo();
-        /// HTML-format the license message from the core
+        std::string copyrightInfo = CopyrightInfo();
```
RandyMcMillan, line 44: The new implementation separates these strings.

hebasto:
NACK on changes of `-version` command-line argument output layout.

jarolrod:
not-aligned
These two sections should be aligned

RandyMcMillan:
Updated Screenshots:


Screen Shot 2020-11-30 at 2 37 28 PM
Screen Shot 2020-11-30 at 2 37 39 PM

RandyMcMillan:
@hebasto 

I ensured that the output is uniform for -version

thanks for the feedback! :) 

Screen Shot 2020-11-30 at 11 45 04 PM

jonasschnelli:
I like the new style.
Though there is still a debate on whether the logo should be orange or black.
https://github.com/bitcoin-core/gui/issues/89

Concept ACK if we keep the color as is (we can always change it later to black if people want this).

RandyMcMillan:
@jonasschnelli - Since it is the about page - I thought using the web page logo might be useful. As it is now, the aboutMessage body is set up for further development - I was thinking of adding a menu bar that would replace the hyperlinks (It will look better). 

A note on color usage and logo specification - keep in mind that most companies/entities create specifications for full color - gray scale and black/white - for different purposes such as web design, print, what ever - As far as I'm concerned - there really isn't a debate - because we should have a specification and media resource files that fit print/web/etc use cases. 

Here is an example of a logo specification

https://ori.hhs.gov/logo-specification-guide

Screen Shot 2020-12-02 at 11 32 02 PM

RandyMcMillan:
@jonasschnelli - Since it is the about page - I thought using the web page logo might be useful. As it is now the aboutMessage body is set up for further development - I was thinking of adding a menu bar that would replace the hyperlinks (It will look better). 

A note on color usage - keep. in mind that most companies create specifications for full color - gray scale and black - for different purposes ass web design, print, what ever - As far as I'm concerned - there really isn't a debate - because we should have a specification and media resource files that fit print/web/etc use cases.

Bosch-0:
Concept ACK

The updated screenshots you posted above, the text should align with the logo such as below. 

image

bitcoinheiro:
I would change that logo. This is Bitcoin, Core is secondary. Only forks want to reinforce the Core over Bitcoin.

- Orange logo
- Bitcoin bold
- Core thin

Bosch-0:
> I would change that logo. This is Bitcoin, Core is secondary. Only forks want to reinforce the Core over Bitcoin.

I disagree with this view. The GUI this repo is for is a product developed by Core that uses the Bitcoin protocol. Other implementations may one day exist that work on Bitcoin that aren't Core.

bitcoinheiro:
What do we want to communicate?
Do we want to have more force on the name Bitcoin or Core?

We cannot be always thinking about other people concerns, the software is what it is and other implementations can exist - and they do exist - without the need to reinforce this concept on the logo and be "apologetic".

I believe users of the Core implementation should feel they are running BITCOIN when running this "product" and not "Core".
When I see the black logo it makes me feel sad, it confuses me instead of reinforcing Bitcoin. Feels like you're putting Bitcoin "in the closet" and we should be ashamed to be Bitcoin instead of proud of it.

Bosch-0:
>  Do we want to have more force on the name Bitcoin or Core?

One or the other at this stage - currently both are used and its confusing. 

> I believe users of the Core implementation should feel they are running BITCOIN

You can still achieve this whilst using the black logo - do people think they are running something else when using products like electrum or bluewallet? I'd say no. This is the designers job to communicate this effectively and use the black and orange logos in their appropriate locations. 

Do most people see Bitcoin Core = Bitcoin?

bitcoinheiro:
> Do most people see Bitcoin Core = Bitcoin?
I believe people do and you're exactly trying to use the design to communicate otherwise.
I don't think people have to be communicated otherwise.

Why Bitcoin so thin and Core so bold? Isn't that trying to communicate Core over Bitcoin?
This project is not Electrum or BlueWallet, it's Bitcoin.

image

Bosch-0:
I've already explained my rationale many times above.

I'm still of the thought that:

Bitcoin Core is a group of developers who work on Bitcoin.

Bitcoin is a network that everyone is apart of. 

Electrum and BlueWallet are part of the Bitcoin ecosystem as is the GUI developed by the Core developers.

jonatack:
I agree with not emphasizing "Core". Let's not institutionalize that emphasis further. Please keep them equal or put more weight on "Bitcoin".

One of my pet peeves is people writing "Core" instead of "Bitcoin Core" ("Core devs", "running Core", and so on).

See the [Bitcoin Optech style guide](https://github.com/bitcoinops/bitcoinops.github.io/blob/master/STYLE.md):

```
Forbidden abbreviations:

Core (use Bitcoin Core)
```

100% agree with our Bitcoin Optech colleagues about this.

jonasschnelli:
Don't get me wrong. I originally designed the black logo and I like it.
Though @gmaxwell made a valid comment [here](https://github.com/bitcoin-core/gui/issues/89#issuecomment-705384143).

Changing the color might have greater impact than originally anticipated.

jonatack:
Concept ACK, sorry about not reviewing the PR itself, in general this looks very nice @RandyMcMillan and is a good improvement.

promag (commented):
I'd be happy to ACK everything but the logo change, maybe you want to split the PR?

RandyMcMillan:
Thank you for the feed back!

RandyMcMillan:
I agree with the statement 

`
The Bitcoin Core Developers
`

 is a group and this is why I've maintained that "Developers" should be capitalized in the branding.

Further more...

From a copyright perspective - THE BITCOIN CORE DEVELOPERS should be in all caps - but this isn't mandated in any legalese that I have read...

But that is outside the scope of this PR. :)

RandyMcMillan:
@promag @jonatack 

This approach is better. I am only using text styling to create "Bitcoin Core" and reusing the existing
bitcoin logo - this doesn't add any weight to the repo.

---

 I've included "Core" to maintain existing branding continuity - but this can easily be removed if/when neccesary.

Screen Shot 2020-12-03 at 7 18 22 PM

---

RandyMcMillan:
I didn't see the link - agree.

https://github.com/bitcoin-core/gui/issues/89#issuecomment-705384143

> The logo should remain orange. This change would inadvertantly play into efforts to create a false equivalence between Bitcoin and various fraudulent clones.
> 
> We now know through leaked emails that even the renaming of Bitcoin Core to that name was dishonestly driven by malicious intent-- the parties that pushed for the name change did so because they were intending to attempt to take over Bitcoin from the community and move it under their control, and they wanted to "other" the community project and disassociate it with Bitcoin. Going along with it was an error. It's water under the bridge now but there is no reason to continue being victimized by past wrongs and compounding a past error.

jarolrod (commented):
Tested ACK 5210082
NACK on the current logo text, it looks really bad IMO. 
logo


Also the `about page` and `command-line options page` don't need to be the same size. There's a lot of unused space in the `about page` because they're set to be the same size.

RandyMcMillan:
@jarolrod - please add some context to the pic you posted.
OS - Screen resolution - dark theme? etc...
Thanks.

jarolrod:
> @jarolrod - please add some context to the pic you posted.
> OS - Screen resolution - dark theme? etc...
> Thanks.

OS: Arch Linux, DE: KDE Plasma 5.19.5, Res:3440x1440
The picture is pixelated because its cropped. Below is the full photo. I was trying to point out that everything looks off. The Bitcoin logo is in a weird position in relation to the logo text and the seperator bar. The logo text appears to have weird spacing. I'm not a designer, but this looks off to me.
comand-line-options-page

RandyMcMillan:
@jarolrod 

It is important to consider the way the text is rendered - keep in mind that we select a font family.
Due to the issues with licensing we don't embed a font in the app - so we select a font family - then your local system selects that actual font that is used. There will be variations.


Screen Shot 2020-12-03 at 9 49 19 PM

RandyMcMillan:
Removed excessive space in the aboutMessage body.
These are the default dimensions when they first appear.

Screen Shot 2020-12-03 at 10 51 00 PM

src/qt/bitcoingui.cpp
```diff
@@ -846,6 +846,7 @@ void BitcoinGUI::showDebugWindowActivateConsole()
 void BitcoinGUI::showHelpMessageClicked()
 {
     helpMessageDialog->show();
+    helpMessageDialog->raise();//bring to top if obscured
```
RandyMcMillan, line 849: If the helpMessageDialog has already been presented - this change brings it to the front if the user re-clicks on the help menu item - otherwise it remains obscured by any other windows.

src/qt/utilitydialog.cpp
```diff
@@ -33,29 +33,46 @@ HelpMessageDialog::HelpMessageDialog(QWidget *parent, bool about) :
     ui(new Ui::HelpMessageDialog)
 {
     ui->setupUi(this);
-
+    this->resize(500,0);//force view height to stretch
```
RandyMcMillan, line 36: The Ui::HelpMessageDialog is shared with the about view.
These changes force the view to stretch to an appropriate height -  this eliminates excessive height dimensions - the user resizes based upon personal preference.

src/qt/utilitydialog.cpp
```diff
@@ -33,29 +33,46 @@ HelpMessageDialog::HelpMessageDialog(QWidget *parent, bool about) :
...
+    ui->aboutMessage->resize(0,0);//force label to stretch
```
RandyMcMillan, line 37: https://github.com/bitcoin-core/gui/pull/140/commits/e0ff8e263dc96199385e31e481e64ad3a242fcd9#r536333863

src/qt/utilitydialog.cpp
```diff
@@ -33,29 +33,46 @@ HelpMessageDialog::HelpMessageDialog(QWidget *parent, bool about) :
...
+        ui->scrollArea->setVerticalScrollBarPolicy(Qt::ScrollBarAsNeeded);
+        ui->copyrightInfo->setText(copyrightInfoHTML); //about bottom left corner
+        ui->aboutMessage->setText(aboutMessageInfoHTML); //aboutMessage bodyy
+        ui->licenseInfo->setText(licenseInfoHTML);
+        ui->versionInfo->setText(versionInfoHTML);
+
+
+    if (about)
+    {
+        setWindowTitle(tr("About %1").arg(PACKAGE_NAME));
```
RandyMcMillan, line 73: Over riding PACKAGE_NAME is outside the scope of this PR.

This PR is implementing changes based on broader community discussions and sentiment.

REF: https://github.com/bitcoin-core/gui/issues/89#issuecomment-705384143

REF: https://github.com/bitcoin-core/gui/pull/140#issuecomment-737773556

src/bitcoind.cpp
```diff
@@ -55,10 +55,10 @@ static bool AppInit(int argc, char* argv[])
 
     // Process help and version before taking care about datadir
     if (HelpRequested(args) || args.IsArgSet("-version")) {
-        std::string strUsage = PACKAGE_NAME " version " + FormatFullVersion() + "\n";
+        std::string strUsage = PACKAGE_NAME " version " + FormatFullVersion() + "\n";//from init.cpp
 
         if (args.IsArgSet("-version")) {
-            strUsage += FormatParagraph(LicenseInfo()) + "\n";
+            strUsage += FormatParagraph(LicenseInfo()) + "\n" + CopyrightInfo() + "\n";//from init.cpp
```
RandyMcMillan, line 61: bitcoind.cpp uses the same strings that help and about use.
LicenseInfo() and CopyrightInfo() are now refactored for reusability.

src/clientversion.cpp
```diff
@@ -47,6 +47,11 @@ static std::string FormatVersion(int nVersion)
     return strprintf("%d.%d.%d", nVersion / 10000, (nVersion / 100) % 100, nVersion % 100);
 }
 
+std::string FormatVersion()
```
RandyMcMillan, line 50: Returns a terse string for reusability.
REF: https://github.com/bitcoin-core/gui/pull/140/commits/e0ff8e263dc96199385e31e481e64ad3a242fcd9#r536344384

src/init.cpp
```diff
@@ -591,23 +591,28 @@ void SetupServerArgs(NodeContext& node)
     argsman.AddHiddenArgs(hidden_args);
 }
 
+std::string VersionInfo()
```
RandyMcMillan, line 594: REF: https://github.com/bitcoin-core/gui/pull/140/commits/e0ff8e263dc96199385e31e481e64ad3a242fcd9#r536344384

REF: https://github.com/bitcoin-core/gui/pull/140/commits/e0ff8e263dc96199385e31e481e64ad3a242fcd9#r536346038

src/qt/bitcoin.qrc
RandyMcMillan: Reuse of bitcoin.png is a better approach. Thanks for the feed back!

src/qt/forms/helpmessagedialog.ui
RandyMcMillan: The view's dimensions are now dynamic and resize based on help/about context.

RandyMcMillan:
Updated default presentations:

Screen Shot 2020-12-04 at 3 19 07 PM

Screen Shot 2020-12-04 at 3 18 56 PM

src/qt/bitcoin.qrc
```diff
@@ -1,4 +1,4 @@
-<!DOCTYPE RCC><RCC version="1.0">
+<RCC>
```
jonasschnelli, line 1: is this change necessary

src/qt/utilitydialog.cpp
```diff
@@ -33,29 +33,46 @@ HelpMessageDialog::HelpMessageDialog(QWidget *parent, bool about) :
...
+        std::string copyrightInfo = CopyrightInfo();
+        // HTML-format aboutMessage elements
+        QString versionInfoHTML = QString::fromStdString(FormatVersion());
+        QString copyrightInfoHTML = QString::fromStdString(CopyrightInfo());
+        QString aboutMessageInfoHTML = QString::fromStdString(AboutMessageInfo());
         QString licenseInfoHTML = QString::fromStdString(LicenseInfo());
         // Make URLs clickable
         QRegExp uri("<(.*)>", Qt::CaseSensitive, QRegExp::RegExp2);
         uri.setMinimal(true); // use non-greedy matching
+        versionInfoHTML.replace(uri, "<a href=\"\\1\">\\1</a>");
```
jonasschnelli, line 50: why does it create a link here?

RandyMcMillan: You are correct! This string no longer needs this transform.
Awesome!

src/qt/bitcoin.qrc
RandyMcMillan: My concern with this kind of Qt Creator generated diff junk is that it forces Gui developers to
manually edit files to eliminate this stuff. This - I suspect - has contributed to bringing Gui dev to a crawl.
Technically this is valid Qt Creator generated code. I would petition members to reevaluate the policy concerning 
Qt Creator diff junk.

--- 

Please consider my comments here:

In this situation Qt Creator generates elements that it is expecting:
When I removed them it logged this error.

https://github.com/bitcoin-core/gui/pull/135#discussion_r535574885

---

PS I am beginning to suspect that when elements are missing that are expected by Qt - it triggers some kind of error handling in the Qt framework.
This is probably wasting cycles and is slowing down the over all performance of the Gui itself.
Only a theory at this point - I haven't been had time to research it further.

RandyMcMillan:
The code that is blocking this is NOT correct and should be considered a bug.
It omits the if logic and outputs bitcoind -h instead of the expected (-v) version string.

RandyMcMillan:
@practicalswift 

The conflicting code is NOT correct.
This change should have never been ACKd

It removed the if logic necessary to select help (-h) versus version (-v) output.

https://github.com/bitcoin-core/gui/commit/6690adba08006739da0060eb4937126bdfa1181a#diff-214c41e31f440626f2b20c3858589d9270c843e254f38f2f97f5098ca83ef886R60

RandyMcMillan:
Posted PR in main repo:

https://github.com/bitcoin/bitcoin/pull/20597

src/bitcoind.cpp
```diff
@@ -55,13 +55,13 @@ static bool AppInit(int argc, char* argv[])
...
-            strUsage += FormatParagraph(LicenseInfo()) + "\n";
-        } else {
-            strUsage += "\nUsage:  bitcoind [options]                     Start " PACKAGE_NAME "\n";
-            strUsage += "\n" + args.GetHelpMessage();
+        if (!args.IsArgSet("-version")) {
+            strUsage += FormatParagraph(LicenseInfo()) + "\n"
+                "\nUsage:  bitcoind [options]                     Start " PACKAGE_NAME "\n"
+                "\n";
+            strUsage += args.GetHelpMessage();
         }
```
RandyMcMillan, line 65: Resolved conflict.

RandyMcMillan: Command Line outputs are the same now.

Screen Shot 2020-12-07 at 10 55 22 PM

RandyMcMillan:
When I started this - I didn't realize how badly the Gui needs to be abstracted away from the command line.
I believe this PR is a good step toward this ideal.

Rspigler:
Whether I agree with the change or not, I don't see how it can be appropriate to slip in a name change.  Either the PR title should be changed, or a new PR created IMO.  It also doesn't make sense to me to have the name be Bitcoin Core on the loading screen, as well as `Help`->`About Bitcoin Core` yet then the about page say Bitcoin (without Core). It's just more inconsistency.

We also have other issues underway on branding discussing this issue that aren't cleared yet (#89, #100).  There's a lot of inconsistency.  Bitcoin.org uses orange; bitcoincore.org uses both; bitcoin repo uses black; bitcoin-core repo uses orange.  And now we're making additional naming changes before any decisions have been made.  Seems very messy.

RandyMcMillan:
@Rspigler - I agree it is a pretty serious issue...

There is a lot to do - and lot of decisions to be made.
This emphasizes the importance of a formal specification for branding/logo usage.

And yes...the scope of this PR has changed.
 
More importantly is - Working on the Gui has led me to believe there is a relatively easy work around to abstracting the Gui variables (PROJECT_NAME, etc..) from the build environment variables so Gui dev can progress and the plumbing can be refactored independently...

So once I know which direction to go concerning naming/branding/logo/etc...
How to proceed should be clearer...

PR #147, opened 2020-12-09
splash: New layout

RandyMcMillan:
Before:
Screen Shot 2020-12-08 at 8 51 03 PM

After:
Screen Shot 2020-12-08 at 8 45 31 PM

### DIFF
--- a/src/qt/splashscreen.cpp
+++ b/src/qt/splashscreen.cpp
@@ -30,23 +30,24 @@
 SplashScreen::SplashScreen(const NetworkStyle* networkStyle)
     : QWidget(), curAlignment(0)
 {
-    // set reference point, paddings
-    int paddingRight            = 50;
-    int paddingTop              = 50;
-    int titleVersionVSpace      = 17;
-    int titleCopyrightVSpace    = 40;
 
     float fontFactor            = 1.0;
     float devicePixelRatio      = 1.0;
+
+    // set reference point, paddings
     devicePixelRatio = static_cast<QGuiApplication*>(QCoreApplication::instance())->devicePixelRatio();
+    int paddingRight            = 180*devicePixelRatio - 100;
+    int paddingTop              = 180*devicePixelRatio + 0;
 
     // define text to place
-    QString titleText       = PACKAGE_NAME;
-    QString versionText     = QString("Version %1").arg(QString::fromStdString(FormatFullVersion()));
-    QString copyrightText   = QString::fromUtf8(CopyrightHolders(strprintf("\xc2\xA9 %u-%u ", 2009, COPYRIGHT_YEAR)).c_str());
-    QString titleAddText    = networkStyle->getTitleAddText();
+    QString titleText           = PACKAGE_NAME;
+    QString packageNameSubStr   = titleText.mid(0,7);
+    QString splashText          = QString("Connecting to the %1 network").arg(packageNameSubStr);
+    QString versionText         = QString("%1").arg(QString::fromStdString(FormatFullVersion()));
+    QString copyrightText       = QString::fromUtf8(CopyrightHolders(strprintf("\xc2\xA9 %u-%u ", 2009, COPYRIGHT_YEAR)).c_str());
+    QString titleAddText        = networkStyle->getTitleAddText();
 
-    QString font            = QApplication::font().toString();
+    QString font                = QApplication::font().toString();
 
     // create a bitmap according to device pixelratio
     QSize splashSize(480*devicePixelRatio,320*devicePixelRatio);
@@ -66,7 +67,8 @@ SplashScreen::SplashScreen(const NetworkStyle* networkStyle)
     pixPaint.fillRect(rGradient, gradient);
 
     // draw the bitcoin icon, expected size of PNG: 1024x1024
-    QRect rectIcon(QPoint(-150,-122), QSize(430,430));
+    int const icon_side = 120;
+    QRect rectIcon(QPoint((splashSize.width() - icon_side)/2,(splashSize.height() - icon_side)/8), QSize(icon_side,icon_side));
 
     const QSize requiredSize(1024,1024);
     QPixmap icon(networkStyle->getAppIcon().pixmap(requiredSize));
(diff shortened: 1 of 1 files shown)


jonasschnelli:
Usually one doesn't mix center alignment (the `Verifing blocks...`-part) with left/right alignment (the rest).
weak-NACK on removing "Core" just here. I suggest keeping it. If you want to propose a renaming, it should be an independent PR that covers a complete re-name.

RandyMcMillan:
@jonatack 
@jonasschnelli 
@Bosch-0
@laanwj 


I thought the 

The Bitcoin Core developers as a group was a strong point and I am
implementing changes based on this idea.

Can you come to some decision on this?
This is past the point of absurdity.

---

> I agree with not emphasizing "Core". Let's not institutionalize that emphasis further. Please keep them equal or put more weight on "Bitcoin".
> 
> One of my pet peeves is people writing "Core" instead of "Bitcoin Core" ("Core devs", "running Core", and so on).
> 
> See the [Bitcoin Optech style guide](https://github.com/bitcoinops/bitcoinops.github.io/blob/master/STYLE.md):
> 
> ```
> Forbidden abbreviations:
> 
> Core (use Bitcoin Core)
> ```
> 
> 100% agree with our Bitcoin Optech colleagues about this.

---

> I've already explained my rationale many times above.
> 
> I'm still of the thought that:
> 
> Bitcoin Core is a group of developers who work on Bitcoin.
> 
> Bitcoin is a network that everyone is apart of.
> 
> Electrum and BlueWallet are part of the Bitcoin ecosystem as is the GUI developed by the Core developers.

promag:
NACK, beside some alignments current splash LGTM.

As suggested by @jonasschnelli try to submit individual pull requests, in this case drop-core-from-splash and change-splash-layout.

RandyMcMillan:
@promag @jonasschnelli @laanwj - I find it strange that there are some many radical changes going on "under the hood" but changes to the Gui need to be split into "incremental" changes that will drag out Gui dev ad infinitum...

promag:
You are trying to change different things in one shot. Also these changes have no motivation, doesn't fix anything,  doesn't improve anything. I've suggested to split in 2 pull requests, or if you prefer, you can split in 2 commits, so reviewers can focus in what's more relevant to them. You can see from git log that there's an effort in having logical commits. I've already suggested this approach in https://github.com/bitcoin-core/gui/pull/135#pullrequestreview-543049890 but didn't got a response, just a 👀  reaction.

> I find it strange that there are some many radical changes going on "under the hood"

Please point what changes you refer that go without review and approval.

Bosch-0:
Here is a WIP design I did for this current splash screen - just my 2c. Planning on returning to this design work soon.

image

jonasschnelli:
I like @Bosch-0's design. Looks nice and clean.

Rspigler:
NACK. I agree with the other commentators. As discussed here (https://github.com/bitcoin-core/gui/pull/140#issuecomment-743607254) you can't slip in a name change, especially in multiple unrelated PRs, creating more confusion/inconsistency.

I like @Bosch-0 's concept, I think it would be a good middle ground for those who have expressed concern that there isn't enough showing that 'Bitcoin Core' connects to the 'Bitcoin' network.

luke-jr:
I don't know where those quotes are from, but this project/software is not "Bitcoin" - it is "Bitcoin Core".

Fine with considering a name change (to a reasonable name, not "Bitcoin"), but that should be a dedicated PR as others have said.

Personally, I like the current splash screen.

RandyMcMillan:
@Bosch-0 - I like the simplicity of your layout - plus it bypasses the whole "Bitcoin" vs "BitcoinCore" issue...
I am going to implement your layout in this PR.

RandyMcMillan:
@luke-jr - had you been a better custodian - maybe controversial issues like this wouldn't even exist?

Bosch-0:
> plus it bypasses the whole "Bitcoin" vs "BitcoinCore" issue...

Yes this was my main intention with this, makes both camps happy

RandyMcMillan:
@Bosch-0 @jonasschnelli 

WIP: Current presentation...

I am not sure why the linting fails...
https://github.com/bitcoin-core/gui/pull/147/commits/be22124d63048b7a362dddade4c17d1bd0a0fc28#diff-f63c3d5094d55f88dbd1967774f85838b3aee5a40540b8c82b924574bca772a0R187


Screen Shot 2020-12-28 at 12 10 15 PM

Screen Shot 2020-12-28 at 12 10 23 PM

hebasto:
> I am not sure why the linting fails...

https://cirrus-ci.com/task/4754423266148352?command=lint#L884:
```
src/qt/splashscreen.cpp: Expected 1 argument(s) after format string but found 0 argument(s): strprintf("%")
src/qt/splashscreen.cpp: Expected 1 argument(s) after format string but found 0 argument(s): strprintf("%")
^---- failure generated from test/lint/lint-format-strings.s
```

Bosch-0:
What's the reason the icon looks off? If you need a new icon the one I included in the above design is here > https://www.figma.com/file/0oqnohjahRtprjRyaetDOL/Bitcoin-Core-Design-System?node-id=1851%3A135

To export just click the icon > go to the menu on the right hand side > go to the export tab > export as PNG.

RandyMcMillan:
@Bosch-0 - I would prefer to limit the scope of this PR to the layout implementation only.
The art work itself - I believe - can be its own PR.
BTW - the work looks excellent! Definitely moving in the right direction! :)

Rspigler:
I think the PR should contain the proper icon before merge

RandyMcMillan:
@Bosch-0 @Rspigler 

the icon needs to be 1024x1024 RGB Alpha:Yes to match the existing artwork.
This ensures proper resolution on high res displays.

The QPainter expects a certain geometry when inserting graphics into the UI.
This requires its own testing. 

### I believe the art work needs its own PR (imo)

Original:
Screen Shot 2021-01-26 at 2 36 34 PM

Updated:
Screen Shot 2021-01-26 at 2 42 14 PM

jarolrod (commented):
Concept ACK

The`Verifying Blocks` prompt glitches on macOS 11.1 with Qt 5.15.2. 
Screen Shot 2021-01-26 at 6 37 34 PM

**Some additional thoughts:**
- Update the OP screenshot. The screenshot currently shown is not what this PR does
- The current layout as of 009b1b68d6a59057ed16667385d3d0599bd09772 doesn't exactly abide to the layout designed by @Bosch-0, but it's close. Perhaps we should get it as close as possible to his design before merge.

src/qt/splashscreen.cpp
```diff
@@ -30,23 +30,24 @@
...
+    int paddingTop              = 180*devicePixelRatio + 0;
 
     // define text to place
-    QString titleText       = PACKAGE_NAME;
-    QString versionText     = QString("Version %1").arg(QString::fromStdString(FormatFullVersion()));
-    QString copyrightText   = QString::fromUtf8(CopyrightHolders(strprintf("\xc2\xA9 %u-%u ", 2009, COPYRIGHT_YEAR)).c_str());
-    QString titleAddText    = networkStyle->getTitleAddText();
+    QString titleText           = PACKAGE_NAME;
+    QString packageNameSubStr   = titleText.mid(0,7);
+    QString splashText          = QString("Connecting to The %1 Network").arg(packageNameSubStr);
```
jarolrod, line 45:
```suggestion
    QString splashText          = QString("Connecting to the %1 network").arg(packageNameSubStr);
```
`the` and `network` should not be capitalized

jarolrod (commented):
Small change on wording

src/qt/splashscreen.cpp
RandyMcMillan: I just kinda like "title case" better - since "The Bitcoin Network" is an entity unto itself. :)
You can define it as a collection of sovereign individuals - and in this way - I believe title case is appropriate.
But I will change it...

laanwj:
Not sure why I'm tagged here, but if you want my opinion: NACK on renaming the software from Bitcoin Core, ACK on graphically more attractive splash screen.

Bosch-0:
> @Bosch-0 - I would prefer to limit the scope of this PR to the layout implementation only.
> The art work itself - I believe - can be its own PR.

I'll get on to this, concept ACK though looks good!

hebasto (commented):
Tested da1ac0f0698b0d9bc6ec00774c5882d7e4ab244c tested on Linux Mint 20.1 (Qt 5.12.8):
DeepinScreenshot_select-area_20210222220848

Hmm. GH shows 2 commits, but the fetched pr branch shows 1 commit on top of the master branch locally.

Concept ACK on @Bosch-0's design.

While this pr is a layout revamping, why do not use a Qt Designer's UI file?
It will distinguish design from coding.

RandyMcMillan:
@hebasto - I will have to search for the comment (not in src/qt/splashscreen.cpp) that stated that the splashscreen was implemented this way to work around an issue with Qt. It may have been a timing issue with launching other services simultaneously if memory serves me.

I agree - it would be optimal to use a .ui file - to eliminate a lot of layout code.

RandyMcMillan:
@Bosch-0 - checkout https://github.com/RandyMcMillan/gui/tree/new-icon for testing.

PR #148, opened 2020-12-09
Bugfix: GUI: Restore SendConfirmationDialog button default to "Yes"

luke-jr:
The SendConfirmationDialog is used for bumping the fee, where "Send" doesn't really make sense

Originally https://github.com/bitcoin/bitcoin/pull/17463, but rewritten here much simpler based on other merged changes.


hebasto (approved):
ACK 8775691383ff394b998232ac8e63fac3a214d18b, tested on Linux Mint 20.1 (x86_64, Qt 5.12.8):

Screenshot from 2021-01-10 20-28-59

Also verified that button text is translated properly (it is not the case on master):

Screenshot from 2021-01-10 20-37-53

PR #152, opened 2020-12-14
Initialise DBus notifications in another thread

luke-jr:
QDBusInterface's constructor can hang if org.freedesktop.Notifications is missing, so avoid delaying startup waiting for a timeout

### DIFF
--- a/src/qt/notificator.cpp
+++ b/src/qt/notificator.cpp
@@ -31,6 +31,18 @@
 #ifdef USE_DBUS
 // https://wiki.ubuntu.com/NotificationDevelopmentGuidelines recommends at least 128
 const int FREEDESKTOP_NOTIFICATION_ICON_SIZE = 128;
+
+void DBusInitThread::run() {
+    auto interface = new QDBusInterface("org.freedesktop.Notifications", "/org/freedesktop/Notifications", "org.freedesktop.Notifications");
+    if (!interface->isValid()) {
+        delete interface;
+        return;
+    }
+    interface->moveToThread(m_notificator.thread());
+    m_notificator.interface = interface;
+    m_notificator.mode = Notificator::Freedesktop;
+}
+
 #endif
 
 Notificator::Notificator(const QString &programName, QSystemTrayIcon *trayicon, QWidget *parent) :
@@ -48,12 +60,8 @@ Notificator::Notificator(const QString &programName, QSystemTrayIcon *trayicon, 
         mode = QSystemTray;
     }
 #ifdef USE_DBUS
-    interface = new QDBusInterface("org.freedesktop.Notifications",
-        "/org/freedesktop/Notifications", "org.freedesktop.Notifications");
-    if(interface->isValid())
-    {
-        mode = Freedesktop;
-    }
+    m_dbus_init_thread = new DBusInitThread(*this);
+    m_dbus_init_thread->start();
 #endif
 #ifdef Q_OS_MAC
     // check if users OS has support for NSUserNotification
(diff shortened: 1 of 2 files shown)


src/qt/notificator.cpp
```diff
@@ -31,6 +31,18 @@
...
+void DBusInitThread::run() {
+    auto interface = new QDBusInterface("org.freedesktop.Notifications", "/org/freedesktop/Notifications", "org.freedesktop.Notifications");
+    if (!interface->isValid()) {
+        delete interface;
+        return;
+    }
+    interface->moveToThread(m_notificator.thread());
+    m_notificator.interface = interface;
+    m_notificator.mode = Notificator::Freedesktop;
+}
```
icota, line 44: Emit [finished](https://doc.qt.io/qt-5/qthread.html#finished) here and clean up instead of keeping the thread for the `Notificator` lifecycle? `DBusInitThread` name implies it's disposable as soon as init is done.

hebasto:
> QDBusInterface's constructor can hang if org.freedesktop.Notifications is missing

[Qt docs](https://doc.qt.io/qt-5/qdbusinterface.html#QDBusInterface-1) says:
> If the remote service _service_ is not present ..., the object created will not be valid...

src/qt/notificator.cpp
luke-jr: Isn't that supposed to be an automatic part of Qt? The docs say the user can't emit it...

icota: You're right. I think it used to be public signal but seems that is no longer the case.

src/qt/notificator.h
```diff
@@ -63,11 +83,15 @@ public slots:
...
     };
     QString programName;
-    Mode mode;
+    std::atomic<Mode> mode;
     QSystemTrayIcon *trayIcon;
 #ifdef USE_DBUS
+    QThread *m_dbus_init_thread{nullptr};
+protected:
     QDBusInterface *interface;
+    friend class DBusInitThread;
```
Talkless, line 92: `friend` could be avoided if `DBusInitThread` would have `interfaceCreated(QDBusInterface* interface)` signal.

Signal can be invoked from within `run()` using:
```
QTimer::singelShot(0, this, [this, interface]() {
  emit interfaceCreated(interface);
});
```
`singleShot` will detect that `this` (context) lives in different thread and lambda will be invoked in that (receiver) context thread (where `DBusInitThread` was created).

In the result, `std::atomic<Mode> mode;` "atomicity" will not be needed, because 
```
  this->interface = interface;
  mode = Notificator::Freedesktop;
```
..would be performed on thread `Notificator` lives, within a slot (or just lambda) that connected to the `DBusInitThread::interfaceCreated(QDBusInterface*)` signal, and so without *any* race condition, as we are setting two variables, and only one is atomic.

`DBusInitThread` constructor will have to get `QThread*` though for `moveToThread().`. Or jus use `DBusInitThread::thread()` to get that "destination" thread? But that's might break if `interface`-receiving thread is not the same that created `DBusInitThread` object...

Also, adding `interface->setParent(this)` after `this->interface = interface;` would remove need for manual `delete`, unless there's deletion ordering is involved...

Talkless (commented):
Concept ACK.

hebasto:
A related discussion on [IRC](https://www.erisian.com.au/bitcoin-core-dev/log-2022-05-17.html#l-314):
> \<luke-jr> oh, I guess Core never merged the dbus timeout fix  - that's probably why -qt is slower
> \<luke-jr> https://github.com/bitcoin-core/gui/pull/152
> \<hebasto> luke-jr: it is not obvious that bug exist, in first place -- https://github.com/bitcoin-core/gui/pull/152#issuecomment-757524350
> \<luke-jr> hebasto: it definitely exists
> \<hebasto> how to reproduce it reliably?
> \<luke-jr> idk, happens every time for me
> \<luke-jr> maybe running bitcoin-qt as another user?
> \<hebasto> ok, could you describe your steps and observations in your PR?
> \<luke-jr> there's no steps, it just happens
> \<hebasto> at least, you could mention your system on which "org.freedesktop.Notifications is missing"

PR #155, opened 2020-12-17
Fix checkbox layout in Create Wallet dialog

hebasto:
From https://github.com/bitcoin/bitcoin/issues/20555#issuecomment-747742627:
> 0.21.0rc3 compiled from source, Ubuntu 20.04 + i3: tested mainnet, testnet, and signet. created a descriptor wallet in qt, tested torv3 with addnode, and checked if anchors.dat is created on shutdown and removed on startup.
> 
> I've noticed the "Advanced options" label being cut-off in the wallet creation dialog.

This PR should fix this issue.

### DIFF
--- a/src/qt/forms/createwalletdialog.ui
+++ b/src/qt/forms/createwalletdialog.ui
@@ -60,7 +60,7 @@
     <rect>
      <x>20</x>
      <y>50</y>
-     <width>171</width>
+     <width>220</width>
      <height>22</height>
     </rect>
    </property>
@@ -79,7 +79,7 @@
     <rect>
      <x>20</x>
      <y>90</y>
-     <width>130</width>
+     <width>220</width>
      <height>21</height>
     </rect>
    </property>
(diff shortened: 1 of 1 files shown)


MarcoFalke:
cc @0xB10C

0xB10C:
This seems to fix it! Thanks.

rc3:
image


pr155:
image

MarcoFalke:
review ACK e71b656f317f38ef0ba0874736f116dae39efc67

didn't test

jonasschnelli:
Shouldn't this be a flexible (stretch layout instead of fixed pixles?

MarcoFalke:
This can be reworked for 0.22, but I think for rc4 we should aim for a minimal fix

hebasto:
> Shouldn't this be a flexible (stretch layout instead of fixed pixles?

The goal of this PR was the minimal fix as @MarcoFalke [said](https://github.com/bitcoin-core/gui/pull/155#issuecomment-747932196). Besides, the issue has spotted on i3 window manager, and the real problem could be much deeper.

jonasschnelli:
post merge utACK the fix. I missed the rc4 issue or 0.21 tag...

PR #157, opened 2020-12-18
Refactor P2PKH filtering, via AddressBookSortFilterProxyModel

luke-jr:
This didn't accomplish what I hoped, but maybe worth including as a refactor anyway.

There's obviously further improvement possible, but this already seems like quite a bit for one refactoring step.

### DIFF
--- a/src/qt/addressbookpage.cpp
+++ b/src/qt/addressbookpage.cpp
@@ -23,11 +23,13 @@
 class AddressBookSortFilterProxyModel final : public QSortFilterProxyModel
 {
     const QString m_type;
+    const bool m_p2pkh_only;
 
 public:
-    AddressBookSortFilterProxyModel(const QString& type, QObject* parent)
+    AddressBookSortFilterProxyModel(const QString& type, const bool p2pkh_only, QObject* parent)
         : QSortFilterProxyModel(parent)
         , m_type(type)
+        , m_p2pkh_only(p2pkh_only)
     {
         setDynamicSortFilter(true);
         setFilterCaseSensitivity(Qt::CaseInsensitive);
@@ -41,6 +43,9 @@ protected:
         auto label = model->index(row, AddressTableModel::Label, parent);
 
         if (model->data(label, AddressTableModel::TypeRole).toString() != m_type) {
+            return false;
+        }
+        if (m_p2pkh_only && !model->data(label, AddressTableModel::IsP2PKHRole).toBool()) {
             return false;
         }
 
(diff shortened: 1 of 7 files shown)


hebasto:
> This didn't accomplish what I hoped...

What do you hope?

luke-jr:
>What do you hope?

I'd like the new "signmessage for p2pkh only" to only show if there are actually non-p2pkh addresses being hidden.

src/qt/addresstablemodel.cpp
```diff
@@ -88,7 +89,8 @@ class AddressTablePriv
                         QString::fromStdString(address.purpose), address.is_mine);
                 cachedAddressTable.append(AddressTableEntry(addressType,
                                   QString::fromStdString(address.name),
-                                  QString::fromStdString(EncodeDestination(address.dest))));
+                                  QString::fromStdString(EncodeDestination(address.dest)),
+                                  (address.dest.type() == typeid(PKHash))));
```
promag, line 91: 6604abe6ab25925b379b3f16630906de3aaf51fb

I think you could do `std::get_if<PKHash>(address.dest) != nullptr`. Same below, and drop <typeinfo> header.

nit, commit prefix should be `qt`?

src/qt/addressbookpage.cpp
```diff
@@ -145,14 +150,14 @@ AddressBookPage::~AddressBookPage()
     delete ui;
 }
 
-void AddressBookPage::setModel(AddressTableModel *_model)
+void AddressBookPage::setModel(AddressTableModel *_model, const bool p2pkh_only)
```
promag, line 153: 86be9e4a353f8c2ab35c3c1177e8ad63b8e50eab

nit, this could be renamed `AddressBookPage::setupModel(AddressTableModel* source_model, const bool p2pkh_only)`. Alternatively these could be passed in `AddressBookPage` constructor and this method would be private.

src/qt/addresstablemodel.cpp
Talkless: There's [std::holds_alternative](https://en.cppreference.com/w/cpp/utility/variant/holds_alternative) for that.

hebasto:
Concept ACK. Rebasing?

src/qt/addressbookpage.cpp
jarolrod: This suggestion looks reasonable to me, @luke-jr will you get back to this?

hebasto:
Closing due to PR author's inactivity. Labeled "Up for grabs".

PR #161, opened 2020-12-23
Add PeerTableModel::StatsRole to prevent data layer violation

hebasto:
This PR allows to access to the `CNodeCombinedStats` instance directly from any view object.

The `PeerTableModel::getNodeStats` member function removed as a kind of layer violation.

No behavior changes.

Also other pulls (bugfixes) are based on this one: #18 and #164.

### DIFF
--- a/src/qt/peertablemodel.h
+++ b/src/qt/peertablemodel.h
@@ -28,6 +28,7 @@ struct CNodeCombinedStats {
     CNodeStateStats nodeStateStats;
     bool fNodeStateStatsAvailable;
 };
+Q_DECLARE_METATYPE(CNodeCombinedStats*)
 
 class NodeLessThan
 {
@@ -65,6 +66,10 @@ public:
         Sent = 4,
         Received = 5,
         Subversion = 6
+    };
+
+    enum {
+        StatsRole = Qt::UserRole,
     };
 
     /** @name Methods overridden from QAbstractTableModel
--- a/src/qt/peertablemodel.h
+++ b/src/qt/peertablemodel.h
@@ -53,7 +53,6 @@ class PeerTableModel : public QAbstractTableModel
 public:
     explicit PeerTableModel(interfaces::Node& node, QObject* parent);
     ~PeerTableModel();
-    const CNodeCombinedStats *getNodeStats(int idx);
     int getRowByNodeId(NodeId nodeid);
     void startAutoRefresh();
     void stopAutoRefresh();
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -1018,11 +1018,9 @@ void RPCConsole::updateTrafficStats(quint64 totalBytesIn, quint64 totalBytesOut)
 
 void RPCConsole::peerLayoutAboutToChange()
 {
-    QModelIndexList selected = ui->peerWidget->selectionModel()->selectedIndexes();
     cachedNodeids.clear();
-    for(int i = 0; i < selected.size(); i++)
-    {
-        const CNodeCombinedStats *stats = clientModel->getPeerTableModel()->getNodeStats(selected.at(i).row());
+    for (const QModelIndex& peer : GUIUtil::getEntryData(ui->peerWidget, PeerTableModel::NetNodeId)) {
+        const auto stats = peer.data(PeerTableModel::StatsRole).value<CNodeCombinedStats*>();
         cachedNodeids.append(stats->nodeStats.nodeid);
     }
 }
@@ -1081,15 +1079,13 @@ void RPCConsole::peerLayoutChanged()
 
 void RPCConsole::updateDetailWidget()
 {
-    QModelIndexList selected_rows;
-    auto selection_model = ui->peerWidget->selectionModel();
-    if (selection_model) selected_rows = selection_model->selectedRows();
-    if (!clientModel || !clientModel->getPeerTableModel() || selected_rows.size() != 1) {
+    const QList<QModelIndex> selected_peers = GUIUtil::getEntryData(ui->peerWidget, PeerTableModel::NetNodeId);
+    if (!clientModel || !clientModel->getPeerTableModel() || selected_peers.size() != 1) {
         ui->detailWidget->hide();
         ui->peerHeading->setText(tr("Select a peer to view detailed information."));
         return;
     }
-    const CNodeCombinedStats *stats = clientModel->getPeerTableModel()->getNodeStats(selected_rows.first().row());
+    const auto stats = selected_peers.first().data(PeerTableModel::StatsRole).value<CNodeCombinedStats*>();
     // update the detail ui with latest node information
     QString peerAddrDetails(QString::fromStdString(stats->nodeStats.addrName) + " ");
     peerAddrDetails += tr("(peer id: %1)").arg(QString::number(stats->nodeStats.nodeid));
@@ -1202,19 +1198,9 @@ void RPCConsole::banSelectedNode(int bantime)
     if (!clientModel)
         return;
 
-    // Get selected peer addresses
-    QList<QModelIndex> nodes = GUIUtil::getEntryData(ui->peerWidget, PeerTableModel::NetNodeId);
-    for(int i = 0; i < nodes.count(); i++)
-    {
-        // Get currently selected peer address
-        NodeId id = nodes.at(i).data().toLongLong();
-
-        // Get currently selected peer address
-        int detailNodeRow = clientModel->getPeerTableModel()->getRowByNodeId(id);
-        if (detailNodeRow < 0) return;
-
+    for (const QModelIndex& peer : GUIUtil::getEntryData(ui->peerWidget, PeerTableModel::NetNodeId)) {
         // Find possible nodes, ban it and clear the selected node
-        const CNodeCombinedStats *stats = clientModel->getPeerTableModel()->getNodeStats(detailNodeRow);
+        const auto stats = peer.data(PeerTableModel::StatsRole).value<CNodeCombinedStats*>();
         if (stats) {
             m_node.ban(stats->nodeStats.addr, bantime);
             m_node.disconnectByAddress(stats->nodeStats.addr);
--- a/src/qt/peertablemodel.cpp
+++ b/src/qt/peertablemodel.cpp
@@ -221,11 +221,6 @@ QModelIndex PeerTableModel::index(int row, int column, const QModelIndex &parent
     return QModelIndex();
 }
 
-const CNodeCombinedStats *PeerTableModel::getNodeStats(int idx)
-{
-    return priv->index(idx);
-}
-
 void PeerTableModel::refresh()
 {
     Q_EMIT layoutAboutToBeChanged();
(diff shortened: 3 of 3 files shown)


hebasto:
Reworked from scratch.

src/qt/rpcconsole.cpp
```diff
@@ -1018,11 +1018,9 @@ void RPCConsole::updateTrafficStats(quint64 totalBytesIn, quint64 totalBytesOut)
 
 void RPCConsole::peerLayoutAboutToChange()
 {
-    QModelIndexList selected = ui->peerWidget->selectionModel()->selectedIndexes();
     cachedNodeids.clear();
-    for(int i = 0; i < selected.size(); i++)
-    {
-        const CNodeCombinedStats *stats = clientModel->getPeerTableModel()->getNodeStats(selected.at(i).row());
+    for (const auto& peer : GUIUtil::getEntryData(ui->peerWidget, PeerTableModel::NetNodeId)) {
+        const auto stats = peer.data(PeerTableModel::StatsRole).value<CNodeCombinedStats*>();
```
jonatack, line 1023: 6e27ee8b6 To me at least, this would be easier to understand what is going on (and not using `auto` allows uniform initialization). The same motivation for the other suggestions below.
```diff
-    for (const auto& peer : GUIUtil::getEntryData(ui->peerWidget, PeerTableModel::NetNodeId)) {
-        const auto stats = peer.data(PeerTableModel::StatsRole).value<CNodeCombinedStats*>();
+    for (const QModelIndex& peer : GUIUtil::getEntryData(ui->peerWidget, PeerTableModel::NetNodeId)) {
+        const CNodeCombinedStats* stats{peer.data(PeerTableModel::StatsRole).value<CNodeCombinedStats*>()};
```

src/qt/rpcconsole.cpp
```diff
@@ -1081,15 +1079,13 @@ void RPCConsole::peerLayoutChanged()
 
 void RPCConsole::updateDetailWidget()
 {
-    QModelIndexList selected_rows;
-    auto selection_model = ui->peerWidget->selectionModel();
-    if (selection_model) selected_rows = selection_model->selectedRows();
-    if (!clientModel || !clientModel->getPeerTableModel() || selected_rows.size() != 1) {
+    const auto selected_peers = GUIUtil::getEntryData(ui->peerWidget, PeerTableModel::NetNodeId);
```
jonatack, line 1082: 6e27ee8b6

```suggestion
    const QList<QModelIndex> selected_peers{GUIUtil::getEntryData(ui->peerWidget, PeerTableModel::NetNodeId)};
```

src/qt/rpcconsole.cpp
```diff
@@ -1081,15 +1079,13 @@ void RPCConsole::peerLayoutChanged()
...
+    if (!clientModel || !clientModel->getPeerTableModel() || selected_peers.size() != 1) {
         ui->detailWidget->hide();
         ui->peerHeading->setText(tr("Select a peer to view detailed information."));
         return;
     }
-    const CNodeCombinedStats *stats = clientModel->getPeerTableModel()->getNodeStats(selected_rows.first().row());
+    const auto stats = selected_peers.first().data(PeerTableModel::StatsRole).value<CNodeCombinedStats*>();
```
jonatack, line 1088: 6e27ee8
```suggestion
    const CNodeCombinedStats* stats{selected_peers.first().data(PeerTableModel::StatsRole).value<CNodeCombinedStats*>()};
```

src/qt/rpcconsole.cpp
```diff
@@ -1201,19 +1197,9 @@ void RPCConsole::banSelectedNode(int bantime)
...
-        NodeId id = nodes.at(i).data().toLongLong();
-
-        // Get currently selected peer address
-        int detailNodeRow = clientModel->getPeerTableModel()->getRowByNodeId(id);
-        if (detailNodeRow < 0) return;
-
+    for (const auto& peer : GUIUtil::getEntryData(ui->peerWidget, PeerTableModel::NetNodeId)) {
         // Find possible nodes, ban it and clear the selected node
-        const CNodeCombinedStats *stats = clientModel->getPeerTableModel()->getNodeStats(detailNodeRow);
+        const auto stats = peer.data(PeerTableModel::StatsRole).value<CNodeCombinedStats*>();
```
jonatack, line 1202: 6e27ee8b67a8d8a81038b58e8503a3ab0d22f881
```diff
-    for (const auto& peer : GUIUtil::getEntryData(ui->peerWidget, PeerTableModel::NetNodeId)) {
+    for (const QModelIndex& peer : GUIUtil::getEntryData(ui->peerWidget, PeerTableModel::NetNodeId)) {
         // Find possible nodes, ban it and clear the selected node
-        const auto stats = peer.data(PeerTableModel::StatsRole).value<CNodeCombinedStats*>();
+        const CNodeCombinedStats* stats{peer.data(PeerTableModel::StatsRole).value<CNodeCombinedStats*>()};
```

src/qt/peertablemodel.h
```diff
@@ -66,6 +66,10 @@ class PeerTableModel : public QAbstractTableModel
         Subversion = 5
     };
 
+    enum {
+        StatsRole = Qt::UserRole
+    };
```
jonatack, line 71: 644f5f5 I'm not very familiar with the gui code, but this enum seems an odd way to do this. Maybe an explanatory comment would be good.

jonatack (commented):
Light ACK 4c05fe01ed05b549558f135cddbe720ba568ecbc with the caveat that I don't know the qt framework or this code well. Saw no regression while manually testing. A few suggestions below to aid in reading and understanding the code.

src/qt/peertablemodel.h
hebasto: As an alternative, a simple `constexpr` could be used (it seems better for me), but here I use the way how we already did such things:
https://github.com/bitcoin-core/gui/blob/f1f26b8d5baec4a45a3a9ba0440cd4eff7af8407/src/qt/addresstablemodel.h#L36-L38

https://github.com/bitcoin-core/gui/blob/f1f26b8d5baec4a45a3a9ba0440cd4eff7af8407/src/qt/bitcoinunits.h#L91-L94

https://github.com/bitcoin-core/gui/blob/f1f26b8d5baec4a45a3a9ba0440cd4eff7af8407/src/qt/coincontroldialog.h#L82-L86

https://github.com/bitcoin-core/gui/blob/f1f26b8d5baec4a45a3a9ba0440cd4eff7af8407/src/qt/transactiontablemodel.h#L46

@promag What do you think?

src/qt/rpcconsole.cpp
hebasto: [AAA Style (Almost Always Auto)](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/)

> ... easier to understand what is going on

Maybe `GUIUtil::getEntryData` requires renaming? `GetModelIndexesByColumn`?

@promag What do you think?

src/qt/rpcconsole.cpp
promag: Type is already in `.value<>()`, IIRC `auto` can be used when the type is clear from the statement.

src/qt/peertablemodel.h
promag: Right, this follow Qt approach, I guess they didn't want to manually assign a value to each role.

With multiple roles `constexpr` looks weird because you'd have to manually increment for each custom role.

src/qt/rpcconsole.cpp
promag: Side node, I'd just ditch `getEntryData` and `hasEntryData`, doesn't improve that much over Qt API. If we keep them then maybe reword to `GetSelectedModelIndex` or something like that with "selection".

In this particular case I think `peer` should have the type, and leave auto for `status`.

src/qt/rpcconsole.cpp
jonatack: True

src/qt/rpcconsole.cpp
jonatack: Re AAA Style, yes, Sutter states the caveat about readability is that people use IDEs. Here, where review is the primary constraint, whatever can aid reviewers in reading and reviewing the code seems like a win, so in practice I've found myself using `auto` only when the type is obvious or when it can avoid a complicated, verbose type. Anyway, just a minor suggestion, feel free to ignore.

src/qt/peertablemodel.h
hebasto: @promag 
> ... this follow Qt approach...

In the [recent example](https://doc.qt.io/qt-5.12/qtwidgets-painting-basicdrawing-example.html) they use
```
const int IdRole = Qt::UserRole;
```


> With multiple roles `constexpr` looks weird because you'd have to manually increment for each custom role.

Agree.

Also https://stackoverflow.com/questions/899917/why-do-people-use-enums-in-c-as-constants-while-they-can-use-const

jonatack: Thanks :)

hebasto:
Updated 4c05fe01ed05b549558f135cddbe720ba568ecbc -> b3e9bcaac85a64a1b41d534b28c8cfb1f08e14e5 ([pr161.02](https://github.com/hebasto/gui/commits/pr161.02) -> [pr161.03](https://github.com/hebasto/gui/commits/pr161.03)):

- addressed @jonatack and @promag's comments
- rebased to prevent conflicts in pulls that based on this one (#18 and #164)

src/qt/rpcconsole.cpp
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui/pull/161#issuecomment-752735925).

src/qt/rpcconsole.cpp
hebasto: [Updated](https://github.com/bitcoin-core/gui/pull/161#issuecomment-752735925).

src/qt/rpcconsole.cpp
hebasto: Marked as "resolved" without changes.

src/qt/rpcconsole.cpp
hebasto: [Updated](https://github.com/bitcoin-core/gui/pull/161#issuecomment-752735925).

jonatack (commented):
Tested re-ACK b3e9bcaac85a64a1b41d534b28c8cfb1f08e14e5 per `git range-diff ae8f797 4c05fe0 b3e9bca`

Thanks for updating.

promag (approved):
Code review ACK b3e9bcaac85a64a1b41d534b28c8cfb1f08e14e5.

As follow ups, maybe it could make sense to split each member of
```cpp
struct CNodeCombinedStats {
    CNodeStats nodeStats;
    CNodeStateStats nodeStateStats;
    bool fNodeStateStatsAvailable;
};
```
on a dedicated role.

Or just add some other roles like:
```diff
- const auto stats = peer.data(PeerTableModel::StatsRole).value<CNodeCombinedStats*>();
- cachedNodeids.append(stats->nodeStats.nodeid);
+ cachedNodeids.append(peer.data(PeerTableModel::NodeIdRole).toInt());
```

jonasschnelli (approved):
utACK b3e9bcaac85a64a1b41d534b28c8cfb1f08e14e5

PR #162, opened 2020-12-24
Add network to peers window and peer details

jonatack:
and rename peers window column headers from NodeId and Node/Service to Peer Id and Address.

Screenshot from 2020-12-27 14-45-31

### DIFF
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -1100,14 +1100,17 @@
                 </widget>
                </item>
                <item row="2" column="0">
-                <widget class="QLabel" name="label_21">
+                <widget class="QLabel" name="peerNetworkLabel">
+                 <property name="toolTip">
+                  <string>The network protocol this peer is connected through: IPv4, IPv6, Onion, I2P, or CJDNS.</string>
+                 </property>
                  <property name="text">
-                  <string>Version</string>
+                  <string>Network</string>
                  </property>
                 </widget>
                </item>
                <item row="2" column="1">
-                <widget class="QLabel" name="peerVersion">
+                <widget class="QLabel" name="peerNetwork">
                  <property name="cursor">
                   <cursorShape>IBeamCursor</cursorShape>
                  </property>
@@ -1123,14 +1126,14 @@
                 </widget>
                </item>
                <item row="3" column="0">
-                <widget class="QLabel" name="label_28">
+                <widget class="QLabel" name="label_21">
                  <property name="text">
-                  <string>User Agent</string>
+                  <string>Version</string>
                  </property>
                 </widget>
                </item>
                <item row="3" column="1">
-                <widget class="QLabel" name="peerSubversion">
+                <widget class="QLabel" name="peerVersion">
                  <property name="cursor">
                   <cursorShape>IBeamCursor</cursorShape>
                  </property>
@@ -1146,14 +1149,14 @@
                 </widget>
                </item>
                <item row="4" column="0">
-                <widget class="QLabel" name="label_4">
+                <widget class="QLabel" name="label_28">
                  <property name="text">
-                  <string>Services</string>
+                  <string>User Agent</string>
                  </property>
                 </widget>
                </item>
                <item row="4" column="1">
-                <widget class="QLabel" name="peerServices">
+                <widget class="QLabel" name="peerSubversion">
                  <property name="cursor">
                   <cursorShape>IBeamCursor</cursorShape>
                  </property>
@@ -1169,14 +1172,14 @@
                 </widget>
                </item>
                <item row="5" column="0">
-                <widget class="QLabel" name="label_29">
+                <widget class="QLabel" name="label_4">
                  <property name="text">
-                  <string>Starting Block</string>
+                  <string>Services</string>
                  </property>
                 </widget>
                </item>
                <item row="5" column="1">
-                <widget class="QLabel" name="peerHeight">
+                <widget class="QLabel" name="peerServices">
                  <property name="cursor">
                   <cursorShape>IBeamCursor</cursorShape>
                  </property>
@@ -1192,14 +1195,14 @@
                 </widget>
                </item>
                <item row="6" column="0">
-                <widget class="QLabel" name="label_27">
+                <widget class="QLabel" name="label_29">
                  <property name="text">
-                  <string>Synced Headers</string>
+                  <string>Starting Block</string>
                  </property>
                 </widget>
                </item>
                <item row="6" column="1">
-                <widget class="QLabel" name="peerSyncHeight">
+                <widget class="QLabel" name="peerHeight">
                  <property name="cursor">
                   <cursorShape>IBeamCursor</cursorShape>
                  </property>
@@ -1215,14 +1218,14 @@
                 </widget>
                </item>
                <item row="7" column="0">
-                <widget class="QLabel" name="label_25">
+                <widget class="QLabel" name="label_27">
                  <property name="text">
-                  <string>Synced Blocks</string>
+                  <string>Synced Headers</string>
                  </property>
                 </widget>
                </item>
                <item row="7" column="1">
-                <widget class="QLabel" name="peerCommonHeight">
+                <widget class="QLabel" name="peerSyncHeight">
                  <property name="cursor">
                   <cursorShape>IBeamCursor</cursorShape>
                  </property>
@@ -1238,14 +1241,14 @@
                 </widget>
                </item>
                <item row="8" column="0">
-                <widget class="QLabel" name="label_22">
+                <widget class="QLabel" name="label_25">
                  <property name="text">
-                  <string>Connection Time</string>
+                  <string>Synced Blocks</string>
                  </property>
                 </widget>
                </item>
                <item row="8" column="1">
-                <widget class="QLabel" name="peerConnTime">
+                <widget class="QLabel" name="peerCommonHeight">
                  <property name="cursor">
                   <cursorShape>IBeamCursor</cursorShape>
                  </property>
@@ -1261,14 +1264,14 @@
                 </widget>
                </item>
                <item row="9" column="0">
-                <widget class="QLabel" name="label_15">
+                <widget class="QLabel" name="label_22">
                  <property name="text">
-                  <string>Last Send</string>
+                  <string>Connection Time</string>
                  </property>
                 </widget>
                </item>
                <item row="9" column="1">
-                <widget class="QLabel" name="peerLastSend">
+                <widget class="QLabel" name="peerConnTime">
                  <property name="cursor">
                   <cursorShape>IBeamCursor</cursorShape>
                  </property>
@@ -1284,14 +1287,14 @@
                 </widget>
                </item>
                <item row="10" column="0">
-                <widget class="QLabel" name="label_19">
+                <widget class="QLabel" name="label_15">
                  <property name="text">
-                  <string>Last Receive</string>
+                  <string>Last Send</string>
                  </property>
                 </widget>
                </item>
                <item row="10" column="1">
-                <widget class="QLabel" name="peerLastRecv">
+                <widget class="QLabel" name="peerLastSend">
                  <property name="cursor">
                   <cursorShape>IBeamCursor</cursorShape>
                  </property>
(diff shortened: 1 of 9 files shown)


hebasto:
Concept ACK.

Sjors:
tACK b1704ead912ed813eaa08e80af5e14a2c087574e

Especially handy because IPv6 addresses are turned into ellipses until you expand it:

Schermafbeelding 2020-12-24 om 12 29 34

jonatack:
> This column might be a more natural place for the inbound/outbound arrow.

Hm, good idea. Am about to open a PR to add the direction+connection type, maybe move the arrow there.

jonatack:
> This column might be a more natural place for the inbound/outbound arrow.

Thanks @Sjors. Good idea, done.

RandyMcMillan:
@jonatack 

https://github.com/bitcoin-core/gui/pull/135

In this PR the direction has its own column. This makes the direction sortable.
Having the network type and the direction individually sortable would be great.

jonatack:
@RandyMcMillan not a problem, will just drop the last commit if that PR is merged

src/qt/forms/debugwindow.ui
```diff
@@ -1100,13 +1100,39 @@
                 </widget>
                </item>
                <item row="2" column="0">
+                <widget class="QLabel" name="peerNetworkLabel">
+                 <property name="toolTip">
+                  <string>The network protocol this peer is connected through.</string>
+                 </property>
```
hebasto, line 1106: b1704ead912ed813eaa08e80af5e14a2c087574e

It is inconsistent to have a tooltip only for this label among others.

src/qt/guiutil.cpp
```diff
@@ -749,6 +749,14 @@ QString boostPathToQString(const fs::path &path)
...
 }
 
+QString NetworkNameToQString(const std::string& network)
+{
+    if (network == "ipv4") return "IPv4";
+    if (network == "ipv6") return "IPv6";
+    if (network == "onion") return "Onion";
+    return "";
+}
+
```
hebasto, line 759: 97d5b82beaf35761bfcc31562f525f4bc52d6529

It seems strange for me to add a new function for the sake of a such functionality. And string interface is fragile.

Why do not expose `Network` in `CNodeStats` (the same approach you use in #163)?

src/qt/peertablemodel.cpp
```diff
@@ -104,7 +106,7 @@ PeerTableModel::PeerTableModel(interfaces::Node& node, QObject* parent) :
     m_node(node),
     timer(nullptr)
 {
-    columns << tr("NodeId") << tr("Node/Service") << tr("Ping") << tr("Sent") << tr("Received") << tr("User Agent");
+    columns << tr("Node Id") << tr("Node/Service") << tr("Network") << tr("Ping") << tr("Sent") << tr("Received") << tr("User Agent");
```
hebasto, line 109: c772533ce5f471b5aa968ac864cf8a2173900fac
As you touching the headers, mind considering the following patch
```diff
--- a/src/qt/peertablemodel.cpp
+++ b/src/qt/peertablemodel.cpp
@@ -106,7 +106,6 @@ PeerTableModel::PeerTableModel(interfaces::Node& node, QObject* parent) :
     m_node(node),
     timer(nullptr)
 {
-    columns << tr("Node Id") << tr("Node/Service") << tr("Network") << tr("Ping") << tr("Sent") << tr("Received") << tr("User Agent");
     priv.reset(new PeerTablePriv());
 
     // set up timer for auto refresh
diff --git a/src/qt/peertablemodel.h b/src/qt/peertablemodel.h
index 39cbe16df..61a0132e0 100644
--- a/src/qt/peertablemodel.h
+++ b/src/qt/peertablemodel.h
@@ -83,7 +83,7 @@ public Q_SLOTS:
 
 private:
     interfaces::Node& m_node;
-    QStringList columns;
+    const QStringList columns{tr("Peer Id"), tr("Address"), tr("Network"), tr("Ping"), tr("Sent"), tr("Received"), tr("User Agent")};
     std::unique_ptr<PeerTablePriv> priv;
     QTimer *timer;
 };
```
?

hebasto (changes requested):
I have reviewed the code (79e997e826b93595896839fd513a052b7a4d001d), and tested it.

**UX notes.**

_Disclaimer_: I'm not a design expert, and my opinions are based exclusively on my experience.

> In this PR the direction has its own column. This makes the direction sortable.
> Having the network type and the direction individually sortable would be great.

I've been waiting for this feature for many years :)

> > This column might be a more natural place for the inbound/outbound arrow.
> 
> Thanks @Sjors. Good idea, done.

I don't think this change is valuable. It is not possible to sort out all connections through a particular network.

src/qt/forms/debugwindow.ui
jonatack: There is one for the Mapped AS field as well. Probably best to add the missing ones but not in this pull.

src/qt/guiutil.cpp
jonatack: Agreed, I wanted to both here and in #163 but I didn't think it would be accepted by reviewers. Updating to switch on the enum instead and add i2p and cjdns.

src/qt/peertablemodel.cpp
jonatack: sure, done in c2cd8267cb21d496bcf66070b3b9b18d5f181ea5 (can drop the clang-formatting changes in that commit but some of the conditionals have dangerous, error-prone formatting)

src/qt/guiutil.cpp
jonatack: done -- I agree it's a shame to add a function but the GUI UI is non all-lowercase like the RPC/CLI interface

jonatack:
Thanks for the feedback. I dropped the arrows commit (I think there is a better approach possible) and took most of @hebasto's suggestions.

hebasto:
> (Can drop the clang-formatting changes in the refactoring commit but some of the conditionals have error-prone formatting.)

The most of the re-formatted code could go away in #18 :)

src/qt/peertablemodel.cpp
```diff
@@ -15,29 +15,29 @@
...
     case PeerTableModel::Ping:
         return pLeft->m_min_ping_usec < pRight->m_min_ping_usec;
     case PeerTableModel::Sent:
         return pLeft->nSendBytes < pRight->nSendBytes;
     case PeerTableModel::Received:
         return pLeft->nRecvBytes < pRight->nRecvBytes;
-    }
+    case PeerTableModel::Subversion:
+        return pLeft->cleanSubVer.compare(pRight->cleanSubVer) < 0;
+    } // no default case, so the compiler can warn about missing cases
```
hebasto, line 42: I think this comment is not true, as the `switch` parameter is an `enum`, not a `enum class`.

src/qt/guiutil.cpp
```diff
@@ -749,6 +749,21 @@ QString boostPathToQString(const fs::path &path)
     return QString::fromStdString(path.string());
 }
 
+QString NetworkToQString(Network net)
+{
+    switch (net) {
+    case NET_UNROUTABLE: return "Unroutable";
```
hebasto, line 755:
```suggestion
    case NET_UNROUTABLE: return tr("Unroutable");
```

src/qt/guiutil.cpp
```diff
@@ -749,6 +749,21 @@ QString boostPathToQString(const fs::path &path)
...
+    case NET_IPV4: return "IPv4";
+    case NET_IPV6: return "IPv6";
+    case NET_ONION: return "Onion";
+    case NET_I2P: return "I2P";
+    case NET_CJDNS: return "CJDNS";
+    case NET_INTERNAL: return "Internal";
```
hebasto, line 761:
```suggestion
    case NET_INTERNAL: return tr("Internal");
```

hebasto (approved):
ACK 67e8b636a88daf47acf01f04e1e1b62b6f200554

src/qt/peertablemodel.cpp
jonatack: You're right--thanks!

hebasto:
@jonatack 
Could this patch makes things look better
```diff
--- a/src/qt/peertablemodel.cpp
+++ b/src/qt/peertablemodel.cpp
@@ -168,6 +168,8 @@ QVariant PeerTableModel::data(const QModelIndex& index, int role) const
         }
     } else if (role == Qt::TextAlignmentRole) {
         switch (index.column()) {
+        case Network:
+            return QVariant(Qt::AlignCenter);
         case Ping:
         case Sent:
         case Received:
```
?

hebasto:
Due to the header titles are changed this patch seems required:
```diff
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -1092,7 +1092,7 @@ void RPCConsole::updateDetailWidget()
     const CNodeCombinedStats *stats = clientModel->getPeerTableModel()->getNodeStats(selected_rows.first().row());
     // update the detail ui with latest node information
     QString peerAddrDetails(QString::fromStdString(stats->nodeStats.addrName) + " ");
-    peerAddrDetails += tr("(node id: %1)").arg(QString::number(stats->nodeStats.nodeid));
+    peerAddrDetails += tr("(peer id: %1)").arg(QString::number(stats->nodeStats.nodeid));
     if (!stats->nodeStats.addrLocal.empty())
         peerAddrDetails += "<br />" + tr("via %1").arg(QString::fromStdString(stats->nodeStats.addrLocal));
     ui->peerHeading->setText(peerAddrDetails);
```

src/qt/guiutil.cpp
jonatack: thanks, done `QObject::tr("Unroutable")`

src/qt/guiutil.cpp
jonatack: thanks, done

jonatack:
> @jonatack
> Could this patch make things look better

Agree; for Peer Id too.

jonatack:
Thanks for the good feedback @hebasto. Added translation, changed the network column align to center, added a commit in your name for the header renaming + initialize `columns` in the .h, fixed the doxygen broken doxygen formatting, dropped the clang formatting.

hebasto (approved):
ACK 12212e2b30b87110f81787d6e10259d02d807279, tested on Linux Mint 20 (x86_64) + Qt 5.12.8.

Looks great:
Screenshot from 2020-12-25 20-26-53

Also I've verified that the "network" field in the `getpeerinfo` RPC has not changed behavior.

With this patch
```diff
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -566,7 +566,7 @@ void CNode::copyStats(CNodeStats &stats, const std::vector<bool> &m_asmap)
     X(nServices);
     X(addr);
     X(addrBind);
-    stats.m_network = ConnectedThroughNetwork();
+    stats.m_network = static_cast<enum Network>(stats.nodeid % 6);
     stats.m_network_name = GetNetworkName(stats.m_network);
     stats.m_mapped_as = addr.GetMappedAS(m_asmap);
     if (m_tx_relay != nullptr) {
```
I've verified that items in the "Network" column are sorted in lexicographical order:
Screenshot from 2020-12-25 20-39-08


TBH, I don't like it, but it could be changed in a followup.

jonatack:
Thanks! My thinking was that users expect this to be sorted by alphabetical order as there isn't an obvious alternative sort expectation, so I went with sort by name rather than by enum id (which may be more what you're thinking of? e.g. ipv4, then ipv6, then onion, then i2p, then cjdns)

hebasto:
After #18 a filter feature could be easy added, and the sort order will do not matter :)

jonatack:
SGTM, will review PRs 161 -> 18 ->164.

jonatack:
Updated the first commit to remove the now-redundant network string (same idea as @promag's suggestion in the connection type pull) and updated the network column ordering by enum id rather than by network name (@hebasto I agree it's better, see screenshot with your patch to show all the Network enum members). Will open this in the main repo as it now touches RPC code.

Screenshot from 2020-12-27 14-21-28

jonatack:
Re-opening here, leaving the main repo pull closed. The changes are minor enough to not warrant moving it to the main repo.

laanwj:
ACK e262a19b0ba07d8d2334551f49ca1577ab2999fc

src/qt/forms/debugwindow.ui
```diff
@@ -1100,13 +1100,39 @@
                 </widget>
                </item>
                <item row="2" column="0">
+                <widget class="QLabel" name="peerNetworkLabel">
+                 <property name="toolTip">
+                  <string>The network protocol this peer is connected through: IPv4, IPv6, Onion, I2P, or CJDNS.</string>
```
hebasto, line 1105: The suggestion for a follow-up is to remove untranslatable network names from the translatable tooltip string.

hebasto (commented):
post-merge ACK e262a19b0ba07d8d2334551f49ca1577ab2999fc

thantkyaw1:
> post-merge ACK [e262a19](https://github.com/bitcoin-core/gui/commit/e262a19b0ba07d8d2334551f49ca1577ab2999fc)

.diff

thantkyaw1:
.patch

thantkyaw1:
Add net to peer

PR #163, opened 2020-12-24
Peer details: replace Direction with Connection Type

### LINKED ISSUE #159: Replace "Direction" with "Connection Type" in Peer Tab
To give more detailed information about the connection type, the legacy type (direction) should be replaced by connection type

Current master:

Screenshot from 2020-12-22 12-37-12

jonatack:
Screenshot from 2021-01-09 11-23-17

Closes #159.

### DIFF
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -1079,7 +1079,7 @@
                <item row="1" column="0">
                 <widget class="QLabel" name="peerConnectionTypeLabel">
                  <property name="toolTip">
-                   <string>The type of peer connection:&lt;ul&gt;&lt;li&gt;Inbound: initiated by peer&lt;/li&gt;&lt;li&gt;Outbound Full Relay: default&lt;/li&gt;&lt;li&gt;Outbound Block Relay: does not relay transactions or addresses&lt;/li&gt;&lt;li&gt;Outbound Manual: added using RPC addnode or -addnode/-connect configuration options&lt;/li&gt;&lt;li&gt;Outbound Feeler: short-lived, for testing addresses&lt;/li&gt;&lt;li&gt;Outbound Address Fetch: short-lived, for soliciting addresses&lt;/li&gt;&lt;/ul&gt;</string>
+                   <string>The type of peer connection:&lt;ul&gt;&lt;li&gt;Inbound: initiated by peer&lt;/li&gt;&lt;li&gt;Outbound Full Relay: default&lt;/li&gt;&lt;li&gt;Outbound Block Relay: does not relay transactions or addresses&lt;/li&gt;&lt;li&gt;Outbound Manual: added using RPC %1 or %2/%3 configuration options&lt;/li&gt;&lt;li&gt;Outbound Feeler: short-lived, for testing addresses&lt;/li&gt;&lt;li&gt;Outbound Address Fetch: short-lived, for soliciting addresses&lt;/li&gt;&lt;/ul&gt;</string>
                  </property>
                  <property name="text">
                   <string>Connection Type</string>
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -1077,14 +1077,17 @@
                 </widget>
                </item>
                <item row="1" column="0">
-                <widget class="QLabel" name="label_23">
+                <widget class="QLabel" name="peerConnectionTypeLabel">
+                 <property name="toolTip">
+                   <string>The type of peer connection:&lt;ul&gt;&lt;li&gt;Inbound: initiated by peer&lt;/li&gt;&lt;li&gt;Outbound Full Relay: default&lt;/li&gt;&lt;li&gt;Outbound Block Relay: does not relay transactions or addresses&lt;/li&gt;&lt;li&gt;Outbound Manual: added using RPC addnode or -addnode/-connect configuration options&lt;/li&gt;&lt;li&gt;Outbound Feeler: short-lived, for testing addresses&lt;/li&gt;&lt;li&gt;Outbound Address Fetch: short-lived, for soliciting addresses&lt;/li&gt;&lt;/ul&gt;</string>
+                 </property>
                  <property name="text">
-                  <string>Direction</string>
+                  <string>Connection Type</string>
                  </property>
                 </widget>
                </item>
                <item row="1" column="1">
-                <widget class="QLabel" name="peerDirection">
+                <widget class="QLabel" name="peerConnectionType">
                  <property name="cursor">
                   <cursorShape>IBeamCursor</cursorShape>
                  </property>
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -764,6 +764,19 @@ QString NetworkToQString(Network net)
     assert(false);
 }
 
+QString ConnectionTypeToQString(ConnectionType conn_type)
+{
+    switch (conn_type) {
+    case ConnectionType::INBOUND: return QObject::tr("Inbound");
+    case ConnectionType::OUTBOUND_FULL_RELAY: return QObject::tr("Outbound Full Relay");
+    case ConnectionType::BLOCK_RELAY: return QObject::tr("Outbound Block Relay");
+    case ConnectionType::MANUAL: return QObject::tr("Outbound Manual");
+    case ConnectionType::FEELER: return QObject::tr("Outbound Feeler");
+    case ConnectionType::ADDR_FETCH: return QObject::tr("Outbound Address Fetch");
+    } // no default case, so the compiler can warn about missing cases
+    assert(false);
+}
+
 QString formatDurationStr(int secs)
 {
     QStringList strList;
--- a/src/qt/guiutil.h
+++ b/src/qt/guiutil.h
@@ -7,17 +7,18 @@
 
 #include <amount.h>
 #include <fs.h>
+#include <net.h>
 #include <netaddress.h>
 
 #include <QEvent>
 #include <QHeaderView>
 #include <QItemDelegate>
+#include <QLabel>
 #include <QMessageBox>
 #include <QObject>
 #include <QProgressBar>
 #include <QString>
 #include <QTableView>
-#include <QLabel>
 
 class QValidatedLineEdit;
 class SendCoinsRecipient;
@@ -227,6 +228,9 @@ namespace GUIUtil
 
     /** Convert enum Network to QString */
     QString NetworkToQString(Network net);
+
+    /** Convert enum ConnectionType to QString */
+    QString ConnectionTypeToQString(ConnectionType conn_type);
 
     /** Convert seconds into a QString with days, hours, mins, secs */
     QString formatDurationStr(int secs);
--- a/src/rpc/net.cpp
+++ b/src/rpc/net.cpp
@@ -128,9 +128,6 @@ static RPCHelpMan getpeerinfo()
                             {RPCResult::Type::BOOL, "inbound", "Inbound (true) or Outbound (false)"},
                             {RPCResult::Type::BOOL, "bip152_hb_to", "Whether we selected peer as (compact blocks) high-bandwidth peer"},
                             {RPCResult::Type::BOOL, "bip152_hb_from", "Whether peer selected us as (compact blocks) high-bandwidth peer"},
-                            {RPCResult::Type::STR, "connection_type", "Type of connection: \n" + Join(CONNECTION_TYPE_DOC, ",\n") + ".\n"
-                                                                      "Please note this output is unlikely to be stable in upcoming releases as we iterate to\n"
-                                                                      "best capture connection behaviors."},
                             {RPCResult::Type::NUM, "startingheight", "The starting height (block) of the peer"},
                             {RPCResult::Type::NUM, "synced_headers", "The last header we have in common with this peer"},
                             {RPCResult::Type::NUM, "synced_blocks", "The last block we have in common with this peer"},
@@ -156,6 +153,9 @@ static RPCHelpMan getpeerinfo()
                                                               "Only known message types can appear as keys in the object and all bytes received\n"
                                                               "of unknown message types are listed under '"+NET_MESSAGE_COMMAND_OTHER+"'."}
                             }},
+                            {RPCResult::Type::STR, "connection_type", "Type of connection: \n" + Join(CONNECTION_TYPE_DOC, ",\n") + ".\n"
+                                                                      "Please note this output is unlikely to be stable in upcoming releases as we iterate to\n"
+                                                                      "best capture connection behaviors."},
                         }},
                     }},
                 },
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -1111,7 +1111,7 @@ void RPCConsole::updateDetailWidget()
     ui->timeoffset->setText(GUIUtil::formatTimeOffset(stats->nodeStats.nTimeOffset));
     ui->peerVersion->setText(QString::number(stats->nodeStats.nVersion));
     ui->peerSubversion->setText(QString::fromStdString(stats->nodeStats.cleanSubVer));
-    ui->peerDirection->setText(stats->nodeStats.fInbound ? tr("Inbound") : tr("Outbound"));
+    ui->peerConnectionType->setText(GUIUtil::ConnectionTypeToQString(stats->nodeStats.m_conn_type));
     ui->peerNetwork->setText(GUIUtil::NetworkToQString(stats->nodeStats.m_network));
     if (stats->nodeStats.m_permissionFlags == PF_NONE) {
         ui->peerPermissions->setText(tr("N/A"));
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -463,6 +463,7 @@ RPCConsole::RPCConsole(interfaces::Node& node, const PlatformStyle *_platformSty
     ui->dataDir->setToolTip(ui->dataDir->toolTip().arg(QString(nonbreaking_hyphen) + "datadir"));
     ui->blocksDir->setToolTip(ui->blocksDir->toolTip().arg(QString(nonbreaking_hyphen) + "blocksdir"));
     ui->openDebugLogfileButton->setToolTip(ui->openDebugLogfileButton->toolTip().arg(PACKAGE_NAME));
+    ui->peerConnectionTypeLabel->setToolTip(ui->peerConnectionTypeLabel->toolTip().arg("addnode").arg(QString(nonbreaking_hyphen) + "addnode").arg(QString(nonbreaking_hyphen) + "connect"));
 
     if (platformStyle->getImagesOnButtons()) {
         ui->openDebugLogfileButton->setIcon(platformStyle->SingleColorIcon(":/icons/export"));


MarcoFalke:
Nice. Concept ACK

src/qt/guiutil.cpp
```diff
@@ -749,6 +749,20 @@ QString boostPathToQString(const fs::path &path)
...
+    // Outbound peer connections
+    if (conn_type == "outbound-full-relay") return "Outbound full relay";
+    if (conn_type == "block-relay-only") return "Outbound block relay";
+    if (conn_type == "manual") return "Outbound manual";
+    if (conn_type == "feeler") return "Outbound feeler";
+    if (conn_type == "addr-fetch") return "Outbound address fetch";
+    // Inbound peer connections
+    if (conn_type == "inbound") return "Inbound full relay";
+    if (conn_type == "inbound-block-relay") return "Inbound block relay";
+    return "";
```
MarcoFalke, line 763: nit: Would be nice if this default fallback was replaced with an assert(false) and the above ifs with an enum-switch-case

jonatack: Agree, but ISTM that only works for integer types.

MarcoFalke: `ConnectionType` is an enum class (implementation defined integer)

jonatack: Right, I wanted to do that but the `ConnectionType` enum wasn't exposed in CNodeStats. Ok, done.

hebasto:
Concept ACK.

src/qt/guiutil.h
```diff
@@ -224,6 +225,9 @@ namespace GUIUtil
     /* Convert OS specific boost path to QString through UTF-8 */
     QString boostPathToQString(const fs::path &path);
 
+    /* Convert ConnectionType enum to QString */
```
hebasto, line 228: 571b0469fa952be98874d1afd772a7fa708dd83f

This comment is not Doxygen-compatible.

src/qt/guiutil.cpp
```diff
@@ -749,6 +749,19 @@ QString boostPathToQString(const fs::path &path)
...
+    case ConnectionType::INBOUND: return "Inbound full relay";
+    case ConnectionType::OUTBOUND_FULL_RELAY: return "Outbound full relay";
+    case ConnectionType::BLOCK_RELAY: return "Outbound block relay";
+    case ConnectionType::MANUAL: return "Outbound manual";
+    case ConnectionType::FEELER: return "Outbound feeler";
+    case ConnectionType::ADDR_FETCH: return "Outbound address fetch";
+    } // no default case, so the compiler can warn about missing cases
+    assert(false);
+}
+
```
hebasto, line 764: 571b0469fa952be98874d1afd772a7fa708dd83f

The returned values should be translated.

UX note: I have a personal preference do not capitalize the first letter of returned values. Feel free to ignore it.

src/qt/guiutil.cpp
```diff
@@ -749,6 +749,19 @@ QString boostPathToQString(const fs::path &path)
     return QString::fromStdString(path.string());
 }
 
+QString ConnectionTypeToQString(ConnectionType conn_type)
+{
+    switch (conn_type) {
+    case ConnectionType::INBOUND: return "Inbound full relay";
```
hebasto, line 755: 571b0469fa952be98874d1afd772a7fa708dd83f

It seems not always true :)

IIUC, you recent changes in `bitcoin-cli` distinguish "block relay" for inbound connections as well.

src/qt/forms/debugwindow.ui
```diff
@@ -1077,14 +1077,17 @@
                 </widget>
                </item>
                <item row="1" column="0">
-                <widget class="QLabel" name="label_23">
+                <widget class="QLabel" name="peerConnectionTypeLabel">
+                 <property name="toolTip">
+                  <string>The connection type of this peer.</string>
+                 </property>
```
hebasto, line 1083: bf8135db2a658aed3129ec82e237f8030a8ec7dd

I've noted your [comment](https://github.com/bitcoin-core/gui/pull/162#discussion_r548865476), but here the tooltip provides no additional info. I don't think that adding such a tooltip for the sake itself is a good approach for the GUI:
- https://uxplanet.org/tooltips-in-ui-design-f63e117aa3d1
- https://www.nngroup.com/articles/tooltip-guidelines/

hebasto (changes requested):
Concept ACK.

I have reviewed the code (bf8135db2a658aed3129ec82e237f8030a8ec7dd), and tested it.

src/qt/guiutil.cpp
jonatack: I also prefer all lowercase (e.g. rpc/cli) but the GUI UI uses capitalization (I hesitated to use titlecase here).

src/qt/forms/debugwindow.ui
jonatack: True. Perhaps it should link to more information about the connection types (like we have in the codebase), maybe in a markdown file in `/doc`.

src/qt/guiutil.cpp
jonatack: > IIUC, you recent changes in `bitcoin-cli` distinguish "block relay" for inbound connections as well.

Yes, I'm anticipating `inbound full relay` and `inbound block relay` connection types (https://github.com/bitcoin/bitcoin/pull/20726 and https://github.com/bitcoin/bitcoin/pull/20729) but I guess it's better to revert to just inbound for now and update it later.

src/qt/forms/debugwindow.ui
jonatack: Maybe I should add more useful info, like in the RPC help (getpeerinfo)

src/qt/guiutil.cpp
jonatack: (in the latest -netinfo update in https://github.com/bitcoin/bitcoin/pull/20764, the issue is avoided by using the connection type only to parse full/block/manual/feeler/addr, but not for the direction)

src/qt/guiutil.cpp
hebasto: I perceive it like "Connection type: Outbound full relay". That's why I prefer the lowercased variant.

If the capitalized variant will be required in some other places in the GUI, it is easy to provide it with a generic function.

I'm not a design expert, then I'm fine with any variant.

src/qt/guiutil.cpp
jonatack: In the latest -netinfo update in https://github.com/bitcoin/bitcoin/pull/20764, the issue is avoided by using the connection type only to parse full/block/manual/feeler/addr, but not for the direction. Maybe best to do the same here, hm.

src/qt/forms/debugwindow.ui
hebasto: Anyway, please keep it useful and _succinct_, or else do not add it at all :)

src/qt/guiutil.h
jonatack: thanks! fixed

src/qt/guiutil.cpp
jonatack: Thanks--translated and changed to titlecase for coherence with the general GUI UI.

src/qt/guiutil.cpp
jonatack: Now "Inbound"

src/qt/forms/debugwindow.ui
jonatack: Added info similar to the getpeerinfo help, see updated screenshot in the PR description.

jonatack:
Thanks for the excellent feedback @hebasto! Updated.

hebasto:
I have reviewed the code (8e7f82a6a84c7b2553f88767dd356f480d1ae4b7), and tested it.

The following patch:
```diff
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -1079,7 +1079,7 @@
                <item row="1" column="0">
                 <widget class="QLabel" name="peerConnectionTypeLabel">
                  <property name="toolTip">
-                  <string>The type of peer connection: Outbound Full Relay (default), Outbound Block Relay (does not relay transactions or addresses), Inbound (initiated by peer), Outbound Manual (added using RPC addnode or -addnode/-connect config options), Outbound Feeler (short-lived, for testing addresses), or Outbound Address Fetch (short-lived, for soliciting addresses).</string>
+                  <string>The type of peer connection: Outbound Full Relay (default), Outbound Block Relay (does not relay transactions or addresses), Inbound (initiated by peer), Outbound Manual (added using RPC %1 or %2/%3 config options), Outbound Feeler (short-lived, for testing addresses), or Outbound Address Fetch (short-lived, for soliciting addresses).</string>
                  </property>
                  <property name="text">
                   <string>Connection Type</string>
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -496,6 +496,8 @@ RPCConsole::RPCConsole(interfaces::Node& node, const PlatformStyle *_platformSty
     clear();
 
     GUIUtil::handleCloseWindowShortcut(this);
+
+    ui->peerConnectionTypeLabel->setToolTip(ui->peerConnectionTypeLabel->toolTip().arg("addnode").arg(QString(nonbreaking_hyphen) + "addnode").arg(QString(nonbreaking_hyphen) + "connect"));
 }
 
 RPCConsole::~RPCConsole()
```

- removes RPC and option names from the translatable string
- prevents effects like this
DeepinScreenshot_select-area_20201225232751

jonatack:
Thanks @hebasto, tested and added your suggestion, updated the screenshot in the PR description.

hebasto (approved):
ACK 6cd242c50f4a8d3f3c509937d80962fe58096ead, tested on Linux Mint 20 (x86_64) + Qt 5.12.8.

Looks great:
Screenshot from 2020-12-25 23-52-27

src/net.h
```diff
@@ -725,6 +725,7 @@ class CNodeStats
     // Name of the network the peer connected through
     std::string m_network;
     uint32_t m_mapped_as;
+    ConnectionType m_conn_type;
```
promag, line 728: a60f67b2fb0eede205db8cce11ad0e6f42dbc90e

This was suggested here https://github.com/bitcoin/bitcoin/pull/19725#discussion_r473361238 and now makes 
`m_conn_type_string` redundant, don't makes sense to keep both TBH. Since `m_conn_type_string` is only used in `getpeerinfo` then drop it and refactor `ConnectionTypeAsString` to take `conn_type` instead?

MarcoFalke: Concept ACK, but RPC changes (refactoring) can only be made in the main repo

jonatack: Done, I think. My initial version of this PR did not touch any non-GUI code. Should I close this PR and re-open in the main repo? (I'm not sure it's worth losing the review above.)

hebasto: Maybe just open a pr in the main repo with the first and last commits?

promag: IMO this isn't a RPC change, it's just a member refactor in a net struct, and the refactor is motivated by this GUI change. Otherwise this can go forward in 2 ways I think, either leave this as it was, and then in main repo drop the redundant member, or 1st refactor and make this dependant of that PR.

jonatack: Yes, this doesn't change the RPC API.

Please let me know which option is best:

1. split it into two pulls, one here, gated by a another one in the other repo
2. move the pull to the other repo
3. revert to the original version that didn't touch main repo code
4. leave this as-is

jonatack: I'm not sure the refactoring changes make sense without the whole. Opened the same pull in https://github.com/bitcoin/bitcoin/pull/20778 and will let the maintainers decide.

promag: Just noting that redundancy is not a blocker, but could be avoided.

jonatack:
It was suggested to re-open here.

MarcoFalke:
Concept re-ACK. Thanks for re-opening.

jonatack:
Rebased.

hebasto (approved):
re-ACK aef969363a7685fc805833a3ec6863c27500d02f, since my [previous](https://github.com/bitcoin-core/gui/pull/163#pullrequestreview-558838006) review only rebased and the _"p2p, rpc, refactor: remove CNodeStats::m_conn_type_string"_ commit added.

jonatack:
Rebased `git range-diff ae8f797 aef9693 7098509`

src/rpc/net.cpp
```diff
@@ -249,7 +249,7 @@ static RPCHelpMan getpeerinfo()
                 recvPerMsgCmd.pushKV(i.first, i.second);
         }
         obj.pushKV("bytesrecv_per_msg", recvPerMsgCmd);
-        obj.pushKV("connection_type", stats.m_conn_type_string);
+        obj.pushKV("connection_type", ConnectionTypeAsString(stats.m_conn_type));
```
hebasto, line 252: 70985097c00827de94a7922620b63e1ee10dd53d

Mind fixing https://github.com/bitcoin/bitcoin/pull/19725/commits/395acfa83a5436790c1a722a5609ac9d48df235f (https://github.com/bitcoin/bitcoin/pull/19725) by aligning the order of the "connection_type" field both in RPC help and in RPC output?

hebasto (approved):
re-ACK 70985097c00827de94a7922620b63e1ee10dd53d, only rebased since my [previous](https://github.com/bitcoin-core/gui/pull/163#pullrequestreview-560126996) review.

src/rpc/net.cpp
jonatack: thanks; done

jonatack:
Thanks @hebasto, updated the last commit to move the getpeerinfo connection_type help to the correct place per review feedback in https://github.com/bitcoin-core/gui/pull/163#discussion_r550863551 and updated the commit message.

hebasto (approved):
re-ACK be8294ff428da7d13043b3f9a27d7ace1c2dadf4

laanwj:
Concept ACK

jonatack:
Rebased after the merge of https://github.com/bitcoin/bitcoin/pull/20786.

hebasto (approved):
re-ACK 20fa97560b2d27f94f556a570f3556356877b79f, only rebased since my [previous](https://github.com/bitcoin-core/gui/pull/163#pullrequestreview-560723519) review due to the conflict with https://github.com/bitcoin/bitcoin/pull/20786, verified with
```
$ git range-diff master be8294ff428da7d13043b3f9a27d7ace1c2dadf4 20fa97560b2d27f94f556a570f3556356877b79f
```

jarolrod (commented):
ACK 20fa97560b2d27f94f556a570f3556356877b79f

Tested on macOS Big Sur 11.1 with qt 5.15.2. Looks great!
Screen Shot 2021-01-08 at 11 01 46 AM

Small NIT: There's a lot of text in the tooltip. It would be nice if all the possible values were put into a list instead of separated by commas. It would look something like [this](https://ptgmedia.pearsoncmg.com/images/chap17_9780132354165/elementLinks/catalogentry.jpg), which to me would be an improvement.

See: [Supported HTML Subset](https://doc.qt.io/qt-5/richtext-html-subset.html),

jonatack:
Better? If yes, bold or no?
Screenshot from 2021-01-09 00-04-04
Screenshot from 2021-01-08 23-57-49

jarolrod:
@jonatack the list looks great! I think not-bold is better. If i may, I would suggest one more thing:

Write the descriptions like this:
`
Outbound Block Relay: does not relay transactions or addresses
`

instead of:
`
Outbound Block Relay (does not relay transactions or addresses)
`

jonatack:
@jarolrod test with inbound first and your colon suggestion (agree, it's more readable)

Screenshot from 2021-01-09 01-10-03

jarolrod:
@jonatack looks awesome 👍

jonatack:
Thanks @jarolrod. Updated in latest push, if you'd like to review/test. @hebasto, can you verify the changed tooltip code?

hebasto:
> @hebasto, can you verify the changed tooltip code?

The used HTML code is [valid](https://doc.qt.io/qt-5/richtext-html-subset.html).

But I'd prefer to keep the tooltip content in the form `*.ui` file. It makes usage of design tools (e.g., Qt Designer) much easier. In `*.cpp` file should be text substitution code only.

jonatack:
@hebasto thanks--done.

hebasto (approved):
re-ACK 06ba9b300866f33e21512af9d7d2891ee1501bf4, the tooltip content is organized as unordered list.

jarolrod (commented):
re-ACK 06ba9b300866f33e21512af9d7d2891ee1501bf4

src/qt/forms/debugwindow.ui
```diff
@@ -1077,14 +1077,17 @@
                 </widget>
                </item>
                <item row="1" column="0">
-                <widget class="QLabel" name="label_23">
+                <widget class="QLabel" name="peerConnectionTypeLabel">
+                 <property name="toolTip">
+                   <string>The type of peer connection:&lt;ul&gt;&lt;li&gt;Inbound: initiated by peer&lt;/li&gt;&lt;li&gt;Outbound Full Relay: default&lt;/li&gt;&lt;li&gt;Outbound Block Relay: does not relay transactions or addresses&lt;/li&gt;&lt;li&gt;Outbound Manual: added using RPC %1 or %2/%3 configuration options&lt;/li&gt;&lt;li&gt;Outbound Feeler: short-lived, for testing addresses&lt;/li&gt;&lt;li&gt;Outbound Address Fetch: short-lived, for soliciting addresses&lt;/li&gt;&lt;/ul&gt;</string>
```
MarcoFalke, line 1082: translation nit: How are translators supposed to translate `:&lt;ul&gt;&lt;li&gt;`? It might be better to say "Can be one of %s" and then use `Join(translated_strings, ", ")` or simply "Refer to the getpeerinfo RPC help for details on the connection types"?

jonatack: @hebasto, if we replace the escaped markup with placeholders (as done here for the nonbreaking hyphen), is that still QT Designer-friendly? Can do it in #179.

hebasto: > translation nit: How are translators supposed to translate `:&lt;ul&gt;&lt;li&gt;`?

There was a discussion in private IRC with @jonatack:
> \<hebasto\>
> 14:24:52 but I expect the translators' burden :)
> 14:25:39 hope that translators are aware of escaping :)

I'm not an expert in translation but I believe that `:&lt;ul&gt;&lt;li&gt;` is not a subject to translate as `%1` or `%2`.

> It might be better to say "Can be one of %s" and then use `Join(translated_strings, ", ")` or simply "Refer to the getpeerinfo RPC help for details on the connection types"?

Either of the suggested variants LGTM.

> if we replace the escaped markup with placeholders (as done here for the nonbreaking hyphen), is that still QT Designer-friendly?

I think it is. But, tbh, I'd prefer more succinct @MarcoFalke's suggestions.

jonatack: Thanks. Helpfulness and readability for users were the reasons for the iterations above and it seems best to not regress on those improvements with referring GUI users to the command line or dropping the readable list layout. I'll propose replacing the escaped markup with placeholders.

hebasto: Why not a single placeholder for the whole unordered list? We still have a (shorter) tooltip in the form file, and make translators' life easier :)

jonatack: SGTM too.

jonatack: Done in #180

PR #165, opened 2020-12-25
Save QSplitter state in QSettings

hebasto:
This PR adds the ability to save the `QSplitter` widget state in `QSettings` during shutdown, and restore it on startup.

A user no longer needs to adjust the splitter every time :)
DeepinScreenshot_select-area_20201225211422

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -457,6 +457,8 @@ RPCConsole::RPCConsole(interfaces::Node& node, const PlatformStyle *_platformSty
         move(QGuiApplication::primaryScreen()->availableGeometry().center() - frameGeometry().center());
     }
 
+    ui->splitter->restoreState(settings.value("PeersTabSplitterSizes").toByteArray());
+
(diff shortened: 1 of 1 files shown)


jonatack:
Great idea. Was complaining about this to myself just again today.

jonatack (approved):
ACK 90f9fc274bf69b33adea593146ff5d6793123781 this sets the "PeersTabSplitterSizes" value in the RPCConsole dtor and restores it in the RPCConsole ctor; tested in Debian with various split settings, tab open/close sequences, and shutdown methods, and the Peers window split state was faithfully maintained.

jonasschnelli:
Good idea and just two lines of code.
utACK 90f9fc274bf69b33adea593146ff5d6793123781

jarolrod (commented):
Tested ACK on macOS Big Sur 11.1 with Qt 5.15.2. Works great and is a useful feature!

luke-jr (approved):
utACK

luke-jr:
Note that we might need to change the setting name for each version that adds/removes columns :/

PR #167, opened 2020-12-29
raise helpMessageDialog

RandyMcMillan:
the raise() method brings the helpMessageDialog to the top if it is obscured by another window.

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -845,7 +845,7 @@ void BitcoinGUI::showDebugWindowActivateConsole()
 
 void BitcoinGUI::showHelpMessageClicked()
 {
-    helpMessageDialog->show();
+    GUIUtil::bringToFront(helpMessageDialog);
 }
 
 #ifdef ENABLE_WALLET


src/qt/bitcoingui.cpp
```diff
@@ -846,6 +846,7 @@ void BitcoinGUI::showDebugWindowActivateConsole()
 void BitcoinGUI::showHelpMessageClicked()
 {
     helpMessageDialog->show();
+    helpMessageDialog->raise();//bring to top if obscured
```
promag, line 849: Not sure if you are aware of `GUIUtil::bringToFront(this)`, seems the right call here.

hebasto:
@RandyMcMillan

Mind providing steps to reproduce `helpMessageDialog` obscuring (as it is a top-level window, and it is activated via a menu item only)?

jarolrod (commented):
No strong opinion, but I do not think this is needed.

The help menu always shows on top, I cannot come up with a scenario where it is obfuscated. Furthermore, If the help window has been moved to the background by clicking on another window, clicking on `Help` -> `Command-line Options` will bring back the help window to the top.

RandyMcMillan:
i will post screen shots - maybe this is a macOS specific issue..

RandyMcMillan:
Maybe this is an "accessibility" issue and warrants an accessibility label - note that if the user is using tab-thru navigation. Once the window is presented and then obfuscated - using the nav bar again will not re-present the window. 

I think is is critical that we incorporate usability/accessibility considerations - I would hope that this gui centric fork allows for these very nuanced issues.

### Not everybody can use a mouse in gui applications. :)

Screen Shot 2021-01-22 at 7 21 41 PM

Screen Shot 2021-01-22 at 7 21 33 PM

hebasto:
Concept ACK.

I can confirm this bug on macOS Big Sur 11.1 (20C69)  + Homebrew's Qt 5.15.2.

@RandyMcMillan Do you mind using a more general approach [suggested](https://github.com/bitcoin-core/gui/pull/167#discussion_r549874197) by @promag:
```diff
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -845,7 +845,7 @@ void BitcoinGUI::showDebugWindowActivateConsole()
 
 void BitcoinGUI::showHelpMessageClicked()
 {
-    helpMessageDialog->show();
+    GUIUtil::bringToFront(helpMessageDialog);
 }
 
 #ifdef ENABLE_WALLET
```

RandyMcMillan:
Thanks for posting the snippet - saved a lot of time!

src/qt/bitcoingui.cpp
RandyMcMillan: Thanks!

promag (commented):
Code review ACK 77114462f2328914b7a918f40776e522a0898e56.

hebasto (approved):
ACK 77114462f2328914b7a918f40776e522a0898e56, tested on:
- macOS Big Sur 11.1 (20C69)  + Homebrew's Qt 5.15.2
- Linux Mint 20.1 + Cinnamon + Qt 5.12.8

MarcoFalke:
Does this need backport?

hebasto:
> Does this need backport?

Yes, I think, as it is a bugfix for macOS builds.

MarcoFalke:
ok, thanks. Will backport.

MarcoFalke:
Backported in https://github.com/bitcoin/bitcoin/pull/20901

PR #173, opened 2021-01-02
Follow Qt docs when implementing rowCount and columnCount

hebasto:
[`QAbstractItemModel::rowCount`](https://doc.qt.io/qt-5/qabstractitemmodel.html#rowCount):
> **Note:** When implementing a table based model, `rowCount()` should return 0 when the parent is valid.

[`QAbstractItemModel::columnCount`](https://doc.qt.io/qt-5/qabstractitemmodel.html#columnCount):
> **Note:** When implementing a table based model, `columnCount()` should return 0 when the parent is valid.

### DIFF
--- a/src/qt/addresstablemodel.cpp
+++ b/src/qt/addresstablemodel.cpp
@@ -177,13 +177,17 @@ AddressTableModel::~AddressTableModel()
 
 int AddressTableModel::rowCount(const QModelIndex &parent) const
 {
-    Q_UNUSED(parent);
+    if (parent.isValid()) {
+        return 0;
+    }
     return priv->size();
 }
 
 int AddressTableModel::columnCount(const QModelIndex &parent) const
 {
-    Q_UNUSED(parent);
+    if (parent.isValid()) {
+        return 0;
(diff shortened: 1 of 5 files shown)


jarolrod:
Tested ACK 195fcb53a09e9b852544778a077727f81d31303e. Compiled and ran on macOS (Big Sur 11.1 and Catalina 10.15.7), Arch Linux, and FreeBSD. visually verified no weird effects with the `Address`, `Ban`, `Peer`, and `Transaction` tables. As already stated, the code change brings us inline with what the QT Docs recommend.

jarolrod (commented):
Tested ACK 195fcb53a09e9b852544778a077727f81d31303e. Compiled and ran on macOS (Big Sur 11.1 and Catalina 10.15.7), Arch Linux, and FreeBSD. visually verified no weird effects with the `Address`, `Ban`, `Peer`, and `Transaction` tables. As already stated, the code change brings us inline with what the QT Docs recommend.

luke-jr (approved):
utACK

PR #174, opened 2021-01-02
refactor: No need to implement index for non-hierarchical models

hebasto:
From Qt docs:

- https://doc.qt.io/qt-5/model-view-programming.html#a-read-only-example-model:
> If our model was hierarchical, we would also have to implement the `index()` and `parent()` functions.

- https://doc.qt.io/qt-5/qabstracttablemodel.html#subclassing:
> Default implementations of the `index()` and `parent()` functions are provided by `QAbstractTableModel`.

### DIFF
--- a/src/qt/addresstablemodel.cpp
+++ b/src/qt/addresstablemodel.cpp
@@ -196,7 +196,7 @@ QVariant AddressTableModel::data(const QModelIndex &index, int role) const
     if(!index.isValid())
         return QVariant();
 
-    AddressTableEntry *rec = static_cast<AddressTableEntry*>(index.internalPointer());
+    const AddressTableEntry* rec{priv->index(index.row())};
 
     if(role == Qt::DisplayRole || role == Qt::EditRole)
     {
@@ -242,7 +242,8 @@ bool AddressTableModel::setData(const QModelIndex &index, const QVariant &value,
 {
     if(!index.isValid())
         return false;
-    AddressTableEntry *rec = static_cast<AddressTableEntry*>(index.internalPointer());
+
+    const AddressTableEntry* rec{priv->index(index.row())};
     std::string strPurpose = (rec->type == AddressTableEntry::Sending ? "send" : "receive");
     editStatus = OK;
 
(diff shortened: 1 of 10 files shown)


jarolrod:
ACK 199830ff920825f5306555e9f6fc08be8534e9bb

Tested on macOS Big Sur 11.1 with Qt 5.15.2. Tested that there were no weird side-effects under `Transactions` and the `Peers Window`. This is a welcome change as it adheres to the Qt docs and removes unnecessary code.

hebasto:
Rebased 199830ff920825f5306555e9f6fc08be8534e9bb -> d64b338bc7ac7f47e4e9921a29d1534f53e9de86 ([pr174.01](https://github.com/hebasto/gui/commits/pr174.01) -> [pr174.02](https://github.com/hebasto/gui/commits/pr174.02)) due to the conflict with #161.

src/qt/addresstablemodel.cpp
```diff
@@ -196,7 +196,7 @@ QVariant AddressTableModel::data(const QModelIndex &index, int role) const
     if(!index.isValid())
         return QVariant();
 
-    AddressTableEntry *rec = static_cast<AddressTableEntry*>(index.internalPointer());
+    const AddressTableEntry* rec{priv->index(index.row())};
```
promag, line 199: Need to check if `rec != nullptr`, `index.isValid()` doesn't check for out of bounds.

Alternatively you could inline `priv->index` and use reference instead `const AddressTableEntry&`. You still need to check for out of bounds.

promag (commented):
I think custom `QModelIndex` with internal pointers are nice when it saves duplicate long lookups on the internal model. Typically, views makes lots of `data(index, role)` calls and this approach increases lookups to the underlaying model. Maybe that's not a concern since views only queries data for visible items and we are using containers with O(1) lookups. Still, for addresses and transactions models it would be nice to make sure we don't make user experience worse.

hebasto:
@promag 
> Still, for addresses and transactions models it would be nice to make sure we don't make user experience worse.

Right. Blind applying such changes to _all_ models is not warranted.

Closing for now.

PR #176, opened 2021-01-03
Fix TxViewDelegate layout

hebasto:
This change:
- prevents overlapping date and amount strings
- guaranties that "eye" sign at the end of the watch-only address/label is always visible

Fix https://github.com/bitcoin/bitcoin/issues/20826

Here are some screenshots with this PR with the _minimum available width_ of the transaction list widget:

Screenshot from 2021-01-03 20-23-56
Screenshot from 2021-01-03 20-24-47
Screenshot from 2021-01-03 20-25-27
Screenshot from 2021-01-03 20-33-20

### DIFF
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -78,6 +78,7 @@ QT_MOC_CPP = \
   qt/moc_transactiondesc.cpp \
   qt/moc_transactiondescdialog.cpp \
   qt/moc_transactionfilterproxy.cpp \
+  qt/moc_transactionoverviewwidget.cpp \
   qt/moc_transactiontablemodel.cpp \
   qt/moc_transactionview.cpp \
   qt/moc_utilitydialog.cpp \
@@ -151,6 +152,7 @@ BITCOIN_QT_H = \
   qt/transactiondesc.h \
   qt/transactiondescdialog.h \
   qt/transactionfilterproxy.h \
+  qt/transactionoverviewwidget.h \
   qt/transactionrecord.h \
   qt/transactiontablemodel.h \
   qt/transactionview.h \
--- a/src/qt/forms/overviewpage.ui
+++ b/src/qt/forms/overviewpage.ui
@@ -504,7 +504,7 @@
            </layout>
           </item>
           <item>
-           <widget class="QListView" name="listTransactions">
+           <widget class="TransactionOverviewWidget" name="listTransactions">
             <property name="styleSheet">
              <string notr="true">QListView { background: transparent; }</string>
             </property>
@@ -519,6 +519,9 @@
             </property>
             <property name="selectionMode">
              <enum>QAbstractItemView::NoSelection</enum>
+            </property>
+            <property name="uniformItemSizes">
+             <bool>true</bool>
             </property>
            </widget>
           </item>
@@ -544,6 +547,13 @@
    </item>
   </layout>
  </widget>
+ <customwidgets>
+  <customwidget>
+   <class>TransactionOverviewWidget</class>
+   <extends>QListView</extends>
+   <header>qt/transactionoverviewwidget.h</header>
+  </customwidget>
+ </customwidgets>
  <resources/>
  <connections/>
 </ui>
--- a/src/qt/forms/overviewpage.ui
+++ b/src/qt/forms/overviewpage.ui
@@ -517,6 +517,9 @@
             <property name="horizontalScrollBarPolicy">
              <enum>Qt::ScrollBarAlwaysOff</enum>
             </property>
+            <property name="sizeAdjustPolicy">
+             <enum>QAbstractScrollArea::AdjustToContents</enum>
+            </property>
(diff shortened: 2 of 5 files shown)


dooglus:
Looks good to me. The window is a lot bigger than it was before, and perhaps bigger than absolutely necessary but this is as small as I can make it:

Screenshot_2021-01-04_07-58-45

For comparison, this is the 'before' shot:

Screenshot_2021-01-01_14-19-16

MarcoFalke:
Is this for 0.21.1?

hebasto:
> Is this for 0.21.1?

Yes, I think so, as it is a bugfix.

hebasto:
@dooglus 
> The window is ... bigger than absolutely necessary

On fresh Debian 10.7 with Xfce 4.12 cannot reproduce such behavior. Did you tweak fonts or themes?

dooglus:
Yes I did, and I've included a screenshot of my settings below, but I ought to describe the issue more fully now that I've played around with it more.

I can reproduce the issue as follows: I have 2 wallets, one with watchonly addresses and one without. I have the one without show up by default when the client first loads, and it makes the client window 1109x842 pixels. If I then use the menu in the top right to switch to the wallet with watchonly addresses then the window resizes itself to 1329x842 to make room for the watchonly address balances as expected.

The problem is when I switch back to the first wallet. The window stays at its increased size and I can't shrink it back. Some screenshots:

1. no watchonly - small window:

new0

2. watchonly - bigger window:

new1

3. back to no watchonly, but the window stays big:

new2

---

I have the following font settings:

Screenshot_2021-01-04_16-49-04

This is [the font I'm using](https://ipfs.io/ipfs/QmbspRzTRimnaRNa183h21vHNDwREawpvMyW5ejj22ibi7?filename=01-iosevka-2.3.0.zip).

hebasto:
@dooglus 

Thanks for testing!

> The problem is when I switch back to the first wallet. The window stays at its increased size and I can't shrink it back.

Confirming this regression. Looking for a fix.

jarolrod (commented):
Tested ACK on macOS 11.1 and Arch Linux both with Qt 5.15.2.

I was able to replicate the bug described by @dooglus on 5cf67f25a7d8cc75e1831aa37125f603f942b811. I can confirm that 1e31eca7bc01cbe68cef14768786a68e783ac11b now fixes this for me.

dooglus:
I cherry-picked your commits back to the v0.21.0rc4 tag and built. I saw a warning:

    qt/walletframe.cpp: In member function ‘void WalletFrame::setCurrentWallet(WalletModel*)’:
    qt/walletframe.cpp:110:9: warning: init-statement in selection statements only available with -std=c++17 or -std=gnu++17
         if (WalletView* view_about_to_hide = currentWalletView(); view_about_to_hide != nullptr) {
             ^~~~~~~~~~

I don't know if I'm supposed to be using a newer compiler but it wasn't enforced when I ran the configure script. If this is intended to be backported to 0.21.1 maybe it shouldn't use c++17 features.

Despite the warning the code works and I can now shrink the window back after switching to and back from the watchonly wallet.

hebasto:
Updated 1e31eca7bc01cbe68cef14768786a68e783ac11b -> af58f5b12cea91467692dd4ae71d8cc916a608ed ([pr176.02](https://github.com/hebasto/gui/commits/pr176.02) -> [pr176.03](https://github.com/hebasto/gui/commits/pr176.03), [diff](https://github.com/hebasto/gui/compare/pr176.02..pr176.03)):

- addressed @dooglus' [comment](https://github.com/bitcoin-core/gui/pull/176#issuecomment-756707868):
> If this is intended to be backported to 0.21.1 maybe it shouldn't use c++17 features.

dooglus:
ACK 1e31eca.

Now it resizes the window to make room for the watchonly addresses, allows me to shrink the window back when viewing a wallet without any such addresses, and builds without warnings when I backport it to v0.21.

jarolrod (commented):
re-ACK af58f5b12cea91467692dd4ae71d8cc916a608ed

Saw the same warning as @dooglus when patching 1e31eca7bc01cbe68cef14768786a68e783ac11b on `0.21rc4`. Confirming that no warnings come up when patching af58f5b12cea91467692dd4ae71d8cc916a608ed on top of `0.21rc5`. Tested on macOS 11.1 and Arch Linux both with Qt 5.15.2

hebasto:
@dooglus 
> ACK [1e31eca](https://github.com/bitcoin-core/gui/commit/1e31eca7bc01cbe68cef14768786a68e783ac11b).
> 
> Now it resizes the window to make room for the watchonly addresses, allows me to shrink the window back when viewing a wallet without any such addresses, and builds without warnings when I backport it to v0.21.

Thanks for reviewing! Is it intentional the previous commit (not the current top) mentioned? :smile:

dooglus:
No, sorry. I meant af58f5b. Have edited it now.

PR #179, opened 2021-01-09
Add `Type` column to peers window, update peer details name/tooltip

jonatack:
This pull:

- adds a sortable `Type` column to the GUI Peers tab window
- updates the peer details row to `Direction/Type`, so the `Type` column without a direction makes sense (the tooltip is also updated)

Screenshot from 2021-02-06 22-53-11

### DIFF
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -766,15 +766,19 @@ QString NetworkToQString(Network net)
     assert(false);
 }
 
-QString ConnectionTypeToQString(ConnectionType conn_type)
+QString ConnectionTypeToQString(ConnectionType conn_type, bool prepend_direction)
 {
+    QString prefix;
+    if (prepend_direction) {
+        prefix = (conn_type == ConnectionType::INBOUND) ? QObject::tr("Inbound") : QObject::tr("Outbound") + " ";
+    }
     switch (conn_type) {
-    case ConnectionType::INBOUND: return QObject::tr("Inbound");
-    case ConnectionType::OUTBOUND_FULL_RELAY: return QObject::tr("Outbound Full Relay");
-    case ConnectionType::BLOCK_RELAY: return QObject::tr("Outbound Block Relay");
-    case ConnectionType::MANUAL: return QObject::tr("Outbound Manual");
-    case ConnectionType::FEELER: return QObject::tr("Outbound Feeler");
-    case ConnectionType::ADDR_FETCH: return QObject::tr("Outbound Address Fetch");
+    case ConnectionType::INBOUND: return prefix;
+    case ConnectionType::OUTBOUND_FULL_RELAY: return prefix + QObject::tr("Full Relay");
+    case ConnectionType::BLOCK_RELAY: return prefix + QObject::tr("Block Relay");
+    case ConnectionType::MANUAL: return prefix + QObject::tr("Manual");
+    case ConnectionType::FEELER: return prefix + QObject::tr("Feeler");
+    case ConnectionType::ADDR_FETCH: return prefix + QObject::tr("Address Fetch");
     } // no default case, so the compiler can warn about missing cases
     assert(false);
 }
--- a/src/qt/peertablemodel.cpp
+++ b/src/qt/peertablemodel.cpp
@@ -29,6 +29,8 @@ bool NodeLessThan::operator()(const CNodeCombinedStats &left, const CNodeCombine
         return pLeft->nodeid < pRight->nodeid;
     case PeerTableModel::Address:
         return pLeft->addrName.compare(pRight->addrName) < 0;
+    case PeerTableModel::ConnectionType:
+        return pLeft->m_conn_type < pRight->m_conn_type;
     case PeerTableModel::Network:
         return pLeft->m_network < pRight->m_network;
     case PeerTableModel::Ping:
@@ -163,6 +165,8 @@ QVariant PeerTableModel::data(const QModelIndex &index, int role) const
         case Address:
             // prepend to peer address down-arrow symbol for inbound connection and up-arrow for outbound connection
             return QString(rec->nodeStats.fInbound ? "↓ " : "↑ ") + QString::fromStdString(rec->nodeStats.addrName);
+        case ConnectionType:
+            return GUIUtil::ConnectionTypeToQString(rec->nodeStats.m_conn_type, /* prepend_direction */ false);
         case Network:
             return GUIUtil::NetworkToQString(rec->nodeStats.m_network);
         case Ping:
@@ -176,6 +180,7 @@ QVariant PeerTableModel::data(const QModelIndex &index, int role) const
         }
     } else if (role == Qt::TextAlignmentRole) {
         switch (index.column()) {
+            case ConnectionType:
             case Network:
                 return QVariant(Qt::AlignCenter);
             case Ping:
--- a/src/qt/peertablemodel.h
+++ b/src/qt/peertablemodel.h
@@ -59,12 +59,13 @@ public:
 
     enum ColumnIndex {
         NetNodeId = 0,
-        Address = 1,
-        Network = 2,
-        Ping = 3,
-        Sent = 4,
-        Received = 5,
-        Subversion = 6
+        Address,
+        ConnectionType,
+        Network,
+        Ping,
+        Sent,
+        Received,
+        Subversion
     };
 
     enum {
@@ -87,7 +88,7 @@ public Q_SLOTS:
 
 private:
     interfaces::Node& m_node;
-    const QStringList columns{tr("Peer Id"), tr("Address"), tr("Network"), tr("Ping"), tr("Sent"), tr("Received"), tr("User Agent")};
+    const QStringList columns{tr("Peer Id"), tr("Address"), tr("Type"), tr("Network"), tr("Ping"), tr("Sent"), tr("Received"), tr("User Agent")};
     std::unique_ptr<PeerTablePriv> priv;
     QTimer *timer;
 };
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -1079,10 +1079,10 @@
                <item row="1" column="0">
                 <widget class="QLabel" name="peerConnectionTypeLabel">
                  <property name="toolTip">
-                   <string>The type of peer connection: %1</string>
+                   <string>The direction and type of peer connection: %1</string>
                  </property>
                  <property name="text">
-                  <string>Connection Type</string>
+                  <string>Direction/Type</string>
                  </property>
                 </widget>
                </item>
--- a/src/qt/guiutil.h
+++ b/src/qt/guiutil.h
@@ -233,7 +233,7 @@ namespace GUIUtil
     QString NetworkToQString(Network net);
 
     /** Convert enum ConnectionType to QString */
-    QString ConnectionTypeToQString(ConnectionType conn_type);
+    QString ConnectionTypeToQString(ConnectionType conn_type, bool prepend_direction);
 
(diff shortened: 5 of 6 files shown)


jonatack:
The image shows the peer details after merge of #163, but this change is independent of it apart from including the same `net.h` header in `qt/guiutil.h`.

jarolrod (commented):
ACK 09aa053dc5a83826265aa575ee61f99f0f76287a

I love this! I was just thinking about how this would be a good feature today, you beat me to it. Code looks good to me. Tested on macOS Big Sur 11.1 with Qt 5.15.2

Screen Shot 2021-01-09 at 8 47 42 PM

jonatack:
Thanks! Though I should drop the relay type from inbound or check if full or block (same in -netinfo).

hebasto:
Concept ACK.

jonatack:
Rebased after merge of #163 and dropped the relay type for inbounds.

jonatack:
Rebased after merge of #163, added relay type checking for inbounds, and inversed the sort for the `Direction` column to be inbound first.

src/qt/peertablemodel.cpp
```diff
@@ -29,6 +29,10 @@ bool NodeLessThan::operator()(const CNodeCombinedStats &left, const CNodeCombine
         return pLeft->nodeid < pRight->nodeid;
     case PeerTableModel::Address:
         return pLeft->addrName.compare(pRight->addrName) < 0;
+    case PeerTableModel::Direction:
+        return pLeft->fInbound > pRight->fInbound;
+    case PeerTableModel::ConnectionType:
+        return pLeft->m_conn_type < pRight->m_conn_type;
```
jonatack, line 35: Memo to myself, if we prefer to sort alphabetically by type name rather than by enum order:
```diff
-        return pLeft->m_conn_type < pRight->m_conn_type;
+        return GUIUtil::ConnectionTypeToShortQString(pLeft->m_conn_type, pLeft->fRelayTxes) < GUIUtil::ConnectionTypeToShortQString(pRight->m_conn_type, pRight->fRelayTxes);
```

src/qt/peertablemodel.h
```diff
@@ -60,11 +60,12 @@ class PeerTableModel : public QAbstractTableModel
...
-        Ping = 3,
-        Sent = 4,
-        Received = 5,
-        Subversion = 6
+        Direction = 2,
+        Network = 3,
+        Ping = 4,
+        Sent = 5,
+        Received = 6,
+        Subversion = 7
```
hebasto: To improve maintainability maybe explicitly assign a value to the first element only?

src/qt/guiutil.cpp
```diff
@@ -777,6 +777,19 @@ QString ConnectionTypeToQString(ConnectionType conn_type)
     assert(false);
 }
 
+QString ConnectionTypeToShortQString(ConnectionType conn_type, bool relay_txes)
+{
+    switch (conn_type) {
+    case ConnectionType::INBOUND: return QObject::tr(relay_txes ? "Full Relay" : "Block Relay");
```
hebasto, line 783:
```suggestion
    case ConnectionType::INBOUND: return relay_txes ? QObject::tr("Full Relay") : QObject::tr("Block Relay");
```

src/qt/peertablemodel.h
jonatack: done

src/qt/guiutil.cpp
jonatack: This seems equivalent to me, but thanks for looking! done.

jonatack:
Thank you @hebasto for your feedback! Updated with your suggestions per `git diff a657b0c 9f76ba6`

src/qt/guiutil.cpp
hebasto: It is [not equivalent](https://doc.qt.io/qt-5/i18n-source-translation.html#using-tr-to-obtain-a-translation) for the `lupdate` tool:
> When Qt's translation tool, lupdate, is used to process a set of source files, the text wrapped in `tr()` calls is stored in a section of the translation file that corresponds to its translation context.

hebasto (approved):
ACK 9f76ba6597cd5d984bac0fa5c076e6b123dba22c, tested on Linux Mint 20.1 (Qt 5.12.8):

Screenshot from 2021-01-10 22-01-41

jarolrod:
re-ACK 9f76ba6597cd5d984bac0fa5c076e6b123dba22c, tested on macOS Big Sur 11.1 (Qt 5.15.2)

luke-jr:
Why are Outbound/Inbound strings better than arrows?

jonatack:
> Why are Outbound/Inbound strings better than arrows?

Good question. I proposed words instead of arrows for these reasons:

1. The peer details area uses words. When it displays, say, "Outbound Full Relay", it seems more coherent for the column data to also display "Outbound" and "Full Relay", than up-arrow and "Full Relay"
2. `-getinfo` and `-netinfo` both also use words for the direction rather than arrows

luke-jr:
Words use a lot more space than arrows... It's not unreasonable to have details show more .. detail :p

hebasto:
> Words use a lot more space than arrows... It's not unreasonable to have details show more .. detail :p

Actually, words simplifiy the GUI for new-comers, as an "arrow up" does not imply "outbound" at all.

jonatack:
Agree. And the GUI uses words as well in the main info tab.

jonatack:
Agree. And the GUI uses words for the direction in the main info tab.

jonatack:
(If space is an issue we could use in/out instead of inbound/outbound, as that is the wording in the GUI info tab, -getinfo, and -netinfo)

hebasto:
> (If space is an issue we could use in/out instead of inbound/outbound, as that is the wording in the GUI info tab, -getinfo, and -netinfo)

I was thinking about it from the beginning, but it won't work for some translations.

jarolrod:
> (If space is an issue we could use in/out instead of inbound/outbound, as that is the wording in the GUI info tab, -getinfo, and -netinfo)

I think we should stick with `inbound/outbound`. Even if you changed to `in/out` wouldn't `Direction` keep the tab at about the same size

luke-jr:
I was thinking something more like this:

https://github.com/jonatack/gui/compare/add-peers-dir-and-type-columns...luke-jr:tmp_gui_peers_dir_arrows

jarolrod:
> I was thinking something more like this:
> 
> [jonatack/gui@add-peers-dir-and-type-columns...luke-jr:tmp_gui_peers_dir_arrows](https://github.com/jonatack/gui/compare/add-peers-dir-and-type-columns...luke-jr:tmp_gui_peers_dir_arrows)

Below is a screenshot of what your branch looks like for me, font issue with the arrows. Anyways, my vote is still for `inbound/outbound`
lukejr-arrows

luke-jr:
I suppose we could also do icons with an arrow and tiny in/out text under it

luke-jr:
There, how does this look? https://github.com/jonatack/gui/compare/add-peers-dir-and-type-columns...luke-jr:tmp_gui_peers_dir_arrows

jarolrod:
Screenshot is taken from [Luke-jr's branch](https://github.com/jonatack/gui/compare/add-peers-dir-and-type-columns...luke-jr:tmp_gui_peers_dir_arrows)

I guess we should get some designer opinion on this: @Bosch-0 what do you think about this icon vs `inbound/outbound` text.

I think this idea has a lot of potential, but as this iteration I would still vote for `inbound/outbound` text. I think it's more informative and fits in better with translations. Also I think `Direction` and `Type` should remain adjacent.
Screen Shot 2021-01-18 at 12 40 44 PM

jonatack:
Thanks @luke-jr for the proposals and to @jarolrod for the screenshot renderings (a thought, no need to hide addresses if you test on signet or testnet, and signet starts up the fastest of all the chains.)

- I agree the direction column should be next to the type and network (but no strong opinion on placement of the address column; maybe after the network column would be better)
- The direction header could be shortened to "Dir", or "<->" like in -netinfo, or use icons/symbols to represent `<->`
- The column header and column values are trivial to change and I suggest leaving that for a follow-up (not by me ;)

This has 2 tested ACKs; @luke-jr if you could ACK as well this could go in and open the door for your proposed follow-up.

luke-jr:
I don't like the text form... As-is, IIRC this creates horizontal scrolling because it doesn't all fit in the window.

jonatack:
I think it depends on your window size state which since #165 is now stored and recalled, so (I believe) this is not an issue after the first resize to your preferred state.

hebasto:
> I suppose we could also do icons with an arrow and tiny in/out text under it

Any text within icons causes translation issues, no?

RandyMcMillan:
@jonatack @hebasto 

This worked for me...

https://github.com/bitcoin-core/gui/pull/135/files#diff-e721b02fefce2a90ae95da0841adfdcc610ffccf93236ef236f81c1baead8e01R163




Most of the issues you may run into - I already figured out...

https://github.com/bitcoin-core/gui/pull/135

jonatack:
This current PR with 2 ACKs is my preferred version among the various proposals I have seen.

jonatack:
I propose that further refinements be done in a follow-up (not by me ;).

RandyMcMillan:
Something I didn't get to yet - was to separate the uncomment and give it its own column.
This could be very useful for some other ideas Ive been thinking about.

luke-jr:
https://github.com/jonatack/gui/compare/add-peers-dir-and-type-columns...luke-jr:tmp_gui_peers_dir_arrows is now translation-friendly...

I'm still weak concept nack on inbound/outbound as strings in the table. I think it's worse than the status quo, so would rather this not be merged as-is.

jonatack:
Moving this back to `Draft` status until #201 is resolved.

jonatack:
Ready for review again. Three changes from the previous version:

- It no longer replaces the arrows with a `Direction` column, which had 2 ACKs but also a light NACK and which can be decided later in another pull
- For now, inbound peers do not display anything in the `Type` column, following the logic of merged #203
- In the details area, the `Connection Type` field is renamed to `Direction/Type`, so the `Type` column without a direction makes sense. Its tooltip is also updated.

I hope the translations work correctly in this latest push.

@hebasto, @jarolrod, @luke-jr, @RandyMcMillan, would you mind having another look?

jonatack:
Updated the PR description and screenshot with this one:

Screenshot from 2021-02-06 22-53-11

jarolrod:
ACK be4cf4832f117ab5f97e1367f7bdcb5361de0dae

Tested with Spanish translations. The text added in the `tooltip` and text associated with the `Type` column will need translations.

pr179-spanishtest

leonardojobim:
ACK https://github.com/bitcoin-core/gui/commit/be4cf4832f117ab5f97e1367f7bdcb5361de0dae. 
Tested on Ubuntu 20.04 on VMWare.

image

leonardojobim (approved):
Tested ACK https://github.com/bitcoin-core/gui/commit/be4cf4832f117ab5f97e1367f7bdcb5361de0dae  on Ubuntu 20.04 on VMWare.

image

RandyMcMillan:
Removing "Peer" from the "Peer Id" column would allow for more prudent use of space.
Otherwise it should be translatable. A simple "ID" or "id" would be better. I think all caps "ID" is better imo. 

Maybe Type and Network should be left justified?

---

NOTE: https://github.com/bitcoin-core/gui/pull/202

PR #202 is going to allow for more info/columns to be added - and more info displayed "at a glance". As more columns are added - the columns can be adjusted for a tighter fit and/or column stretching in a resize event. 


--- 

ACK 6fc72bd6f030c8b49e8b9f68188413766a239d73

macOS 10.14.6 (18G103)

Screen Shot 2021-02-21 at 7 22 01 PM

PR #183, opened 2021-01-13
Add include for std::bind.

sinetek:
Hi, this patch adds in <functional> because the GUI code makes use of std::bind.
That's all.

### DIFF
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -35,6 +35,7 @@
 
 #include <algorithm>
 #include <cstdint>
+#include <functional>
 #include <unordered_map>
 
 #include <math.h>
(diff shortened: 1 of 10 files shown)


MarcoFalke:
Not sure if it is worth to do one-off fixes like this. How would reviewers check that the patch is both necessary and complete?

luke-jr:
As opposed to what? It's a bug and should be fixed...

MarcoFalke:
There are also missing `<cmath>` includes all over the place. What I am saying is that we need an automated tool like icwyu to do this in a reproducible and automated manner. Otherwise this will be a waste of our precious reviewers' time.

luke-jr (approved):
utACK (all files with include added do use std::bind, and std::bind requires the include)

jonasschnelli:
This touches more non GUI files, thus should perhaps have been opened at the main repository (but let's keep it here for now).
utACK 2a39ccf1334ef3c48c6f9969a0fc916b9e10aae1

PR #186, opened 2021-01-16
Add information to "Confirm fee bump" window

prayank23:
+ Add information in bump fee confirmation box according to the documentation: https://bitcoincore.org/en/doc/0.20.0/rpc/wallet/bumpfee/

+ Workaround to fix issue: https://github.com/bitcoin/bitcoin/issues/20795 in which user isn't aware of new inputs, outputs added to replacement transaction before broadcasting. Initial transaction had used coin control features and custom change address. Until the issue is fixed by change in coin selection algorithm we can add this warning.

+ Waiting for comments from devs who are working on coin selection algorithm PRs or involved in related research. However got two comments from Luke Dashjr and Pieter Wuille:

_luke-jr: Reducing the change output also could be a privacy problem, since it identifies which output was change._ 

_sipa: Wallet doesn't know the original transaction was using coin control. So I think its expected that if you use automatic fee bumping, you'll get whatever the coin selection algorithm decides. As for why its not decreasing the change and instead adding another input, that may be a bug._  (IRC: #bitcoin-core-dev)

### DIFF
--- a/src/qt/walletmodel.cpp
+++ b/src/qt/walletmodel.cpp
@@ -514,6 +514,13 @@ bool WalletModel::bumpFee(uint256 hash, uint256& new_hash)
     questionString.append("</td><td>");
     questionString.append(BitcoinUnits::formatHtmlWithUnit(getOptionsModel()->getDisplayUnit(), new_fee));
     questionString.append("</td></tr></table>");
+
+    // Display warning in the "Confirm fee bump" window if the "Coin Control Features" option is enabled
+    if (getOptionsModel()->getCoinControlFeatures()) {
+        questionString.append("<br><br>");
+        questionString.append(tr("Warning: This may pay the additional fee by reducing change outputs or adding inputs, when necessary. It may add a new change output if one does not already exist. These changes may potentially leak privacy."));
+    }
+
     SendConfirmationDialog confirmationDialog(tr("Confirm fee bump"), questionString);
     confirmationDialog.exec();
     QMessageBox::StandardButton retval = static_cast<QMessageBox::StandardButton>(confirmationDialog.result());


percy-g2 (changes requested):
Concept ACK

Compiled successfully on Ubuntu 20.04.1 LTS and tested with below details:

Initial Tx: `64e2c03527b50a934f9cdedf87813ab534023afd161c907f4380dd24506f9a71`
Selected one of the Inputs, used one address from wallet as change address and broadcasted RBF enabled Tx.

Replacement Tx: `02a48410d61974125ed145c465004e7e7c681b2398108ed4c4a5acea261b160c`
New inputs and Outputs were added automatically.

User was informed about the changes in the "Confirm fee bump" dialog:
image

Can add one more `<br />` in Line 517

prayank23:
> Can add one more <br /> in Line 517

Done. Added in https://github.com/bitcoin-core/gui/pull/186/commits/014498e9df2f8df7b1592d73dcf9d8d2fd5921a6

percy-g2 (approved):
ACK https://github.com/bitcoin-core/gui/pull/186/commits/014498e9df2f8df7b1592d73dcf9d8d2fd5921a6

hebasto (commented):
IIUC, the purpose of this change is to warn users about privacy issues while bumping a fee, right?

Why not to add this statement explicitly (however, not sure about exact wording) with an appropriate subtitle? Smth like _"WARNING: fee bumping could leak privacy if new inputs and change output will be added to the transaction"_.

@prayank23
Do you mind changing this PR title s/Add information in RBF pop-up/Add information to "Confirm fee bump" window/ ?

prayank23:
> IIUC, the purpose of this change is to warn users about privacy issues while bumping a fee, right?

Inform user about changes in replacement transaction apart from fees which is already mentioned. Adding inputs and outputs can be considered good or bad for privacy or fees depending on the user.

As luke-jr mentioned reducing change output is also a privacy issue 

> Why not to add this statement explicitly (however, not sure about exact wording) with an appropriate subtitle? Smth like "WARNING: fee bumping could leak privacy if new inputs and change output will be added to the transaction".

I was considering adding the word "warning" but didn't. It can be added. Mentioning about privacy looks scary and as I explained above it may not be a privacy issue for some users. Let me explain with one example:

Alice uses coin control features to select one input, custom change address and broadcasts RBF enabled tx. It is not confirmed for hours and she needs this amount urgently on one exchange for trading. She increases the fee with a replacement transaction not aware of other inputs and outputs being added, broadcasts it. There were few UTXOs in the wallet which were associated with darknet transactions and some of them were used in replacement transaction. In this case it's a privacy issue.

Bob has a business and uses coin control features to select one input, custom change address and broadcasts RBF enabled tx. Not confirmed for hours, few users related to this business are waiting. Bob increases the fee with replacement transaction which adds few extra inputs and outputs but Bob doesn't care as this wallet has all transactions related to business. Fees could have been less if less inputs/outputs but priority in this case is to complete the users requests asap.

hebasto:
> Adding inputs and outputs can be considered good or bad for privacy or fees depending on the user.

If a txout, that was not correlated to a user/wallet in any way, is added to a transaction via a new input, it is always bad for privacy. However, it is true that some users do not care about it.

prayank23:
Makes sense. Users may not care but it affects privacy. Will rephrase the sentence and add "warning" and "privacy".

src/qt/walletmodel.cpp
```diff
@@ -514,6 +514,8 @@ bool WalletModel::bumpFee(uint256 hash, uint256& new_hash)
     questionString.append("</td><td>");
     questionString.append(BitcoinUnits::formatHtmlWithUnit(getOptionsModel()->getDisplayUnit(), new_fee));
     questionString.append("</td></tr></table>");
+    questionString.append("<br /><br />");
```
hebasto, line 517: nit: Just `<br>` is used in the most codebase in this repository.

src/qt/walletmodel.cpp
```diff
@@ -514,6 +514,8 @@ bool WalletModel::bumpFee(uint256 hash, uint256& new_hash)
...
+    questionString.append(tr("WARNING: New inputs and outputs can be added in replacement transaction if necessary which affects privacy"));
```
hebasto, line 518: nit: Could a full stop at the sentence end make it better?

hebasto (commented):
Approach ACK 6be2bff70f93c48807bf3620cdef6fe0b40fbe89

@prayank23 
Do you mind squashing all commits into the single one?

hebasto:
@achow101 @meshcollider  @luke-jr @Sjors @sipa 

Do you mind looking into this PR?

src/qt/walletmodel.cpp
luke-jr: This seems to confusing to normal people. They don't know what inputs or outputs are, nor necessarily care how Bitcoin works internally.

hebasto: > This seems to confusing to normal people.

:D

> They don't know what inputs or outputs are, nor necessarily care how Bitcoin works internally.

Educational purpose, no?

Could it be re-worded without revealing of internal tx details?

percy-g2: IMO best way to rework this is, if expert options for coin control features is enabled then

-  Show proper internal TX details or a warning message in dialog like this.

@prayank23  @luke-jr @hebasto

src/qt/walletmodel.cpp
percy-g2: @hebasto  yeah by just adding that  before table row will fix it  instead of putting it in append.

@prayank23 below line should fix it

` questionString.append<br/>(tr("WARNING: New inputs and outputs can be added in replacement transaction if necessary which affects privacy"));`

prayank23: Done. Replaced `<br />` with `<br>`

src/qt/walletmodel.cpp
prayank23: > nit: Could a full stop at the sentence end make it better?

Done

> This seems to confusing to normal people. They don't know what inputs or outputs are, nor necessarily care how Bitcoin works internally.

"Inputs" is already mentioned in GUI in 'Coin Control Features", user who did the initial transaction by selecting inputs obviously cares about it and "Outputs" can be replaced with "Change Address" although it can mean different things but easier to understand.

prayank23:
> Do you mind squashing all commits into the single one?

@hebasto Will do once there are enough ACKs and we are sure about the statement that can be used in this warning.

hebasto:
@prayank23 
> > Do you mind squashing all commits into the single one?
> 
> @hebasto Will do once there are enough ACKs and we are sure about the statement that can be used in this warning.

Doing in such a way all ACKs will be invalidated. It will require additional re-reviewing that increases the review burden.

prayank23:
> Doing in such a way all ACKs will be invalidated. It will require additional re-reviewing that increases the review burden.

Done. Squashed commits.

hebasto (approved):
ACK 3d93f27dc2361428dbbd57a11d99b7681faf0a2d, tested on Linux Mint 20.1 (Qt 5.12.8):

Screenshot from 2021-01-21 03-08-17

More sophisticated solutions (e.g., opening a coin control dialog) could be added in follow ups.

src/qt/walletmodel.cpp
luke-jr: Users also shouldn't *need* to know about internals.

Coin Control Features are an advanced feature.

Maybe just only append this when it's enabled?

prayank23: > Maybe just only append this when it's enabled?

This makes sense. What is the best way to know if coin control was used for a transaction? So that it can be used in `if` statement for this warning.

src/qt/walletmodel.cpp
```diff
@@ -514,6 +514,8 @@ bool WalletModel::bumpFee(uint256 hash, uint256& new_hash)
     questionString.append("</td><td>");
     questionString.append(BitcoinUnits::formatHtmlWithUnit(getOptionsModel()->getDisplayUnit(), new_fee));
     questionString.append("</td></tr></table>");
+    questionString.append("<br><br>");
+    questionString.append(tr("WARNING: New inputs and outputs can be added in replacement transaction if necessary which affects privacy."));
```
jonatack, line 518: grammar nit, "which" needs a preceding comma
```suggestion
    questionString.append(tr("WARNING: New inputs and outputs can be added in replacement transaction if necessary, which affects privacy."));
```

jonatack: An article ("a", etc.) is also missing before the word "replacement"

jonatack: "can be added...if necessary" is vague, would suggest being more precise: added by what, why would it be necessary.

prayank23: > An article ("a", "the", etc.) is also missing before the word "replacement"

Will add "the"

> "can be added...if necessary" is vague, would suggest being more precise: added by what, why would it be necessary.

I rephrased _'adding inputs when necessary. It may add a new change output if one does not already exist.'_ from https://bitcoincore.org/en/doc/0.20.0/rpc/wallet/bumpfee/ Not sure what else can I add which even normal users can understand

jonatack: Thanks for updating! (very kind but no need to credit me for a drive-by comment) can you squash to one commit?

jonatack: Oh I see, sorry, this is based somewhat on the bumpfee help.
```
The command will pay the additional fee by reducing change outputs or adding inputs when necessary.
It may add a new change output if one does not already exist.
All inputs in the original transaction will be included in the replacement transaction.
```

jonatack: I'm not sure this should have a screaming WARNING.

Maybe:

```
This will pay the additional fee by reducing change outputs
or adding inputs when necessary. It may add a new change output
if one does not already exist. These changes may affect privacy.
```

prayank23: > can you squash to one commit?

Sure will do once we finalize the warning message. Does the below mentioned thing look okay?

**WARNING:** This will pay the additional fee by reducing change outputs
or adding inputs when necessary. It may add a new change output
if one does not already exist. These may potentially leak privacy.

prayank23: But the message will be incorrect because the issue is more than just user not aware of how RBF works in bitcoin core wallet. It is also the coin selections algorithm when using a custom change address in the initial transaction. Original Issue: https://github.com/bitcoin/bitcoin/issues/20795

prayank23: No extra inputs or outputs were added in this: https://tbtc.bitaps.com/f35a69b066d1ff79cb13bc6dba2eb345f239b2bfa268e3fb0ee06b28748bea99 when I didn't use "custom change address" for initial transaction

dooglus: If you `setlabel` the change address before bumping the fee it causes Core not to recognize the change address as a change address and to therefore add a new input. I think it makes sense to label all your change UTXOs if you plan to use coin control for privacy. Without labelling your change how do you hope to prevent undesired UTXO linkage?

I wrote about this in bitcoin/bitcoin#20935.

src/qt/walletmodel.cpp
luke-jr: AFAIK it isn't possible right now, we'd just need to assume the current checkbox state is what the user wants behaviour based on

src/qt/walletmodel.cpp
jonatack: I could fix the issue but I doubt it would see enough review; my fee rate work has just been sitting there for the past couple months.

As for the message, could s/will/may/, e.g.:

"This may pay the additional fee by reducing change outputs or adding inputs, when necessary. It may add a new change output if one does not already exist. These may potentially leak privacy."

prayank23: @dooglus Thanks for sharing. Will do some test and respond in https://github.com/bitcoin/bitcoin/issues/20935

@jonatack Made changes in https://github.com/bitcoin-core/gui/commit/570ebc6238639d40c7d54bbf39857de9416f2ba2

src/qt/walletmodel.cpp
prayank23: IMO a generic warning message based on `bumpfee` docs is good enough for now. In long term we need to fix few other issues related to coin selection algorithm, coin control for RBF etc. It can add random inputs even when coin control is disabled.

luke-jr: With CC disabled, it was *created* with random inputs already...

luke-jr (changes requested):
(NACK until message is either CoinControl-only or understandable for non-technical folks)

jonatack:
utACK 570ebc6238639d40c7d54bbf39857de9416f2ba2 in terms of the message content

I understand @luke-jr's disagreement with the change but I don't know enough about GUI user understanding or expertise level to have an opinion.

prayank23:
> (NACK until message is either CoinControl-only or understandable for non-technical folks)

@luke-jr 

Done. Added an `if` statement in https://github.com/bitcoin-core/gui/commit/0ae1ce0eb7fb559cccbbca961082dc4e506187ff


"Confirm fee bump" Windows with Coin Control Features option disabled:
image



"Confirm fee bump" Windows with Coin Control Features option enabled:
image

src/qt/walletmodel.cpp
prayank23: Added `if` statement to check if CC is enabled

src/qt/walletmodel.cpp
```diff
@@ -514,6 +514,13 @@ bool WalletModel::bumpFee(uint256 hash, uint256& new_hash)
     questionString.append("</td><td>");
     questionString.append(BitcoinUnits::formatHtmlWithUnit(getOptionsModel()->getDisplayUnit(), new_fee));
     questionString.append("</td></tr></table>");
+
+    //Check if "Coin Control Features" options is enabled to display warning in "Confirm fee bump" Window
+    if (getOptionsModel()->getCoinControlFeatures() == true) {
+    questionString.append("<br><br>");
+    questionString.append(tr("Warning: This may pay the additional fee by reducing change outputs or adding inputs, when necessary. It may add a new change output if one does not already exist. These may potentially leak privacy."));
+    }
```
jonatack, line 522: you can run ` clang-format -i src/qt/walletmodel.cpp --lines=518:522` to format your code automatically
```diff
     if (getOptionsModel()->getCoinControlFeatures() == true) {
-    questionString.append("<br><br>");
-    questionString.append(tr("Warning: This may pay the additional fee by reducing change outputs or adding inputs, when necessary. It may add a new change output if one does not already exist. These may potentially leak privacy."));
+        questionString.append("<br><br>");
+        questionString.append(tr("Warning: This may pay the additional fee by reducing change outputs or adding inputs, when necessary. It may add a new change output if one does not already exist. These may potentially leak privacy."));
     }
```

src/qt/walletmodel.cpp
```diff
@@ -514,6 +514,13 @@ bool WalletModel::bumpFee(uint256 hash, uint256& new_hash)
     questionString.append("</td><td>");
     questionString.append(BitcoinUnits::formatHtmlWithUnit(getOptionsModel()->getDisplayUnit(), new_fee));
     questionString.append("</td></tr></table>");
+
+    //Check if "Coin Control Features" options is enabled to display warning in "Confirm fee bump" Window
```
jonatack, line 518: nit, not sure this comment is needed, but if you keep it
```suggestion
    // Display warning in the "Confirm fee bump" window if the "Coin Control Features" option is enabled
```

src/qt/walletmodel.cpp
```diff
@@ -514,6 +514,13 @@ bool WalletModel::bumpFee(uint256 hash, uint256& new_hash)
...
+    if (getOptionsModel()->getCoinControlFeatures() == true) {
```
jonatack, line 519:
```suggestion
    if (getOptionsModel()->getCoinControlFeatures()) {
```

jonatack (commented):
A few suggestions below; thanks for updating.

src/qt/walletmodel.cpp
prayank23: Thanks

jonatack:
utACK 84add0d8dc23b44a5ca8f7089653cc7ad5081404

src/qt/walletmodel.cpp
```diff
@@ -514,6 +514,13 @@ bool WalletModel::bumpFee(uint256 hash, uint256& new_hash)
     questionString.append("</td><td>");
     questionString.append(BitcoinUnits::formatHtmlWithUnit(getOptionsModel()->getDisplayUnit(), new_fee));
     questionString.append("</td></tr></table>");
+
+    // Display warning in the "Confirm fee bump" window if the "Coin Control Features" option is enabled
+    if (getOptionsModel()->getCoinControlFeatures()) {
+        questionString.append("<br><br>");
+        questionString.append(tr("Warning: This may pay the additional fee by reducing change outputs or adding inputs, when necessary. It may add a new change output if one does not already exist. These may potentially leak privacy."));
```
luke-jr, line 521:
```suggestion
        questionString.append(tr("Warning: This may pay the additional fee by reducing change outputs or adding inputs, when necessary. It may add a new change output if one does not already exist. These changes may potentially leak privacy."));
```

luke-jr (approved):
utACK, minor language nit

hebasto (approved):
ACK 84add0d8dc23b44a5ca8f7089653cc7ad5081404

Happy to re-ACK after addressing the recenet @luke-jr's [suggestion](https://github.com/bitcoin-core/gui/pull/186#discussion_r564054798).

jonasschnelli:
Tested ACK - 232d1f92bb4c99ce0f5d210b17562c96c32ab61a

jonasschnelli:
Bildschirmfoto 2021-01-26 um 10 26 59

(nit: not sure if it needs to be bold)

PR #188, opened 2021-01-19
Write PSBTs to file with binary mode

achow101:
As noted in https://github.com/bitcoin/bitcoin/issues/20959, PSBT files should be opened in binary mode as on windows, all newlines are turned into CRLF which produces invalid PSBTs.

Fixes https://github.com/bitcoin/bitcoin/issues/20959

### DIFF
--- a/src/qt/psbtoperationsdialog.cpp
+++ b/src/qt/psbtoperationsdialog.cpp
@@ -145,7 +145,7 @@ void PSBTOperationsDialog::saveTransaction() {
     if (filename.isEmpty()) {
         return;
     }
-    std::ofstream out(filename.toLocal8Bit().data());
+    std::ofstream out(filename.toLocal8Bit().data(), std::ofstream::out | std::ofstream::binary);
     out << ssTx.str();
     out.close();
     showStatus(tr("PSBT saved to disk."), StatusLevel::INFO);
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -434,7 +434,7 @@ void SendCoinsDialog::on_sendButton_clicked()
             if (filename.isEmpty()) {
                 return;
             }
-            std::ofstream out(filename.toLocal8Bit().data());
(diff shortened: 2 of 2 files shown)


MarcoFalke:
Approach ACK. Is there a way to test this on non-windows?

hebasto:
Concept ACK.

For reference: https://en.cppreference.com/w/cpp/io/c/FILE#Binary_and_text_modes

hebasto:
@Quantris Does this pr fix https://github.com/bitcoin/bitcoin/issues/20959 for you?

Talkless (approved):
utACK cc3971c9ff538a924c1a76ca1352bcaeb24f579f. 

I DID test, but only on Linux (Debian Sid with Qt 5.15.2). I've created `.psbt` with PR version, imported into master version, and was able to sing and broadcast (regtest), so at least I don't see regressions in Linux.

MarcoFalke:
Backported in https://github.com/bitcoin/bitcoin/pull/20901

Quantris:
> @Quantris Does this pr fix [bitcoin/bitcoin#20959](https://github.com/bitcoin/bitcoin/issues/20959) for you?

Sorry I missed replying to this at the time. I'm not really set up to build the project myself so I wouldn't have been able to test this without someone providing a binary.

I can affirm that the issue seems fixed in 0.21.1 which includes this change; I tested in Win10 with the following steps in a watch-only wallet:
1) create a partial transaction & save to file test.psbt (during this process it is also copied to clipboard in base64)
2) load from clipboard, confirm it displays as expected & save to another file test2.psbt
3) load from file test.psbt, confirm it displays as expected
4) load from file test2.psbt, confirm it displays as expected

The same process in 0.21.0 fails in steps 3 & 4.

I also looked around a bit for info on testing this kind of issue. Unfortunately I didn't find anything that makes text-mode vs. binary-mode distinguishable in linux (I was thinking maybe there would be some kind of env var or build flag for the stdlib). IOW seems like the only way to test this would be to run the tests in a Windows env.

Thanks for the fix @achow101 !

Issue #101, opened 2020-10-08
Drop workaround for fixed QTBUG-42503


hebasto:
In the code
https://github.com/bitcoin-core/gui/blob/9dd4de2832e298398436b957375d62de67f0a928/src/qt/overviewpage.cpp#L128

we should drop the workaround for [QTBUG-42503](https://bugreports.qt.io/browse/QTBUG-42503) which is already fixed since Qt 5.5.0.

Disclaimer: I'm not able to test a patch as testing requires retina display.

prusnak:
PR in https://github.com/bitcoin-core/gui/pull/189

jonasschnelli:
I keep this PR open since we still use a `QPushButton` for those icons (rather than a `QLabel`).

prusnak:
How should a QLabel with an icon work?

Via the html img hack?

``` C
QLabel("<html><img src=':/resource.png'></html>")
```

Or should we use 2 QLabels for each QPushButton (one with an icon, the other with a text)?

jonasschnelli:
AFAIK you can just use `QLabel::setPixmap()` (along with `setText()`).

laanwj:
I am fairly sure you need two labels in that case, one with the pixmap one with the text. Otherwise, the label doesn't know how you want to position the icon and text relative to each other.

When it's possible to avoid using HTML that's generally preferable.

prusnak:
I also think it's not possible to set both icon and text for a label.

Here's the trick how to solve it via VBoxLayout containing two labels https://forum.qt.io/topic/64020/settext-and-setpixmap-using

```
QWidget * container = new QWidget(this);
QVBoxLayout * layout = new QVBoxLayout(container);
QLabel * textLabel = new QLabel(container);
QLabel * imageLabel = new QLabel(container);

layout->addWidget(imageLabel);
layout->addWidget(textLabel);

imageLabel->setPixmap(...);
textLabel->setText(...);
```

Honestly, I am not so sure that's better than using a QPushButton

jonasschnelli:
Let's keep the QPushButton then. Closing.

PR #189, opened 2021-01-19
qt: drop workaround for QTBUG-42503 which was fixed in Qt 5.5.0

prusnak:
Fixes https://github.com/bitcoin-core/gui/issues/101

### DIFF
--- a/src/qt/overviewpage.cpp
+++ b/src/qt/overviewpage.cpp
@@ -126,7 +126,6 @@ OverviewPage::OverviewPage(const PlatformStyle *platformStyle, QWidget *parent) 
 
     // use a SingleColorIcon for the "out of sync warning" icon
     QIcon icon = platformStyle->SingleColorIcon(":/icons/warning");
-    icon.addPixmap(icon.pixmap(QSize(64,64), QIcon::Normal), QIcon::Disabled); // also set the disabled icon because we are using a disabled QPushButton to work around missing HiDPI support of QLabel (https://bugreports.qt.io/browse/QTBUG-42503)
     ui->labelTransactionsStatus->setIcon(icon);
     ui->labelWalletStatus->setIcon(icon);
 


hebasto:
@prusnak 

Thanks! Do you mind providing a relevant screenshot from a retina display?

jarolrod (commented):
Concept ACK

The workaround was actually introduced in 7247d103ffc2772b6dd997b7aa4d7147a8044d73 where we moved from a warning text to a warning icon. The `QPushButton` was used since at the time of that PR the [QTBUG-42503](https://bugreports.qt.io/browse/QTBUG-42503) was already known. Otherwise, going by the comment `"to work around missing HiDPI support of QLabel"` a `QLabel` would have been used.

Since `QLabel` is no longer affected by this bug, I guess fixing the workaround would mean refactoring the icon from a QPushButton to a QLabel. But, since we have additional logic with the modal overlay, maybe it makes more sense to keep it a `QPushButton`.

prusnak:
> Thanks! Do you mind providing a relevant screenshot from a retina display?

Unfortunately, I have issues with linking the bitcoin-qt binary on my system with retina display due to unrelated error.

jarolrod:
@prusnak @hebasto Here are screenshots taken on a Retina (HiDPI) display (2020 MacBook Pro).

While the line that is removed in this PR is no longer needed, I fail to see how it "fixes" the workaround. The workaround was to use a disabled `QPushButton` because `QLabel` had a regression where it failed to scale by the `devicePixelRatio` (this is no longer the case, see: [patch](https://codereview.qt-project.org/c/qt/qtbase/+/127002)). The fix would not be to continue to use the `QPushButton` unless it is decided to keep it.

**Master**
master

**PR**
pr

jonasschnelli:
Tested ACK a2a3f4cd8d53e33b0e1b1aeff073b096ade034e9 -   
Bildschirmfoto 2021-01-26 um 10 40 43

Issue #191, opened 2021-01-20
User Agent header flickers when trying to resize any column


hebasto:
![header_bug](https://user-images.githubusercontent.com/32963518/105186299-3cba3880-5b3a-11eb-92dc-42f408ce88da.gif)

System: Linux Mint 20.1 + Cinnamon DE + Qt 5.12.8

jarolrod:
Confirming the same happens on macOS 11.1 + Qt 5.15.2

RandyMcMillan:
Appending a "bumper" as the last column of the peertablemodel should fix this issue.

In this stale PR - I found that was a reasonable solution.

https://github.com/bitcoin-core/gui/pull/135/files#diff-e721b02fefce2a90ae95da0841adfdcc610ffccf93236ef236f81c1baead8e01R112

hebasto:
> Appending a "bumper" as the last column of the peertablemodel should fix this issue.
> 
> In this stale PR - I found that was a reasonable solution.
> 
> https://github.com/bitcoin-core/gui/pull/135/files#diff-e721b02fefce2a90ae95da0841adfdcc610ffccf93236ef236f81c1baead8e01R112

@RandyMcMillan 
Could that change be separated from #135 into its own bugfix PR?

RandyMcMillan:
I will take a look - the PR may be too stale.

RandyMcMillan:
I should have a fix in a few minutes.

hebasto:
@RandyMcMillan 
Another option is to fix the `TableViewLastColumnResizingFixer` class.

hebasto:
Interesting that the `banlistWidget` has the same properties as the `peerWidget`, but the former works without flickering.

hebasto:
It appears that this bug is fixed in f73d340abeedf7e583fe9b2a8b3685a745a9199e (#164).

@jarolrod Could you check it please?

jarolrod:
@hebasto testing now

One note about this bug is that it happens if that `User Agent` column is a certain size. If I resize the `User Agent` column to be smaller, then it doesn't happen when I got to resize other columns. But, if I bring it back to its original size, the bug shows up.

![bug-gif](https://user-images.githubusercontent.com/23396902/105268249-d811d480-5b5f-11eb-807e-035866e70ddd.gif)

jarolrod:
@hebasto confirming that #164 fixes this bug, seems that your refactoring of `PeerTableModel::refresh()` did the job

PR #192, opened 2021-01-20
fix for #191

RandyMcMillan:
@hebasto - here is a barebones fix for #191 
We can tweak the presentation in a follow up PR.

### DIFF
--- a/src/qt/peertablemodel.h
+++ b/src/qt/peertablemodel.h
@@ -64,7 +64,8 @@ public:
         Ping = 3,
         Sent = 4,
         Received = 5,
-        Subversion = 6
+        Subversion = 6,
+        Bumper = 7
     };
 
     enum {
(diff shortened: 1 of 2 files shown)


hebasto:
@RandyMcMillan 

Thanks for fixing!

1. Did you find out the reason of head flickering?
2. Could using of `GUIUtil::TableViewLastColumnResizingFixer` be a better option as it is used in other places of the GUI?
3. Do you mind dropping `@` from the PR description?

jarolrod (commented):
This PR does not fix #191 

It instead creates a new column in the Peer Table model. A fix to this should not hack on a new column so that the `User Agent` column does not glitch. Perhaps a refactor that creates a `PeerTableModelView` which makes use of `TableViewLastColumnResizingFixer` is more appropriate.

**PR:** Creates a new column to the right of `User Agent`
Screen Shot 2021-01-20 at 3 51 23 PM

RandyMcMillan:
@jarolrod - It sounds like you have the motivation to implement a proper solution.
I will be excited to see how you refactor this into a proper MVC implementation.
I will be ecstatic to see it actually get merged!
Happy coding!

jarolrod:
@RandyMcMillan was just relaying what I found while reviewing the PR. Wasn't trying to discourage you in anyway or come off as rude.

PR #180, opened 2021-01-10
Peer details: connection type follow-ups

jonatack:
Three follow-ups to #163:
- return relay type for inbound peers
- improve markup handling in the tooltip to facilitate translations
- update ConnectionType doxygen documentation
 
Screenshot from 2021-01-11 08-37-44

PR #194, opened 2021-01-21
Save/restore RPCConsole geometry only for window

hebasto:
After using the GUI with `-disablewallet` the "Node window" inherits the geometry of the main window, that could be unexpected for users.

This PR provides independent geometry settings for `RPCConsole` in both modes:
- window sizes and `QSplitter` sizes when `-disablewallet=0`
- only `QSplitter` sizes when `-disablewallet=1`

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -453,12 +453,20 @@ RPCConsole::RPCConsole(interfaces::Node& node, const PlatformStyle *_platformSty
 {
     ui->setupUi(this);
     QSettings settings;
-    if (!restoreGeometry(settings.value("RPCConsoleWindowGeometry").toByteArray())) {
-        // Restore failed (perhaps missing setting), center the window
-        move(QGuiApplication::primaryScreen()->availableGeometry().center() - frameGeometry().center());
+#ifdef ENABLE_WALLET
+    if (WalletModel::isWalletEnabled()) {
+        // RPCConsole widget is a window.
+        if (!restoreGeometry(settings.value("RPCConsoleWindowGeometry").toByteArray())) {
+            // Restore failed (perhaps missing setting), center the window
+            move(QGuiApplication::primaryScreen()->availableGeometry().center() - frameGeometry().center());
+        }
+        ui->splitter->restoreState(settings.value("RPCConsoleWindowPeersTabSplitterSizes").toByteArray());
+    } else
+#endif // ENABLE_WALLET
+    {
+        // RPCConsole is a child widget.
+        ui->splitter->restoreState(settings.value("RPCConsoleWidgetPeersTabSplitterSizes").toByteArray());
     }
-
-    ui->splitter->restoreState(settings.value("PeersTabSplitterSizes").toByteArray());
 
     constexpr QChar nonbreaking_hyphen(8209);
     const std::vector<QString> CONNECTION_TYPE_DOC{
@@ -516,8 +524,18 @@ RPCConsole::RPCConsole(interfaces::Node& node, const PlatformStyle *_platformSty
 RPCConsole::~RPCConsole()
 {
     QSettings settings;
-    settings.setValue("RPCConsoleWindowGeometry", saveGeometry());
-    settings.setValue("PeersTabSplitterSizes", ui->splitter->saveState());
+#ifdef ENABLE_WALLET
+    if (WalletModel::isWalletEnabled()) {
+        // RPCConsole widget is a window.
+        settings.setValue("RPCConsoleWindowGeometry", saveGeometry());
+        settings.setValue("RPCConsoleWindowPeersTabSplitterSizes", ui->splitter->saveState());
+    } else
+#endif // ENABLE_WALLET
+    {
+        // RPCConsole is a child widget.
+        settings.setValue("RPCConsoleWidgetPeersTabSplitterSizes", ui->splitter->saveState());
+    }
+
     m_node.rpcUnsetTimerInterface(rpcTimerInterface);
     delete rpcTimerInterface;
     delete ui;


MarcoFalke:
```
qt/rpcconsole.cpp:(.text+0x77ed): undefined reference to `WalletModel::isWalletEnabled()'

**Comment by hebasto** [comment]
Fixed.

**Review by luke-jr — COMMENTED** [review]
Concept ACK, though the code seems uglier than I would like (ignore this nit if we can't think of a better way)

Perhaps add a version number or comment to the splitter settings while you're at it, since adding/changing columns likely shouldn't use old sizes... Or at least a comment reminding us to do so later when/if we change the columns again.

**Review by promag — COMMENTED** [review]
Code review ACK b5523f1aaa526b49a0678be6abe7270bdb2fc497.

I think this `#ifdef` madness could go away by moving `WalletModel::isWalletEnabled()` to another place where it always compiled in.

**Comment by luke-jr** [comment]
>I think this #ifdef madness could go away by moving WalletModel::isWalletEnabled() to another place where it always compiled in.

But then the code here will always be compiled in... the whole point of not building wallet support is to omit the code for it :p

**Comment by promag** [comment]
@luke-jr I understand that but my proposal doesn't mean wallet support is compiled in, only downside is a runtime check:
```cpp
bool IsWalletEnabled()
{
#ifdef ENABLE_WALLET
    return !gArgs.GetBoolArg("-disablewallet", DEFAULT_DISABLE_WALLET);
#else
    return false;
#endif
}
```

luke-jr:
But with this, everything conditional on `IsWalletEnabled()` (in this case, the settings code) is being compiled in to a non-wallet build.

(Unless perhaps IsWalletEnabled is carefully made always-inlined so the compiler knows to remove the branch?)

hebasto:
Rebased b5523f1aaa526b49a0678be6abe7270bdb2fc497 -> 01d9586ae85e49efaa00f11c1f26c24c9b82b278 ([pr194.02](https://github.com/hebasto/gui/commits/pr194.02) -> [pr194.03](https://github.com/hebasto/gui/commits/pr194.03)) due to the conflict with #180.

jarolrod (commented):
Can you show a screenshot of the behavior this `PR` is supposed to exhibit and what it is supposed to fix? Comparing with `master`, I don't see what this is supposed to do.

Testing on macOS 11.1 with Qt 5.15.2

**Master**
---

**Wallet Enabled**
Run `bitcoin-qt` with wallet enabled. The main window is resized to be wider.

| Main Window      | Node Window |
| ----------- | ----------- |
|Screen Shot 2021-02-19 at 3 51 23 PM|Screen Shot 2021-02-19 at 3 51 52 PM|

**Wallet Disabled**
run `bitcoin-qt` with `disablewallet=1`. The `Node Window` has inherited the geometry of the `Main Window` when `bitcoin-qt` was run with wallet enabled.

| Node Window      | 
| ----------- | 
| Screen Shot 2021-02-19 at 3 52 20 PM| 

Now, I'll resize the `Node Window` to be smaller. 

| Node Window      | 
| ----------- | 
|Screen Shot 2021-02-19 at 3 53 22 PM| 

**Restart With Wallet Re-Enabled**
The `Main Window` inherits the geometry of the `Node Window` when I resized the `Node Window` while `bitcoin-qt` was run with the `-disablewallet=1`.

| Main Window      | 
| ----------- | 
|Screen Shot 2021-02-19 at 3 54 26 PM| 






**PR**
---
**Wallet Enabled**
Run `bitcoin-qt` with wallet enabled. The main window is resized to be wider.

| Main Window      | Node Window |
| ----------- | ----------- |
|Screen Shot 2021-02-19 at 4 01 08 PM | Screen Shot 2021-02-19 at 4 01 22 PM |

**Wallet Disabled**
run `bitcoin-qt` with `disablewallet=1`. The `Node Window` has inherited the geometry of the `Main Window` when `bitcoin-qt` was run with wallet enabled. This is the same behavior as `master`.

| Node Window      | 
| ----------- | 
|Screen Shot 2021-02-19 at 4 02 04 PM| 

Now, I'll resize the `Node Window` to be smaller. 

| Node Window      | 
| ----------- | 
|Screen Shot 2021-02-19 at 4 02 18 PM| 

**Restart With Wallet Re-Enabled**
The `Main Window` inherits the geometry of the `Node Window` when I resized the `Node Window` while `bitcoin-qt` was run with the `-disablewallet=1`. This is the same behavior as `master`.

| Main Window      | 
| ----------- | 
| Screen Shot 2021-02-19 at 4 02 39 PM|

hebasto:
@jarolrod 
> Comparing with `master`, I don't see what this is supposed to do.

Initially, I tested this pr in the Cinnamon DE only.
Here are the steps to reproduce the issue on macOS BigSur 11.2.1 (20D74) on the master branch:
1. run `bitcoin-qt -disablewallet=0`, go to the menu -> `Window` -> `Information`, note the size/location of the "Node window", quit
2. run `bitcoin-qt -disablewallet=1`, resize/move the window and note its size/location, quit
3. run `bitcoin-qt -disablewallet=0`, go to the menu -> `Window` -> `Information`, note the size/location of the "Node window" is inherited from the the step 2

Talkless:
> I think this `#ifdef` madness could go away by moving `WalletModel::isWalletEnabled()` to another place where it always compiled in.

`isWalletEnabled` could be `constexpr` function (that implies inline) that returns `false` if `ENABLE_WALLET=0`, and out-of-line function checking program arguments at runtime otherwise.

Talkless (approved):
tACK 01d9586ae85e49efaa00f11c1f26c24c9b82b278, tested on Debian Sid with Qt 5.15.2. I've managed to reproduce issue using https://github.com/bitcoin-core/gui/pull/194#issuecomment-782822663 instructions, and I see that this PR does detach main window and information window sizes. Built with `--enable-wallet` and `--disable-wallet`.

jarolrod (commented):
ACK 01d9586ae85e49efaa00f11c1f26c24c9b82b278, tested on macOS 11.2 Qt 5.15.2

Reproduced the steps thanks to @hebasto's instructions. Confirming the behavior described on `master` and this `PR` fixes that

promag (commented):
Code review ACK 01d9586ae85e49efaa00f11c1f26c24c9b82b278.

This is a straightforward solution to handle both cases (with and without wallet support).

A different path would be to move load/save window state out of `RPCConsole` and also use `QSettings::beginGroup` to distinguish each case.

PR #201, opened 2021-01-28
gui: ensure inbound block relay peers have relevant services

jonatack:
Responding to https://github.com/bitcoin-core/gui/pull/180#discussion_r565863674, this ensures that inbound block relay peers have relevant services and returns all others as inbound.

### DIFF
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -766,10 +766,10 @@ QString NetworkToQString(Network net)
     assert(false);
 }
 
-QString ConnectionTypeToQString(ConnectionType conn_type, bool relay_txes)
+QString ConnectionTypeToQString(ConnectionType conn_type, bool relay_txes, bool relevant_services)
 {
     switch (conn_type) {
-    case ConnectionType::INBOUND: return relay_txes ? QObject::tr("Inbound Full Relay") : QObject::tr("Inbound Block Relay");
+    case ConnectionType::INBOUND: return !relay_txes && relevant_services ? QObject::tr("Inbound Block Relay") : QObject::tr("Inbound");
     case ConnectionType::OUTBOUND_FULL_RELAY: return QObject::tr("Outbound Full Relay");
     case ConnectionType::BLOCK_RELAY: return QObject::tr("Outbound Block Relay");
     case ConnectionType::MANUAL: return QObject::tr("Outbound Manual");
--- a/src/qt/guiutil.h
+++ b/src/qt/guiutil.h
@@ -9,6 +9,7 @@
 #include <fs.h>
 #include <net.h>
 #include <netaddress.h>
+#include <protocol.h>
 
 #include <QEvent>
 #include <QHeaderView>
@@ -233,7 +234,7 @@ namespace GUIUtil
     QString NetworkToQString(Network net);
 
     /** Convert enum ConnectionType to QString */
-    QString ConnectionTypeToQString(ConnectionType conn_type, bool relay_txes);
+    QString ConnectionTypeToQString(ConnectionType conn_type, bool relay_txes, bool relevant_services);
 
     /** Convert seconds into a QString with days, hours, mins, secs */
     QString formatDurationStr(int secs);
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -462,7 +462,7 @@ RPCConsole::RPCConsole(interfaces::Node& node, const PlatformStyle *_platformSty
 
     constexpr QChar nonbreaking_hyphen(8209);
     const std::vector<QString> CONNECTION_TYPE_DOC{
-        tr("Inbound Full/Block Relay: initiated by peer"),
+        tr("Inbound/Inbound Block Relay: initiated by peer"),
         tr("Outbound Full Relay: default"),
         tr("Outbound Block Relay: does not relay transactions or addresses"),
         tr("Outbound Manual: added using RPC %1 or %2/%3 configuration options")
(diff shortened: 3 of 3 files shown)


src/qt/guiutil.cpp
```diff
@@ -766,10 +766,10 @@ QString NetworkToQString(Network net)
     assert(false);
 }
 
-QString ConnectionTypeToQString(ConnectionType conn_type, bool relay_txes)
+QString ConnectionTypeToQString(ConnectionType conn_type, bool relay_txes, bool relevant_services)
 {
     switch (conn_type) {
-    case ConnectionType::INBOUND: return relay_txes ? QObject::tr("Inbound Full Relay") : QObject::tr("Inbound Block Relay");
+    case ConnectionType::INBOUND: return !relay_txes && relevant_services ? QObject::tr("Inbound Block Relay") : QObject::tr("Inbound Full Relay");
```
MarcoFalke, line 772: spv nodes aren't full relay peers either if they have tx relay disabled

jonatack: thanks for having a look...simplified to "Inbound" and "Inbound Block Relay", open to other suggestions you have

src/qt/guiutil.cpp
```diff
@@ -766,10 +766,10 @@ QString NetworkToQString(Network net)
     assert(false);
 }
 
-QString ConnectionTypeToQString(ConnectionType conn_type, bool relay_txes)
+QString ConnectionTypeToQString(ConnectionType conn_type, bool relay_txes, bool relevant_services)
 {
     switch (conn_type) {
-    case ConnectionType::INBOUND: return relay_txes ? QObject::tr("Inbound Full Relay") : QObject::tr("Inbound Block Relay");
+    case ConnectionType::INBOUND: return !relay_txes && relevant_services ? QObject::tr("Inbound Block Relay") : QObject::tr("Inbound");
```
MarcoFalke, line 772: This heuristic seems to be working fine. At least I can't find a counter example right now that can be observed on today's network. Though, it still seems fragile to assume that the type of the inbound connection can be deduced based on fuzzy logic.

If it was fine to assume so, then BIP https://github.com/bitcoin/bips/pull/1052 wouldn't be needed.

jonatack: If helpful, the logic is from https://github.com/bitcoin/bitcoin/blob/16b784d953365bb2d7ae65acd2b20a79ef8ba7b6/src/net.cpp#L872 in our inbound eviction logic.

jonatack: WDYT, display "Inbound" for all inbounds or add this info? (-netinfo currently shows it)

MarcoFalke: I think plain "Inbound" makes most sense because it is neither possible to enumerate all inbound connection types, nor is it possible to determine them (if they were enumerable).

jonatack: Yes, it seems we will need to deduce some types of connections from other properties, as currently done to obtain the inbound block relay peers for eviction protection. Take for instance this https://github.com/bitcoin/bitcoin/pull/20726#issuecomment-754802089:

>> #19670 introduced a workaround to deduce inbound block-relay-only peers and protect them from eviction. After this PR, do you think we could revisit to make it more direct?

> I think we would wait until this has been deployed for a while and then (potentially) change the eviction logic. Two things occur to me on this point:
> - Older software will continue to use the old way of (not) communicating their intentions, via the fRelayTxes flag. So to accommodate older software we may not want to switch to keying off this p2p message for a little while at least.
> - It seems most of the feedback has been to shy away from encoding overly broad semantics into a single p2p message; while I think that is fine it also means that from our peer's point of view, they will have to infer the connection "type" from the properties of the peer, rather than from the peer declaring its full intentions in a single message. I think for now we can treat "will never relay txes" as essentially synonymous with these block-relay connections, but perhaps in the future there will be some distinction, and all a peer can do is rely on some set of flags that it might use to preference some peers over others -- which might put us exactly back to where we are today, with us just keying off of whether fRelayTxes is false, anyway.

So while I'd like to see this type of peers easily, yes for now it may be more prudent to just use plain "Inbound" and display other properties useful for inferring the connection status, like fRelayTxes, separately.

MarcoFalke: Even with BIP https://github.com/bitcoin/bips/pull/1052 you couldn't enumerate all possible inbound connection types. This is simply impossible.

Using properties of the connection like service bits or relay flags (and maybe later a `disabletx` message) or other statistics like last sent tx/block to aid eviction can be done without exactly enumerating the connection types.

jonatack: I don't disagree, though as a user I'd like to see this information...either directly as proposed here, or indirectly by manually looking at the `inbound` + `relaytxes` + `services`/`servicesnames` getpeerinfo boolean fields. Anyway, closing for now, replaced by #203.

MarcoFalke: Services and inbound status should be shown in the gui. I don't recall whether version.fRelay is shown

jonatack: Yes, inbound and services are. Proposing now to add frelaytxes and bip152 high bandwidth.

hebasto:
Sorry for being late...

It's unfortunate that it is closed for the following reasons:
- no need to enumerate all possible inbound connection types, only block-relay-only are important to highlight as they form a subnetwork with a node graph that is untraceable for spies; other inbound connection types differ between each other only by requested resources;
- using heuristic to detect "potential block-relay only peers" seems fine to me

Concept ACK (and ready to test if it'll be reopened).

Nevertheless, will review #203.

jonatack:
@hebasto Thanks--I'm happy to re-open if desired.

src/qt/guiutil.cpp
hebasto, line 772: pico-nit:
> If helpful, the logic is from https://github.com/bitcoin/bitcoin/blob/16b784d953365bb2d7ae65acd2b20a79ef8ba7b6/src/net.cpp#L872 in our inbound eviction protection logic...

Maybe add a relevant comment here?

hebasto (approved):
ACK 89c20ee42014cebdff8604ae09820ef5c2a5d045, tested on Linux Mint 20.1 (Qt 5.12.8).

MarcoFalke:
> only block-relay-only are important to highlight as they form a subnetwork with a node graph that is untraceable for spies; [...] using heuristic to detect "potential block-relay only peers" seems fine to me

My point is that what is displayed in the gui is not the "untraceable subnetwork". Whether a node asked you to send txs has nothing to do of *their* behaviour to send you txs (or other peers they have sent a version.fRalay=false), so I think incorrectly classifying them as "block-relay-only" is confusing at best. It would be good to describe your use case, but if your use case is to determine the "untraceable subnetwork", this heuristic is neither accurate nor precise. https://en.wikipedia.org/wiki/Accuracy_and_precision

Even when ignoring all different p2p clients on the network and only considering vanilla Bitcoin Core nodes (at the commit of this pull) wont give precise results for `-blocksonly` peers:

Screenshot from 2021-02-01 10-11-44
Screenshot from 2021-02-01 10-12-52

jonatack:
I agree this can be deferred for now and have aligned `-netinfo` (in merged pull https://github.com/bitcoin/bitcoin/pull/20764) in the same direction as #203.

jonatack:
Thanks for the explanations @MarcoFalke.

jonatack:
Thanks for the discussions @MarcoFalke and @hebasto. I agree this can be deferred for now and have aligned `-netinfo` (in merged pull https://github.com/bitcoin/bitcoin/pull/20764) in the same direction as #203.

PR #202, opened 2021-01-29
peers-tab: bug fix right panel toggle

RandyMcMillan:
Initial Presentation:

Screen Shot 2021-01-28 at 8 36 15 PM

When node row selected - panel is presented:

Screen Shot 2021-01-28 at 8 36 22 PM

When network disabled - right panel is hidden:

Screen Shot 2021-01-28 at 8 36 32 PM

### DIFF
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -988,7 +988,7 @@
            </item>
           </layout>
          </widget>
-         <widget class="QWidget" name="widget_2" native="true">
+         <widget class="QWidget" name="peersTabRightPanel" native="true">
           <property name="sizePolicy">
            <sizepolicy hsizetype="Minimum" vsizetype="Preferred">
             <horstretch>0</horstretch>
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -1097,7 +1097,7 @@ void RPCConsole::updateDetailWidget()
 {
     const QList<QModelIndex> selected_peers = GUIUtil::getEntryData(ui->peerWidget, PeerTableModel::NetNodeId);
     if (!clientModel || !clientModel->getPeerTableModel() || selected_peers.size() != 1) {
-        ui->detailWidget->hide();
+        ui->peersTabRightPanel->hide();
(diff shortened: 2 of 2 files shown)


luke-jr:
What is this fixing?

RandyMcMillan:
Hiding the detailWidget is incorrect - hiding the right panel of the splitter view is the correct usage of the splitter view. This change uses the splitter view correctly.

Running the code will make what the change addresses readily apparent.

jarolrod (commented):
ACK 8353e8cecc3e34a8699b53849df0adab622c1c14

This is a welcome change as it makes more efficient use of the peers tab window. On `master` the peer detail widget is always shown even if no peer has been selected. This `pr` introduces a toggle feature where the peer detail widget will only pop up when a peer has been selected. 

Checked and confirmed that when multiple peers are selected the peer widget will still hide (introduced in #13)

jonatack (commented):
ACK 8353e8cecc3e34a8699b53849df0adab622c1c14 tested rebased to current master. Behavior is initially a bit surprising but this would allow more columns to be added to the peers tab window. Verified that selecting more than one peer, clicking on a column header, or running`disconnectnode "" <currently-selected-peer-id>` in the console (or on the CLI with the `-server` startup option) returns the window to its full size. If this is merged, it might be nice to have an obvious way to close the details area like a clickable "close this" icon at the upper left of the area.

RandyMcMillan:
@jonatack - thank you! yes - this is the goal! Part of a larger common initiative to make the peers tab contain more info and be more useful - a better use of GUI real estate. :) This PR is sparse - as to not break other work being done. :)

Talkless (approved):
tACK 8353e8cecc3e34a8699b53849df0adab622c1c14, tested on Debian Sid. Made `bitcoind` connect to `bitcoin-qt` with the PR changes, and after I quit the `bitcoind` instance, right panel do disappear, compared to the previous commit where it didn't.

hebasto:
A UX note: to clear selection use `Ctrl` + click (actually, it inverses selection).

hebasto:
When no peer is selected, and one calls a context menu, it causes the peer selection with a layout change.

I found such a behavior distracting.

PR #204, opened 2021-01-31
Drop buggy TableViewLastColumnResizingFixer class

hebasto:
In Qt 5 the last column resizing with dragging its left edge works out-of-the-box.

The current `TableViewLastColumnResizingFixer` implementation could put the last column content out of the view port and confuse a user:
Screenshot from 2021-01-31 18-04-32

Historical context:
- https://github.com/bitcoin/bitcoin/pull/2862
- https://github.com/bitcoin/bitcoin/pull/3626
- https://github.com/bitcoin/bitcoin/pull/3738
- https://github.com/bitcoin/bitcoin/pull/3920

#205 is a nice addition.

PR #205, opened 2021-01-31
Save/restore TransactionView and recentRequestsView tables column sizes

hebasto:
Save/restore TransactionView and recentRequestsView tables column sizes.
Sorting order is not saved/restored intentionally.

Based on #204 (the first commit).


jonatack:
Concept ACK if the idea is to save the column sizes if adjusted by the user and restore them on next use (is that the goal of this pull?), ATM for me on Debian the Transactions window column widths are indeed not persisted after shutdown. (Why not save the sorting order too?)

hebasto:
> ... if the idea is to save the column sizes if adjusted by the user and restore them on next use (is that the goal of this pull?)...

Correct.

> Why not save the sorting order too?

After a while users could forget/unnotice another sort order, and that could cause their panic (when they are not looking at the recent made transaction).

jonatack:
Tested, it seems to work well. Sort is not preserved but column widths are, as described.

jonasschnelli:
Concept ACK (I guess this is also resettable by `-resetguisettings`?)

hebasto:
@jonasschnelli 
> I guess this is also resettable by `-resetguisettings`?

Correct. `-resetguisettings` affects all of the settings those use a `QSettings` class instance.

jarolrod (commented):
ACK 964885d04801c6ab77ce4705cff01c9d83bc3ed8, tested on macOS 11.1 Qt 5.15.2

PR works well and does what it says it will do. Confirmed that the column width's of transactions and recent requests are saved and then restored upon node restart. Code looks good and it removes some complexity (yay, less lines of code).

Below are screenshots comparing the behavior between `master` and this `pr`

**Master**: Start point
MASTER-Start

**Master**: Resize
MASTER-resize

**Master**: Restart, column width does not persist
MASTER-does-not-persist

**PR**: Start
PR-start

**PR:** Resize
PR-resize

**PR**: Restart, column width persists
PR-persists

src/qt/transactionview.cpp
```diff
@@ -126,27 +127,40 @@ TransactionView::TransactionView(const PlatformStyle *platformStyle, QWidget *pa
...
+    transactionView->setContextMenuPolicy(Qt::CustomContextMenu);
+    transactionView->installEventFilter(this);
+    transactionView->setAlternatingRowColors(true);
+    transactionView->setSelectionBehavior(QAbstractItemView::SelectRows);
+    transactionView->setSelectionMode(QAbstractItemView::ExtendedSelection);
+    transactionView->setSortingEnabled(true);
+    transactionView->verticalHeader()->hide();
+
+    QSettings settings;
+    if (!transactionView->horizontalHeader()->restoreState(settings.value("TransactionViewHeaderState").toByteArray())) {
```
Talkless, line 154: tiny nit:
`TransactionViewHeaderState` string literal is used twice in same .cpp file. Maybe we can then consider this as "magic string" (as magic constant) and avoid duplicating, using some local .cpp-file constant?

Talkless (approved):
tACK 964885d04801c6ab77ce4705cff01c9d83bc3ed8, tested on Debian Sid, saving/restoring and resetting (with `-resetguisettings`) works as expected.

src/qt/transactionview.cpp
hebasto: This is the current way to work with setting names in the codebase.
Your suggestion is nice to apply to _all_ of the setting names in a follow up pr.

Talkless: > Your suggestion is nice to apply to all of the setting names in a follow up pr.

True. Maybe even `GUISettings` class to wrap calls to `QSettings` with methods like `getTransactionViewHeaderState()`.

Geremia:
This fixed the column widths issue #21306; but now, after opening bitcoin-qt, the transactions aren't chronologically sorted until I click the date column header twice.

hebasto:
@Geremia 
> but now, after opening bitcoin-qt, the transactions aren't chronologically sorted until I click the date column header twice.

Thank you for reporting. Fixed in #229.

> Also, this doesn't fix "Requested payments history" and "Peers" column widths issues.

Are you sure about the "Requested payments history" tab?

The "Peers" tab is WIP :)

Geremia:
@hebasto 
>Are you sure about the "Requested payments history" tab?

Yes, widths aren't remembered.

hebasto:
While working on #256, two flaws were noted in this PR.

1. In multi-wallet environment settings of the transaction table of the last closed wallet are saved only.
2. `QTableView::setColumnWidth` does not work without model set (but it is not documented by Qt).

:disappointed:

PR #206, opened 2021-02-01
Display fRelayTxes and bip152_highbandwidth_{to, from} in peer details

jonatack:
This pull adds two fields to the peer details, "Wants Tx Relay" (fRelayTxes) and "High Bandwidth" (bip152_highbandwidth to/from). See the added tooltips for more info.

### DIFF
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -1224,14 +1224,17 @@
                 </widget>
                </item>
                <item row="7" column="0">
-                <widget class="QLabel" name="label_29">
+                <widget class="QLabel" name="peerHighBandwidthLabel">
+                 <property name="toolTip">
+                  <string>High bandwidth BIP152 compact block relay: %1</string>
+                 </property>
                  <property name="text">
-                  <string>Starting Block</string>
+                  <string>High Bandwidth</string>
                  </property>
                 </widget>
                </item>
                <item row="7" column="1">
-                <widget class="QLabel" name="peerHeight">
+                <widget class="QLabel" name="peerHighBandwidth">
                  <property name="cursor">
                   <cursorShape>IBeamCursor</cursorShape>
                  </property>
@@ -1247,14 +1250,14 @@
                 </widget>
                </item>
                <item row="8" column="0">
-                <widget class="QLabel" name="label_27">
+                <widget class="QLabel" name="label_29">
                  <property name="text">
-                  <string>Synced Headers</string>
+                  <string>Starting Block</string>
                  </property>
                 </widget>
                </item>
                <item row="8" column="1">
-                <widget class="QLabel" name="peerSyncHeight">
+                <widget class="QLabel" name="peerHeight">
                  <property name="cursor">
                   <cursorShape>IBeamCursor</cursorShape>
                  </property>
@@ -1270,14 +1273,14 @@
                 </widget>
                </item>
                <item row="9" column="0">
-                <widget class="QLabel" name="label_25">
+                <widget class="QLabel" name="label_27">
                  <property name="text">
-                  <string>Synced Blocks</string>
+                  <string>Synced Headers</string>
                  </property>
                 </widget>
                </item>
                <item row="9" column="1">
-                <widget class="QLabel" name="peerCommonHeight">
+                <widget class="QLabel" name="peerSyncHeight">
                  <property name="cursor">
                   <cursorShape>IBeamCursor</cursorShape>
                  </property>
@@ -1293,14 +1296,14 @@
                 </widget>
                </item>
                <item row="10" column="0">
-                <widget class="QLabel" name="label_22">
+                <widget class="QLabel" name="label_25">
                  <property name="text">
-                  <string>Connection Time</string>
+                  <string>Synced Blocks</string>
                  </property>
                 </widget>
                </item>
                <item row="10" column="1">
-                <widget class="QLabel" name="peerConnTime">
+                <widget class="QLabel" name="peerCommonHeight">
                  <property name="cursor">
                   <cursorShape>IBeamCursor</cursorShape>
                  </property>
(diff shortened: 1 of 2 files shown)


jonatack:
I think this is ready for review :rocket:  Will add last block and last transaction fields in a separate pull to keep this change small.

jarolrod (commented):
Concept ACK, just a few thoughts

The usage of text symbols to indicate something makes sense in a CLI environment, as used in the netinfo dashboard:
netinfo-cli

In adding these new fields, I don't think we should continue to use the same text symbols used in a CLI environment in a GUI environment.

**Suggestions:**

`Relay Transactions`
- I don't think it's great UX to show a checkmark when the peer is relaying transactions and nothing when it isn't. It would be better to have a simple `TRUE` or `FALSE` for each case.
- The tooltip should also be modified to have bullet points just like the high-bandwidth field's tooltip has

`High Bandwidth`
- The `.` and `*` symbols are small. I think a `TO` and `FROM` key-word would be a better fit, given the context of the GUI environment, and provide a better UX experience.

`When Not Applicable`
- Currently, if any of these fields are not applicable they will still show up and show up empty. Perhaps we can fill in the space with `N/A` or not show the field at all. Below is a screenshot showing this behavior.
empty-fields

src/qt/forms/debugwindow.ui
```diff
@@ -1198,13 +1198,65 @@
                 </widget>
                </item>
                <item row="6" column="0">
+                <widget class="QLabel" name="peerRelayTxesLabel">
+                 <property name="toolTip">
+                   <string>Whether the peer relays transactions: "✔" if yes, empty if no.</string>
```
MarcoFalke, line 1203: fRelay doesn't say whether the peer relays transactions. fRelay says whether the peer asked *us* to relay transactions or not.

MarcoFalke (commented):
Concept ACK

src/qt/forms/debugwindow.ui
jonatack: Thanks -- fixed

jonatack:
Thanks @jarolrod and @MarcoFalke. Updated to "Wants Tx Relay" (Yes/No) and "High Bandwidth" (To/From/No). Updated the tooltips accordingly.

jarolrod (commented):
ACK 142807af8b82e2372a03df893c50df4f4a96aca4

Screen Shot 2021-02-18 at 7 57 38 AM

The `Yes/No` for `Relay Transactions` and the `To/From/No` for `High Bandwidth` is certainly an improvement

MarcoFalke:
review ACK 142807af8b82e2372a03df893c50df4f4a96aca4

src/qt/rpcconsole.cpp
```diff
@@ -1109,6 +1114,12 @@ void RPCConsole::updateDetailWidget()
         peerAddrDetails += "<br />" + tr("via %1").arg(QString::fromStdString(stats->nodeStats.addrLocal));
     ui->peerHeading->setText(peerAddrDetails);
     ui->peerServices->setText(GUIUtil::formatServicesStr(stats->nodeStats.nServices));
+    ui->peerRelayTxes->setText(stats->nodeStats.fRelayTxes ? "Yes" : "No");
```
Talkless, line 1117: Yes, No, To, From not translatable?

I've launched bitcoin-qt in my language and I see some values translated (like `N/A` is translated to appropriate alternative), I assume these should be translated too.

src/qt/rpcconsole.cpp
```diff
@@ -1109,6 +1114,12 @@ void RPCConsole::updateDetailWidget()
...
+    QString bip152_hb_settings;
+    if (stats->nodeStats.m_bip152_highbandwidth_to) bip152_hb_settings += "To";
+    if (stats->nodeStats.m_bip152_highbandwidth_from) bip152_hb_settings += (bip152_hb_settings == "" ? "From" : "/From");
```
Talkless, line 1120: nit: Consider using `.isEmpty()` instead of `== ""`, some times it might be costly:
https://wiki.qt.io/Using_QString_Effectively

Talkless (commented):
Concept ACK.

src/qt/rpcconsole.cpp
jarolrod: Those are new strings being introduced in the PR. They will need translations which will be done when they are placed on Transifex.

Talkless: I mean, these strings are not wrapped in `tr()`.

src/qt/rpcconsole.cpp
MarcoFalke: Slightly prefer isEmpty as well, but this shouldn't have any impact on performance and should be only a stylistic matter

jonatack:
Good points @Talkless, will retouch those in the next patch adding last block and last transaction fields.

src/qt/rpcconsole.cpp
Talkless: @MarcoFalke for std::string, a template class that has probably all code inlined - maybe (though `== ""` for `std::string` is also generating more code - see https://godbolt.org/z/h8r1Wq).

But QString invokes out-of-line`QString::compare_helper` with size set to `-1`, so it will, in the runtime, call `strlen` to determine lenght.. while `isEmpty()` is just inline `{ return d->size == 0; }`. 

Using `== ""` to check empty string bloats code for a else trivial check.

I agree that we don't need any (premature-)optimizations here (not performance critical code at all), but _not_ using `isEmpty()` (or std::string::empty() for that matter) is just a needless _pessimization_ and (relative) code bloat.

src/qt/rpcconsole.cpp
jonatack: Thanks @Talkless, done in #221.

src/qt/rpcconsole.cpp
jonatack: Agree, thanks @Talkless, done in #221.

PR #207, opened 2021-02-01
scripted-diff: Use Courier New as tabular figures font in Overview page

hebasto:
The https://github.com/bitcoin/bitcoin/pull/16432/commits/8d75115844baefe5cad4d82ec8dce001dd16eb9c (https://github.com/bitcoin/bitcoin/pull/16432) introduced a monospaced font for tabulated figures.

But the Qt's font matching algorithm does not guarantee that:
- the monospaced font will be chosen at all (see https://github.com/bitcoin/bitcoin/pull/16432#issuecomment-514486077, the actually used font on macOS 11.1 is `.AppleSystemUIFont` which is not monospaced, i.e., `QFontInfo::fixedPitch()` returns `false`)
- the chosen monospaced font has expected style and weight (see https://github.com/bitcoin-core/gui/pull/87#issuecomment-768920736 _"I think the correct monospace font should be bigger and eventually bold"_)

This change suggests to use Courier New explicitly which is available on Windows and macOS by default.
On Linux systems the Qt substitutes the missed Courier New font (in case it is not installed) quite well. For example, the Liberation Mono is used on my system.

This is an alternative to #87. But [personally](https://github.com/bitcoin-core/gui/pull/79#issuecomment-770792758) I still lean to embedding font (#79).

### DIFF
--- a/src/qt/forms/overviewpage.ui
+++ b/src/qt/forms/overviewpage.ui
@@ -118,7 +118,7 @@
              <widget class="QLabel" name="labelWatchPending">
               <property name="font">
                <font>
-                <family>Monospace</family>
+                <family>Courier New</family>
                 <weight>75</weight>
                 <bold>true</bold>
                </font>
@@ -144,7 +144,7 @@
              <widget class="QLabel" name="labelUnconfirmed">
               <property name="font">
                <font>
-                <family>Monospace</family>
+                <family>Courier New</family>
                 <weight>75</weight>
                 <bold>true</bold>
                </font>
@@ -170,7 +170,7 @@
              <widget class="QLabel" name="labelWatchImmature">
               <property name="font">
                <font>
-                <family>Monospace</family>
+                <family>Courier New</family>
                 <weight>75</weight>
                 <bold>true</bold>
                </font>
@@ -229,7 +229,7 @@
              <widget class="QLabel" name="labelImmature">
               <property name="font">
                <font>
-                <family>Monospace</family>
+                <family>Courier New</family>
                 <weight>75</weight>
                 <bold>true</bold>
                </font>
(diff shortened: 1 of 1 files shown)


luke-jr:
NACK forcing a specific font. We have `GUIUtil::fixedPitchFont` to workaround Qt limitations...

Furthermore, numbers shouldn't need a monospace font at all.
Replacing the spaces with U+2007 and absent decimal point with U+2008 seems like the correct approach there.

luke-jr:
>Replacing the spaces with U+2007 and absent decimal point with U+2008 seems like the correct approach there.

Problem with this approach is that we're masking with '#' which isn't the same width as digits, and there doesn't seem to be a good substitute. Best I could come up with was https://dpaste.com/2H7QFU9BZ

laanwj:
Wouldn't this just have to be chanaged all over again when we decide to embed the fonts? (#79)

hebasto:
> Wouldn't this just have to be chanaged all over again when we decide to embed the fonts? (#79)

I really hope we'll move with #79 :)

jarolrod (commented):
Concept 0 on this

Embedding the font as proposed in #79 is the way to go. In embedding the font we can effectively eliminate the current issues surrounding the representation of monospace characters. This change will not directly address the issue where the font is not available, we assume that linux will choose an appropriate monospace font.

 My opinion is that we should focus on #79 instead of devoting review time to this

luke-jr:
@jarolrod #79 is NACK. And even if we embed a font for backward platforms like Windows/macOS, *especially* NACK on forcing it by neglecting support of non-embedded builds.

hebasto:
@luke-jr 
> @jarolrod #79 is NACK. And even if we embed a font for backward platforms like Windows/macOS, _especially_ NACK on forcing it by neglecting support of non-embedded builds.

Hopefully your comment is addressed in https://github.com/bitcoin-core/gui/pull/79#issuecomment-782909149

hebasto:
Closing in favor of #79.

Issue #209, opened 2021-02-02
"Label" for receiving addresses, and its position in context menu is misleading


surak:
Transactions have a field "label". It makes it look like each transaction can be labeled (which is bitcoin parlay seems to be a `comment`), when it's actually a label of the receiving address. This is misleading and can be clarified. 

This is not a new issue: https://github.com/bitcoin/bitcoin/pull/5905 , but a new but was created (and closed) on bitcoin-core about that recently: https://github.com/bitcoin/bitcoin/issues/21057

Even though a `comment` field would be very welcome, this issue here is simply about the confusion created by the understanding of the term label on the GUI.

**Actual behavior**

When one sets a label, it sets a label for the receiving address. It's not clear until the moment when one clicks "edit label" and a dialog appears what that means.

**To reproduce**

Right-click a transaction, choose "edit label" - it seems a unfortunate place for such a thing, since other menu items are for that transaction and not for the receiving address.

**System information**

Bitcoin-QT 0.21.0 for Mac OS.

jonasschnelli:
Agree that this is confusing.
Would removing the "edit label" action in the transaction context menu be more an option? I guess one trying to edit the labels can still use the address books

MarcoFalke:
Indeed, as a first step it could make sense to remove the per-tx "edit label" because it is really per-address. As a second step, https://github.com/bitcoin/bitcoin/pull/5905 could be rebased.

jarolrod:
When you click `Edit Label` you are really editing the label for the address type. Looking at the code, the `Edit Label` action determines what the address type is and brings up the associated mechanism. If we are to remove the `Edit Label` action from the `transactionview`, we must ensure there's another way to edit the label

For a `Sending` address, you can currently edit the label in the `Address Book` of the send tab.

For a `Receiving Address`, it seems the only place to edit the label would be in the `transactionview` of the `Transactions` tab as the received address view lacks an `Edit Label` action in its context menu. An `Edit Label` action should be added to the `Receive` Tab.

jarolrod:
Documenting that this issue is related to https://github.com/bitcoin/bitcoin/issues/1168

MarcoFalke:
Fixed in #223

PR #211, opened 2021-02-18
qt: Remove Transactionview Edit Label Action

jarolrod:
This PR removes the `Edit Label` action from the `transactionview` context menu. Since the `Edit Label` action will no longer be utilized in the `transactionview`, the `Edit Label` function logic is also removed.

| Master        |        PR        |
| ----------- | ----------- |
|Screen Shot 2021-02-17 at 8 34 34 PM|Screen Shot 2021-02-17 at 8 35 10 PM|

Among the context menu actions for each transaction in the `transactionview` is the `Edit Label` action.
While all other actions apply directly to the selected transaction, the `Edit Label` action applies to the selected transaction's address. As documented in issue #209 and [#1168](https://github.com/bitcoin/bitcoin/issues/1168) , this is an "unfortunate" placement for such an action. The current placement creates a confusing UX scenario where the outcome of the action is ambiguous. 

**Example of Ambiguous Behavior:**
The context menu gives the wrong impression that the `Edit Label` action will edit a `Label` for the specific transaction that has been right-clicked on. This impression can be because all other actions in this menu will relate to the specific transaction and the misconception between `Comment` and `Label`.
editlabel-start

Let's say I wanted to give the transaction selected in the screenshot above a comment of "2-17[17:43]". Given all the context clues, it will be reasonable to assume that the `Edit Label` function will give a label to this transaction. Instead, it edits the `Label` for the address behind this transaction. Thus, changing the `Label` for all transactions associated with this address.
editlabel-end

**Maintaining `Edit Label` Functionality:**
The action of Editing a Label should instead be reserved for the respective address tables of the `Send` and `Receive` tabs. As documented in this [comment](https://github.com/bitcoin-core/gui/issues/209#issuecomment-780922101), `Edit Label` is currently implemented in the `Send` tab and is missing in the `Receive` tab. A follow-up PR can add the `Edit Label` functionality to the `Receive` tab.

### DIFF
--- a/src/qt/transactionview.cpp
+++ b/src/qt/transactionview.cpp
@@ -158,7 +158,6 @@ TransactionView::TransactionView(const PlatformStyle *platformStyle, QWidget *pa
     QAction *copyTxIDAction = new QAction(tr("Copy transaction ID"), this);
     QAction *copyTxHexAction = new QAction(tr("Copy raw transaction"), this);
     QAction *copyTxPlainText = new QAction(tr("Copy full transaction details"), this);
-    QAction *editLabelAction = new QAction(tr("Edit label"), this);
     QAction *showDetailsAction = new QAction(tr("Show transaction details"), this);
 
     contextMenu = new QMenu(this);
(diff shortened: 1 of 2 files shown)


MarcoFalke:
Nice. Concept ACK

Talkless (approved):
tACK 8f9644890a167a093d95ecef1f12a20dce1bc581, tested on Debian Sid.

MarcoFalke:
review ACK 8f9644890a167a093d95ecef1f12a20dce1bc581

luke-jr:
Uh, Concept NACK! I use this functionality regularly.

jarolrod:
@luke-jr you can edit a sending address in the send tab Address Book

luke-jr:
Shouldn't need to go to all that trouble just to edit a label...

laanwj:
I disagree with this change, I use the functionality a lot, it was the primary way I used to organize and edit transactions labels and am sad to see this go.

Using the address book should not be encouraged at all! A while ago it was even proposed to remove it. There's rarely need to scroll through all historical addresses used and unused.

> While all other actions apply directly to the selected transaction, the Edit Label action applies to the selected transaction's address.

**In practice** when bitcoin is used in the commonly advised way, generate a new address for each transaction, those are equivalent though.

PR #213, opened 2021-02-19
qt: Add Copy Address Action to Payment Requests

jarolrod:
Currently, the only way to copy the address of a payment request is to double-click on the payment request and then click on the copy address button. This PR adds a convenient context menu action to copy the address of a payment request.

| Master        | PR                 |
| ----------- | ------------ |
|Screen Shot 2021-02-18 at 8 33 08 PM |Screen Shot 2021-02-18 at 8 33 50 PM|

### DIFF
--- a/src/qt/receivecoinsdialog.cpp
+++ b/src/qt/receivecoinsdialog.cpp
@@ -44,6 +44,7 @@ ReceiveCoinsDialog::ReceiveCoinsDialog(const PlatformStyle *_platformStyle, QWid
 
     // context menu actions
     QAction *copyURIAction = new QAction(tr("Copy URI"), this);
+    QAction* copyAddressAction = new QAction(tr("Copy address"), this);
     QAction *copyLabelAction = new QAction(tr("Copy label"), this);
     QAction *copyMessageAction = new QAction(tr("Copy message"), this);
     QAction *copyAmountAction = new QAction(tr("Copy amount"), this);
@@ -51,6 +52,7 @@ ReceiveCoinsDialog::ReceiveCoinsDialog(const PlatformStyle *_platformStyle, QWid
     // context menu
     contextMenu = new QMenu(this);
     contextMenu->addAction(copyURIAction);
+    contextMenu->addAction(copyAddressAction);
     contextMenu->addAction(copyLabelAction);
     contextMenu->addAction(copyMessageAction);
     contextMenu->addAction(copyAmountAction);
@@ -58,6 +60,7 @@ ReceiveCoinsDialog::ReceiveCoinsDialog(const PlatformStyle *_platformStyle, QWid
     // context menu signals
     connect(ui->recentRequestsView, &QWidget::customContextMenuRequested, this, &ReceiveCoinsDialog::showMenu);
     connect(copyURIAction, &QAction::triggered, this, &ReceiveCoinsDialog::copyURI);
+    connect(copyAddressAction, &QAction::triggered, this, &ReceiveCoinsDialog::copyAddress);
     connect(copyLabelAction, &QAction::triggered, this, &ReceiveCoinsDialog::copyLabel);
     connect(copyMessageAction, &QAction::triggered, this, &ReceiveCoinsDialog::copyMessage);
     connect(copyAmountAction, &QAction::triggered, this, &ReceiveCoinsDialog::copyAmount);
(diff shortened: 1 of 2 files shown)


src/qt/receivecoinsdialog.cpp
```diff
@@ -286,6 +289,19 @@ void ReceiveCoinsDialog::copyURI()
     GUIUtil::setClipboard(uri);
 }
 
+// context menu action: copy address
+void ReceiveCoinsDialog::copyAddress()
+{
+    QModelIndex sel = selectedRow();
```
Talkless, line 295: nit: `sel` could be const

Talkless (changes requested):

src/qt/receivecoinsdialog.cpp
jarolrod: addressed in 24fcd97

jarolrod:
Updated from 405215f -> 24fcd9, addressed @Talkless suggestion

Changes:
- `sel` is now `const`

hebasto:
Concept ACK. I've noted the same inconvenience :)

hebasto:
@jarolrod

This pr, which is submitted on 2021-02-19, is based on the branch commit from 2020-11-05. While it has no ACKs, maybe rebase it to eliminate CI glitches?

src/qt/receivecoinsdialog.cpp
```diff
@@ -286,6 +289,19 @@ void ReceiveCoinsDialog::copyURI()
     GUIUtil::setClipboard(uri);
 }
 
+// context menu action: copy address
```
hebasto, line 293: I understand that this comment follows the surrounding code style.
But isn't it worth to add a Doxygen comment into the header instead of such an approach?

jarolrod:
Updated from 24fcd97 -> 8871d9a

Changes: Rebased to a more recent master branch in order to avoid CI glitches on @hebasto recommendation

src/qt/receivecoinsdialog.cpp
```diff
@@ -287,6 +290,19 @@ void ReceiveCoinsDialog::copyURI()
...
+void ReceiveCoinsDialog::copyAddress()
+{
+    const QModelIndex sel = selectedRow();
+    if (!sel.isValid()) {
+        return;
+    }
+
+    const RecentRequestsTableModel * const submodel = model->getRecentRequestsTableModel();
```
hebasto, line 301: style nit (see [clang-format-diff.py](https://github.com/bitcoin/bitcoin/tree/master/contrib/devtools#clang-format-diffpy)):
```suggestion
    const RecentRequestsTableModel* const submodel = model->getRecentRequestsTableModel();
```

---

btw, why `submodel` variable is required at all? It does not improve readability, at least for me :)
Maybe just
```
    const QString address = model->getRecentRequestsTableModel()->entry(sel.row()).recipient.address;
```

This line could be [shorter](https://github.com/bitcoin/bitcoin/pull/20986) then 100 characters, if s/`sel`/`r`

Feel free to ignore the second part of this comment.

src/qt/receivecoinsdialog.cpp
```diff
@@ -44,20 +44,23 @@ ReceiveCoinsDialog::ReceiveCoinsDialog(const PlatformStyle *_platformStyle, QWid
 
     // context menu actions
     QAction *copyURIAction = new QAction(tr("Copy URI"), this);
+    QAction *copyAddressAction = new QAction(tr("Copy address"), this);
```
hebasto, line 47: style nit (see [clang-format-diff.py](https://github.com/bitcoin/bitcoin/tree/master/contrib/devtools#clang-format-diffpy)):
```suggestion
    QAction* copyAddressAction = new QAction(tr("Copy address"), this);
```

hebasto (approved):
ACK 8871d9a6a98d44934c634ddfaa3e4de4e45092fc

src/qt/receivecoinsdialog.cpp
jarolrod: addressed in e348d7e

src/qt/receivecoinsdialog.cpp
jarolrod: addressed the first half of your comment in e348d7e

jarolrod:
Updated from 8871d9a -> e348d7e

Changes:
- Closer adherence to [clang-format](https://github.com/bitcoin/bitcoin/blob/master/src/.clang-format)

hebasto (approved):
re-ACK e348d7ea2c0b5d2eb8039dab33b0b9a48655885f, only suggested changes since my [previous](https://github.com/bitcoin-core/gui/pull/213#pullrequestreview-595520204) review.

PR #214, opened 2021-02-19
qt: Disable requests context menu actions when appropriate

jarolrod:
The recent requests table will allow you to copy data points even if they do not exist. This PR implements checks to disable the `copy label`, `copy message`, and `copy amount` context menu actions if the respective fields are empty. This brings the recent requests table context menu behavior closer to the behavior seen in the transaction view.

On a payment request entry which does not have a value for label, message, or amount:
| ----------- | ----------- |
|Screen Shot 2021-02-19 at 1 22 28 AM| Screen Shot 2021-02-19 at 1 21 49 AM|

`copy URI` never needs to be disabled as an entry in the recent requests table must have a URI even if it doesn't have a label, message, or amount. #213 will add a `copy address` context menu action. This also does not need a check as an entry must be associated with an address.

Below are some more examples of how this PR will behave:
**Has Label, Message, and Amount**
Screen Shot 2021-02-19 at 12 05 38 AM

**Has Label and Amount, but no Message**
Screen Shot 2021-02-19 at 12 05 58 AM

### DIFF
--- a/src/qt/receivecoinsdialog.cpp
+++ b/src/qt/receivecoinsdialog.cpp
@@ -45,9 +45,9 @@ ReceiveCoinsDialog::ReceiveCoinsDialog(const PlatformStyle *_platformStyle, QWid
     // context menu actions
     QAction *copyURIAction = new QAction(tr("Copy URI"), this);
     QAction* copyAddressAction = new QAction(tr("Copy address"), this);
-    QAction *copyLabelAction = new QAction(tr("Copy label"), this);
-    QAction *copyMessageAction = new QAction(tr("Copy message"), this);
-    QAction *copyAmountAction = new QAction(tr("Copy amount"), this);
+    copyLabelAction = new QAction(tr("Copy label"), this);
+    copyMessageAction = new QAction(tr("Copy message"), this);
+    copyAmountAction = new QAction(tr("Copy amount"), this);
 
     // context menu
     contextMenu = new QMenu(this);
@@ -269,9 +269,18 @@ void ReceiveCoinsDialog::copyColumnToClipboard(int column)
 // context menu
 void ReceiveCoinsDialog::showMenu(const QPoint &point)
 {
-    if (!selectedRow().isValid()) {
+    const QModelIndex sel = selectedRow();
+    if (!sel.isValid()) {
         return;
     }
+
+    // disable context menu actions when appropriate
+    const RecentRequestsTableModel* const submodel = model->getRecentRequestsTableModel();
+    const RecentRequestEntry& req = submodel->entry(sel.row());
+    copyLabelAction->setDisabled(req.recipient.label.isEmpty());
+    copyMessageAction->setDisabled(req.recipient.message.isEmpty());
+    copyAmountAction->setDisabled(req.recipient.amount == 0);
+
     contextMenu->exec(QCursor::pos());
 }
 
--- a/src/qt/receivecoinsdialog.h
+++ b/src/qt/receivecoinsdialog.h
@@ -53,6 +53,9 @@ private:
     Ui::ReceiveCoinsDialog *ui;
     WalletModel *model;
     QMenu *contextMenu;
+    QAction* copyLabelAction;
+    QAction* copyMessageAction;
+    QAction* copyAmountAction;
     const PlatformStyle *platformStyle;
 
     QModelIndex selectedRow();


jonasschnelli:
utACK 69915b80ccf853b66345b3d612d24eddd3411a2a - I guess comparing against zero for an "empty" amount is acceptable here.

jarolrod:
> utACK [69915b8](https://github.com/bitcoin-core/gui/commit/69915b80ccf853b66345b3d612d24eddd3411a2a) - I guess comparing against zero for an "empty" amount is acceptable here.

Comparing against zero is what's used to determine `(no amount requested)` in [recentrequeststablemodel.cpp](https://github.com/bitcoin-core/gui/blob/master/src/qt/recentrequeststablemodel.cpp) Line 84

Talkless:
Concept ACK.

src/qt/receivecoinsdialog.cpp
```diff
@@ -267,9 +267,18 @@ void ReceiveCoinsDialog::copyColumnToClipboard(int column)
...
-    if (!selectedRow().isValid()) {
+    QModelIndex sel = selectedRow();
+    if (!sel.isValid()) {
         return;
     }
+
+    // disable context menu actions when appropriate
+    const RecentRequestsTableModel * const submodel = model->getRecentRequestsTableModel();
+    const RecentRequestEntry *req = &submodel->entry(sel.row());
+    copyLabelAction->setEnabled(!req->recipient.label.isEmpty());
```
Talkless, line 278: nit: negations could be avoided (simplified) by using `setDisabled` instead (same for all three similar lines).

src/qt/receivecoinsdialog.cpp
```diff
@@ -267,9 +267,18 @@ void ReceiveCoinsDialog::copyColumnToClipboard(int column)
 // context menu
 void ReceiveCoinsDialog::showMenu(const QPoint &point)
 {
-    if (!selectedRow().isValid()) {
+    QModelIndex sel = selectedRow();
```
Talkless, line 270: nit: `sel` Could be const.

src/qt/receivecoinsdialog.cpp
```diff
@@ -267,9 +267,18 @@ void ReceiveCoinsDialog::copyColumnToClipboard(int column)
...
+    if (!sel.isValid()) {
         return;
     }
+
+    // disable context menu actions when appropriate
+    const RecentRequestsTableModel * const submodel = model->getRecentRequestsTableModel();
+    const RecentRequestEntry *req = &submodel->entry(sel.row());
```
Talkless, line 277: nit: Pointer `req` is not change within function, could be  `const RecentRequestEntry * const req` (same as `submodel` above).

Talkless (changes requested):

src/qt/receivecoinsdialog.cpp
jarolrod: addressed in 336a643

src/qt/receivecoinsdialog.cpp
jarolrod: addressed in 336a643

src/qt/receivecoinsdialog.cpp
jarolrod: addressed in 336a643

jarolrod:
updated 69915b8 -> 336a643

Addressed @Talkless suggestions, thanks for the review!

Changes:
- variables that can be `const` are made `const`
- refactor context menu action disable logic from using `setEnabled` with negations to using `setDisabled` without negations

src/qt/receivecoinsdialog.cpp
```diff
@@ -267,9 +267,18 @@ void ReceiveCoinsDialog::copyColumnToClipboard(int column)
...
 {
-    if (!selectedRow().isValid()) {
+    const QModelIndex sel = selectedRow();
+    if (!sel.isValid()) {
         return;
     }
+
+    // disable context menu actions when appropriate
+    const RecentRequestsTableModel * const submodel = model->getRecentRequestsTableModel();
+    const RecentRequestEntry * const req = &submodel->entry(sel.row());
```
MarcoFalke, line 277: wouldn't a const reference make more sense?
```suggestion
    const RecentRequestEntry& req = submodel->entry(sel.row());
```

MarcoFalke (commented):
tested ACK 336a6435965dc813bdc2bf11398bce564e52c984

didn't review code

hebasto:
Concept ACK.

src/qt/receivecoinsdialog.h
```diff
@@ -54,6 +54,9 @@ public Q_SLOTS:
     GUIUtil::TableViewLastColumnResizingFixer *columnResizingFixer;
     WalletModel *model;
     QMenu *contextMenu;
+    QAction *copyLabelAction;
+    QAction *copyMessageAction;
+    QAction *copyAmountAction;
```
hebasto, line 59: style nit (see [clang-format-diff.py](https://github.com/bitcoin/bitcoin/tree/master/contrib/devtools#clang-format-diffpy)):
```suggestion
    QAction* copyLabelAction;
    QAction* copyMessageAction;
    QAction* copyAmountAction;
```

src/qt/receivecoinsdialog.cpp
```diff
@@ -267,9 +267,18 @@ void ReceiveCoinsDialog::copyColumnToClipboard(int column)
...
 void ReceiveCoinsDialog::showMenu(const QPoint &point)
 {
-    if (!selectedRow().isValid()) {
+    const QModelIndex sel = selectedRow();
+    if (!sel.isValid()) {
         return;
     }
+
+    // disable context menu actions when appropriate
+    const RecentRequestsTableModel * const submodel = model->getRecentRequestsTableModel();
```
hebasto, line 276: style nit (see [clang-format-diff.py](https://github.com/bitcoin/bitcoin/tree/master/contrib/devtools#clang-format-diffpy)):
```suggestion
    const RecentRequestsTableModel* const submodel = model->getRecentRequestsTableModel();
```

hebasto (approved):
ACK 336a6435965dc813bdc2bf11398bce564e52c984, tested on Linux Mint 20.1 (Qt 5.12.8).

src/qt/receivecoinsdialog.h
jarolrod: Addressed in 3f83f05f52be8f3bac2267d447b5037409cde17d

src/qt/receivecoinsdialog.cpp
jarolrod: addressed in 3f83f05f52be8f3bac2267d447b5037409cde17d

jarolrod:
updated from 336a643  -> 3f83f05f52be8f3bac2267d447b5037409cde17d, thanks @hebasto @MarcoFalke for the review

Changes:
- Adhere to [clang-format](https://github.com/bitcoin/bitcoin/blob/master/src/.clang-format)
- `req` is made to be a const reference as that makes more sense

src/qt/receivecoinsdialog.cpp
```diff
@@ -266,9 +266,18 @@ void ReceiveCoinsDialog::copyColumnToClipboard(int column)
...
+    if (!sel.isValid()) {
         return;
     }
+
+    // disable context menu actions when appropriate
+    const RecentRequestsTableModel* const submodel = model->getRecentRequestsTableModel();
+    const RecentRequestEntry& req = submodel->entry(sel.row());
+    copyLabelAction->setDisabled(req.recipient.label.isEmpty());
+    copyMessageAction->setDisabled(req.recipient.message.isEmpty());
+    copyAmountAction->setDisabled(req.recipient.amount == 0);
```
hebasto, line 279:
```suggestion
    const auto& r = model->getRecentRequestsTableModel()->entry(sel.row()).recipient;
    copyLabelAction->setDisabled(r.label.isEmpty());
    copyMessageAction->setDisabled(r.message.isEmpty());
    copyAmountAction->setDisabled(r.amount == 0);
```

hebasto (approved):
re-ACK 3f83f05f52be8f3bac2267d447b5037409cde17d, only suggested changes and rebased since my [previous](https://github.com/bitcoin-core/gui/pull/214#pullrequestreview-595579186) review, verified with
```
$ git range-diff master 336a6435965dc813bdc2bf11398bce564e52c984 3f83f05f52be8f3bac2267d447b5037409cde17d
```

src/qt/receivecoinsdialog.cpp
jarolrod: Since 213 was merged with the style used in this PR, I'll close this for consistency

jarolrod:
Updated from 3f83f05 -> bb3da8f

Changes: Resolve conflict with #213

hebasto (approved):
re-ACK bb3da8fe410285887f22679c6f08a1f40bcfac8f

luke-jr (approved):
utACK

PR #216, opened 2021-02-22
Support building without embedded monospace font

luke-jr:


hebasto:
Concept ACK. Maybe a bit more motivation for the second commit description?

luke-jr:
A few motivations for the 2nd commit:

* Most non-macOS/Windows systems already have a decent monospace font installed.
* Only macOS/Windows appear to have denied users the option to configure their monospace font.
* Embedding fonts is generally undesirable and non-standard on *nix platforms.

hebasto:
As this PR changes the build stuff, should it be moved into the main repo?

(The GUI repo doesn't even have the "Build system" label)

PR #219, opened 2021-02-23
qt: Prevent the main window popup menu

hebasto:
https://github.com/bitcoin/bitcoin/issues/11168 is not fixed by https://github.com/bitcoin/bitcoin/pull/11169 completely, as users are allowed to right click on the menu bar:

Screenshot from 2021-02-23 14-18-24

This PR moves the context menu prohibition from `QToolBar` instance to its parent `BitcoinGUI` instance, which is derived from `QMainWindow`.

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -88,6 +88,8 @@ BitcoinGUI::BitcoinGUI(interfaces::Node& node, const PlatformStyle *_platformSty
         move(QGuiApplication::primaryScreen()->availableGeometry().center() - frameGeometry().center());
     }
 
+    setContextMenuPolicy(Qt::PreventContextMenu);
+
 #ifdef ENABLE_WALLET
     enableWallet = WalletModel::isWalletEnabled();
 #endif // ENABLE_WALLET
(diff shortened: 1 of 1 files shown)


jarolrod (commented):
ACK ca5bd1c8e5fee29a3c2f0a3fbc879ee74f892266, tested on Debian 20.04 Qt 5.12. Confirming that I can replicate the behavior described on `master` and this `PR` fixes it.

The question is: "Is there a use case for disabling tabs?"
I would lean towards no. To me, it doesn't make sense to allow a user to toggle the visibility of the tabs.

leonardojobim (approved):
tACK https://github.com/bitcoin-core/gui/pull/219/commits/ca5bd1c8e5fee29a3c2f0a3fbc879ee74f892266

Tested on Ubuntu 20.04 Qt 5.12.8.
I also confirm that context menu is displayed on `master` branch but it does not happen on the branch of this PR.

luke-jr:
Concept ACK, but I don't understand why the toolbar inherits this prohibition and other widgets (where we want a context menu) don't...

hebasto:
> I don't understand why the toolbar inherits this prohibition and other widgets (where we want a context menu) don't...

Probably, due to the fact that `QMainWindow` has its _own_ layout which a `QMenuBar` belongs to.

PR #221, opened 2021-02-23
qt, refactor: rpcconsole translatable string fixes and improvements

jonatack:
- fixups from #206 review feedback (thanks!), see commit message for details
- hoists repeatedly used translatable strings to the `RPCConsole` class for reuse

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -474,9 +474,9 @@ RPCConsole::RPCConsole(interfaces::Node& node, const PlatformStyle *_platformSty
     const QString list{"<ul><li>" + Join(CONNECTION_TYPE_DOC, QString("</li><li>")) + "</li></ul>"};
     ui->peerConnectionTypeLabel->setToolTip(ui->peerConnectionTypeLabel->toolTip().arg(list));
     const QString hb_list{"<ul><li>\""
-        + tr("To") + "\" – " + tr("we selected the peer for high bandwidth relay") + "</li><li>\""
-        + tr("From") + "\" – " + tr("the peer selected us for high bandwidth relay") + "</li><li>\""
-        + tr("No") + "\" – " + tr("no high bandwidth relay selected") + "</li></ul>"};
+        + ts.to + "\" – " + tr("we selected the peer for high bandwidth relay") + "</li><li>\""
+        + ts.from + "\" – " + tr("the peer selected us for high bandwidth relay") + "</li><li>\""
+        + ts.no + "\" – " + tr("no high bandwidth relay selected") + "</li></ul>"};
     ui->peerHighBandwidthLabel->setToolTip(ui->peerHighBandwidthLabel->toolTip().arg(hb_list));
     ui->dataDir->setToolTip(ui->dataDir->toolTip().arg(QString(nonbreaking_hyphen) + "datadir"));
     ui->blocksDir->setToolTip(ui->blocksDir->toolTip().arg(QString(nonbreaking_hyphen) + "blocksdir"));
@@ -619,10 +619,10 @@ void RPCConsole::setClientModel(ClientModel *model, int bestblock_height, int64_
 
         // create peer table context menu actions
         QAction* disconnectAction = new QAction(tr("&Disconnect"), this);
-        QAction* banAction1h      = new QAction(tr("Ban for") + " " + tr("1 &hour"), this);
-        QAction* banAction24h     = new QAction(tr("Ban for") + " " + tr("1 &day"), this);
-        QAction* banAction7d      = new QAction(tr("Ban for") + " " + tr("1 &week"), this);
-        QAction* banAction365d    = new QAction(tr("Ban for") + " " + tr("1 &year"), this);
+        QAction* banAction1h = new QAction(ts.ban_for + " " + tr("1 &hour"), this);
+        QAction* banAction24h = new QAction(ts.ban_for + " " + tr("1 &day"), this);
+        QAction* banAction7d = new QAction(ts.ban_for + " " + tr("1 &week"), this);
+        QAction* banAction365d = new QAction(ts.ban_for + " " + tr("1 &year"), this);
 
         // create peer table context menu
         peersTableContextMenu = new QMenu(this);
@@ -1114,11 +1114,11 @@ void RPCConsole::updateDetailWidget()
         peerAddrDetails += "<br />" + tr("via %1").arg(QString::fromStdString(stats->nodeStats.addrLocal));
     ui->peerHeading->setText(peerAddrDetails);
     ui->peerServices->setText(GUIUtil::formatServicesStr(stats->nodeStats.nServices));
-    ui->peerRelayTxes->setText(stats->nodeStats.fRelayTxes ? tr("Yes") : tr("No"));
+    ui->peerRelayTxes->setText(stats->nodeStats.fRelayTxes ? ts.yes : ts.no);
     QString bip152_hb_settings;
-    if (stats->nodeStats.m_bip152_highbandwidth_to) bip152_hb_settings = tr("To");
-    if (stats->nodeStats.m_bip152_highbandwidth_from) bip152_hb_settings += (bip152_hb_settings.isEmpty() ? tr("From") : QLatin1Char('/') + tr("From"));
-    if (bip152_hb_settings.isEmpty()) bip152_hb_settings = tr("No");
+    if (stats->nodeStats.m_bip152_highbandwidth_to) bip152_hb_settings = ts.to;
+    if (stats->nodeStats.m_bip152_highbandwidth_from) bip152_hb_settings += (bip152_hb_settings.isEmpty() ? ts.from : QLatin1Char('/') + ts.from);
+    if (bip152_hb_settings.isEmpty()) bip152_hb_settings = ts.no;
     ui->peerHighBandwidth->setText(bip152_hb_settings);
     const int64_t time_now{GetSystemTimeInSeconds()};
     ui->peerConnTime->setText(GUIUtil::formatDurationStr(time_now - stats->nodeStats.nTimeConnected));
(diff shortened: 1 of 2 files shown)


src/qt/rpcconsole.cpp
```diff
@@ -474,9 +474,9 @@ RPCConsole::RPCConsole(interfaces::Node& node, const PlatformStyle *_platformSty
     const QString list{"<ul><li>" + Join(CONNECTION_TYPE_DOC, QString("</li><li>")) + "</li></ul>"};
     ui->peerConnectionTypeLabel->setToolTip(ui->peerConnectionTypeLabel->toolTip().arg(list));
     const QString hb_list{"<ul><li>\""
-        + tr("To") + "\" – " + tr("we selected the peer for high bandwidth relay") + "</li><li>\""
-        + tr("From") + "\" – " + tr("the peer selected us for high bandwidth relay") + "</li><li>\""
-        + tr("No") + "\" – " + tr("no high bandwidth relay selected") + "</li></ul>"};
+        + tr_to   + "\" – " + tr("we selected the peer for high bandwidth relay") + "</li><li>\""
+        + tr_from + "\" – " + tr("the peer selected us for high bandwidth relay") + "</li><li>\""
+        + tr_no   + "\" – " + tr("no high bandwidth relay selected")              + "</li></ul>"};
```
hebasto, line 479: Does [`clang-format-diff.py`](https://github.com/bitcoin/bitcoin/tree/master/contrib/devtools#clang-format-diffpy) agree with extra spaces?

src/qt/rpcconsole.cpp
```diff
@@ -1114,11 +1114,11 @@ void RPCConsole::updateDetailWidget()
...
     ui->peerHeading->setText(peerAddrDetails);
     ui->peerServices->setText(GUIUtil::formatServicesStr(stats->nodeStats.nServices));
-    ui->peerRelayTxes->setText(stats->nodeStats.fRelayTxes ? tr("Yes") : tr("No"));
+    ui->peerRelayTxes->setText(stats->nodeStats.fRelayTxes ? tr_yes : tr_no);
     QString bip152_hb_settings;
-    if (stats->nodeStats.m_bip152_highbandwidth_to) bip152_hb_settings = tr("To");
-    if (stats->nodeStats.m_bip152_highbandwidth_from) bip152_hb_settings += (bip152_hb_settings.isEmpty() ? tr("From") : "/" + tr("From"));
-    if (bip152_hb_settings.isEmpty()) bip152_hb_settings = tr("No");
+    if (stats->nodeStats.m_bip152_highbandwidth_to) bip152_hb_settings = tr_to;
+    if (stats->nodeStats.m_bip152_highbandwidth_from) bip152_hb_settings += (bip152_hb_settings.isEmpty() ? tr_from : "/" + tr_from);
```
hebasto, line 1120: While here, why not s/`"/"`/`QLatin1String("/")`/ ?

hebasto (commented):
Concept ACK.

src/qt/rpcconsole.cpp
jonatack: Dropped (a bit of a shame though, seems less readable) and did the same for other touched lines.

src/qt/rpcconsole.cpp
jonatack: Done, and also hoisted repeated translatables "never", "unknown" and "Ban for".

src/qt/rpcconsole.h
```diff
@@ -70,6 +70,8 @@ class RPCConsole: public QWidget
 
     QString tabTitle(TabTypes tab_type) const;
     QKeySequence tabShortcut(TabTypes tab_type) const;
+    QString tr_yes{tr("Yes")}, tr_no{tr("No")}, tr_to{tr("To")}, tr_from{tr("From")},
+        tr_never{tr("never")}, tr_unknown{tr("Unknown")}, tr_na{tr("N/A")}, tr_ban_for{tr("Ban for")};
```
jonatack, line 74: maybe "never" should be capitalized like the others
```suggestion
        tr_never{tr("Never")}, tr_unknown{tr("Unknown")}, tr_na{tr("N/A")}, tr_ban_for{tr("Ban for")};
```

src/qt/rpcconsole.h
```diff
@@ -70,6 +70,8 @@ class RPCConsole: public QWidget
 
     QString tabTitle(TabTypes tab_type) const;
     QKeySequence tabShortcut(TabTypes tab_type) const;
+    QString tr_yes{tr("Yes")}, tr_no{tr("No")}, tr_to{tr("To")}, tr_from{tr("From")},
```
hebasto, line 73: nit:
```suggestion
    const QString tr_yes{tr("Yes")}, tr_no{tr("No")}, tr_to{tr("To")}, tr_from{tr("From")},
```

hebasto (approved):
ACK 17edc9a9c8ca90f13f81ec8dcb949a5dbb5b3d32, tested on Linux Mint 20.1 (Qt 5.12.8).

src/qt/rpcconsole.h
jonatack: Thanks, good idea. Updating s/never/Never/ too.

jonatack: done

hebasto (approved):
re-ACK 52d10ee3cf59371b54cd4c9455f6e5b68ce07682

src/qt/rpcconsole.h
```diff
@@ -70,6 +70,8 @@ class RPCConsole: public QWidget
 
     QString tabTitle(TabTypes tab_type) const;
     QKeySequence tabShortcut(TabTypes tab_type) const;
+    const QString tr_yes{tr("Yes")}, tr_no{tr("No")}, tr_to{tr("To")}, tr_from{tr("From")},
+        tr_ban_for{tr("Ban for")}, tr_na{tr("N/A")}, tr_never{tr("Never")}, tr_unknown{tr("Unknown")};
```
hebasto, line 74: Maybe introduce a namespace instead of `tr_` prefix?

jonatack: Sure. Would you provide a working example of what you're thinking of?

hebasto: https://github.com/hebasto/gui/commit/9c7668050b872cbf2c755e3f945cc2a07d763b44

hebasto: If that idea is not that you like, could you, at least, make these variables `private`?

jonatack: Thanks! I would not have guessed that version, so your branch saved time. LGTM, done, added co-credit.

hebasto (approved):
re-ACK 2e2fe50ec1f3e3e4bf4b1b17d0a5351b9682601b

jonatack:
Rebased following merge of #226, dropped "Never" from the struct as it is now only used once.

hebasto (approved):
re-ACK 3d85e9229f074127c41eea43b4829c4d5bfc115a

src/qt/rpcconsole.cpp
```diff
@@ -1114,11 +1114,11 @@ void RPCConsole::updateDetailWidget()
...
     ui->peerHeading->setText(peerAddrDetails);
     ui->peerServices->setText(GUIUtil::formatServicesStr(stats->nodeStats.nServices));
-    ui->peerRelayTxes->setText(stats->nodeStats.fRelayTxes ? "Yes" : "No");
+    ui->peerRelayTxes->setText(stats->nodeStats.fRelayTxes ? ts.yes : ts.no);
     QString bip152_hb_settings;
-    if (stats->nodeStats.m_bip152_highbandwidth_to) bip152_hb_settings += "To";
-    if (stats->nodeStats.m_bip152_highbandwidth_from) bip152_hb_settings += (bip152_hb_settings == "" ? "From" : "/From");
-    if (bip152_hb_settings == "") bip152_hb_settings = "No";
+    if (stats->nodeStats.m_bip152_highbandwidth_to) bip152_hb_settings = ts.to;
+    if (stats->nodeStats.m_bip152_highbandwidth_from) bip152_hb_settings += (bip152_hb_settings.isEmpty() ? ts.from : QLatin1String("/") + ts.from);
```
Talkless, line 1120: nit: `QLatin1String("/")` could be `QLatin1Char('/')` instead. A bit simpler code path, no looping through source string (even if it's size 1) needed (peeking into Qt implementation).

src/qt/rpcconsole.h
```diff
@@ -136,6 +136,11 @@ public Q_SLOTS:
     void cmdRequest(const QString &command, const WalletModel* wallet_model);
 
 private:
+    struct TranslatedStrings {
```
Talkless, line 139: Interesting technique. I guess translators will see `TranslatedStrings` as location, not `rpcconsole.h`? But I guess that doesn't matter, not much context needed?

Talkless (commented):
Concept ACK

src/qt/rpcconsole.h
Talkless: I've tried with some small test project, and `lupdate`/`linguist` uses parent class as context, so no issues here.

src/qt/rpcconsole.cpp
jonatack: Thanks @Talkless! Done.

src/qt/rpcconsole.h
jonatack: @Talkless thank you for checking!

jonatack:
Updated with suggestion from @Talkless (thanks!) to replace `QLatin1String("/")` with `QLatin1Char('/')`

hebasto (approved):
re-ACK 6242beeb067139c01dd27c63ebcd24df5808cb15

Talkless (approved):
tACK 6242beeb067139c01dd27c63ebcd24df5808cb15, tested on Debian Sid with Qt 5.15.2. I see "Ban for.." translated to my native language as before, "To/From/Yes/No" are not but that's expected, as `.ts` files are not updated.

jarolrod (commented):
ACK 6242beeb067139c01dd27c63ebcd24df5808cb15

Good idea! These strings are/will be super common. This allows them to be used in other places when appropriate.

PR #223, opened 2021-02-24
qt: Re-add and rename transaction "Edit Label" action

laanwj:
This reverts PR #211.

I disagree with this change, I use the functionality a lot, it was the primary way I used to organize and edit transactions labels and am sad to see this go.

> you can edit a sending address in the send tab Address Book

Using the address book should not be encouraged at all! A while ago it was even proposed to remove it. There's rarely need to scroll through all historical addresses used and unused. The transaction list does just fine for this.

> While all other actions apply directly to the selected transaction, the Edit Label action applies to the selected transaction's address.

**In practice** when bitcoin is used in the commonly advised way, generate a new address for each transaction, those are equivalent though.

I doubt I (and **luke-jr**) will be the only users that will stumblle on this. Further discussion here: https://github.com/bitcoin-core/gui/pull/211#issuecomment-784755998

### DIFF
--- a/src/qt/transactionview.cpp
+++ b/src/qt/transactionview.cpp
@@ -172,6 +172,7 @@ TransactionView::TransactionView(const PlatformStyle *platformStyle, QWidget *pa
     QAction *copyTxIDAction = new QAction(tr("Copy transaction ID"), this);
     QAction *copyTxHexAction = new QAction(tr("Copy raw transaction"), this);
     QAction *copyTxPlainText = new QAction(tr("Copy full transaction details"), this);
+    QAction *editLabelAction = new QAction(tr("Edit label"), this);
     QAction *showDetailsAction = new QAction(tr("Show transaction details"), this);
 
     contextMenu = new QMenu(this);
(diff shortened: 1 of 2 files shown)


sipa:
I have little opinion on whether the functionality should exist at all, but I think the original change was motivated by confusion about what the feature does.

Just renaming it to "Change *address* label" would mitigate that just as well.

laanwj:
> Just renaming it to "Change address label" would mitigate that just as well.

Sounds good to me, I can include that change here if there's agreement on that.

hebasto:
Original suggestions:
- https://github.com/bitcoin-core/gui/issues/209#issuecomment-773098197 by @jonasschnelli 
- https://github.com/bitcoin-core/gui/issues/209#issuecomment-773100889 by @MarcoFalke 

> Just renaming it to "Change _address_ label" would mitigate that just as well.

I like this suggestion.

jarolrod:
> Just renaming it to "Change *address* label" would mitigate that just as well.

This is good, keeps the functionality while being more informative as to the outcome. Very democratic :)

hebasto (approved):
ACK 8f9644890a167a093d95ecef1f12a20dce1bc581, verified that 22664d6287dda243ebf06df4409b50ab509603e1 is a clean revert of 8f9644890a167a093d95ecef1f12a20dce1bc581.

luke-jr:
Concept ACK, thanks

MarcoFalke:
(edited OP before merge to remove `@`)

PR #225, opened 2021-02-24
Fix PaymentServer::handleURIOrFile warning message

hebasto:
`GUIUtil::parseBitcoinURI` does not check address validity. Therefore, an invalid address cannot cause parsing failure.

An invalid address case is covered by its own message.

This PR fixes the warning messages.

### DIFF
--- a/src/qt/paymentserver.cpp
+++ b/src/qt/paymentserver.cpp
@@ -251,7 +251,7 @@ void PaymentServer::handleURIOrFile(const QString& s)
             }
             else
                 Q_EMIT message(tr("URI handling"),
-                    tr("URI cannot be parsed! This can be caused by an invalid Bitcoin address or malformed URI parameters."),
+                    tr("URI cannot be parsed! This can be caused by malformed URI parameters."),
                     CClientUIInterface::ICON_WARNING);
 
             return;


Talkless (approved):
utACK 2434d004219f90f8d5408d22f222cfb7060565ac, no any testing, just built on Debian Sid with Qt 5.15.2

hebasto:
Updated 2434d004219f90f8d5408d22f222cfb7060565ac -> 45ebd9e1825886767efc38ddc3e300c638fcae6c ([pr225.01](https://github.com/hebasto/gui/commits/pr225.01) -> [pr225.02](https://github.com/hebasto/gui/commits/pr225.02)):
- rebased on top of the #280
- dropped all refactoring changes

prayank23:
> `GUIUtil::parseBitcoinURI` does not check address validity. Therefore, an invalid address cannot cause parsing failure.

If address validity is **not*" checked, can it be the cause for parsing failure?

> An invalid address case is covered by its own message "Invalid payment address %1".

Invalid address case is covered by `IsValidDestination()` and updated in https://github.com/bitcoin-core/gui/pull/280/ however, if code for `else` statement is executed can we assume address was never checked for validity?

https://github.com/prayank23/gui/blob/3bad0b3fada9ab7c5b03d31dc33d72654c1ba2be/src/qt/paymentserver.cpp#L252

hebasto:
Rebased 45ebd9e1825886767efc38ddc3e300c638fcae6c -> 5b20d46f8be6b7fc08012d80a9fc42e9dbc812ff ([pr225.02](https://github.com/hebasto/gui/commits/pr225.02) -> [pr225.03](https://github.com/hebasto/gui/commits/pr225.03)) on top of the recent CI changes.

jarolrod (commented):
Code Review ACK 5b20d46f8be6b7fc08012d80a9fc42e9dbc812ff 🥃

prayank23:
@hebasto Can you share any examples that I can try for which `GUIUtil::parseBitcoinURI` will return false and https://github.com/hebasto/gui/blob/5b20d46f8be6b7fc08012d80a9fc42e9dbc812ff/src/qt/paymentserver.cpp#L254 will be executed ?

PR #226, opened 2021-02-24
Add "Last Block" and "Last Tx" rows to peer details area

jonatack:
- add `RPCConsole::TimeDurationField` helper to replace repeated code and call system time only once in `RPCConsole::updateDetailWidget`
- add "Last Tx" (`CNodeStats::nLastTXTime`) field to peer details
- add "Last Block" (`CNodeStats::nLastBlockTime`) field to peer details

### DIFF
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -1079,7 +1079,7 @@
                <item row="1" column="0">
                 <widget class="QLabel" name="peerConnectionTypeLabel">
                  <property name="toolTip">
-                   <string>The direction and type of peer connection: %1</string>
+                  <string>The direction and type of peer connection: %1</string>
                  </property>
                  <property name="text">
                   <string>Direction/Type</string>
@@ -1342,17 +1342,17 @@
                 </widget>
                </item>
                <item row="12" column="0">
-                <widget class="QLabel" name="peerLastTxLabel">
+                <widget class="QLabel" name="peerLastBlockLabel">
                  <property name="toolTip">
-                  <string>Elapsed time since a novel transaction accepted into our mempool was received from this peer.</string>
+                  <string>Elapsed time since a novel block passing initial validity checks was received from this peer.</string>
                  </property>
                  <property name="text">
-                  <string>Last Tx</string>
+                  <string>Last Block</string>
                  </property>
                 </widget>
                </item>
                <item row="12" column="1">
-                <widget class="QLabel" name="peerLastTx">
+                <widget class="QLabel" name="peerLastBlock">
                  <property name="cursor">
                   <cursorShape>IBeamCursor</cursorShape>
                  </property>
@@ -1368,14 +1368,17 @@
                 </widget>
                </item>
                <item row="13" column="0">
-                <widget class="QLabel" name="label_15">
+                <widget class="QLabel" name="peerLastTxLabel">
+                 <property name="toolTip">
+                  <string>Elapsed time since a novel transaction accepted into our mempool was received from this peer.</string>
+                 </property>
                  <property name="text">
-                  <string>Last Send</string>
+                  <string>Last Tx</string>
                  </property>
                 </widget>
                </item>
                <item row="13" column="1">
-                <widget class="QLabel" name="peerLastSend">
+                <widget class="QLabel" name="peerLastTx">
                  <property name="cursor">
                   <cursorShape>IBeamCursor</cursorShape>
                  </property>
@@ -1391,14 +1394,14 @@
                 </widget>
                </item>
                <item row="14" column="0">
-                <widget class="QLabel" name="label_19">
+                <widget class="QLabel" name="label_15">
                  <property name="text">
-                  <string>Last Receive</string>
+                  <string>Last Send</string>
                  </property>
                 </widget>
                </item>
                <item row="14" column="1">
-                <widget class="QLabel" name="peerLastRecv">
+                <widget class="QLabel" name="peerLastSend">
                  <property name="cursor">
                   <cursorShape>IBeamCursor</cursorShape>
                  </property>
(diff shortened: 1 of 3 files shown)


jonatack:
Screenshot from 2021-02-24 17-03-13
Screenshot from 2021-02-24 17-02-58

src/qt/forms/debugwindow.ui
```diff
@@ -1342,13 +1342,65 @@
                 </widget>
                </item>
                <item row="12" column="0">
+                <widget class="QLabel" name="peerLastBlockLabel">
+                 <property name="toolTip">
+                  <string>Elapsed time since a novel block passing initial validity checks was received from this peer.</string>
```
jarolrod, line 1347:
```suggestion
                  <string>Elapsed time since a novel block was received from this peer.</string>
```

src/qt/forms/debugwindow.ui
```diff
@@ -1342,13 +1342,65 @@
...
+                 </property>
+                 <property name="textInteractionFlags">
+                  <set>Qt::LinksAccessibleByMouse|Qt::TextSelectableByKeyboard|Qt::TextSelectableByMouse</set>
+                 </property>
+                </widget>
+               </item>
+               <item row="13" column="0">
+                <widget class="QLabel" name="peerLastTxLabel">
+                 <property name="toolTip">
+                  <string>Elapsed time since a novel transaction accepted into our mempool was received from this peer.</string>
```
jarolrod, line 1373:
```suggestion
                  <string>Elapsed time since a novel transaction was received from this peer.</string>
```

jarolrod: alternative
```suggestion
                  <string>Elapsed time since a novel transaction, accepted into our mempool, was received from this peer.</string>
```

jarolrod (commented):
ACK c2758983f21415adde73f6948009ff3ef7f9bdb2

Small nits, maybe the extra explanation is not needed. Feel free to ignore.

jonatack:
Thanks for having a look, @jarolrod! Much appreciated.

The docs echo the ones I added in the `-netinfo help` and are a compact version of those I added to `src/net.h`:
```
    /** UNIX epoch time of the last block received from this peer that we had
     * not yet seen (e.g. not already received from another peer), that passed
     * preliminary validity checks and was saved to disk, even if we don't
     * connect the block or it eventually fails connection. Used as an inbound
     * peer eviction criterium in CConnman::AttemptToEvictConnection. */

    /** UNIX epoch time of the last transaction received from this peer that we
     * had not yet seen (e.g. not already received from another peer) and that
     * was accepted into our mempool. Used as an inbound peer eviction criterium
     * in CConnman::AttemptToEvictConnection. */
```
which I think are a bit more useful to users than the equivalent ones I added in the getpeerinfo rpc help (I've been adding these docs everywhere, hehe). For example, if we receive a novel block that does not pass initial validity checks, or a novel transaction that isn't accepted into our mempool, it won't appear. A user might wonder why, but the answers are difficult for casual users to glean from reading the code, and while writing the docs it took some review discussion with others to get this info cleared up. So it seems worth documenting when we have the space to do so.

jarolrod:
@jonatack thanks for the context, looks good to me then 👍

hebasto:
Concept ACK. Will review in the morning :)

src/qt/rpcconsole.cpp
```diff
@@ -1120,11 +1120,12 @@ void RPCConsole::updateDetailWidget()
     if (stats->nodeStats.m_bip152_highbandwidth_from) bip152_hb_settings += (bip152_hb_settings == "" ? "From" : "/From");
     if (bip152_hb_settings == "") bip152_hb_settings = "No";
     ui->peerHighBandwidth->setText(bip152_hb_settings);
-    ui->peerLastSend->setText(stats->nodeStats.nLastSend ? GUIUtil::formatDurationStr(GetSystemTimeInSeconds() - stats->nodeStats.nLastSend) : tr("never"));
-    ui->peerLastRecv->setText(stats->nodeStats.nLastRecv ? GUIUtil::formatDurationStr(GetSystemTimeInSeconds() - stats->nodeStats.nLastRecv) : tr("never"));
+    const int64_t time_now{GetSystemTimeInSeconds()};
+    ui->peerConnTime->setText(TimeDurationField(stats->nodeStats.nTimeConnected, time_now));
```
hebasto, line 1124: This refactoring changes the logic of the `peerConnTime` value setting, as now "Never" is a possible option.
Does it makes sense? Is it guaranteed to be safe?

jonatack: Good point. I was seeing output of `0s` and not "Never" but no reason to change the logic. Changed.

jonatack:
Updated per `git diff c275898 a5b349` to not use the helper for `peerConnTime` (and inverted the order of the helper arguments).

hebasto (approved):
ACK a5b349af5878c8127e89abb21d96863a8debbc40, tested on Linux Mint 20.1 (Qt 5.12.8):

Screenshot from 2021-02-25 13-15-21

Screenshot from 2021-02-25 13-14-58


piconit: while here why not to fix indentation from #179:
```diff
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -1079,7 +1079,7 @@
                <item row="1" column="0">
                 <widget class="QLabel" name="peerConnectionTypeLabel">
                  <property name="toolTip">
-                   <string>The direction and type of peer connection: %1</string>
+                  <string>The direction and type of peer connection: %1</string>
                  </property>
                  <property name="text">
                   <string>Direction/Type</string>
```

jonatack:
Thanks @hebasto, done.

hebasto (approved):
re-ACK 70d3c5d0b987411873d557440607aa4f3f65b1ec

Thank you @jonatack !

jarolrod (commented):
re-ack 70d3c5d0b987411873d557440607aa4f3f65b1ec

PR #227, opened 2021-02-24
Add URI info into -help output

hebasto:
When wallet is enabled `-help` shows the info about passing an URI to the `bitcoin-qt`:

```
$ src/qt/bitcoin-qt -? | head
Bitcoin Core version v21.99.0-82f37decdbba

Usage:  bitcoin-qt [options] [URI]

URI must follow BIP 21

Options:

  -?
       Print this help message and exit
```

Screenshot from 2021-02-24 20-47-00

### DIFF
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -510,7 +510,12 @@ int GuiMain(int argc, char* argv[])
     // Show help message immediately after parsing command-line options (for "-lang") and setting locale,
     // but before showing splash screen.
     if (HelpRequested(gArgs) || gArgs.IsArgSet("-version")) {
-        HelpMessageDialog help(nullptr, gArgs.IsArgSet("-version"));
+#ifdef ENABLE_WALLET
+        const bool wallet_enabled{WalletModel::isWalletEnabled()};
+#else
+        constexpr bool wallet_enabled{false};
+#endif // ENABLE_WALLET
+        HelpMessageDialog help(nullptr, gArgs.IsArgSet("-version"), wallet_enabled);
         help.showOrPrint();
         return EXIT_SUCCESS;
     }
--- a/src/qt/utilitydialog.cpp
+++ b/src/qt/utilitydialog.cpp
@@ -23,14 +23,15 @@
 #include <QLabel>
 #include <QMainWindow>
 #include <QRegExp>
+#include <QStringBuilder>
 #include <QTextCursor>
 #include <QTextTable>
 #include <QVBoxLayout>
 
 /** "Help message" or "About" dialog box */
-HelpMessageDialog::HelpMessageDialog(QWidget *parent, bool about) :
-    QDialog(parent, GUIUtil::dialog_flags),
-    ui(new Ui::HelpMessageDialog)
+HelpMessageDialog::HelpMessageDialog(QWidget* parent, bool about, bool wallet_enabled)
+    : QDialog(parent, GUIUtil::dialog_flags),
+      ui(new Ui::HelpMessageDialog)
 {
     ui->setupUi(this);
 
@@ -58,7 +59,8 @@ HelpMessageDialog::HelpMessageDialog(QWidget *parent, bool about) :
         ui->helpMessage->setVisible(false);
     } else {
         setWindowTitle(tr("Command-line options"));
-        QString header = "Usage:  bitcoin-qt [command-line options]                     \n";
+        const QString uri_info = wallet_enabled ? QStringLiteral(" [URI]\n\nURI must follow BIP 21") : QString();
+        const QString header(QLatin1String("Usage:  bitcoin-qt [options]") % uri_info % QLatin1Char('\n'));
         QTextCursor cursor(ui->helpMessage->document());
         cursor.insertText(version);
         cursor.insertBlock();
--- a/src/qt/utilitydialog.h
+++ b/src/qt/utilitydialog.h
@@ -22,7 +22,7 @@ class HelpMessageDialog : public QDialog
     Q_OBJECT
 
 public:
-    explicit HelpMessageDialog(QWidget *parent, bool about);
+    explicit HelpMessageDialog(QWidget* parent, bool about, bool wallet_enabled = false);
     ~HelpMessageDialog();
 
     void printToConsole();
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -96,7 +96,7 @@ BitcoinGUI::BitcoinGUI(interfaces::Node& node, const PlatformStyle *_platformSty
     updateWindowTitle();
 
     rpcConsole = new RPCConsole(node, _platformStyle, nullptr);
-    helpMessageDialog = new HelpMessageDialog(this, false);
+    helpMessageDialog = new HelpMessageDialog(this, false, enableWallet);
 #ifdef ENABLE_WALLET
     if(enableWallet)
     {


jarolrod (commented):
ACK 82f37decdbba340ea9798e0c82ae7b3cc4100248

I did not know you can provide a URI when running bitcoin-qt, so this is definitely helpful information.

One thing is that it would leave the user wondering if they aren't already aware, "what is BIP 21?" An internet search suffices, but what If a hyper-link was used?

hebasto:
> ... what If a hyper-link was used?

Not sure if it is worth to add complexity, but I will do it if there will be reviewers' consensus.

src/qt/bitcoin.cpp
```diff
@@ -510,7 +510,11 @@ int GuiMain(int argc, char* argv[])
     // Show help message immediately after parsing command-line options (for "-lang") and setting locale,
     // but before showing splash screen.
     if (HelpRequested(gArgs) || gArgs.IsArgSet("-version")) {
-        HelpMessageDialog help(nullptr, gArgs.IsArgSet("-version"));
+        bool wallet_enabled{false};
```
Talkless, line 513: nit: welp, we could go full const by using complete set of `#ifdef / #else / #endif`:

```
#ifdef ENABLE_WALLET
        const bool wallet_enabled{WalletModel::isWalletEnabled()};
#else
        constexpr bool wallet_enabled{false};
#endif // ENABLE_WALLET
```

src/qt/utilitydialog.h
```diff
@@ -22,7 +22,7 @@ class HelpMessageDialog : public QDialog
     Q_OBJECT
 
 public:
-    explicit HelpMessageDialog(QWidget *parent, bool about);
+    explicit HelpMessageDialog(QWidget *parent, bool about, bool wallet_enabled = true);
```
Talkless, line 25: Not sure if we need `= true` default. `HelpMessageDialog` is not something that is used a lot (only twice?), and if it was, assuming `= true` as default would probably be error prone, as it has to actually depend on the build.

Also, `bitcoingui.h` has `bool enableWallet = false;` defined, so at least it's inconsistency.

src/qt/utilitydialog.cpp
```diff
@@ -58,7 +58,8 @@ HelpMessageDialog::HelpMessageDialog(QWidget *parent, bool about) :
         ui->helpMessage->setVisible(false);
     } else {
         setWindowTitle(tr("Command-line options"));
-        QString header = "Usage:  bitcoin-qt [command-line options]                     \n";
+        const QString uri_info = wallet_enabled ? QStringLiteral(" [URI]\n\nURI must follow BIP 21") : QString();
+        const QString header(QLatin1String("Usage:  bitcoin-qt [options]") + uri_info + QLatin1Char('\n'));
```
Talkless, line 62: nit: well, if care was taken to use lean `QLatin1String`, including `<QStringBuilder>` and using `%` instead of `+` would make it even beter. Using `QStringBuilder` avoids that quadratic re-copying of string buffer with every `+`.

https://wiki.qt.io/Using_QString_Effectively#QStringBuilder_:_Fast_QString_concatenation

https://doc.qt.io/qt-5/qstring.html#more-efficient-string-construction

Talkless (changes requested):
Concept ACK

src/qt/utilitydialog.h
hebasto: btw, I think that combining two independent dialogs in one widget is a bad implementation.

hebasto:
Updated 82f37decdbba340ea9798e0c82ae7b3cc4100248 -> 0135c2a1f05f64522de9a00cdfa1c33fe668a0bb ([pr227.01](https://github.com/hebasto/gui/commits/pr227.01) -> [pr227.02](https://github.com/hebasto/gui/commits/pr227.02), [diff](https://github.com/hebasto/gui/compare/pr227.01..pr227.02)):

- addressed all of @Talkless' comments

src/qt/bitcoin.cpp
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui/pull/227#issuecomment-786746698).

src/qt/utilitydialog.h
hebasto: [Updated](https://github.com/bitcoin-core/gui/pull/227#issuecomment-786746698).

src/qt/utilitydialog.cpp
hebasto: [Updated](https://github.com/bitcoin-core/gui/pull/227#issuecomment-786746698).

src/qt/bitcoin.cpp
```diff
@@ -510,7 +510,12 @@ int GuiMain(int argc, char* argv[])
     // Show help message immediately after parsing command-line options (for "-lang") and setting locale,
     // but before showing splash screen.
     if (HelpRequested(gArgs) || gArgs.IsArgSet("-version")) {
-        HelpMessageDialog help(nullptr, gArgs.IsArgSet("-version"));
+#ifdef ENABLE_WALLET
+        const bool wallet_enabled{WalletModel::isWalletEnabled()};
```
luke-jr, line 514: Whether this running instance has the wallet disabled at runtime or not is irrelevant to a new execution... Can fold the `#ifdef` deep into utilitydialog.cpp

luke-jr (changes requested):

src/qt/utilitydialog.h
```diff
@@ -22,7 +22,7 @@ class HelpMessageDialog : public QDialog
     Q_OBJECT
 
 public:
-    explicit HelpMessageDialog(QWidget *parent, bool about);
+    explicit HelpMessageDialog(QWidget* parent, bool about, bool wallet_enabled = false);
```
Talkless, line 25: I see default changed to `= false`, but do we really need it (default value) at all?

I mean, why some "abstract" user of this class would allow himself to skip passing explicit argument and let default be assumed as `false`? Maybe there's argument for it and I am missing it.

hebasto: When `about` is `true` it does not matter what value `wallet_enabled` has. It seems weird to choose any its value on the caller site.

Talkless: Oh, so it depends on `about` too. OK got it.

src/qt/bitcoin.cpp
Talkless: @luke-jr you mean that there's no need for `HelpMessageDialog` to have that third constructor argument at all? It could decide in it's implementation if wallet is compiled or not?

luke-jr: Correct. The help message is about the build, not the current runtime.

src/qt/utilitydialog.cpp
```diff
@@ -58,7 +59,8 @@ HelpMessageDialog::HelpMessageDialog(QWidget *parent, bool about) :
         ui->helpMessage->setVisible(false);
     } else {
         setWindowTitle(tr("Command-line options"));
-        QString header = "Usage:  bitcoin-qt [command-line options]                     \n";
+        const QString uri_info = wallet_enabled ? QStringLiteral(" [URI]\n\nURI must follow BIP 21") : QString();
```
jarolrod, line 62: The URI has to be enclosed in single/double quotation marks. Maybe this should be added to avoid people wondering why they get to an exit message in their terminal.
```suggestion
        const QString uri_info = wallet_enabled ? QStringLiteral(" ["URI"]\n\nURI must follow BIP 21") : QString();
```

src/qt/bitcoin.cpp
Talkless: Ping @hebasto , @luke-jr has a point, or will your argue that run time (not compile time) changes should change help too?

hebasto:
Closing and leaving it up for grabs.

PR #228, opened 2021-02-24
Drop unused parameter in GUIUtil::get{Save|Open}FileName

hebasto:


jonatack (commented):
Code review ACK c3c9d8db30a4148eeb14db0ac10f0f27fced1087 (but should this feature be used instead?)

hebasto:
@jonasnick 
> but should this feature be used instead?

From https://doc.qt.io/qt-5/qfiledialog.html#getOpenFileName and https://doc.qt.io/qt-5/qfiledialog.html#getSaveFileName it follows that the `selectedFilter` out parameter is useful only in cases when multiple filters are passed to the `filter` parameter. That is not our case, right?

Talkless (approved):
utACK c3c9d8db30a4148eeb14db0ac10f0f27fced1087. Haven't even built, don't see anything to test.

luke-jr (changes requested):
Concept NACK to removing the option entirely. This looks useful for allowing the user to save a plaintext backup file from the GUI, or a fully signed transaction in normal transaction format...

Maybe make it default to nullptr and omit in callers that don't use it, though..

hebasto:
@luke-jr 
> This looks useful for allowing the user to save a plaintext backup file from the GUI...

Do you mean providing filters like `"Binary Wallet Data (*.dat);;Plain Text Wallet Data (*.txt)"` ?

luke-jr:
Yes, see #230 :)

hebasto:
@jonatack 
> but should this feature be used instead?

@luke-jr 
> Yes, see #230 :)

Closing.

PR #230, opened 2021-02-27
Support backup to new text-based database dump format

luke-jr:
Note: Leaving out the RPC dumpwallet format because it stores an encrypted wallet as unencrypted. If that's desirable to support at all, it can be a followup PR (perhaps with a strong warning).

### DIFF
--- a/src/interfaces/wallet.h
+++ b/src/interfaces/wallet.h
@@ -36,6 +36,10 @@ struct WalletContext;
 struct bilingual_str;
 typedef uint8_t isminefilter;
 
+enum class WalletBackupFormat {
+    Raw,   // Literal db copy
+};
+
 namespace interfaces {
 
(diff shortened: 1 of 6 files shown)


hebasto:
What use cases do you see for such a feature?

luke-jr:
Personal preference? Dunno, @achow101, what use cases did you have in mind for this dump format?

(I'm not adding the new format here, mind you, simply making it available to GUI users)

achow101:
I'm not sure how useful this is. There were two primary reasons for adding the `dump` and `createfromdump` commands: 1) To make it easier to debug low level database record issues and 2) to make it possible to change the database storage type without changing the wallet level logic. In the second case, this is not something that we really want people to do but I wanted to be able to do it at least just for testing.

PR #199, opened 2021-01-27
Added updated bitcoin PNG / SVG icon

Bosch-0:
In reference to https://github.com/bitcoin-core/gui/pull/147#issuecomment-767788887 Added the visually updated bitcoin icon (also discussed https://github.com/bitcoin-core/gui/issues/89), both the production PNG and source SVG version. It was requested that a seperate PR be open for the artwork change so here it is.

Images are the same dimensions as previous bitcoin.png file so should render fine in the GUI.

Also made some significant file size improvements without effecting image quality. bitcoin.png was optimized using [optimize-png.py](https://github.com/bitcoin-core/bitcoin-maintainer-tools/blob/master/optimise-pngs.py)

Original bitcoin.png file size: 306 KB

New updated bitcoin.png file after optimizing: 21.4 KB 

An overall 92.81% reduction.

The vector used is also much more simple (the blur on the original SVG added some extra complexity). 

Left is original right is updated icon.

Group 133

PR #231, opened 2021-02-28
splash: new layout and new icon

RandyMcMillan:
Testing  #147 and #199


jarolrod:
Is this just for testing how these two PR's would work when combined? You shouldn't open a new PR to do so ...

PR #233, opened 2021-03-01
qt test: Don't bind to regtest port

### LINKED ISSUE #10: Use random free port for test_bitcoin-qt
Currently the `test_bitcoin-qt` unit test uses the standard regtest port. It will fail to start if something else is using the regtest port, like a regtest instance of bitcoind or bitcoin-qt. Or another running `test_bitcoin-qt` (this happens if you are updating multiple branches at the same time). Instead of using the standard regtest port, this unit test should be using a random free port like we do in the functional tests.

#### Want to work on this issue?

For guidance on contributing, please read [CONTRIBUTING.md](https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md) before opening your pull request.

achow101:
The qt tests don't need to bind to the regtest port. By not binding, it will no longer conflict with existing regtest instances and the tests will run as normal.

Fixes #10

### DIFF
--- a/src/qt/test/test_main.cpp
+++ b/src/qt/test/test_main.cpp
@@ -54,6 +54,13 @@ int main(int argc, char* argv[])
 
     NodeContext node_context;
     std::unique_ptr<interfaces::Node> node = interfaces::MakeNode(&node_context);
+    gArgs.ForceSetArg("-listen", "0");
+    gArgs.ForceSetArg("-listenonion", "0");
+    gArgs.ForceSetArg("-discover", "0");
+    gArgs.ForceSetArg("-dnsseed", "0");
+    gArgs.ForceSetArg("-fixedseeds", "0");
+    gArgs.ForceSetArg("-upnp", "0");
+    gArgs.ForceSetArg("-natpmp", "0");
 
     bool fInvalid = false;
 


jarolrod (commented):
ACK a9d1ddd91dc20332033d09869afee76ff6f281b8

Confirming the behavior cited in #10 on `master`. Also confirming that this `PR` fixes it.

Tested on macOS 11.2 Qt 5.15.2 using the `QT_QPA_PLATFORM=cocoa ./test_bitcoin-qt` work around. 

I setup a node running `bitcoin-qt -regtest`, then setup two instances of the test to run at the same time. Everything ran smoothly and the blocking behavior was not encountered.

src/qt/test/test_main.cpp
```diff
@@ -54,6 +54,7 @@ int main(int argc, char* argv[])
 
     NodeContext node_context;
     std::unique_ptr<interfaces::Node> node = interfaces::MakeNode(&node_context);
+    gArgs.ForceSetArg("-listen", "0");
```
MarcoFalke, line 57: Same for listenonion?

        f.write("listenonion=0\n")
        f.write("discover=0\n")
        f.write("dnsseed=0\n")
        f.write("fixedseeds=0\n")
        f.write("upnp=0\n")
        f.write("natpmp=0\n")

hebasto:
I think this change prevents an opportunity to test Qt signals and UI response when new block/tx arrives.

(that, of course, will require to spin up another node)

src/qt/test/test_main.cpp
achow101: Added those overrides as well.

achow101:
> I think this change prevents an opportunity to test Qt signals and UI response when new block/tx arrives.
> 
> (that, of course, will require to spin up another node)

In that case, `-port` can be overridden. In any case, I don't think that we should reject his just because there might be a change in the future that conflicts with it.

MarcoFalke:
cr ACK e21276a82a9996c73e43990ccf927397f71399ea

didn't test

src/qt/test/test_main.cpp
```diff
@@ -54,6 +54,13 @@ int main(int argc, char* argv[])
...
```
jarolrod, line 57: Passing 0 to `port` lets the system assign a free port number. This fixes the issue while allowing for future GUI tests that involve interactions with another node.

```suggestion
    gArgs.ForceSetArg("-port", "0");
```

achow101: I don't think that is necessary at this time. Additionally, I don't think doing this would be useful. In this hypothetical test, the test framework needs to know what port is being used, so I don't think this would help.

jarolrod: true

jarolrod (commented):
re-ACK e21276a82a9996c73e43990ccf927397f71399ea, tested on macOS 11.2

Issue #95, opened 2020-09-17
Don't abort launch when failing to load a future wallet


Sjors:
The new settings.json introduced in bitcoin/bitcoin#15935 creates a problem when downgrading a node. E.g. try creating a hardware wallet enabled wallet with #4 and then downgrade back to master. It will fail at launch with "Wallet requires newer version of Bitcoin Core". You then have to manually delete the `settings.json` file.

It would be better if QT continues to load, but without the wallet, when this happens.

Sjors:
cc @ryanofsky @achow101

jonasschnelli:
This could be potentially harmful for pruned peers where syncing the chain may cause the wallet to fall behind the prune depth and thus require a fresh sync of the whole chain.

So I’m unsure about this. I think the current behavior (stop on wallet load failure) is preferable and the saner option. One option would be to threat pruned peers different... but sounds meh.

promag:
@jonasschnelli that's already a problem with multiwallet? When a wallet is needed and loaded it can already require a sync.

Sjors:
If could be a Yes / No prompt.

jonasschnelli:
@promag with multiwallet, it is an issue if you manually load or unload a wallet. The GUI should probably warn about that (it gets less critical once we automatically load manual loaded wallets after a restart).

But wallets set to load via the startup arguments should probably never silently unload.

promag:
Agreed, unnecessary source of panic. A warning or so would do the trick as @Sjors suggest.

ryanofsky:
Good catch. A similar problem was discussed in https://github.com/bitcoin/bitcoin/pull/19754#issuecomment-685858578 and https://github.com/bitcoin/bitcoin/pull/15454#issuecomment-686168009
about what should happen if a previously loaded wallet in settings.json could no longer be loaded because it was deleted/moved/unmounted. Different options mentioned there were:

1. Quietly not load the wallet, just logging a warning
2. Quietly not load the wallet, logging a warning, and removing it from `settings.json` so it doesn't get automatically loaded again in the future.
3. Show a modal prompt like "Wallet <wallet name> could not be loaded because of <reason>, do you want to try to load it next time Bitcoin Core is started?"
4. Load the wallet, but show indications it is having problems, and limit interactions disabling the transaction list or controls that won't work. Allow switching to the wallet in the GUI dropdown and unloading it with the "File -> Close Wallet...", like other wallets to avoid loading it again in the future.

[...]

PR #236, opened 2021-03-03
Bugfix: Allow the user to start anyway when loading a wallet errors

luke-jr:
Right now, if a GUI user leaves a wallet loaded and something happens to make it error at startup, they're stuck.

This turns the error dialog into a question about starting without the wallet instead.

### DIFF
--- a/src/wallet/load.cpp
+++ b/src/wallet/load.cpp
@@ -7,6 +7,7 @@
 
 #include <fs.h>
 #include <interfaces/chain.h>
+#include <node/interface_ui.h>
 #include <scheduler.h>
 #include <util/check.h>
 #include <util/string.h>
@@ -22,6 +23,17 @@
 #include <system_error>
 
 namespace wallet {
+
+bool HandleWalletLoadError(interfaces::Chain& chain, const std::string& wallet_file, const bilingual_str& error_string)
+{
+    if (!chain.initQuestion(error_string + Untranslated("\n\n") + _("Continue without this wallet?"), error_string, _("Error"), CClientUIInterface::MSG_ERROR | CClientUIInterface::MODAL | CClientUIInterface::BTN_OK | CClientUIInterface::BTN_ABORT)) {
+        return false;
+    }
+
+    RemoveWalletSetting(chain, wallet_file);
+    return true;
+}
+
 bool VerifyWallets(WalletContext& context)
 {
     interfaces::Chain& chain = *context.chain;
@@ -73,6 +85,7 @@ bool VerifyWallets(WalletContext& context)
     // Keep track of each wallet absolute path to detect duplicates.
     std::set<fs::path> wallet_paths;
 
+    bool modified_wallet_list = false;
     for (const auto& wallet : chain.getSettingsList("wallet")) {
         const auto& wallet_file = wallet.get_str();
         const fs::path path = fsbridge::AbsPathJoin(GetWalletDir(), fs::PathFromString(wallet_file));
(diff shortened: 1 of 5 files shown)


luke-jr:
Not sure why GitHub thinks this has conflicts - it doesn't.

Mainly this is a WIP because it uses a hacky global to avoid trying to load the wallets the user has chosen to skip... Suggestions for a good way to deal with that?

luke-jr:
Rebased, which made it practical to drop the global hack. This seems ready for review now.

kristapsk:
Concept ACK, will review / test.

src/wallet/load.cpp
```diff
@@ -22,6 +23,17 @@
 #include <system_error>
 
 namespace wallet {
+
+bool HandleWalletLoadError(interfaces::Chain& chain, const std::string& wallet_file, const bilingual_str& error_string)
+{
+    if (!chain.initQuestion(error_string + Untranslated("\n\n") + _("Continue without wallet?"), error_string, _("Error"), CClientUIInterface::MSG_ERROR | CClientUIInterface::MODAL | CClientUIInterface::BTN_OK | CClientUIInterface::BTN_ABORT)) {
```
kristapsk, line 29: Maybe better question message would be "Continue without this wallet?", so that is clear that only wallet having error is not being loaded, not all other wallets or wallet functionality disabled altogether?

kristapsk:
Apart from review comment above, ACK, working for me (tested scenario with trying to load external signer wallet without external signer support compiled into Bitcoin Core).

src/wallet/load.cpp
luke-jr: Sounds good, done

kristapsk (approved):
ACK cc85951352a4ebf8464f415b2a2ec66b7773fbd5

ryanofsky (commented):
Nice fix! I think there is a minor problem in the implementation, where if a user chooses not to load any wallet, none of the `bitcoin.conf` or command line wallets will be loaded even if they were present and could be verified. I suggested a fix for this below.

ryanofsky:
@luke-jr do you want to respond to suggestions here? Or would you be open to a someone making a new PR?

Status of the PR seems to be that it is up to date, but has a bug that could cause command line and bitcoin.conf wallets to be ignored. I suggested a fix for the bug and some cleanups in https://github.com/bitcoin-core/gui/pull/236#discussion_r842066839.

I think there is also a usability problem with this PR in the case of temporary errors where a wallet can't be loaded due to a temporary issue like: background assumeutxo download #23997, pruning, a missing external signer #22173, an encrypted drive not being mounted, a removable drive not being plugged in, or a version incompatibility that will be resolved by upgrading or downgrading. In these cases it would be useful to have the option to continue starting the GUI and letting the node sync, while temporarily not loading individual wallets that are unavailable, and the current dialog doesn't provide an option to temporarily not load a wallet. The only options are quit or continue without loading the wallet next time. I think it would be a improvement to change the dialog to "Wallet <wallet name> could not be loaded because of <reason>, do you want to try to load it next time Bitcoin Core is started?" with  "Yes" "No" buttons (and maybe "Details" and "Abort" buttons off to the side like #379). This was one of four alternatives suggested in https://github.com/bitcoin-core/gui/issues/95, and there is more discussion about this issue there.

hebasto:
> @luke-jr do you want to respond to suggestions here? Or would you be open to a someone making a new PR?

I'm going to close this, and mark "Up for grabs".

Issue #151, opened 2020-12-14
Issues with new create wallet dialogue


Rspigler:
As commented here (https://github.com/bitcoin-core/gui/pull/96#issuecomment-727259580)

`Encrypt Wallet`: "Encrypt the wallet.  The wallet will be encrypted with a passphrase of your choice." This is incorrect. The wallet is not encrypted, only the private keys are.

Also, this inconsistency:

Checking `Disable Private Keys` checks `Make Blank Wallet` (and disables `Encrypt Wallet`). This is correct.  `Disable Private Keys` should only be checked when `Make Blank Wallet` is also checked.  However, there is a way to have `Disable Private Keys` checked without having `Make Blank Wallet` checked (which doesn't make sense).

Check `Disable Private Keys`, then uncheck the auto-check `Make Blank Wallet`.  You are now left with the settings:

Disabled `Encrypt Wallet`, checked `Disable Private Keys`.

This is a setting that you can't enable yourself. (Checking `Disable Private Keys` also checks `Make Blank Wallet` normally).

To fix, unchecking the auto-enabled `Make Blank Wallet` should also uncheck `Disable Private Keys` and re-enable `Encrypt Wallet`.

hebasto:
cc @achow101 @Sjors

hebasto:
@Rspigler 
> `Encrypt Wallet`: "Encrypt the wallet. The wallet will be encrypted with a passphrase of your choice." This is incorrect. The wallet is not encrypted, only the private keys are.

This wording seems ubiquitous:

![Screenshot from 2021-03-05 03-21-59](https://user-images.githubusercontent.com/32963518/110053572-5a4a0700-7d62-11eb-8520-14ad14fa12f9.png)

Rspigler:
Where is that screenshot from? Some wallets do encrypt the entire wallet.  Bitcoin Core doesn't.

Compiling on master, I still see for the tooltip "Encrypt the wallet. The wallet will be encrypted with a passphrase of your choice", which is incorrect.  The wallet is not encrypted, the private keys are, which are two very different things https://github.com/bitcoin/bitcoin/issues/18085

There are also various text boxes after encrypting the wallet which are incorrect "You wallet is now encrypted" etc.

hebasto:
@Rspigler 
> Where is that screenshot from?

The Bitcoin Core.

Create a wallet without encryption. then in menu go to "Settings" -> "Encrypt Wallet".

Rspigler:
[...]

PR #243, opened 2021-03-10
fix issue when disabling the auto-enabled blank wallet checkbox

jarolrod:
As detailed by #151, On `master` a user can create the confusing scenario where you have a disabled `Encrypt Wallet` checkbox and a selected `Disable Private Keys` checkbox after unselecting the auto-enabled `Blank Wallet` checkbox.

This commit makes it so that when the `Blank Wallet` checkbox is auto-selected after the user selects `Disable Private keys`, unselecting it will also unselect the `Disable Private Keys` checkbox, which in turn re-enables the `Encrypt Wallet` checkbox.

Below are screenshots comparing the behavior of selecting `Disable Private Keys` then unselecting the `Blank Wallet` between `master` and this `PR`:

**Master:**
| Select `Disable Private Keys` | Unselect `Blank Wallet` |
| ----------------------------- | ------------------------ |
| Screen Shot 2021-03-09 at 7 57 14 PM |   Screen Shot 2021-03-09 at 7 57 31 PM    |

**PR:**
| Select `Disable Private Keys` | Unselect `Blank Wallet` |
| ----------------------------- | ------------------------ |
| Screen Shot 2021-03-09 at 7 34 12 PM | Screen Shot 2021-03-09 at 7 34 20 PM |

### DIFF
--- a/src/qt/createwalletdialog.cpp
+++ b/src/qt/createwalletdialog.cpp
@@ -53,6 +53,12 @@ CreateWalletDialog::CreateWalletDialog(QWidget* parent) :
         }
     });
 
+    connect(ui->blank_wallet_checkbox, &QCheckBox::toggled, [this](bool checked) {
+        if (!checked) {
+          ui->disable_privkeys_checkbox->setChecked(false);
+        }
+    });
+
 #ifndef USE_SQLITE
         ui->descriptor_checkbox->setToolTip(tr("Compiled without sqlite support (required for descriptor wallets)"));
         ui->descriptor_checkbox->setEnabled(false);


src/qt/createwalletdialog.cpp
```diff
@@ -53,6 +53,13 @@ CreateWalletDialog::CreateWalletDialog(QWidget* parent) :
         }
     });
 
+    connect(ui->blank_wallet_checkbox, &QCheckBox::toggled, [this](bool checked) {
+        // Disable the disable_privkeys_checkbox when blank_wallet_checkbox is false
```
Talkless, line 57: Not sure if these kind of comments are helpful. You can see that stated fact by just looking at rather self-explanatory code, with descriptive variable names.

Maybe this message would be more informative:
```
//Having "Disable Private Keys" checked after unchecking "Make Blank Wallet" makes no sense. Disabling keys means wallet has to be blank.
```
Or similar.

leonardojobim (approved):
Tested ACK https://github.com/bitcoin-core/gui/pull/243/commits/ca6bb0dac80bf970db9f42109f2304a2fdc5d1ee on Ubuntu 20.04

Rspigler:
Tested ACK commit ca6bb0dac80bf970db9f42109f2304a2fdc5d1ee 

Looks good!

Doesn't fix all the issues I described in https://github.com/bitcoin-core/gui/issues/151, since the encryption tooltip is still wrong.  But this annoying inconsistency is fixed

hebasto:
Concept ACK. The logic of checkboxes interaction is correct.

src/qt/createwalletdialog.cpp
hebasto: Actually, the comment is wrong as the `disable_privkeys_checkbox` is to be unchecked, not disabled :)

I think just removing it is fine.

hebasto (commented):
Approach ACK ca6bb0dac80bf970db9f42109f2304a2fdc5d1ee, please remove comment.

jarolrod:
Updated from ca6bb0d -> 915e341, addressed @hebasto's comment

**Changes**: remove the redundant comment. Code is self-explanatory.

src/qt/createwalletdialog.cpp
jarolrod: Removed the comment in 915e341

hebasto (approved):
ACK 915e34112b5a4c2ef391d7e3706603bcd6f62a2a

Talkless (approved):
ACK 915e34112b5a4c2ef391d7e3706603bcd6f62a2a

PR #248, opened 2021-03-13
Fix: For values of "Bytes transferred" and "Bytes/s" with 1000-based prefix names use 1000-based divisor instead of 1024-based

wodry:
v.0.21.0

I saw in the GUI peer window in the "received" column `1007 KB`, and after increasing to >=1024 I guess, it switched to `1 MB`. I would have expected the display unit to change from KB to MB already at value >=1000.

I looked into the code, and the values appear to be power-of-2 byte values, so the switching at >=1024 and not >=1000 seems correct.
But the unit display is not precisely correct, binary prefixes should be used for power-of-2 byte values.

To be correct, this PR changes ~~KB/MB/GB to KiB/MiB/GiB.~~ KB to kB and the divisor from 1024 to 1000.

### DIFF
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -759,14 +759,14 @@ QString formatNiceTimeOffset(qint64 secs)
 
 QString formatBytes(uint64_t bytes)
 {
-    if(bytes < 1024)
+    if (bytes < 1'000)
         return QObject::tr("%1 B").arg(bytes);
-    if(bytes < 1024 * 1024)
-        return QObject::tr("%1 KB").arg(bytes / 1024);
-    if(bytes < 1024 * 1024 * 1024)
-        return QObject::tr("%1 MB").arg(bytes / 1024 / 1024);
+    if (bytes < 1'000'000)
+        return QObject::tr("%1 kB").arg(bytes / 1'000);
+    if (bytes < 1'000'000'000)
+        return QObject::tr("%1 MB").arg(bytes / 1'000'000);
 
-    return QObject::tr("%1 GB").arg(bytes / 1024 / 1024 / 1024);
+    return QObject::tr("%1 GB").arg(bytes / 1'000'000'000);
 }
 
 qreal calculateIdealFontSize(int width, const QString& text, QFont font, qreal minPointSize, qreal font_size) {
--- a/src/qt/trafficgraphwidget.cpp
+++ b/src/qt/trafficgraphwidget.cpp
@@ -79,7 +79,7 @@ void TrafficGraphWidget::paintEvent(QPaintEvent *)
     int base = floor(log10(fMax));
     float val = pow(10.0f, base);
 
-    const QString units     = tr("KB/s");
+    const QString units = tr("kB/s");
     const float yMarginText = 2.0;
 
     // draw lines
@@ -128,10 +128,10 @@ void TrafficGraphWidget::updateRates()
 
     quint64 bytesIn = clientModel->node().getTotalBytesRecv(),
             bytesOut = clientModel->node().getTotalBytesSent();
-    float inRate = (bytesIn - nLastBytesIn) / 1024.0f * 1000 / timer->interval();
-    float outRate = (bytesOut - nLastBytesOut) / 1024.0f * 1000 / timer->interval();
-    vSamplesIn.push_front(inRate);
-    vSamplesOut.push_front(outRate);
+    float in_rate_kilobytes_per_sec = static_cast<float>(bytesIn - nLastBytesIn) / timer->interval();
+    float out_rate_kilobytes_per_sec = static_cast<float>(bytesOut - nLastBytesOut) / timer->interval();
+    vSamplesIn.push_front(in_rate_kilobytes_per_sec);
+    vSamplesOut.push_front(out_rate_kilobytes_per_sec);
     nLastBytesIn = bytesIn;
     nLastBytesOut = bytesOut;
 


wodry:
This is how it looks with this PR applied:
bildschirmfoto

src/qt/guiutil.cpp
```diff
@@ -762,11 +762,11 @@ QString formatBytes(uint64_t bytes)
     if(bytes < 1024)
         return QString(QObject::tr("%1 B")).arg(bytes);
     if(bytes < 1024 * 1024)
-        return QString(QObject::tr("%1 KB")).arg(bytes / 1024);
+        return QString(QObject::tr("%1 KiB")).arg(bytes / 1024);
```
jarolrod, line 765:
```suggestion
        return QString(QObject::tr("%1 KB")).arg(bytes / 1000);
```

Shouldn't we instead refactor to abide by the IEC 80000-13 standard where `1 KB=1000 bytes (base 10)`? macOS and Unix use this IEC standard, Windows is one of the only systems still treating `1KB=1024 bytes (base 2)`.

Or determine the system type and use its standard. So for Unix/macOS systems use base 10, for Windows use base 2.

src/qt/guiutil.cpp
```diff
@@ -762,11 +762,11 @@ QString formatBytes(uint64_t bytes)
...
     if(bytes < 1024 * 1024 * 1024)
-        return QString(QObject::tr("%1 MB")).arg(bytes / 1024 / 1024);
+        return QString(QObject::tr("%1 MiB")).arg(bytes / 1024 / 1024);
 
-    return QString(QObject::tr("%1 GB")).arg(bytes / 1024 / 1024 / 1024);
+    return QString(QObject::tr("%1 GiB")).arg(bytes / 1024 / 1024 / 1024);
```
jarolrod, line 769: I think this mebibyte, gibibyte naming scheme can be confusing to non-technical users

jarolrod (commented):
I think this `mebibyte`/`gibibyte` naming scheme can be confusing to non-technical users

src/qt/guiutil.cpp
wodry: I would be ok with that. Also with KB/s and change the divisor there accordingly form 1024 to 1000 (see e.g. [Ubuntu units policy](https://wiki.ubuntu.com/UnitsPolicy), which says for network bandwidth base-10 should be used). Whatever approach to solve, having unit KB with divisor 1024 is not correct I think.

jarolrod: @wodry yeah, I would say change this to base-10 instead and update the pr/commit description

wodry: Changed the title to be open for either approach. Would like to see more opinions.

hebasto:
Concept ACK on being consistent.

My vote is for decimal prefixes for data transfer rates and amounts of the transferred data, that means usage of kB/s, kilobyte per second, and kB, kilobyte. Note, that the small letter "k" is a standard prefix that means "kilo".

https://en.wikipedia.org/wiki/Data-rate_units#Unit_prefixes:
> The letter K is often used as a non-standard abbreviation for 1,024, especially in "KB" to mean KiB, the kilobyte in its binary sense.

wodry:
> My vote is for decimal prefixes

Thanks, so you would vote to change the divisor from 1024 to 1000 together with renaming KB to kB, right?

hebasto:
> > My vote is for decimal prefixes
> 
> Thanks, so you would vote to change the divisor from 1024 to 1000 together with renaming KB to kB, right?

Yes. That is my suggestion.

sipa:
No strong opinion on 1000-based vs 1024-based, but please use unit names consistent with what is chosen.

wodry:
I updated so the prefixes are unchanged (aside of of KB->kB as @hebasto suggested), and just the divisor is corrected from 1024 to 1000.

wodry:
Not sure if this `* 1.0f` is really needed or if there is a nicer solution. Just took it over.

wodry:
This is how it looks now with the updated PR applied. The 1000-based prefixes have the advantage to be consistent with the 10-based scaling of the network throughput graph horizontal marker lines (e.g. 100 kB/s, 1000 kB/s like in the screenshot).

bildschirmfoto
bildschirmfoto2

src/qt/guiutil.cpp
```diff
@@ -759,14 +759,14 @@ QString formatNiceTimeOffset(qint64 secs)
...
 {
-    if(bytes < 1024)
+    if(bytes < 1000)
         return QObject::tr("%1 B").arg(bytes);
-    if(bytes < 1024 * 1024)
-        return QObject::tr("%1 KB").arg(bytes / 1024);
-    if(bytes < 1024 * 1024 * 1024)
-        return QObject::tr("%1 MB").arg(bytes / 1024 / 1024);
+    if(bytes < 1000 * 1000)
+        return QObject::tr("%1 kB").arg(bytes / 1000);
```
jarolrod, line 765: @hebasto: This seems like a good place for some context to translators. For example, look at the French translation below.
Screen Shot 2021-03-18 at 7 49 37 PM
 I don't know how the decimal unit `kB` is written in French, but how do we let a translator know that they are supposed to translate as the`decimal unit` and not the `binary unit`?

hebasto: Yes. We could add more translation context later, when https://github.com/bitcoin/bitcoin/issues/21465 approach is implemented.

hebasto (commented):
Approach ACK 200b503733ab2b78c023571f31157cb81597c3e2

@wodry 

Do you mind applying [`clang-format-diff.py`](https://github.com/bitcoin/bitcoin/tree/master/contrib/devtools#clang-format-diffpy) to your commit?

src/qt/trafficgraphwidget.cpp
```diff
@@ -128,8 +128,9 @@ void TrafficGraphWidget::updateRates()
 
     quint64 bytesIn = clientModel->node().getTotalBytesRecv(),
             bytesOut = clientModel->node().getTotalBytesSent();
-    float inRate = (bytesIn - nLastBytesIn) / 1024.0f * 1000 / timer->interval();
-    float outRate = (bytesOut - nLastBytesOut) / 1024.0f * 1000 / timer->interval();
+    // Rates are in bytes/millisecond which is equal to kilobytes/second
+    float inRate = (bytesIn - nLastBytesIn) * 1.0f / timer->interval();
+    float outRate = (bytesOut - nLastBytesOut) * 1.0f / timer->interval();
```
hebasto, line 133: Why `  * 1.0f` is required?

src/qt/trafficgraphwidget.cpp
```diff
@@ -128,8 +128,9 @@ void TrafficGraphWidget::updateRates()
 
     quint64 bytesIn = clientModel->node().getTotalBytesRecv(),
             bytesOut = clientModel->node().getTotalBytesSent();
-    float inRate = (bytesIn - nLastBytesIn) / 1024.0f * 1000 / timer->interval();
-    float outRate = (bytesOut - nLastBytesOut) / 1024.0f * 1000 / timer->interval();
+    // Rates are in bytes/millisecond which is equal to kilobytes/second
```
hebasto, line 131: Instead of adding a comment, is it better just choose proper names for variables?
E.g., `in_rate_kilobytes_per_sec` ?

src/qt/trafficgraphwidget.cpp
wodry: I am no C++ expert, just took it over from `/ 1024.0f * 1000` before. If the `1024.0f` was also not needed (I imagined it might have had some reason of C++ type conversion), I will remove the `* 1.0f`

hebasto: If type conversion is supposed, it would better to apply explicit conversion.

wodry:
> Approach ACK [200b503](https://github.com/bitcoin-core/gui/commit/200b503733ab2b78c023571f31157cb81597c3e2)
> 
> @wodry
> 
> Do you mind applying [`clang-format-diff.py`](https://github.com/bitcoin/bitcoin/tree/master/contrib/devtools#clang-format-diffpy) to your commit?

Thanks for the tipp, have done that, will update.

hebasto:
May I suggest a change:

```diff
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -759,14 +759,14 @@ QString formatNiceTimeOffset(qint64 secs)
 
 QString formatBytes(uint64_t bytes)
 {
-    if (bytes < 1000)
+    if (bytes < 1'000)
         return QObject::tr("%1 B").arg(bytes);
-    if (bytes < 1000 * 1000)
-        return QObject::tr("%1 kB").arg(bytes / 1000);
-    if (bytes < 1000 * 1000 * 1000)
-        return QObject::tr("%1 MB").arg(bytes / 1000 / 1000);
+    if (bytes < 1'000'000)
+        return QObject::tr("%1 kB").arg(bytes / 1'000);
+    if (bytes < 1'000'000'000)
+        return QObject::tr("%1 MB").arg(bytes / 1'000'000);
 
-    return QObject::tr("%1 GB").arg(bytes / 1000 / 1000 / 1000);
+    return QObject::tr("%1 GB").arg(bytes / 1'000'000'000);
 }
 
 qreal calculateIdealFontSize(int width, const QString& text, QFont font, qreal minPointSize, qreal font_size) {
diff --git a/src/qt/trafficgraphwidget.cpp b/src/qt/trafficgraphwidget.cpp
index 9e0b67917..7e12410c8 100644
--- a/src/qt/trafficgraphwidget.cpp
+++ b/src/qt/trafficgraphwidget.cpp
@@ -128,8 +128,8 @@ void TrafficGraphWidget::updateRates()
 
     quint64 bytesIn = clientModel->node().getTotalBytesRecv(),
             bytesOut = clientModel->node().getTotalBytesSent();
-    float in_rate_kilobytes_per_sec = (bytesIn - nLastBytesIn) / timer->interval();
-    float out_rate_kilobytes_per_sec = (bytesOut - nLastBytesOut) / timer->interval();
+    float in_rate_kilobytes_per_sec = static_cast<float>(bytesIn - nLastBytesIn) / timer->interval();
+    float out_rate_kilobytes_per_sec = static_cast<float>(bytesOut - nLastBytesOut) / timer->interval();
     vSamplesIn.push_front(in_rate_kilobytes_per_sec);
     vSamplesOut.push_front(out_rate_kilobytes_per_sec);
     nLastBytesIn = bytesIn;
```

?

wodry:
Thanks very much for the high quality coding hints! Updated commit.

hebasto (approved):
ACK d09ebc47233187ab8dd5a70df4d261353958978c, tested on Linux Mint 20.1 (Qt 5.12.8) the both "Network Traffic" and "Peers" tabs of the "Node Window".

jarolrod (commented):
ACK d09ebc47233187ab8dd5a70df4d261353958978c

Tested on macOS 11.1 Qt 5.15.2

hebasto:
@jonatack Do you mind looking into this PR?

jonatack:
> @jonatack Do you mind looking into this PR?

I've been trying to stay away from changes where my opinion would be the minority. I grew up with powers of two and prefer them.

leonardojobim (commented):
Tested ACK https://github.com/bitcoin-core/gui/pull/248/commits/d09ebc47233187ab8dd5a70df4d261353958978c on Ubuntu 20.04 Qt 5.12.8

hebasto:
Additional discussion from [IRC](http://www.erisian.com.au/bitcoin-core-dev/log-2021-03-18.html#l-200):
> \<Arvidt> Any preference for having base-10 (kB/MB/GB) or base-2 (KiB/MiB/GiB) prefix for GUI network peer tab data transferred/throughput value? See https://github.com/bitcoin-core/gui/pull/248 Would be thankful to get some feedback.
> \<luke-jr> Arvidt: base 10 sucks ;)
> \<hebasto> luke-jr: what reasons for?
> \<luke-jr> hebasto: 10 is a multiple of 5 and 2
> \<luke-jr> but more importantly, networks tend to use base-1024 measurements
> \<sipa> my gigabit ethernet does 1000000000 bits/s
> \<luke-jr> sipa: it does? :o
> \<sipa> yes
> \<luke-jr> hmm
> \<sipa> all network stuff is 1000-based, as far as i know'
> \<sipa> only disk sizes use 1024-based units conventionally
> \<sipa> or on-disk sizes
> \<luke-jr> RAM typically does even now
> \<sipa> that's true, RAM is also typically 1024-based

jnewbery:
Concept ACK. Using powers of 10 and correctly calling them kB and MB seems reasonable for a user-facing GUI.

PR #250, opened 2021-03-14
scripted-diff: Drop redundant QString calls

hebasto:
The return type of `QObject::tr` function _is_ `QString` :tiger2:


jarolrod (commented):
ACK def1e64bb4fc902bfc19d6487b33cd940f0e0f34, tested on macOS 10.14.6 Qt 5.15.2

This change is correct according to the Qt docs: https://doc.qt.io/qt-5/qobject.html#tr

Verified that these are all the occurrences of this case. Ran `bitcoin-qt` and verified there are no alterations of behavior with this change.

PR #251, opened 2021-03-17
Improve URI/file handling message

hebasto:
This PR:
- fixes missing spaces after full stops
- makes the translation context much bigger

The latter is the main motivation for this PR, as I became a translator :tiger2: 

Screenshots:
- master (a9d1b40d53ec417eefbe767aa66701ef8e1801d5)

DeepinScreenshot_select-area_20210317211750

- this PR:

Screenshot from 2021-03-17 21-13-36


jarolrod (commented):
ACK ef3e1d7272d294c65cc9904079571cf67dc6c463

Looking on `transifex`. I See that on `master` all of these strings would be broken up into seperate translations:

Screen Shot 2021-03-24 at 11 39 37 AM

This would unify these strings under one translation. This added context should be a benefit to translators

PR #254, opened 2021-03-23
refactor: Drop redundant setEditTriggers(NoEditTriggers) calls

hebasto:
The models of the both views have no `Qt::ItemIsEditable` flag:
https://github.com/bitcoin-core/gui/blob/3c87dbe95c925274f80234ad4a88beb5a05fdfff/src/qt/peertablemodel.cpp#L218-L224
https://github.com/bitcoin-core/gui/blob/3c87dbe95c925274f80234ad4a88beb5a05fdfff/src/qt/bantablemodel.cpp#L148-L154

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -608,7 +608,6 @@ void RPCConsole::setClientModel(ClientModel *model, int bestblock_height, int64_
         // set up peer table
         ui->peerWidget->setModel(model->getPeerTableModel());
         ui->peerWidget->verticalHeader()->hide();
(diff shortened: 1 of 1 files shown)


Talkless (approved):
utACK 257f55c119c2c63245f3a84a9cd8f7aaeaf2d129, seems reasonable.

jarolrod (commented):
ACK 257f55c119c2c63245f3a84a9cd8f7aaeaf2d129, looks correct.

Tested for any weird behavior on macOS 11.2.3 Qt 5.15.2

PR #260, opened 2021-03-26
Handle exceptions instead of crash

hebasto:
This PR is an alternative to https://github.com/bitcoin/bitcoin/pull/18897, and is based on Russ' [idea](https://github.com/bitcoin/bitcoin/pull/18897#pullrequestreview-418703664):
> IMO it would be nice to have a followup PR that eliminated the one-line forwarding methods ...

Related issues
- #91 
- https://github.com/bitcoin/bitcoin/issues/18643

Qt docs: https://doc.qt.io/qt-5.12/exceptionsafety.html#exceptions-in-client-code

With this PR the GUI handles the wallet-related exception, and:
- display it to a user:

Screenshot from 2021-04-01 02-55-59

- prints a message to `stderr`:
```


************************
EXCEPTION: 18NonFatalCheckError       
wallet/wallet.cpp:2677 (IsCurrentForAntiFeeSniping)
Internal bug detected: '!chain.findBlock(block_hash, FoundBlock().time(block_time))'
You may report this issue here: https://github.com/bitcoin/bitcoin/issues
       
bitcoin in QPushButton->SendCoinsDialog       

```

- writes a message to the `debug.log`
- and, if the exception is a non-fatal error, leaves the main window running.

### DIFF
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -45,10 +45,12 @@
 #include <QApplication>
 #include <QDebug>
 #include <QFontDatabase>
+#include <QLatin1String>
 #include <QLibraryInfo>
 #include <QLocale>
 #include <QMessageBox>
 #include <QSettings>
+#include <QStringBuilder>
 #include <QThread>
 #include <QTimer>
 #include <QTranslator>
@@ -417,7 +419,10 @@ void BitcoinApplication::shutdownResult()
 
 void BitcoinApplication::handleRunawayException(const QString &message)
 {
-    QMessageBox::critical(nullptr, "Runaway exception", BitcoinGUI::tr("A fatal error occurred. %1 can no longer continue safely and will quit.").arg(PACKAGE_NAME) + QString("<br><br>") + message);
+    QMessageBox::critical(
+        nullptr, tr("Runaway exception"),
+        tr("A fatal error occurred. %1 can no longer continue safely and will quit.").arg(PACKAGE_NAME) %
+        QLatin1String("<br><br>") % GUIUtil::MakeHtmlLink(message, PACKAGE_BUGREPORT));
     ::exit(EXIT_FAILURE);
 }
 
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -426,6 +426,16 @@ void BitcoinApplication::handleRunawayException(const QString &message)
     ::exit(EXIT_FAILURE);
 }
 
+void BitcoinApplication::handleNonFatalException(const QString& message)
+{
+    assert(QThread::currentThread() == thread());
+    QMessageBox::warning(
+        nullptr, tr("Internal error"),
+        tr("An internal error occurred. %1 will attempt to continue safely. This is "
+           "an unexpected bug which can be reported as described below.").arg(PACKAGE_NAME) %
+        QLatin1String("<br><br>") % GUIUtil::MakeHtmlLink(message, PACKAGE_BUGREPORT));
+}
+
 WId BitcoinApplication::getMainWinId() const
 {
     if (!window)
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -42,6 +42,7 @@
 #include <QGuiApplication>
 #include <QJsonObject>
 #include <QKeyEvent>
+#include <QLatin1String>
 #include <QLineEdit>
 #include <QList>
 #include <QLocale>
@@ -54,6 +55,7 @@
 #include <QShortcut>
 #include <QSize>
 #include <QString>
+#include <QStringBuilder>
 #include <QTextDocument> // for Qt::mightBeRichText
 #include <QThread>
 #include <QUrlQuery>
@@ -891,6 +893,13 @@ QImage GetImage(const QLabel* label)
 #else
     return label->pixmap()->toImage();
 #endif
+}
+
+QString MakeHtmlLink(const QString& source, const QString& link)
+{
+    return QString(source).replace(
+        link,
+        QLatin1String("<a href=\"") % link % QLatin1String("\">") % link % QLatin1String("</a>"));
 }
 
 } // namespace GUIUtil
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -902,4 +902,15 @@ QString MakeHtmlLink(const QString& source, const QString& link)
         QLatin1String("<a href=\"") % link % QLatin1String("\">") % link % QLatin1String("</a>"));
 }
 
+void PrintSlotException(
+    const std::exception* exception,
+    const QObject* sender,
+    const QObject* receiver)
+{
+    std::string description = sender->metaObject()->className();
+    description += "->";
+    description += receiver->metaObject()->className();
+    PrintExceptionContinue(exception, description.c_str());
+}
+
 } // namespace GUIUtil
--- a/src/qt/guiutil.h
+++ b/src/qt/guiutil.h
@@ -9,18 +9,23 @@
 #include <fs.h>
 #include <net.h>
 #include <netaddress.h>
+#include <util/check.h>
 
+#include <QApplication>
 #include <QEvent>
 #include <QHeaderView>
 #include <QItemDelegate>
 #include <QLabel>
 #include <QMessageBox>
+#include <QMetaObject>
 #include <QObject>
 #include <QProgressBar>
 #include <QString>
 #include <QTableView>
 
+#include <cassert>
 #include <chrono>
+#include <utility>
 
 class QValidatedLineEdit;
 class SendCoinsRecipient;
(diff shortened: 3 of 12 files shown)


hebasto:
Draft for now as @ryanofsky's [comments](https://github.com/bitcoin/bitcoin/pull/18897#pullrequestreview-418703664) should be addressed soon.

hebasto:
#### Note to reviewers

To test this PR you could:
1. apply this patch
```diff
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -2674,7 +2674,7 @@ static bool IsCurrentForAntiFeeSniping(interfaces::Chain& chain, const uint256&
     }
     constexpr int64_t MAX_ANTI_FEE_SNIPING_TIP_AGE = 8 * 60 * 60; // in seconds
     int64_t block_time;
-    CHECK_NONFATAL(chain.findBlock(block_hash, FoundBlock().time(block_time)));
+    CHECK_NONFATAL(!chain.findBlock(block_hash, FoundBlock().time(block_time)));
     if (block_time < (GetTime() - MAX_ANTI_FEE_SNIPING_TIP_AGE)) {
         return false;
     }
```

and initiate a new transaction creation, or

2. apply this patch
```diff

--- a/src/wallet/interfaces.cpp
+++ b/src/wallet/interfaces.cpp
@@ -330,7 +330,7 @@ public:
         }
         num_blocks = m_wallet->GetLastBlockHeight();
         block_time = -1;
-        CHECK_NONFATAL(m_wallet->chain().findBlock(m_wallet->GetLastBlockHash(), FoundBlock().time(block_time)));
+        CHECK_NONFATAL(!m_wallet->chain().findBlock(m_wallet->GetLastBlockHash(), FoundBlock().time(block_time)));
         tx_status = MakeWalletTxStatus(*m_wallet, mi->second);
         return true;
     }
```
and initiate opening a wallet.

hebasto:
@ryanofsky 
re [comment](https://github.com/bitcoin/bitcoin/pull/18897#discussion_r430727242):
> Would drop this special case for NonFatalCheckError and simply catch std::exception below. NonFatalCheckError is really an internal implementation of the CHECK macro and most useful to unit tests. Exceptions aren't valid in qt callbacks so NonFatalCheckError is no different here than any other exception. Just another unexpected condition that should never happen

The idea is to distinguish non-fatal errors when no need to stop the client; [described](https://github.com/bitcoin/bitcoin/issues/18643#issue-599952770) by @MarcoFalke:
> ... it is more user friendly for a GUI to catch the exception and display it to the user, then abort the current action and leave the main window running.

hebasto:
@promag @ryanofsky @jonasschnelli

Kindly asking for reviewing :)

:tiger2:

leonardojobim (approved):
Tested ACK https://github.com/bitcoin-core/gui/pull/260/commits/63cca51a9956b4333237416833722d93df43c9ff on Ubuntu 20.04 Qt Qt 5.12.8.

On master branch:
ss01

On this PR branch:
ss02

laanwj:
Personally I like the "A fatal error occured. Bitcoin Core can no longer  …" message with a seperator and then the detailed error message much better than an "Internal error" popup with some developer text. 

I think we should keep that.

> prints a message to stderr:

Is it logged to the debug log? Anything printed to stderr tends to get lost (or is logged to a really hard to find place) on graphical operating systems.

ryanofsky (approved):
Code review ACK 63cca51a9956b4333237416833722d93df43c9ff. Agree with laanwj that error text could be improved and I made a specific suggestion below, but this is already better than crashing, and I like the approach of catching exceptions in the connect handler so they don't cause problems for Qt.

ryanofsky:
re: https://github.com/bitcoin-core/gui/pull/260#pullrequestreview-625682268

> Is it logged to the debug log?

Yep, PrintExceptionContinue logs it.

src/qt/guiutil.h
```diff
@@ -327,6 +332,54 @@ namespace GUIUtil
...
+                } catch (const NonFatalCheckError& e) {
+                    PrintSlotException(&e, sender, receiver);
+                    ok = QMetaObject::invokeMethod(
+                        qApp, "handleNonFatalException",
+                        blockingGUIThreadConnection(),
+                        Q_ARG(QString, QString::fromStdString(e.what())));
+                } catch (const std::exception& e) {
+                    PrintSlotException(&e, sender, receiver);
+                    ok = QMetaObject::invokeMethod(
+                        qApp, "handleRunawayException",
```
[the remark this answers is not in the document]
ryanofsky: re: https://github.com/bitcoin-core/gui/pull/260#discussion_r605195191

Thanks, you're right. I didn't realize these methods had the same name. If I am confirmed as [Official Namer Of Stuff](https://github.com/bitcoin/bitcoin/pull/20861#discussion_r596389178) I will decree that gui code must use different names for things that are not the same. Might start with [`addNewRequest`'s](https://github.com/bitcoin/bitcoin/blob/2c0fc856a6b0c82c5dddbbaee417171577514507/src/qt/recentrequeststablemodel.h#L80-L82) first, though.

hebasto:
@laanwj @ryanofsky 

Thank you for your reviews and suggestions!

Updated 63cca51a9956b4333237416833722d93df43c9ff -> f81672e65b443c00a650b2d920996ea738fa2d7f ([pr260.10](https://github.com/hebasto/gui/commits/pr260.10) -> [pr260.02](https://github.com/hebasto/gui/commits/pr260.11), [diff](https://github.com/hebasto/gui/compare/pr260.10..pr260.11)):

- addressed reviewers' comments
- small improvements

src/qt/bitcoin.cpp
```diff
@@ -421,6 +421,12 @@ void BitcoinApplication::handleRunawayException(const QString &message)
     ::exit(EXIT_FAILURE);
 }
 
+void BitcoinApplication::handleNonFatalException(const QString& exception_message)
+{
+    assert(QThread::currentThread() == qApp->thread());
+    QMessageBox::warning(nullptr, "Internal error", exception_message);
```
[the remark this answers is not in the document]
hebasto: Thanks! [Done](https://github.com/bitcoin-core/gui/pull/260#issuecomment-811546172).

src/qt/walletmodel.cpp
```diff
@@ -65,7 +65,8 @@ WalletModel::~WalletModel()
 void WalletModel::startPollBalance()
 {
     // This timer will be fired repeatedly to update the balance
-    connect(timer, &QTimer::timeout, this, &WalletModel::pollBalanceChanged);
+    connect(timer, &QTimer::timeout, this, &WalletModel::timerTimeout);
```
laanwj, line 68: Might want to add a comment here why the signal is not directly connected (e.g. because `QTimer::timeout` is a private signal)

hebasto:
Updated f81672e65b443c00a650b2d920996ea738fa2d7f -> b8e5d0d3fe3386807d47f50d13ac34fcd2a538fd ([pr260.11](https://github.com/hebasto/gui/commits/pr260.11) -> [pr260.12](https://github.com/hebasto/gui/commits/pr260.12), [diff](https://github.com/hebasto/gui/compare/pr260.11..pr260.12)):

- addressed @laanwj's [comment](https://github.com/bitcoin-core/gui/pull/260#discussion_r607066469):
> Might want to add a comment here why the signal is not directly connected (e.g. because `QTimer::timeout` is a private signal)

src/qt/walletmodel.cpp
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui/pull/260#issuecomment-813403971).

src/qt/bitcoin.cpp
```diff
@@ -417,7 +419,10 @@ void BitcoinApplication::shutdownResult()
 
 void BitcoinApplication::handleRunawayException(const QString &message)
 {
-    QMessageBox::critical(nullptr, "Runaway exception", BitcoinGUI::tr("A fatal error occurred. %1 can no longer continue safely and will quit.").arg(PACKAGE_NAME) + QString("<br><br>") + message);
+    QMessageBox::critical(
+        nullptr, tr("Runaway exception"),
+        tr("A fatal error occurred. %1 can no longer continue safely and will quit.").arg(PACKAGE_NAME) %
+        QLatin1String("<br><br>") % GUIUtil::MakeHtmlLink(message, PACKAGE_BUGREPORT));
```
ryanofsky, line 425: In commit "qt: Make PACKAGE_BUGREPORT link clickable" (af7e365b1516d660d271475fdfe0c20ae09e66a8)

Since our project only uses the issue tracker for development, not user support, it would be good in the future to link to something like stack exchange or a support resources page if a fatal error occurs, instead of the issue tracker.

But this commit is only making existing issue tracker links clickable, not adding new links, so it doesn't cause problems and is good because it keeps the links rendered in consistent way.

In the future it might be be good stop having "Runaway exception" dialogs and instead have distinct "Internal bug" (e.g. unhandled preconditions or postconditions) and "Fatal error" (e.g. failing disk writes) dialogs, treating the first kind like bugs and the second kind like support issues. But it seems good to keep the scope of this PR narrow and focused on the nonfatal checks like it is now.

src/qt/guiutil.cpp
```diff
@@ -902,4 +902,15 @@ QString MakeHtmlLink(const QString& source, const QString& link)
         QLatin1String("<a href=\"") % link % QLatin1String("\">") % link % QLatin1String("</a>"));
 }
 
+void PrintSlotException(
+    const std::exception* exception,
+    const QObject* sender,
+    const QObject* receiver)
+{
+    std::string description = sender->metaObject()->className();
```
ryanofsky, line 910: In commit "qt: Add GUIUtil::ExceptionSafeConnect function" (f7e260a471010e2d656fbc5ea8c310f6d94c26b9)

Not important, but just in case you revisit this, I think it might be a little clearer to call this variable "location" instead of "description". (I was a little confused at first about why it didn't seem to include information from the exception.)

ryanofsky (approved):
Code review ACK b8e5d0d3fe3386807d47f50d13ac34fcd2a538fd. This is great! I think more improvements are possible but implementation is very clean and I love how targeted each commit is. Changes since last review: adding more explanatory text, making links clickable, reorganizing.

hebasto:
@MarcoFalke

Do you mind looking into this PR as it resolves your https://github.com/bitcoin/bitcoin/issues/18643?

laanwj:
Code review ACK b8e5d0d3fe3386807d47f50d13ac34fcd2a538fd

PR #271, opened 2021-04-05
Don't clear console prompt when font resizing

### LINKED ISSUE #269: Don't Clear Console Prompt When Resizing Text
Any text left in the `Console Window` prompt will get cleared upon clicking on any of the resize buttons. There is no reason to clear what is in there, and it can be surprising behavior for a user.

Example of when this can be an annoyance: You have a long command written out but not yet completed. You know the output will be large, so you resize text to be smaller. The command you were writing out gets cleared, and you have to start over.

Below are screenshots showing this behavior:

| Have Text in Prompt      | Click Resize Button |
| ----------------------- | ------------------- |
| have-text | resize |

jarolrod:
On master, a console resize event will clear the prompt. To fix this, we store the content of the prompt and re-set it upon a resize. This preserves the prompt text throughout resizes. The text will still clear when you click the clear button, as it should.

**Master**

| Before Resize      | After Resize |
| ----------------- | ------------ |
|  master-beforeresize  |  master-afterresize |

**PR**
| Before Resize      | After Resize |
| ----------------- | ------------ |
| pr-beforeresize | pr-afterresize |

Closes #269

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -494,7 +494,7 @@ RPCConsole::RPCConsole(interfaces::Node& node, const PlatformStyle *_platformSty
     ui->lineEdit->setMaxLength(16 * 1024 * 1024);
     ui->messagesWidget->installEventFilter(this);
 
-    connect(ui->clearButton, &QPushButton::clicked, [this] { clear(/* clearHistory */ false); });
+    connect(ui->clearButton, &QPushButton::clicked, [this] { clear(); });
     connect(ui->fontBiggerButton, &QPushButton::clicked, this, &RPCConsole::fontBigger);
     connect(ui->fontSmallerButton, &QPushButton::clicked, this, &RPCConsole::fontSmaller);
     connect(ui->btnClearTrafficGraph, &QPushButton::clicked, ui->trafficGraph, &TrafficGraphWidget::clear);
@@ -776,19 +776,14 @@ void RPCConsole::setFontSize(int newSize)
 
     // clear console (reset icon sizes, default stylesheet) and re-add the content
     float oldPosFactor = 1.0 / ui->messagesWidget->verticalScrollBar()->maximum() * ui->messagesWidget->verticalScrollBar()->value();
-    clear(false);
+    clear();
     ui->messagesWidget->setHtml(str);
     ui->messagesWidget->verticalScrollBar()->setValue(oldPosFactor * ui->messagesWidget->verticalScrollBar()->maximum());
 }
 
-void RPCConsole::clear(bool clearHistory)
+void RPCConsole::clear()
 {
     ui->messagesWidget->clear();
-    if(clearHistory)
-    {
-        history.clear();
-        historyPtr = 0;
-    }
     ui->lineEdit->clear();
     ui->lineEdit->setFocus();
 
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -494,7 +494,7 @@ RPCConsole::RPCConsole(interfaces::Node& node, const PlatformStyle *_platformSty
     ui->lineEdit->setMaxLength(16 * 1024 * 1024);
     ui->messagesWidget->installEventFilter(this);
 
-    connect(ui->clearButton, &QPushButton::clicked, this, &RPCConsole::clear);
+    connect(ui->clearButton, &QPushButton::clicked, [this] { clear(/* clearHistory */ false); });
     connect(ui->fontBiggerButton, &QPushButton::clicked, this, &RPCConsole::fontBigger);
     connect(ui->fontSmallerButton, &QPushButton::clicked, this, &RPCConsole::fontSmaller);
     connect(ui->btnClearTrafficGraph, &QPushButton::clicked, ui->trafficGraph, &TrafficGraphWidget::clear);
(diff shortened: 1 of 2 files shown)


jonatack (commented):
Tested ACK dcd7a8c3db4a037c1085616b136267e7059980f0

Naming nit, I'm not sure this applies to the GUI but per doc/developer-notes.md:

- **Symbol naming conventions**. These are preferred in new code, but are not
required when doing so would need changes to significant pieces of existing
code.
  - Variable (including function arguments) and namespace names are all lowercase and may use `_` to separate words (snake_case).

jarolrod:
> Naming nit, I'm not sure this applies to the GUI but per doc/developer-notes.md:

@jonatack Good point, I was following the style of the surrounding code as well as the [Qt Coding Style](https://wiki.qt.io/Qt_Coding_Style). @hebasto what do you think? This is a good topic for the [Qt Dev Notes](https://github.com/bitcoin-core/bitcoin-devwiki/wiki/Developer-Notes-for-Qt-Code)

hebasto:
Concept ACK.

> This preserves the prompt text throughout resizes.

Why not to implement this idea directly: https://github.com/hebasto/gui/commits/pr271-alt ?

promag (commented):
Concept ACK.

promag:
> Concept ACK.
> 
> > This preserves the prompt text throughout resizes.
> 
> Why not to implement this idea directly: https://github.com/hebasto/gui/commits/pr271-alt ?

I prefer this approach.

Bosch-0:
This doesn't appear to be an issue on Windows 10

hebasto:
@Bosch-0 
> This doesn't appear to be an issue on Windows 10

Do you mind elaborating?

Testing [bitcoin-0.21.1rc1](https://bitcoincore.org/bin/bitcoin-core-0.21.1/test.rc1/bitcoin-0.21.1rc1-win64.zip) (not master) on Windows Pro 20H2 (build 19042.928), and clicking on the "A+" or "A-" buttons indeed clears the console prompt. I see this PR as a UX improvement, no?

Bosch-0:
Apologies I originally misread, thought this was resizing the window not font. Yes I can replicate this, will test this PR today

jarolrod:
updated from dcd7a8c -> 7962e0d

Changes:
- use @hebasto's version from branch [pr271-alt]. This version is the superior implementation.

hebasto (approved):
ACK 7962e0dde8bbd0fa3dd702e2224774f1edaadcb6

src/qt/rpcconsole.cpp
```diff
@@ -776,20 +776,15 @@ void RPCConsole::setFontSize(int newSize)
 
     // clear console (reset icon sizes, default stylesheet) and re-add the content
     float oldPosFactor = 1.0 / ui->messagesWidget->verticalScrollBar()->maximum() * ui->messagesWidget->verticalScrollBar()->value();
-    clear(false);
+    clear(/* keep_prompt */ true);
```
Talkless, line 779: Offtopic: I feel it would be better to use enums, avoiding `/* comments explaining what are we dogin */` I've seen here and there. `clear(CearMode::KeePrompt)` would be self-docummenting, less error prone. But again, that's offtopic for this PR.

Talkless (approved):
tACK d2cc3390054616c73f72a59f864700f6de14067b, tested on Debian Sid with Qt 5.15.2

hebasto:
@Talkless 
> tACK [d2cc339](https://github.com/bitcoin-core/gui/commit/d2cc3390054616c73f72a59f864700f6de14067b), tested on Debian Sid with Qt 5.15.2

Could you use the top commit hash which is 7962e0dde8bbd0fa3dd702e2224774f1edaadcb6 for the current branch state? Just edit your comment will be ok :)

src/qt/rpcconsole.cpp
hebasto: The relevant discussion is here https://github.com/bitcoin/bitcoin/issues/21684.

Talkless:
> @Talkless
> 
> > tACK [d2cc339](https://github.com/bitcoin-core/gui/commit/d2cc3390054616c73f72a59f864700f6de14067b), tested on Debian Sid with Qt 5.15.2
> 
> Could you use the top commit hash which is [7962e0d](https://github.com/bitcoin-core/gui/commit/7962e0dde8bbd0fa3dd702e2224774f1edaadcb6) for the current branch state? Just edit your comment will be ok :)

Ouch, sorry, fixed.

src/qt/rpcconsole.cpp
```diff
@@ -776,20 +776,15 @@ void RPCConsole::setFontSize(int newSize)
...
     ui->messagesWidget->setHtml(str);
     ui->messagesWidget->verticalScrollBar()->setValue(oldPosFactor * ui->messagesWidget->verticalScrollBar()->maximum());
 }
 
-void RPCConsole::clear(bool clearHistory)
```
luke-jr, line 784: In the future, please keep the existing param so calls to eg `clear(true)` don't silently change behaviour...

hebasto: I don't follow... In d2cc3390054616c73f72a59f864700f6de14067b `RPCConsole::clear()` has no parameters.
Why keep the existing useless param?

luke-jr: In case another PR were to use the bool param, this change would silently result in different behaviour when the two are merged together. Using an enum class would also solve it.

hebasto: > Using an enum class would also solve it.

Agree. https://github.com/bitcoin/bitcoin/issues/21684

PR #277, opened 2021-04-10
Do not use QClipboard::Selection on Windows and macOS.

### LINKED ISSUE #258: GUI: Data set on unsupported clipboard mode. QMimeData object will be deleted.
Right-mouse-button clicking on any transaction in Transactions tab and copying address, label, and so on, results in the following message in the debug.log:

`GUI: Data set on unsupported clipboard mode. QMimeData object will be deleted.`

The copied content is actually getting copied to the clipboard, so no problem with the feature itself, just with the log getting filled with unnecessary messages. I don't know if the issue was present in previous versions or if it's system specific, just noticed that today when I was observing debug.log live when submitting a new transaction.

Bitcoin Core v0.21.0
MacOS 11.2.3

hebasto:
Windows and macOS do [not support](https://doc.qt.io/qt-5/qclipboard.html#notes-for-windows-and-macos-users) the global mouse selection.

Fixes #258.

### DIFF
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -628,8 +628,11 @@ bool SetStartOnSystemStartup(bool fAutoStart) { return false; }
 
 void setClipboard(const QString& str)
 {
-    QApplication::clipboard()->setText(str, QClipboard::Clipboard);
-    QApplication::clipboard()->setText(str, QClipboard::Selection);
+    QClipboard* clipboard = QApplication::clipboard();
+    clipboard->setText(str, QClipboard::Clipboard);
+    if (clipboard->supportsSelection()) {
+        clipboard->setText(str, QClipboard::Selection);
+    }
 }
 
 fs::path qstringToBoostPath(const QString &path)


src/qt/guiutil.cpp
```diff
@@ -629,7 +629,10 @@ bool SetStartOnSystemStartup(bool fAutoStart) { return false; }
 void setClipboard(const QString& str)
 {
     QApplication::clipboard()->setText(str, QClipboard::Clipboard);
+#if !defined(Q_OS_WIN) && !defined(Q_OS_MAC)
```
promag, line 632: Use `QClipboard::supportsSelection()` instead.

promag (commented):
Concept ACK.

promag (commented):
Code review ACK 7f3a5980c1d54988a707b961fd2ef647cebb4c5b.

src/qt/guiutil.cpp
hebasto: Thanks!

jarolrod (commented):
Concept ACK, I think this change is ok

But I cannot replicate the issue described in #258 when compiling master with Qt 5.15.2 or building master with depends which uses Qt 5.12.10

I **can** replicate the issue when building the 0.21 tag with depends (Qt 5.9.8). I think the issue linked is not a problem anymore going forward

hebasto:
@streamofstars Does this work for you?

MarcoFalke:
Is this for backport to 0.21?

hebasto:
> Is this for backport to 0.21?

Yes, I guess.

hebasto:
Backported in https://github.com/bitcoin/bitcoin/pull/22427.

PR #286, opened 2021-04-21
qt: Bump transifex slug for 22.x

laanwj:
Opening the 22.x translations early because of experimentation with the new xliff translations format. So change the slug so that the `tx` tool will fetch the right translation.

In this context, change `file_filter` to use `xlf` as well as the files pulled with `tx pull` are that format now. The setting only affects the naming not the format of the files.

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -1,7 +1,7 @@
 [main]
 host = https://www.transifex.com
 
-[bitcoin.qt-translation-021x]
-file_filter = src/qt/locale/bitcoin_<lang>.ts
+[bitcoin.qt-translation-022x]
+file_filter = src/qt/locale/bitcoin_<lang>.xlf
 source_file = src/qt/locale/bitcoin_en.xlf
 source_lang = en


hebasto:
Concept ACK.

Are additional settings required on Transifex side? Or it will open 22.x translations automatically?

laanwj:
> Are additional settings required on Transifex side? Or it will open 22.x translations automatically?

I've already done that (as you've noticed).

laanwj:
After this I'll add an additional step to the [fetch script](https://github.com/bitcoin-core/bitcoin-maintainer-tools/blob/master/update-translations.py) to use `lconvert` to convert the fetched `.xlf` files to `.ts`. I think that is all that is necessary.

hebasto (approved):
ACK 417305991a0573484f4aa3820103d8b991cb8f81, I have reviewed the code and it looks OK, I agree it can be merged. All changes are in accordance with the Transifex [docs](https://docs.transifex.com/client/client-configuration).

Issue #287, opened 2021-04-21
Translation using XLIFF file


hebasto:
TL;DR

1. _Use_ [Qt Translator Comments](https://doc.qt.io/qt-5/i18n-source-translation.html#translator-comments) to annotate [`tr()`](https://doc.qt.io/qt-5/qobject.html#tr) calls
2. Do _not_ use disambiguation strings in [`tr()`](https://doc.qt.io/qt-5/qobject.html#tr) calls, except in item 3
3. If two or more identical strings have _different_ meanings and could be translated in different ways, _use_ disambiguation strings

---

Since https://github.com/bitcoin/bitcoin/pull/21694 is merged there is an opportunity for developers to provide an exhaustive context for translatable strings to translators who are working via the [Transifex](https://www.transifex.com/bitcoin/bitcoin/dashboard/).

### Strings in Qt code

Using a Qt **TS** translation file (pre-21694), v0.21.x and earlier:
![DeepinScreenshot_select-area_20210421165407](https://user-images.githubusercontent.com/32963518/115598009-9a4d5500-a2e2-11eb-8053-6d43705a09e4.png)

Using a **XLIFF** translation file (post-21694), v22.x:
![DeepinScreenshot_select-area_20210421165641](https://user-images.githubusercontent.com/32963518/115598350-fd3eec00-a2e2-11eb-8723-5ec33fc1c7d7.png)

The added "DEVELOPER NOTES" are the main means to provide context to translators in the Transifex editor.
See:
- https://docs.transifex.com/getting-started-as-a-localization-manager/4-provide-context#developer-notes-&-instructions
- https://docs.transifex.com/translation/tools-in-the-editor#string-instruction-and-developer-notes

To provide such notes in C++ code developers must use [Qt Translator Comments](https://doc.qt.io/qt-5/i18n-source-translation.html#translator-comments) that annotate [`tr()`](https://doc.qt.io/qt-5/qobject.html#tr) calls: https://github.com/bitcoin-core/gui/blob/e16f8720dca2de1040478968c9f3ca07644a04b7/src/qt/addressbookpage.cpp#L298-L299

Please note, that despite documented support for `<context-group>` and `<context>` XML elements of XLIFF file by Transifex, disambiguation strings, i.e., the second argument of `tr()` call, are _not_  rendered in Transifex editor. Therefore, they add no value to translation process.

[...]

PR #289, opened 2021-04-23
Add "Direction" column to peers tab

jonatack:
Screenshot from 2021-04-24 13-15-54

This was initially proposed in #179 and saw an enthusiastic reception with several ACKs. One contributor was a weak NACK due to the horizontal space and so this was temporarily dropped, but since then #202 (and soon #256 and #293) resolve the issue. Moreover, the current state of displaying only a connection `Type` column without a connection `Direction` column is confusing (e.g. https://github.com/bitcoin/bitcoin/issues/21747#issuecomment-825072559), particularly as both the direction and the type are displayed in the peer details area with the `Direction/Type` row.

This patch fixes it and completes the initial proposal in #179 by adding a `Direction` column, making the peers tab the same as the `Direction/Type` row in the peer details and the direction and type columns in our other user-facing peer connections table in `-netinfo`.

Users can now sort the peers table by direction. The default sort is set to inbound, then outbound.

This patch also makes the direction and type translations the same for the 3 places where they are used (the peers column, the peer details row, and its tooltip), adds translator comments to link them together, and adds QStringBuilder fast QString concatenation with the `%` operator.

### DIFF
--- a/src/qt/addressbookpage.cpp
+++ b/src/qt/addressbookpage.cpp
@@ -16,6 +16,7 @@
 #include <qt/platformstyle.h>
 
 #include <QIcon>
+#include <QLatin1String>
 #include <QMenu>
 #include <QMessageBox>
 #include <QSortFilterProxyModel>
--- a/src/qt/peertablemodel.cpp
+++ b/src/qt/peertablemodel.cpp
@@ -113,8 +113,10 @@ QVariant PeerTableModel::data(const QModelIndex &index, int role) const
         case NetNodeId:
             return (qint64)rec->nodeStats.nodeid;
         case Address:
-            // prepend to peer address down-arrow symbol for inbound connection and up-arrow for outbound connection
-            return QString(rec->nodeStats.fInbound ? "↓ " : "↑ ") + QString::fromStdString(rec->nodeStats.addrName);
+            return QString::fromStdString(rec->nodeStats.addrName);
+        case Direction:
+            //: Connection direction. Also used in ConnectionTypeToQString() and CONNECTION_TYPE_DOC.
+            return QString(rec->nodeStats.fInbound ? tr("Inbound") : tr("Outbound"));
         case ConnectionType:
             return GUIUtil::ConnectionTypeToQString(rec->nodeStats.m_conn_type, /* prepend_direction */ false);
         case Network:
@@ -134,6 +136,7 @@ QVariant PeerTableModel::data(const QModelIndex &index, int role) const
         case NetNodeId:
         case Address:
             return {};
+        case Direction:
         case ConnectionType:
         case Network:
             return QVariant(Qt::AlignCenter);
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -37,12 +37,14 @@
 #include <QDateTime>
 #include <QFont>
 #include <QKeyEvent>
+#include <QLatin1String>
 #include <QMenu>
 #include <QMessageBox>
 #include <QScreen>
 #include <QScrollBar>
 #include <QSettings>
 #include <QString>
+#include <QStringBuilder>
 #include <QStringList>
 #include <QTime>
 #include <QTimer>
@@ -52,6 +54,9 @@ const int CONSOLE_HISTORY = 50;
 const int INITIAL_TRAFFIC_GRAPH_MINS = 30;
 const QSize FONT_RANGE(4, 40);
 const char fontSizeSettingsKey[] = "consoleFontSize";
+
+const QLatin1String COLON_SPACE{": "};
+const QLatin1String SPACE{" "};
 
 const struct {
     const char *url;
@@ -463,16 +468,18 @@ RPCConsole::RPCConsole(interfaces::Node& node, const PlatformStyle *_platformSty
 
     constexpr QChar nonbreaking_hyphen(8209);
     const std::vector<QString> CONNECTION_TYPE_DOC{
-        tr("Inbound: initiated by peer"),
-        tr("Outbound Full Relay: default"),
-        tr("Outbound Block Relay: does not relay transactions or addresses"),
-        tr("Outbound Manual: added using RPC %1 or %2/%3 configuration options")
+        //: The connection direction (Inbound/Outbound) is also used in ConnectionTypeToQString() and PeerTableModel::data().
+        //: The connection types (Full Relay, Block Relay, Manual, Feeler, Address Fetch) are also used in ConnectionTypeToQString().
+        tr("Inbound") % COLON_SPACE % tr("initiated by peer"),
+        tr("Outbound") % SPACE % tr("Full Relay") % COLON_SPACE % tr("default"),
+        tr("Outbound") % SPACE % tr("Block Relay") % COLON_SPACE % tr("does not relay transactions or addresses"),
+        tr("Outbound") % SPACE % tr("Manual") % COLON_SPACE % tr("added using RPC %1 or %2/%3 configuration options")
             .arg("addnode")
-            .arg(QString(nonbreaking_hyphen) + "addnode")
-            .arg(QString(nonbreaking_hyphen) + "connect"),
-        tr("Outbound Feeler: short-lived, for testing addresses"),
-        tr("Outbound Address Fetch: short-lived, for soliciting addresses")};
-    const QString list{"<ul><li>" + Join(CONNECTION_TYPE_DOC, QString("</li><li>")) + "</li></ul>"};
+            .arg(QString(nonbreaking_hyphen) % "addnode")
+            .arg(QString(nonbreaking_hyphen) % "connect"),
+        tr("Outbound") % SPACE % tr("Feeler") % COLON_SPACE % tr("short-lived, for testing addresses"),
+        tr("Outbound") % SPACE % tr("Address Fetch") % COLON_SPACE % tr("short-lived, for soliciting addresses")};
+    const QString list{"<ul><li>" % Join(CONNECTION_TYPE_DOC, QString("</li><li>")) % "</li></ul>"};
     ui->peerConnectionTypeLabel->setToolTip(ui->peerConnectionTypeLabel->toolTip().arg(list));
     const QString hb_list{"<ul><li>\""
         + ts.to + "\" – " + tr("we selected the peer for high bandwidth relay") + "</li><li>\""
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -482,9 +482,9 @@ RPCConsole::RPCConsole(interfaces::Node& node, const PlatformStyle *_platformSty
     const QString list{"<ul><li>" % Join(CONNECTION_TYPE_DOC, QString("</li><li>")) % "</li></ul>"};
     ui->peerConnectionTypeLabel->setToolTip(ui->peerConnectionTypeLabel->toolTip().arg(list));
     const QString hb_list{"<ul><li>\""
-        + ts.to + "\" – " + tr("we selected the peer for high bandwidth relay") + "</li><li>\""
-        + ts.from + "\" – " + tr("the peer selected us for high bandwidth relay") + "</li><li>\""
-        + ts.no + "\" – " + tr("no high bandwidth relay selected") + "</li></ul>"};
+        % ts.to % "\" – " % tr("we selected the peer for high bandwidth relay") % "</li><li>\""
+        % ts.from % "\" – " % tr("the peer selected us for high bandwidth relay") % "</li><li>\""
+        % ts.no % "\" – " % tr("no high bandwidth relay selected") % "</li></ul>"};
     ui->peerHighBandwidthLabel->setToolTip(ui->peerHighBandwidthLabel->toolTip().arg(hb_list));
     ui->dataDir->setToolTip(ui->dataDir->toolTip().arg(QString(nonbreaking_hyphen) + "datadir"));
     ui->blocksDir->setToolTip(ui->blocksDir->toolTip().arg(QString(nonbreaking_hyphen) + "blocksdir"));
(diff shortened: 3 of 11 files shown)


jarolrod (commented):
Concept ACK

I think this is an improvement over the current little arrows next to the address. But, this improvement is dependent on getting translations done. 

What if we kept the arrow next to the `Inbound` and `Outbound` text. That way, if there is no translation available, the arrows provide some context to go on.

src/qt/peertablemodel.cpp
```diff
@@ -161,8 +163,9 @@ QVariant PeerTableModel::data(const QModelIndex &index, int role) const
         case NetNodeId:
             return (qint64)rec->nodeStats.nodeid;
         case Address:
-            // prepend to peer address down-arrow symbol for inbound connection and up-arrow for outbound connection
-            return QString(rec->nodeStats.fInbound ? "↓ " : "↑ ") + QString::fromStdString(rec->nodeStats.addrName);
+            return QString::fromStdString(rec->nodeStats.addrName);
+        case Direction:
+            return QString(rec->nodeStats.fInbound ? "Inbound" : "Outbound");
```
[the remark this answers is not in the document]
hebasto: Also we could add translator comments now (see #287 for details).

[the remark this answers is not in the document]
jonatack: Done.

jonatack:
> What if we kept the arrow next to the `Inbound` and `Outbound` text. That way, if there is no translation available, the arrows provide some context to go on.

Thanks for reviewing! Pushed with this change. WDYT?

before

Screenshot from 2021-04-24 13-15-54

after

Screenshot from 2021-04-24 12-44-27

jonatack:
> What if we kept the arrow next to the `Inbound` and `Outbound` text. That way, if there is no translation available, the arrows provide some context to go on.

Thanks for reviewing! Pushed with this change. WDYT? If it was just me I would probably leave out the arrow as it's cleaner to my eye, but I don't have a strong opinion. 

before

Screenshot from 2021-04-24 13-15-51

after

Screenshot from 2021-04-24 13-51-46

RandyMcMillan:
@jonatack - In this previous PR #135 I did something similar...



The goal was to actually embed unicode arrows in the column and not use text for direction at all.

This may be better visually and save column space.

The column sorting function should sort by unicode value with little to no additional work.

I just wanted to mention that since this PR is focusing on the direction column.

jonatack:
Thanks -- yes, I know but don't prefer a column with just arrows.

More objectively, there is an issue described in the PR description above: the Inbound/Outbound direction is displayed in the peer details area (the `Direction/Type` row) but not in the peers windows itself, which is confusing to users. Now that space is less of an issue thanks to #202 and #256, we can complete what was started in #179 and properly display the Direction and the Type next to each other as we are already doing in the peer details.

RandyMcMillan:
#293 should fit with #202 #256 & #179 as well...

What do you think leave the Ampersands or take them out?

jonatack:
> What do you think leave the Ampersands or take them out?

The current space-ampersand-space, or comma-space, seem like good separators.

jonatack:
Update: after adding the arrows yesterday, I removed them today because the `Direction` and `Type` columns should display the same thing as the `Direction/Type` details row. In a new second commit, I made the direction and type translations the same for the three places where they are used (peers column, peer details row, and its tooltip) and added translator comments to link them together. It's not clear to me, however, if the translator comments should refer to places in the codebase or to where the translations are displayed to users.

src/qt/rpcconsole.cpp
```diff
@@ -462,15 +462,17 @@ RPCConsole::RPCConsole(interfaces::Node& node, const PlatformStyle *_platformSty
 
     constexpr QChar nonbreaking_hyphen(8209);
     const std::vector<QString> CONNECTION_TYPE_DOC{
-        tr("Inbound: initiated by peer"),
-        tr("Outbound Full Relay: default"),
-        tr("Outbound Block Relay: does not relay transactions or addresses"),
-        tr("Outbound Manual: added using RPC %1 or %2/%3 configuration options")
+        //: The connection direction (Inbound/Outbound) is also used in ConnectionTypeToQString() and PeerTableModel::data().
+        //: The connection types (Full Relay, Block Relay, Manual, Feeler, Address Fetch) are also used in ConnectionTypeToQString().
+        tr("Inbound") + ": " + tr("initiated by peer"),
```
Talkless, line 467: nit: Now there's lot's of string concatenations, maybe worth using QStringBuilder's `%` and wrapping " " and ": " with `QLatin1String` ?

Talkless (approved):
tACK d04ebdaab74493bb0cba79fbb81b8225e8bbb8de, tested on Debian Sid with Qt 5.15.2.

src/qt/rpcconsole.cpp
jonatack: Thanks @Talkless. Added QLatin1String constants in 73cf0fdc8edacf9294f9facc8248fb078298b2d7. Let me know if you think using `%` would still improve the diff and what you'd suggest.

src/qt/addressbookpage.cpp
```diff
@@ -16,6 +16,7 @@
 #include <qt/platformstyle.h>
 
 #include <QIcon>
+#include <QLatin1String>
```
Talkless, line 19: Uhm, I'm a bit lost, why do we have QLatin1String is included a lot in this PR?

src/qt/rpcconsole.cpp
```diff
@@ -52,6 +54,9 @@ const int INITIAL_TRAFFIC_GRAPH_MINS = 30;
 const QSize FONT_RANGE(4, 40);
 const char fontSizeSettingsKey[] = "consoleFontSize";
 
+const QLatin1String COLON_SPACE{": "};
```
Talkless, line 57: Interesting idea. It actually could be `constexpr`.

src/qt/rpcconsole.cpp
```diff
@@ -36,12 +36,14 @@
...
 #include <QFont>
 #include <QKeyEvent>
+#include <QLatin1String>
 #include <QMenu>
 #include <QMessageBox>
 #include <QScreen>
 #include <QScrollBar>
 #include <QSettings>
 #include <QString>
+#include <QStringBuilder>
```
Talkless, line 46: Not sure if it helps only by including (without using `operator %`), unless whole bitcoin-qt would be built with `QT_USE_QSTRINGBUILDER`.

jonatack: Added because this file already uses `%`.

src/qt/addressbookpage.cpp
jonatack: Added the header where it is used but header is missing.

src/qt/rpcconsole.cpp
jonatack: That was the first thing I tried but the compiler doesn't like it.

src/qt/rpcconsole.cpp
Talkless: I see, QLatin1String and QStringBuilder included because where simply _missing_, not because of new usage. New usage would be nice though, but I don't want to be too-QString-nazi..

src/qt/rpcconsole.cpp
Talkless: That's strange, QLatin1String's constructors are constexpr-marked since at least 5.3: https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/tools/qstring.h?h=5.3#n89

Sure that wasn't type or something? What kind of message was that? And what compiler?

jonatack: I'm using Qt version 5.15.2 and according to doc/dependencies.md Bitcoin Core uses 5.12.10 and supports back to 5.9.5.

jonatack:
Rebased due to the merge of #18. Thanks for the reviews @Talkless!

src/qt/rpcconsole.cpp
jonatack: This seemed relevant: https://stackoverflow.com/questions/56201976/qt-vs-constexpr-string-literal

Clang 9 and GCC 10.2.1:
```
  CXX      qt/libbitcoinqt_a-rpcconsole.o
qt/rpcconsole.cpp:58:32: error: constexpr variable 'COLON_SPACE' must be initialized by a constant expression
static constexpr QLatin1String COLON_SPACE{": "};
                               ^~~~~~~~~~~~~~~~~
/usr/include/x86_64-linux-gnu/qt5/QtCore/qstring.h:91:93: note: non-constexpr function 'strlen' cannot be used in a constant expression
    Q_DECL_CONSTEXPR inline explicit QLatin1String(const char *s) noexcept : m_size(s ? int(strlen(s)) : 0), m_data(s) {}
                                                                                            ^
qt/rpcconsole.cpp:58:32: note: in call to 'QLatin1String(&": "[0])'
static constexpr QLatin1String COLON_SPACE{": "};
                               ^
qt/rpcconsole.cpp:59:32: error: constexpr variable 'SPACE' must be initialized by a constant expression
static constexpr QLatin1String SPACE{" "};
                               ^~~~~~~~~~
/usr/include/x86_64-linux-gnu/qt5/QtCore/qstring.h:91:93: note: non-constexpr function 'strlen' cannot be used in a constant expression
    Q_DECL_CONSTEXPR inline explicit QLatin1String(const char *s) noexcept : m_size(s ? int(strlen(s)) : 0), m_data(s) {}
                                                                                            ^
qt/rpcconsole.cpp:59:32: note: in call to 'QLatin1String(&" "[0])'
static constexpr QLatin1String SPACE{" "};
                               ^
2 errors generated.
```

src/qt/rpcconsole.cpp
jonatack: Thanks for the nudge, @Talkless. Looked up https://wiki.qt.io/Using_QString_Effectively and updated. Please let me know what you think.

prayank23:
Concept ACK

src/qt/peertablemodel.cpp
```diff
@@ -113,8 +113,10 @@ QVariant PeerTableModel::data(const QModelIndex &index, int role) const
         case NetNodeId:
             return (qint64)rec->nodeStats.nodeid;
         case Address:
-            // prepend to peer address down-arrow symbol for inbound connection and up-arrow for outbound connection
-            return QString(rec->nodeStats.fInbound ? "↓ " : "↑ ") + QString::fromStdString(rec->nodeStats.addrName);
+            return QString::fromStdString(rec->nodeStats.addrName);
+        case Direction:
+            //: Connection direction. Also used in ConnectionTypeToQString() and CONNECTION_TYPE_DOC.
+            return QString(rec->nodeStats.fInbound ? tr("Inbound") : tr("Outbound"));
```
prayank23, line 119: Can we use `IN` and `OUT` in caps instead of Inbound and Outbound?

prayank23:
> Screenshot from 2021-04-24 13-15-54

How do you have 3 Block Relay connections in this screenshot? Isn't [MAX 2](https://github.com/bitcoin/bitcoin/blob/66fd3b28e85c167f3955b5603496daf8d91abcad/src/net.h#L68)?

jarolrod (commented):
ACK 2bf3f5aee014ab4e0ff72e6bdbb33ada02dec314

This patch works well and is a UX improvement. Tested on macOS 11.3 Qt 5.15.2

Thanks for working on this and adding in translator comments.

Screen Shot 2021-04-30 at 2 59 31 AM

src/qt/peertablemodel.cpp
jarolrod: Here is how that would look, not a bad idea. But, a) I think this change is ok as is b) would like to hear other opinions.

Screen Shot 2021-04-30 at 3 09 34 AM

jonatack: The goal is to unify the peers tab Direction and Type columns with the Direction/Type row in the peer details sidebar, so if the side bar displays "Outbound Full Relay", the main table also displays "Outbound Full Relay" with the ability to sort by either Direction or Type.

jonatack: I tried to communicate it in the PR description.

jonatack:
> How do you have 3 Block Relay connections in this screenshot? Isn't [MAX 2](https://github.com/bitcoin/bitcoin/blob/66fd3b28e85c167f3955b5603496daf8d91abcad/src/net.h#L68)?

We very often have a short-lived third extra block relay connection since the merge of https://github.com/bitcoin/bitcoin/pull/19858. 

See this open PR that updates our docs about this: https://github.com/bitcoin/bitcoin/pull/21709/files.

src/qt/peertablemodel.cpp
jonatack: (I also tried "In/Out" while working on this PR, and the horizontal space isn't reduced unless the column header is "Dir" instead of "Direction". But the other headers are full words.)

hebasto:
Concept ACK on removing direction info from the "Address" column.

@luke-jr wrt your comment https://github.com/bitcoin-core/gui/pull/179#issuecomment-762417066 and new [circumstances](https://github.com/bitcoin-core/gui/pull/289#issue-622087016), did you change your mind?

jonatack:
Up for grabs.

prayank23:
I can work on this. But I am not sure why is this closed and "Up for grabs" because @jonatack already made changes required and reviews looked positive as well.

hebasto:
@prayank23 
> I can work on this.

You're also welcome to join the `##bitcoin-core-gui` (with two #) channel on IRC :)

jonatack:
Hi @prayank23, sorry for the confusion.  I opened this to finish what was started with #179 but decided to not spend further time on GUI/design/QT/UI/translation things.  Everything takes time to do well, and discussing each word, arrow, indentation, translation, and consistency question and explanation until everyone is happy is completely normal, but it takes bandwidth in an area that isn't my focus.  Cheers.

jarolrod:
@hebasto this is no longer up for grabs as it was picked up in #317

PR #290, opened 2021-04-23
Peers window `Peer id` improvements

jonatack:
This patch:

- renames the peers tab column header from `Peer Id` to `Peer` to allow resizing the column more tightly (this will be particularly useful after #256) and does the same for the peer details area.

- center-aligns the entries in the `Peer` column to align them under the column header rather than to the left (I also tried right-alignment but the values are unreadably close to the `Address` values and it seems more esthetic to be aligned under the header :file_cabinet:)

Screenshot from 2021-04-23 16-43-21

### DIFF
--- a/src/qt/peertablemodel.cpp
+++ b/src/qt/peertablemodel.cpp
@@ -179,9 +179,9 @@ QVariant PeerTableModel::data(const QModelIndex &index, int role) const
         assert(false);
     } else if (role == Qt::TextAlignmentRole) {
         switch (column) {
-        case NetNodeId:
         case Address:
             return {};
+        case NetNodeId:
         case ConnectionType:
         case Network:
             return QVariant(Qt::AlignCenter);
--- a/src/qt/peertablemodel.h
+++ b/src/qt/peertablemodel.h
@@ -88,7 +88,7 @@ public Q_SLOTS:
 
 private:
     interfaces::Node& m_node;
-    const QStringList columns{tr("Peer Id"), tr("Address"), tr("Type"), tr("Network"), tr("Ping"), tr("Sent"), tr("Received"), tr("User Agent")};
+    const QStringList columns{tr("Peer"), tr("Address"), tr("Type"), tr("Network"), tr("Ping"), tr("Sent"), tr("Received"), tr("User Agent")};
     std::unique_ptr<PeerTablePriv> priv;
     QTimer *timer;
 };
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -1086,7 +1086,7 @@ void RPCConsole::updateDetailWidget()
     const auto stats = selected_peers.first().data(PeerTableModel::StatsRole).value<CNodeCombinedStats*>();
     // update the detail ui with latest node information
     QString peerAddrDetails(QString::fromStdString(stats->nodeStats.addrName) + " ");
-    peerAddrDetails += tr("(peer id: %1)").arg(QString::number(stats->nodeStats.nodeid));
+    peerAddrDetails += tr("(peer: %1)").arg(QString::number(stats->nodeStats.nodeid));
     if (!stats->nodeStats.addrLocal.empty())
         peerAddrDetails += "<br />" + tr("via %1").arg(QString::fromStdString(stats->nodeStats.addrLocal));
     ui->peerHeading->setText(peerAddrDetails);


RandyMcMillan:
Tested on MacOS 10.14.6 (18G103)

Screen Shot 2021-04-23 at 2 39 25 PM

Talkless (approved):
tACK e14d79f11eaf208b81fc90216304e1e0ea3d673e, tested on Debian Sid with Qt 5.15.2.

Talkless:
> renaming to Id instead of Peer would work also but I'm not sure about translateability

What are the concerns exactly?

kristapsk:
> renaming to `Id` instead of `Peer` would work also but I'm not sure about translateability

"Id" seems to be a short for "Identifier" in most of the languages I checked. Are there exceptions for that and languages where "Id" can't be translated to something similar? https://en.wikipedia.org/wiki/ID

jonatack:
"Id" is actually incorrect (I think you are referring to "ID"). Outside of databases and developers, "Id" is a somewhat confusing word to use because it has many meanings. See https://www.thefreedictionary.com/id. The link you gave also lists many definitions. It is also an acronym for many different things. For these reasons, I proposed "Peer" to see what reviewers think.

jonatack:
(Updated the PR description to "ID" instead of "Id".)

src/qt/peertablemodel.cpp
```diff
@@ -179,9 +179,9 @@ QVariant PeerTableModel::data(const QModelIndex &index, int role) const
         assert(false);
     } else if (role == Qt::TextAlignmentRole) {
         switch (column) {
-        case NetNodeId:
         case Address:
             return {};
+        case NetNodeId:
```
jarolrod, line 184: In e14d79f11eaf208b81fc90216304e1e0ea3d673e

Why move the position of `NetNodeId` here? In the other areas of the code, `NetNodeId` always comes before `Address` in our switch statements. I don't think we should move it here.

You can keep it where it originally is and do:

```suggestion
        case NetNodeId:
            return QVariant(Qt::AlignCenter);
```

jonatack: Generally I prefer to propose the smaller, simpler diff. The order doesn't matter here.

jarolrod: It's not much of a difference, plus this makes it easier to reason about. You are explicitly stating that you are making it `AlignCenter`. I'd prefer we keep the same order as our other switch statements, but it's not a strong opinion.

jarolrod:
ACK e14d79f11eaf208b81fc90216304e1e0ea3d673e

% nit about switch statement order

hebasto:
> renaming to Id instead of Peer would work also but I'm not sure about translateability

btw, a good translation of a noun "peer" that means "a participant of the peer-to-peer network" could be a real challenge, at least in some languages :smiley:

MarcoFalke:
Is there a way to tell translators synonyms? "Connection" might be easier to translate.

jonatack:
Heh, interesting.  Should I mention "ID" in a translation comment? (I'm not against using "ID" instead if reviewers prefer but to my mind it reminds me of "papers, please" :police_officer:)

jonatack:
> Is there a way to tell translators synonyms? "Connection" might be easier to translate.

Yes, could mention synonyms like "connection, number" in the translator comment.

hebasto:
One more [opinion](https://github.com/bitcoin-core/gui/pull/179#issuecomment-782962121):
> Removing "Peer" from the "Peer Id" column would allow for more prudent use of space.
> Otherwise it should be translatable. A simple "ID" or "id" would be better. I think all caps "ID" is better imo.

hebasto:
> > Is there a way to tell translators synonyms? "Connection" might be easier to translate.
> 
> Yes, could mention synonyms like "connection, number" in the translator comment.

We could mention the _meaning_ of a term :)

hebasto:
> Heh, interesting. Should I mention "ID" in a translation comment? (I'm not against using "ID" instead if reviewers prefer but to my mind it reminds me of "papers, please" police_officer)

If we concern brevity, that the "Peer" word is redundant because the tab is called "Peers".
If we concern consistency, the `getpeerinfo` output has the `id` field.

Anyway, translator comments are always welcome.

IIRC, there were some comments about aligning of the "Peer Id" tab in this repo, especially when numbers of id go up. But I cannot find them :)

Maybe our fellow designers @Bosch-0 and @GBKS could chare their opinion about this PR?

src/qt/peertablemodel.cpp
hebasto: Always keeping the same order of enum elements in `switch` statement could decrease mental burden for some ppl :)

jonatack:
It looks like "peers" and "peer" are already translated in `qt/locale` (if I understand them correctly).
```
    <message>
        <source>&amp;Peers</source>
        <translation>&amp;Gegenstellen</translation>
    </message>
    <message>
        <source>Banned peers</source>
        <translation>Gesperrte Gegenstellen</translation>
    </message>
    <message>
        <source>Select a peer to view detailed information.</source>
        <translation>Gegenstelle auswählen, um detaillierte Informationen zu erhalten.</translation>
    </message>
```
```
    <message>
        <source>&amp;Peers</source>
        <translation>&amp;Пиры</translation>
    </message>
    <message>
        <source>Banned peers</source>
        <translation>Заблокированные пиры</translation>
    </message>
    <message>
        <source>Select a peer to view detailed information.</source>
        <translation>Выберите пира для просмотра детальной информации.</translation>
    </message>
```

jonatack:
I've decided to close this.

jonatack:
> Maybe our fellow designers @Bosch-0 and @GBKS could share their opinions about this PR?

One thought: designer, UI, and UX can be different specialties, a bit like front end / back end / dev ops / DBA / pentest / embedded, etc. There are designers (say, of icons, logos, typography or illustrations) who wouldn't expect to be asked to make a call about a UI question, and others who might be more comfortable with that.

jonatack:
To clarify, I think this is a good change but it's too much time involved on a very tiny change that may never be merged in a repo that isn't my main focus. That said, I'd be happy for someone more focused on the GUI to fix it.

src/qt/peertablemodel.cpp
Talkless: Ditto, diff is only for "once" (for review), while people will read this switch multiple times in the future, so I'm also for same ordering.

jarolrod:
@hebasto This is no longer up for grabs as the idea of changing `Peer id` to `Peer` was picked up in #311 and the idea of improving the alignment of the `Peer Number` was picked up in #325

PR #291, opened 2021-04-23
Network Graph layout - debug window improvement. 

RandyMcMillan:
#### Rearrange the views for (imo) better layout and debug window resizing options.

Screen Shot 2021-04-23 at 6 29 30 PM

Removing "lblGraphRange" minimumSize enables the view to be reduced to the size of a desktop "widget" :)

https://github.com/bitcoin-core/gui/pull/291/files#diff-a24601363160c5ffd048f45a763e702c988b067f96e48a816c36f855f9820826L681

Screen Shot 2021-04-23 at 6 28 00 PM

The peers tab resizes if the detail view is revealed.

Screen Shot 2021-04-23 at 6 29 15 PM

Screen Shot 2021-04-23 at 6 29 21 PM

Info tab still useful at the "widget" size.

Screen Shot 2021-04-23 at 6 28 11 PM

Console tab still functional at "widget" size.

Screen Shot 2021-04-23 at 6 29 04 PM

### DIFF
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -612,79 +612,32 @@
       <attribute name="title">
        <string>&amp;Network Traffic</string>
       </attribute>
-      <layout class="QHBoxLayout" name="horizontalLayout_3">
+      <layout class="QVBoxLayout" name="verticalLayout_4">
        <item>
-        <layout class="QVBoxLayout" name="verticalLayout_4">
-         <item>
-          <widget class="TrafficGraphWidget" name="trafficGraph" native="true">
-           <property name="sizePolicy">
-            <sizepolicy hsizetype="Expanding" vsizetype="Expanding">
-             <horstretch>0</horstretch>
-             <verstretch>0</verstretch>
-            </sizepolicy>
-           </property>
-          </widget>
-         </item>
-         <item>
-          <layout class="QHBoxLayout" name="horizontalLayout_2">
-           <item>
-            <widget class="QSlider" name="sldGraphRange">
-             <property name="minimum">
-              <number>1</number>
-             </property>
-             <property name="maximum">
-              <number>288</number>
-             </property>
-             <property name="pageStep">
-              <number>12</number>
-             </property>
-             <property name="value">
-              <number>6</number>
-             </property>
-             <property name="orientation">
-              <enum>Qt::Horizontal</enum>
-             </property>
-            </widget>
-           </item>
-           <item>
-            <widget class="QLabel" name="lblGraphRange">
-             <property name="minimumSize">
-              <size>
-               <width>100</width>
-               <height>0</height>
-              </size>
-             </property>
-             <property name="alignment">
-              <set>Qt::AlignCenter</set>
-             </property>
-            </widget>
-           </item>
-           <item>
-            <widget class="QPushButton" name="btnClearTrafficGraph">
-             <property name="text">
-              <string>&amp;Reset</string>
-             </property>
-             <property name="autoDefault">
-              <bool>false</bool>
-             </property>
-            </widget>
-           </item>
-          </layout>
-         </item>
-        </layout>
+        <widget class="TrafficGraphWidget" name="trafficGraph" native="true">
+         <property name="sizePolicy">
+          <sizepolicy hsizetype="Expanding" vsizetype="Expanding">
+           <horstretch>0</horstretch>
+           <verstretch>0</verstretch>
+          </sizepolicy>
+         </property>
+        </widget>
        </item>
        <item>
-        <layout class="QVBoxLayout" name="verticalLayout">
+        <layout class="QHBoxLayout" name="bottomFrame">
          <item>
           <widget class="QGroupBox" name="groupBox">
            <property name="title">
-            <string>Totals</string>
+            <string/>
            </property>
-           <layout class="QVBoxLayout" name="verticalLayout_5">
+           <property name="flat">
+            <bool>true</bool>
+           </property>
+           <layout class="QVBoxLayout" name="verticalLayout_9">
             <item>
-             <layout class="QHBoxLayout" name="horizontalLayout_4">
+             <layout class="QHBoxLayout" name="ReceivedBox">
               <item>
-               <widget class="Line" name="line">
+               <widget class="Line" name="ReceivedLine">
                 <property name="sizePolicy">
                  <sizepolicy hsizetype="Fixed" vsizetype="Fixed">
                   <horstretch>0</horstretch>
@@ -740,7 +693,7 @@
                </widget>
               </item>
               <item>
-               <widget class="QLabel" name="label_16">
+               <widget class="QLabel" name="Received">
                 <property name="text">
                  <string>Received</string>
                 </property>
@@ -762,9 +715,9 @@
              </layout>
             </item>
             <item>
-             <layout class="QHBoxLayout" name="horizontalLayout_5">
+             <layout class="QHBoxLayout" name="SentBox">
               <item>
-               <widget class="Line" name="line_2">
+               <widget class="Line" name="SentLine">
                 <property name="sizePolicy">
                  <sizepolicy hsizetype="Fixed" vsizetype="Fixed">
                   <horstretch>0</horstretch>
@@ -820,7 +773,7 @@
                </widget>
               </item>
               <item>
-               <widget class="QLabel" name="label_17">
+               <widget class="QLabel" name="Sent">
                 <property name="text">
                  <string>Sent</string>
                 </property>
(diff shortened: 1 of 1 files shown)


RandyMcMillan:
Tested with #290 #256

https://github.com/RandyMcMillan/gui/commits/net-graph-layout-290-256

RandyMcMillan:
Tested with #271 

https://github.com/RandyMcMillan/gui/commits/net-graph-layout-271

Bosch-0:
tACK https://github.com/bitcoin-core/gui/pull/291/commits/99f6f43bdd05ed83fca462c0514d6157bad1025b on 

Windows 10. 

Looks good to me, agree this does look slightly more clean. Other windows seem fine. Before / After screens: 

Frame 35

src/qt/forms/debugwindow.ui
```diff
@@ -657,6 +657,181 @@
          </item>
          <item>
           <layout class="QHBoxLayout" name="horizontalLayout_2">
+           <item>
+            <widget class="QGroupBox" name="groupBox_2">
+             <property name="toolTip">
+              <string extracomment="Totals"/>
```
jonatack, line 663: How does a user see this tooltip?

jonatack (commented):
Tested Concept ACK. This seems like a nice design improvement.

Suggest reviewers look at the diff with `colorMoved = dimmed-zebra` and `colorMovedWs = allow-indentation-change`.

Is the slider working correctly? It seems like I lose the traffic history if I move it.

src/qt/forms/debugwindow.ui
RandyMcMillan: thanks - removed

RandyMcMillan: The traffic history always took time to repaint.
I doubt there is an easy fix - it would require caching excessive data in anticipation of a resolution change.
I would lean toward prudent disk space usage over this inconvenience.

jonatack: Thanks, makes sense and is unrelated to this change. Will re-review tomorrow.

Talkless:
Idunno about Network Info tab. It's now inconsistent with Peers tab, where details are shown on the right.
What if we get more Network metrics in the future, not just received/sent? Keeping right pane allows for future growth.

RandyMcMillan:
there is plenty of room for more tabs. 😄

jonatack:
ACK 1cdcdb30ce57ab3847a839dcb71570c2c5ef0e36

Talkless:
> there is plenty of room for more tabs. smile

What I mean is if there's more data series needed (in addition to "Sent", and "Received") on the _same_ graph, not about new tabs.

jarolrod (commented):
tACK 1cdcdb30ce57ab3847a839dcb71570c2c5ef0e36

Screen Shot 2021-04-30 at 12 52 39 AM


Tested on macOS 11.3 Qt 5.15.2. I agree that this is a better use of space.

@RandyMcMillan 
You may want to fix your commit author; it's currently authored by `git`. You may want to change that back to you.

@Talkless 
> What I mean is if there's more data series needed (in addition to "Sent", and "Received") on the same graph, not about new tabs.

That can be figured out when new data series are proposed

hebasto:
Concept ACK. IIUC this is a reincarnation of #90.

```
$ git log -1
commit 1cdcdb30ce57ab3847a839dcb71570c2c5ef0e36 (HEAD -> pr291-0430)
Author: git <git@gits-Air.deepspace.dynalias.org>
Date:   Mon Sep 14 17:21:44 2020 -0400

    Network Graph layout
```
^ this (the authorship) should be fixed before merging.

src/qt/forms/debugwindow.ui
```diff
@@ -678,14 +850,8 @@
...
-             <property name="minimumSize">
-              <size>
-               <width>100</width>
-               <height>0</height>
-              </size>
-             </property>
-             <property name="alignment">
-              <set>Qt::AlignCenter</set>
+             <property name="text">
+              <string>Range</string>
```
hebasto, line 854: Why this change?

This label will be overwritten in https://github.com/bitcoin-core/gui/blob/480bf01c295527bd212964efe4df3bb886db5654/src/qt/rpcconsole.cpp#L1006

and it is unnecessary burden to translate the "Range" string.

hebasto:
`horizontalLayout_3` is useless:

Screenshot from 2021-04-30 14-45-45

With this change
```diff
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -642,9 +642,7 @@
       <attribute name="title">
        <string>&amp;Network Traffic</string>
       </attribute>
-      <layout class="QHBoxLayout" name="horizontalLayout_3">
-       <item>
-        <layout class="QVBoxLayout" name="verticalLayout_4">
+      <layout class="QVBoxLayout" name="verticalLayout_4">
          <item>
           <widget class="TrafficGraphWidget" name="trafficGraph" native="true">
            <property name="sizePolicy">
@@ -867,8 +865,6 @@
            </item>
           </layout>
          </item>
-        </layout>
-       </item>
       </layout>
      </widget>
      <widget class="QWidget" name="tab_peers">
```

we have

Screenshot from 2021-04-30 14-51-15

hebasto (commented):
Approach ACK 0174e68fd59460b53a187bf396cb54e523185eb1

A slider glitching was introduced:
gui291-0430-1

I think the root of this bug is changed layout of the `sldGraphRange`, `lblGraphRange` and `btnClearTrafficGraph` widgets.

On master:
```xml
            <widget class="QLabel" name="lblGraphRange">
             <property name="minimumSize">
              <size>
               <width>100</width>
               <height>0</height>
              </size>
             </property>
             <property name="alignment">
              <set>Qt::AlignCenter</set>
             </property>
            </widget>
```

This PR:
```xml
            <widget class="QLabel" name="lblGraphRange">
            </widget>
```

RandyMcMillan:
Great catch! In #90 the peers tab toggle detailView wasn't merged yet! So this wasn't an issue.
Re-adding the properties now.

NOTE: I would have reused and rebased PR #90 - For some reason gh wouldn't let me reopen it.

Talkless:
> > What I mean is if there's more data series needed (in addition to "Sent", and "Received") on the same graph, not about new tabs.
> 
> That can be figured out when new data series are proposed

So we gonna move back to the right? :) . Maybe adding a splitter, making right pane collapsible would help if someone wants "full screen like" graph?

RandyMcMillan:
yes @Talkless - that is an option. I will take a look at that.

RandyMcMillan:
Update: 797b1a5

Landscape Layout:

Screen Shot 2021-05-16 at 3 32 09 PM

Widget Layout: 

Screen Shot 2021-05-16 at 3 32 00 PM

src/qt/forms/debugwindow.ui
```diff
@@ -871,22 +824,57 @@
...
-               </size>
-              </property>
-             </spacer>
-            </item>
            </layout>
           </widget>
          </item>
+         <item>
+          <widget class="QSlider" name="sldGraphRange">
+           <property name="minimumSize">
```
RandyMcMillan, line 832: (re) setting minimum size to address sizing issue when time interval changes

src/qt/forms/debugwindow.ui
```diff
@@ -871,22 +824,57 @@
...
+            <number>6</number>
+           </property>
+           <property name="orientation">
+            <enum>Qt::Horizontal</enum>
+           </property>
+          </widget>
+         </item>
+         <item>
+          <widget class="QLabel" name="lblGraphRange">
+           <property name="minimumSize">
```
RandyMcMillan, line 857: (re) setting minimum size to address sizing issue when time interval changes

RandyMcMillan:
Slider Issue Fixed:

slider

ShaMan239 (commented):
Concept ACK
Tested Successfully on Ubuntu 20.04

The New Layout for the Network Traffic Window looks much better than the current master.
It is also not affecting the functionality of other Node Windows.
Nice Work!
The bug discussed [here](https://github.com/bitcoin-core/gui/pull/291#pullrequestreview-649577005) is also fixed. This was done by op by setting a minimum size of 60px for the lblGraphRange.
But I think op should increase the minimum size a little more because it is causing some overflow issues. Adding a screenshot of the problem:
Screenshot from 2021-08-07 01-06-56(1)
Overflowing of time (in the blue box near the slider)

Other than this the PR seems to work excellently
Adding some screenshots for comparison.

**On Maximum Size:**
|Master|PR|
|---------|----|
|Screenshot from 2021-08-07 00-11-07|Screenshot from 2021-08-07 00-10-02|

**On Minimum Size:**

|Master|PR|
|---------|----|
|Screenshot from 2021-08-07 00-11-00(2)|Screenshot from 2021-08-07 00-09-48|

**Difference in Layout (Screenshot from Qt Designer):**

|Master|PR|
|---------|----|
|Screenshot from 2021-08-07 00-52-00|Screenshot from 2021-08-07 00-47-58|
<p>

hebasto:
I've reconsidered my opinion, and withdraw my ACKs (https://github.com/bitcoin-core/gui/pull/291#issuecomment-830029755, https://github.com/bitcoin-core/gui/pull/291#pullrequestreview-649577005).

Agree with @Talkless's point (https://github.com/bitcoin-core/gui/pull/291#issuecomment-826322546, https://github.com/bitcoin-core/gui/pull/291#issuecomment-827027675, https://github.com/bitcoin-core/gui/pull/291#issuecomment-831420253).

One of such changes, that require the current layout, are "Received" and "Send" counters separated by data type: blocks, transactions, addresses, service messages. Especially, this could be helpful with upcoming Erlay protocol (at least for me, of course).

So, for now, Concept NACK from me.

rebroad:
Can we make it so the default is 3 hours not 30 minutes? Or make it so that this can be configured somewhere?

Also, can we make it so that when the time range is changed that the data is not lost?

src/qt/forms/debugwindow.ui
```diff
@@ -642,79 +642,32 @@
...
+            <string/>
            </property>
-           <layout class="QVBoxLayout" name="verticalLayout_5">
+           <property name="flat">
+            <bool>true</bool>
+           </property>
+           <layout class="QVBoxLayout" name="verticalLayout_9">
             <item>
-             <layout class="QHBoxLayout" name="horizontalLayout_4">
+             <layout class="QHBoxLayout" name="RecievedBox">
```
luke-jr, line 668: typo here and 2 more

RandyMcMillan:
Looks good with PR #473 as well...

Screen Shot 2021-11-20 at 7 39 55 PM

RandyMcMillan:
Minimum size presentation: 

Screen Shot 2021-11-20 at 7 41 06 PM

rebroad:
> Minimum size presentation:
> 
> Screen Shot 2021-11-20 at 7 41 06 PM

I like this new format - also, it makes it more possible to add a button above/below reset which could be used to toggle between linear and non-linear (rather than clicking on the graph).

Issue #288, opened 2021-04-22
Replace disambiguation strings with translator comments


hebasto:
In the Qt code we used disambiguation strings, i.e., the second argument in [`tr()`](https://doc.qt.io/qt-5/qobject.html#tr) call, to provide additional information to translators. All such cases must be replaced with [translator comments](https://doc.qt.io/qt-5/i18n-source-translation.html#translator-comments) to make things work with the Transifex (see #287).

#### Useful skills:

- basic C++17 skills
- basic understanding of Bitcoin Core [translation process](https://github.com/bitcoin-core/gui/blob/master/doc/translation_process.md)

#### Want to work on this issue?

For guidance on contributing, please read [CONTRIBUTING.md](https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md) before opening your pull request.

PR #292, opened 2021-04-24
WIP: translate strings (#288)

RandyMcMillan:
Some guidance on this would be appreciated until I get comfortable with this.

### DIFF
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -710,7 +710,7 @@ QString formatServicesStr(quint64 mask)
     if (strList.size())
         return strList.join(" & ");
     else
-        return QObject::tr("None");
+        return QObject::tr("None", "absence of value");
 }
 
 QString formatPingTime(std::chrono::microseconds ping_time)
--- a/src/qt/addressbookpage.cpp
+++ b/src/qt/addressbookpage.cpp
@@ -81,32 +81,32 @@ AddressBookPage::AddressBookPage(const PlatformStyle *platformStyle, Mode _mode,
     case ForSelection:
         switch(tab)
         {
-        case SendingTab: setWindowTitle(tr("Choose the address to send coins to")); break;
-        case ReceivingTab: setWindowTitle(tr("Choose the address to receive coins with")); break;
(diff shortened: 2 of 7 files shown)


jonatack:
The PR description in #287 looks like helpful guidance for this.

(PS - don't  put @-prefixed usernames in commits and PR descriptions; the latter because usernames in the description are copied into the merge commit by the merge script. This can cause endless annoying notifications for those concerned. An update to the merge script now warns maintainers if a merge message contains a @username, but it's better not to add them to begin with. For more, see https://jonatack.github.io/articles/how-to-contribute-pull-requests-to-bitcoin-core#pr-descriptions.)

hebasto:
@RandyMcMillan 

Thanks for working on this!

> The PR description in #287 looks like helpful guidance for this.

Indeed, for cases in this PR the translator comments should be used.

To ensure, that they are working properly, you could use `make -C src translate`.

Also for such changes I'd prefer a pr with a wide scope :)

hebasto:
@RandyMcMillan 

If your goal is resolving #288, you could look into #220 where some disambiguation strings were introduced, and they should be replaced with translator comments.

src/qt/guiutil.cpp
```diff
@@ -710,7 +710,7 @@ QString formatServicesStr(quint64 mask)
     if (strList.size())
         return strList.join(" & ");
     else
-        return QObject::tr("None");
+        return QObject::tr("None", "absence of value");
```
hebasto, line 713: Our goal is to replace useless (in most cases) disambiguation strings (the second argument in `tr` call) with translator comments:
```suggestion
        //: Absence of value.
        return QObject::tr("None");
```

RandyMcMillan:
NOTE: Currently I am just trying to get an idea of how much work needs to be done.

hebasto:
In its current state this PR does not follow our the most recent translation policy, as we are not using the second argument of the [`QObject:tr`](https://doc.qt.io/qt-5/qobject.html#tr) function.

#311 is a good example of the mentioned approach.

Closing for now.

PR #293, opened 2021-04-24
peers-tab: enable wordWrap for Services

RandyMcMillan:
#### Before:
Screen Shot 2021-04-24 at 7 30 25 PM

#### After:
Screen Shot 2021-04-24 at 7 21 43 PM

### DIFF
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -1198,6 +1198,9 @@
                  <property name="textFormat">
                   <enum>Qt::PlainText</enum>
                  </property>
+                 <property name="wordWrap">
+                  <bool>true</bool>
+                 </property>
                  <property name="textInteractionFlags">
                   <set>Qt::LinksAccessibleByMouse|Qt::TextSelectableByKeyboard|Qt::TextSelectableByMouse</set>
                  </property>


RandyMcMillan:
#### Dots instead:

Screen Shot 2021-04-24 at 7 51 44 PM

#### Blanks instead:
Screen Shot 2021-04-24 at 7 55 47 PM (2)

#### Blanks with stretched detail view:
Screen Shot 2021-04-24 at 7 56 42 PM (2)

Talkless:
tACK 84612fad4585b4fd6b12b37b9c7349958bb4362d, tested on Debian Sid amd64.
PR seems reasonable, that line is really usually long. 
I don't care about changing to dots and spaces, IMO good enough as it is, but wouldn't NACK other variants.

Talkless (approved):
tACK 84612fa, tested on Debian Sid amd64.
PR seems reasonable, that line is really usually long.
I don't care about changing to dots and spaces, IMO good enough as it is, but wouldn't NACK other variants.

jonatack:
Concept ACK, this seems very useful. I think either the current " & " or a ", " separator would be good. Will test.

kristapsk (approved):
ACK 84612fad4585b4fd6b12b37b9c7349958bb4362d. I agree that current "&" separator is ok.

jonatack (commented):
ACK 84612fad4585b4fd6b12b37b9c7349958bb4362d

Here or in a follow-up, it may be a good idea to use either a non-breaking space before the ampersand or a comma-space to avoid the separator splitting to a line by itself. Otherwise this is a great screen space-saver. Thanks!

Screenshot from 2021-04-25 15-59-44

RandyMcMillan:
I experimented with different ways to try to control where it breaks. I will take another look.

jonatack:
@RandyMcMillan you have two tested ACKs here and it's a nice compact self-contained diff that adds value, so if you find something for that I think you could propose it separately.

hebasto:
Concept ACK.

@RandyMcMillan 

Thanks for working on this! It has been in my TODO list for a long time :)

What if to add another tiny change:
```diff
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -708,7 +708,7 @@ QString formatServicesStr(quint64 mask)
     }
 
     if (strList.size())
-        return strList.join(" & ");
+        return strList.join("\n");
     else
         return QObject::tr("None");
 }
```
?

We could get:
Screenshot from 2021-04-25 19-04-21

Btw, @jonatack's alternative suggestions
> Here or in a follow-up, it may be a good idea to use either a non-breaking space before the ampersand or a comma-space to avoid the separator splitting to a line by itself.

also could be implemented in the same line :)

hebasto:
@RandyMcMillan For nice commit history do you mind changing the prefix in the _commit_ message s/`peers-tab:`/`qt:`/ ?

See: https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#creating-the-pull-request

hebasto (approved):
ACK bd04efe84069cdbee4f4b80150613a3bec26dda1, tested on Linux Mint 20.1 (Qt 5.12.8):

DeepinScreenshot_select-area_20210425193959

jonatack:
Now it will always use multiple lines even when there is enough horizontal space to only use one?

hebasto:
> Now it will always use multiple lines even when there is enough horizontal space to only use one?

Right. Is it bad?

kristapsk (approved):
re-ACK bd04efe84069cdbee4f4b80150613a3bec26dda1. I like this approach with always splitting by newline better than any separator character.

image

jonatack:
> Right. Is it bad?

It just trades one dimension for the other. The flexibility of optimizing the available space was what appealed to me.

jarolrod (commented):
ACK bd04efe84069cdbee4f4b80150613a3bec26dda1

Tested on macOS 11.2.3 Qt 5.15.2

Screen Shot 2021-04-25 at 1 03 11 PM

@jonatack
> Now it will always use multiple lines even when there is enough horizontal space to only use one?

I think this is ok

src/qt/forms/debugwindow.ui
```diff
@@ -1198,6 +1198,9 @@
                  <property name="textFormat">
                   <enum>Qt::PlainText</enum>
                  </property>
+                 <property name="wordWrap">
+                  <bool>true</bool>
+                 </property>
```
jonatack, line 1203: Is this still needed if we hardcode it to use one line per flag?

hebasto:
> The flexibility of optimizing the available space was what appealed to me.

You're right, and I agree with you, ceteris paribus. But in this case no other row requires wider space. Therefore, making the column wider is not an optimal way to use the available area.

jonatack:
If the user prefers a wider space to a higher one, the flexible version allows the user to do it with no downside.

src/qt/forms/debugwindow.ui
hebasto: I think, @jonatack is right, and this change seems redundant.

RandyMcMillan:
Great suggestion with using "\n" - it eliminates the issue. :)

We can always set a max width later...once the column flow gets stable.
We can leave it fluid for now.

Screen Shot 2021-04-24 at 7 45 20 PM

Screen Shot 2021-04-25 at 1 22 33 PM

jonatack:
> Great suggestion with using "\n" - it eliminates the issue. :)

It only adds vertical size to reduce the horizontal size, I don't see why it's better than optimising the available space :shrugging:

hebasto:
> If the user prefers a wider space to a higher one, the flexible version allows the user to do it with no downside.
> 
> Edit: I'd suggest the comma-space separator as it requires fewer characters and is a well-understood separator.

The downside of such approach is that some services, that are not placed in the beginning of the line, could just miss user's attention. Of course, it's a minor one.

RandyMcMillan:
I lean toward a UI that avoids horizontal scroll bars.
It becomes a usability issue because the horz scroll bar over lays textual information
that a person may try to copy and paste. (This is an issue with the table view itself that I was planning on addressing soon)

RandyMcMillan:
Comma separated for comparison

Screen Shot 2021-04-25 at 1 54 52 PM

Talkless:
I agree with @jonatack, flexible approach would be better IMO, i.e. without "forced" newlines.

RandyMcMillan:
Since @hebasto approved the PR as is - I will wait on his feedback.

RandyMcMillan:
I agree with @jonatack and @Talkless
I just wanted to provide examples to prompt discussion on these nuanced decisions. :)

src/qt/forms/debugwindow.ui
RandyMcMillan: This was referring to the force new line version.
Since we are concluding that forcing a new line is not preferred we should leave it.

src/qt/guiutil.cpp
```diff
@@ -455,8 +454,8 @@ bool ToolTipToRichTextFilter::eventFilter(QObject *obj, QEvent *evt)
     return QObject::eventFilter(obj, evt);
 }
 
-LabelOutOfFocusEventFilter::LabelOutOfFocusEventFilter(QObject* parent)
-    : QObject(parent)
+LabelOutOfFocusEventFilter::LabelOutOfFocusEventFilter(QObject* parent) :
+    QObject(parent)
```
hebasto, line 458: This change looks unrelated to this PR.

src/qt/guiutil.cpp
```diff
@@ -434,7 +434,6 @@ ToolTipToRichTextFilter::ToolTipToRichTextFilter(int _size_threshold, QObject *p
     QObject(parent),
     size_threshold(_size_threshold)
 {
-
```
hebasto, line 437: This change looks unrelated to this PR.

src/qt/guiutil.cpp
```diff
@@ -708,7 +707,7 @@ QString formatServicesStr(quint64 mask)
     }
 
     if (strList.size())
-        return strList.join(" & ");
+        return strList.join(" ");
```
hebasto, line 710: > I agree with @jonatack and @Talkless

I have an impression that @jonatack' [suggestion](https://github.com/bitcoin-core/gui/pull/293#issuecomment-826357677) is:
> I'd suggest the comma-space separator as it requires fewer characters and is a well-understood separator.

hebasto (changes requested):

hebasto:
fb5e880a090fd2586e52c564e8d6a32995639ea9 -- looking at the diff lines count I see +7/-3. If drop unrelated changes, it could be +4/-1.

hebasto (approved):
ACK a0f797867433b0d3e8d7851ddf05743fe70d320a, tested on Linux Mint 20.1 (Qt 5.12.8):

Node A:

Screenshot from 2021-05-10 10-04-19
Screenshot from 2021-05-10 10-04-53

Node B:

Screenshot from 2021-05-10 10-05-46
Screenshot from 2021-05-10 10-06-16

Talkless:
tACK a0f797867433b0d3e8d7851ddf05743fe70d320a on same environment as previously.

kristapsk (approved):
re-ACK a0f797867433b0d3e8d7851ddf05743fe70d320a. Tested under Gentoo Linux with Xfce4 (Qt 5.15.2).

PR #296, opened 2021-04-25
Do not use QObject::tr plural syntax for numbers with a unit symbol

hebasto:
Working on translation, I found this is useless and unnecessarily burdensome for translators. I guess, this statement is correct internationally wide :)

### DIFF
--- a/src/qt/intro.cpp
+++ b/src/qt/intro.cpp
@@ -278,12 +278,12 @@ void Intro::setStatus(int status, const QString &message, quint64 bytesAvailable
 
 void Intro::UpdateFreeSpaceLabel()
 {
-    QString freeString = tr("%n GB of free space available", "", m_bytes_available / GB_BYTES);
+    QString freeString = tr("%1 GB of free space available").arg(m_bytes_available / GB_BYTES);
     if (m_bytes_available < m_required_space_gb * GB_BYTES) {
-        freeString += " " + tr("(of %n GB needed)", "", m_required_space_gb);
+        freeString += " " + tr("(of %1 GB needed)").arg(m_required_space_gb);
         ui->freeSpace->setStyleSheet("QLabel { color: #800000 }");
     } else if (m_bytes_available / GB_BYTES - m_required_space_gb < 10) {
-        freeString += " " + tr("(%n GB needed for full chain)", "", m_required_space_gb);
+        freeString += " " + tr("(%1 GB needed for full chain)").arg(m_required_space_gb);
         ui->freeSpace->setStyleSheet("QLabel { color: #999900 }");
     } else {
         ui->freeSpace->setStyleSheet("");


promag (commented):
Code review ACK 3adde72bc99215062c8dabd38f8c34ad093452b5. Agree with OP, looks reasonable to me.

luke-jr:
Hopefully this won't disrupt some non-Latin language...

hebasto:
> Hopefully this won't disrupt some non-Latin language...

In which way?

luke-jr:
eg, if 300 GB and 400 GB are written using different characters.

hebasto:
> eg, if 300 GB and 400 GB are written using different characters.

In such case we could expect a comment from a translator here or on transifex.com, right?

luke-jr:
I'm not sure. Have we ever had any feedback of that sort from translators?

hebasto:
> I'm not sure. Have we ever had any feedback of that sort from translators?

Translators can open issues on transifex.com. There are some of them open now.

luke-jr:
I think we should revert this...

`bg` had different strings: `(%n GB е нужен)` vs `(%n GB са нужни)`

`ca`: `Un GB d'espai lliure disponible` vs `%1 GB d'espai lliure disponibles`; and `(Un GB necessari)` vs `(de %n GB necessàris)`

`eo`: `%n gigabajto de libera loko disponeble` vs `%n gigabajtoj de libera loko disponebla.`

`es_CL`: `(de %n GB requerido)` vs `(de %n GB requeridos)`

`fr`: `%n Go d’espace libre disponible` vs `%n Go d’espace libre disponibles`; and `(%n Go nécessaire pour la chaîne complète)` vs `(%n Go nécessaires pour la chaîne complète)`

`he`: `ג״ב של מקום פנוי זמין` vs `%n ג״ב של מקום פנוי זמינים`

`hr`: `(od potrebnog prostora od %n GB)` vs `(od potrebnog %n GB)`; and `(potreban je %n GB za cijeli lanac)` vs `(potrebna su %n GB-a za cijeli lanac)` vs `(potrebno je %n GB-a za cijeli lanac)`

`pl`: `(%n GB potrzebny na pełny łańcuch)` vs `(%n GB potrzebne na pełny łańcuch)` vs `(%n GB potrzebnych na pełny łańcuch)`

`ro`: `(din %n GB necesar)` vs `(din %n GB necesari)`

`sk`: `(z %n GB potrebného)` vs `(z %n GB potrebných)`; and `(%n GB potrebný pre plný reťazec)` vs `(%n GB potrebné pre plný reťazec)` vs `(%n GB potrebných pre plný reťazec)`

`sl`: `(%n GB potreben za celotno verigo blokov)` vs `(%n GB potrebna za celotno verigo blokov)` vs `(%n GB potrebni za celotno verigo blokov)` vs `(%n GB potrebnih za celotno verigo blokov)`

`sv`: `%n GB fritt utrymme kvar` vs `%n GB ledigt utrymme kvar`; and `(av %n GB behövs)` vs `(av de %n GB som behövs)`

`szl`: `(z %n GB przidajnego)` vs `(z %n GB przidajnych)`

flack:
just squinting at the above examples, it seems the logic of the ones I can somewhat decipher is like this:

```
1 GB free space is available
2 GB free space are available

1 GB is necessary for (...)
2 GB are necessary for (...)
```

I guess the question then is if all those languages have a way that covers both in cases in one phrase.

luke-jr:
> I guess the question then is if all those languages have a way that covers both in cases in one phrase.

Even if possible, we should go back to `%n` if forcing `%1` would make the translation weirder.

kybl:
From the languages above, I can understand Slavic ones. There, it matters not even if the number is 1 or more, but if the number is 1, or (2, 3, or 4), or (5 and more, or 0).

For example, in the Slovak language case:
* `0 GB potrebných` <-- 0 and 5+ case
* `1 GB potrebný` <-- 1 case
* `2 GB potrebné` <-- 2-4 case
* `3 GB potrebné` <-- 2-4 case
* `4 GB potrebné` <-- 2-4 case
* `5 GB potrebných` <-- 0 and 5+ case

I found a very informative source here: https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/language_plural_rules.html

Not sure if it can be somehow easily implementable in Transifex. In the current case, the translator probably tries to guess the typical number it will be there (impossible) and put form of such number.

westoleaboat:
In Spanish you may differentiate as follows: (similar as difference between 'is' and 'are' in English) 

- `0 GB requeridos` for zero it is plural
- `1 GB requerido` for one it is singular
- `2 GB requeridos` for >1 follows plural, same as zero

hebasto:
If a subject or an object is measured in some units, it is wrong to substitute it with units.

Therefore, a phrase _"%n GB needed for full chain"_ is bad worded, because a free space needed, not gigabytes.

Anyway, if this PR appeared so controversial, why do not just to open a new PR and revert these changes?

hebasto:
A relevant [announcement](https://www.transifex.com/bitcoin/bitcoin/announcements/12955/) on Transifex.com

GChuf:
I can confirm for slovenian and other slavic languages that we do have 3 or 4 different ways of saying "%n GB needed%, depending on the actual number of gigabytes. Similar to english and "is/are". There's no way to cover all cases ... this is exactly why transifex allows you to have more than 2 options.

Maybe? you could cover all cases in all languages by saying something like "Amount of GB needed: %n", but I don't think this was ever broken to begin with, and you only translate this once.

PR #297, opened 2021-04-25
Avoid unnecessary translations

hebasto:
Working on translation, I found these translations introduced in #79, that are unnecessary (assuming the universal nature of the "BTC" string).

### DIFF
--- a/src/qt/forms/optionsdialog.ui
+++ b/src/qt/forms/optionsdialog.ui
@@ -738,14 +738,14 @@
               <item>
                <widget class="QLabel" name="embeddedFont_label_1">
                 <property name="text">
-                 <string>111.11111111 BTC</string>
+                 <string notr="true">111.11111111 BTC</string>
                 </property>
                </widget>
               </item>
               <item>
                <widget class="QLabel" name="embeddedFont_label_9">
                 <property name="text">
-                 <string>909.09090909 BTC</string>
+                 <string notr="true">909.09090909 BTC</string>
                 </property>
                </widget>
               </item>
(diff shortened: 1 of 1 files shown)


jarolrod (commented):
ACK 19d51a2907cc00244510460a658c482d1b4bdfda

Tested that these are all the occurrences in this file, and cross-checked for any occurrences in other files. We already set the `notr` attribute in other areas of the code such as [coincontroldialog.ui](https://github.com/bitcoin-core/gui/blob/master/src/qt/forms/coincontroldialog.ui#L215), [overviewpage.ui](https://github.com/bitcoin-core/gui/blob/master/src/qt/forms/overviewpage.ui#L142), and [sendcoinsdialog.ui](https://github.com/bitcoin-core/gui/blob/master/src/qt/forms/sendcoinsdialog.ui#L407).

Furthermore, `BTC` is treated as a symbol or unit of Bitcoin and is untranslatable by nature.

Sjors:
Concept ACK

promag (commented):
ACK.

luke-jr:
Is it really untranslated? What about languages that don't use the Latin alphabet at all?

hebasto:
> Is it really untranslated? What about languages that don't use the Latin alphabet at all?

For example? (not non-Latin languages, but abbreviated name "BTC" in non-Latin languages)

My three non-Latin languages use "BTC", btw.

luke-jr:
I would expect Russian to use a Cyrillic abbreviation? Possibly Chinese and Japanese to get single-character units at some point?

hebasto:
> I would expect Russian to use a Cyrillic abbreviation? Possibly Chinese and Japanese to get single-character units at some point?

I've made a quick re-check of some multi-language online services. All of them use "BTC" in non-Latin languages.

PR #298, opened 2021-04-25
Peertableview alternating row colors

RandyMcMillan:
peers-tab: enable alternating row colors for peer table and banned table

### DIFF
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -923,6 +923,9 @@
              <property name="tabKeyNavigation">
               <bool>false</bool>
              </property>
+             <property name="alternatingRowColors">
+              <bool>true</bool>
+             </property>
              <property name="textElideMode">
               <enum>Qt::ElideMiddle</enum>
              </property>
@@ -983,6 +986,9 @@
             <widget class="QTableView" name="banlistWidget">
              <property name="tabKeyNavigation">
               <bool>false</bool>
+             </property>
+             <property name="alternatingRowColors">
+              <bool>true</bool>
              </property>
              <property name="sortingEnabled">
               <bool>true</bool>


src/qt/forms/debugwindow.ui
```diff
@@ -925,6 +925,9 @@
              </property>
              <property name="textElideMode">
               <enum>Qt::ElideMiddle</enum>
+              </property>
```
jarolrod, line 928: You're introducing this property within the `textElideMode` property. 

You should be introducing this after line `928` on the original file, not after line `927`. That way you don't have to introduce this closing tag to then introduce the property. 

Also, the indentation is off here.

jarolrod (commented):
Concept ACK

The other table views such as in the `Receive` and `Transactions` tab use alternating row colors. Seems appropriate that we also enable this on the `Peers` table.

src/qt/forms/debugwindow.ui
RandyMcMillan: Just trying to edit .ui files without qt-creator - the diff junk qt-creator makes is annoying.

RandyMcMillan:
e94920a0bb859d557dd978febde0a65d46fabb68

Screen Shot 2021-04-26 at 7 31 35 PM (2)

jarolrod (commented):
tACK e94920a0bb859d557dd978febde0a65d46fabb68

Can you edit your PR description to include a sentence(s) explaining this change? Normally we don't leave a PR description empty or only fill it with images. Additionally, edit the PR title to remove the leading `qt:`. This is redundant as it's known this is in the GUI repo and the commit title already includes this.

Thanks for also adding this to the `Banned Peers` table
Screen Shot 2021-04-26 at 10 36 06 PM

jonatack:
I really appreciate @RandyMcMillan's initiatives to improve the GUI and want to encourage them. In this case, it's maybe a case of "just because we can may not mean that we should": (a) separating info for a single peer not only isn't as critical as it might be for a transaction, but the peers detail area already exists for this, (b) the peers table columns can be as useful (or more useful) information as the the rows (in the peers table, I tend to read the columns) and adding horizontal lines would make that more difficult, and (c) it's more cluttered and less minimal a design without a real need AFAICT.

Bosch-0:
tACK https://github.com/bitcoin-core/gui/commit/e94920a0bb859d557dd978febde0a65d46fabb68 on Windows 10 - works as intended. Before / after below:

Frame 43

I would agree that this is a slight UI/UX improvement though I understand Jon's concerns. The peers tab isn't something I've used too often so I can not attest to the value of reading info horizontally or vertically, but horizontal alternating row colors is a pretty common design for tables that people will be familiar with and it makes things less visually cluttered.

src/qt/forms/debugwindow.ui
hebasto: I'm using Qt Designer to edit UI files.

Yeah, in most cases diff junk is unavoidable, but it is easy for me to revert it in Sublime Merge :)

jonatack:
I think this change makes this window needlessly noisy, more cluttered, and more difficult to use. It was the main tab I used. Oh well.

RandyMcMillan:
@jonatack - Maybe we can add more gui options to the user preferences - This may be a great direction to go. I like @hebasto's addition of the monospace font option - I have plenty of ideas for accessibility that would be great to be able to toggle on/off.

rebroad:
Please can this change be reverted? I agree with @jonatack  - there's way too much into in the right pane now, and I don't see the need for the alternating colours - I never had an issue following the lines - did anyone else?

rebroad:
> tACK [e94920a](https://github.com/bitcoin-core/gui/commit/e94920a0bb859d557dd978febde0a65d46fabb68) on Windows 10 - works as intended. Before / after below:
> 
> Frame 43
> 
> I would agree that this is a slight UI/UX improvement though I understand Jon's concerns. The peers tab isn't something I've used too often so I can not attest to the value of reading info horizontally or vertically, but horizontal alternating row colors is a pretty common design for tables that people will be familiar with and it makes things less visually cluttered.

Hold on a moment.. Those two screenshots aren't a comparison of before and after just this change, are they? Looks like many changes between those two examples...

promag:
ACK, this makes table presentation consistent.

PR #305, opened 2021-04-29
Display send/recv in Bps instead of totals in the debug window

rebroad:
Moved from https://github.com/bitcoin/bitcoin/pull/21806

I find this far more useful for than the totals. For example, I can easily see which nodes are not really talking with this metric, and not so much with the totals (as is current functionality).

Example:-

image

### DIFF
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -781,6 +781,40 @@ QString formatBytes(uint64_t bytes)
     return QObject::tr("%1 GB").arg(bytes / 1'000'000'000);
 }
 
+QString formatBps(float bits)
+{
+    if (bits < 10)
+        //: "Bits per second"
+        return QObject::tr("%1 bps").arg(0.01 * int(bits * 100));
+    if (bits < 100)
+        //: "Bits per second"
+        return QObject::tr("%1 bps").arg(0.1 * int(bits * 10));
+    if (bits < 1'000)
+        //: "Bits per second"
+        return QObject::tr("%1 bps").arg((int)bits);
+    if (bits < 10'000)
+        //: "Kilobits per second"
+        return QObject::tr("%1 kbps").arg(0.01 * ((int)bits / 10));
+    if (bits < 100'000)
+        //: "Kilobits per second"
+        return QObject::tr("%1 kbps").arg(0.1 * ((int)bits / 100));
+    if (bits < 1'000'000)
+        //: "Kilobits per second"
+        return QObject::tr("%1 kbps").arg((int)bits / 1'000);
+    if (bits < 10'000'000)
+        //: "Megabits per second"
+        return QObject::tr("%1 Mbps").arg(0.01 * ((int)bits / 10'000));
+    if (bits < 100'000'000)
+        //: "Megabits per second"
+        return QObject::tr("%1 Mbps").arg(0.1 * ((int)bits / 100'000));
+    if (bits < 10'000'000'000)
+        //: "Megabits per second"
+        return QObject::tr("%1 Mbps").arg((long)bits / 1'000'000);
+
+    //: "Gigabits per second"
+    return QObject::tr("%1 Gbps").arg((long)bits / 1'000'000'000);
+}
+
 qreal calculateIdealFontSize(int width, const QString& text, QFont font, qreal minPointSize, qreal font_size) {
     while(font_size >= minPointSize) {
         font.setPointSizeF(font_size);
--- a/src/qt/guiutil.h
+++ b/src/qt/guiutil.h
@@ -229,6 +229,7 @@ namespace GUIUtil
     QString formatNiceTimeOffset(qint64 secs);
 
     QString formatBytes(uint64_t bytes);
+    QString formatBps(float bits);
 
     qreal calculateIdealFontSize(int width, const QString& text, QFont font, qreal minPointSize = 4, qreal startPointSize = 14);
 
--- a/src/qt/peertablemodel.cpp
+++ b/src/qt/peertablemodel.cpp
@@ -85,10 +85,20 @@ QVariant PeerTableModel::data(const QModelIndex& index, int role) const
             return GUIUtil::NetworkToQString(rec->nodeStats.m_network);
         case Ping:
             return GUIUtil::formatPingTime(rec->nodeStats.m_min_ping_time);
-        case Sent:
-            return GUIUtil::formatBytes(rec->nodeStats.nSendBytes);
-        case Received:
-            return GUIUtil::formatBytes(rec->nodeStats.nRecvBytes);
+        case Sent: {
+            int64_t now = GetTimeSeconds();
+            if (now != rec->nodeStats.nTimeConnected) // Avoid division by zero
+                return GUIUtil::formatBps(rec->nodeStats.nSendBytes * 8.0 / (now - rec->nodeStats.nTimeConnected));
+            else
+                return QString::fromStdString("");
+        }
+        case Recv: {
+            int64_t now = GetTimeSeconds();
+            if (rec->nodeStats.nTimeConnected != now)
+                return GUIUtil::formatBps(rec->nodeStats.nRecvBytes * 8.0 / (now - rec->nodeStats.nTimeConnected));
+            else
+                return QString::fromStdString("");
+        }
         case Subversion:
             return QString::fromStdString(rec->nodeStats.cleanSubVer);
         } // no default case, so the compiler can warn about missing cases
@@ -105,7 +115,7 @@ QVariant PeerTableModel::data(const QModelIndex& index, int role) const
             return QVariant(Qt::AlignCenter);
         case Ping:
         case Sent:
-        case Received:
+        case Recv:
             return QVariant(Qt::AlignRight | Qt::AlignVCenter);
         case Subversion:
             return {};
--- a/src/net.h
+++ b/src/net.h
@@ -419,7 +419,6 @@ public:
     RecursiveMutex cs_sendProcessing;
 
     uint64_t nRecvBytes GUARDED_BY(cs_vRecv){0};
-
     std::atomic<int64_t> nLastSend{0};
     std::atomic<int64_t> nLastRecv{0};
     //! Unix epoch time at peer connection, in seconds.
--- a/src/qt/peertablemodel.h
+++ b/src/qt/peertablemodel.h
@@ -53,7 +53,7 @@ public:
         Network,
         Ping,
         Sent,
-        Received,
+        Recv,
         Subversion
     };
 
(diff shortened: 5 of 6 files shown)


src/qt/guiutil.h
```diff
@@ -218,6 +218,7 @@ namespace GUIUtil
     QString formatNiceTimeOffset(qint64 secs);
 
     QString formatBytes(uint64_t bytes);
+    QString formatBps(uint64_t bytes);
```
jarolrod, line 221: Should add doxygen documentation

see: https://github.com/bitcoin/bitcoin/blob/master/doc/developer-notes.md#coding-style-doxygen-compatible-comments

src/qt/guiutil.cpp
```diff
@@ -774,6 +774,18 @@ QString formatBytes(uint64_t bytes)
     return QObject::tr("%1 GB").arg(bytes / 1'000'000'000);
 }
 
+QString formatBps(uint64_t bytes)
+{
+    if (bytes < 1'000)
+        return QObject::tr("%1 B/s").arg(bytes);
```
jarolrod, line 780: should add Translator comments

see: https://doc.qt.io/qt-5/i18n-source-translation.html#translator-comments

jarolrod (commented):
Concept ACK

I agree that this is more useful than showing the totals. One can still see the totals by clicking on the peer.

One issue is that `Sent` and `Received` are now used in two different contexts. In the peer table, it's denoting how many bytes are ***currently*** being sent/received. In the sidebar it's denoting the ***total*** sent/received.

My suggestion is to change the `Sent`/`Received` text in the sidebar to `Total Sent`/`Total Received`.

305-test

src/qt/peertablemodel.cpp
```diff
@@ -122,9 +122,9 @@ QVariant PeerTableModel::data(const QModelIndex &index, int role) const
         case Ping:
             return GUIUtil::formatPingTime(rec->nodeStats.m_min_ping_time);
         case Sent:
-            return GUIUtil::formatBytes(rec->nodeStats.nSendBytes);
+            return GUIUtil::formatBps(rec->nodeStats.nSendBytes / (GetSystemTimeInSeconds() + 1 - rec->nodeStats.nTimeConnected));
         case Received:
-            return GUIUtil::formatBytes(rec->nodeStats.nRecvBytes);
+            return GUIUtil::formatBps(rec->nodeStats.nRecvBytes / (GetSystemTimeInSeconds() + 1 - rec->nodeStats.nTimeConnected));
```
jonatack, line 127: ISTM this calculation will be useful shortly after the connection time but for long-lived peers would provide little to no real-time feedback. Maybe use the time in seconds since last message sent/received like in `-netinfo`?

jonatack:
Worth exploring. The sidebar probably should be updated to display the same data, possibly in addition to the totals.

hebasto:
@rebroad 
> Changed to WIP as I just noticed the column sorting is still using the old values.

You also could convert it to a draft:
DeepinScreenshot_select-area_20210430132714

src/qt/peertablemodel.cpp
rebroad: done

rebroad:
> @rebroad
> 
> > Changed to WIP as I just noticed the column sorting is still using the old values.
> 
> You also could convert it to a draft:
> DeepinScreenshot_select-area_20210430132714

ah, thank you - I was wondering if that was possible... the terminology confuses me though as I'd say it's ready for review, but not perhaps ready to be merged.

src/qt/guiutil.h
rebroad: hmmm. well, the units indicate that it is a different metric, so I'm not sure if the addition of the word "total" is needed. What do other people think?

rebroad: @jarolrod I don't think doxygen docs exist for formatBytes, so perhaps that can be done in another pull request to do it for a number of functions.

rebroad:
> Worth exploring. The sidebar probably should be updated to display the same data, possibly in addition to the totals.

I considered doing this, but I think it's an overall advantage to not have too much data there. It's more useful to have bandwidth in the columns, and totals in the sidebar, IMHO.

src/qt/guiutil.h
jarolrod: @rebroad 
Since you're adding this function (`formatBps`) you should add the Doxygen comment with it. `formatBytes` not including a Doxygen when it was introduced is another issue that can be addressed in another PR.

rebroad: @jarolrod given that formatBps is so similar to formatBytes, I'm averse to doing the Doxygen for formatBytes as part of this pull, and I'm also averse to doing Doxygen for just formatBps without formatBytes in this pull, but I'm fine with doing a later pull request to add the Doxygen for both formatBytes and formatBps.

hebasto:
What is the base period to calculate data-transfer rate?

src/qt/guiutil.cpp
```diff
@@ -774,6 +774,18 @@ QString formatBytes(uint64_t bytes)
...
 }
 
+QString formatBps(uint64_t bytes)
+{
+    if (bytes < 10'000)
+        return QObject::tr("%1 bps").arg(bytes);
+    if (bytes < 10'000'000)
+        return QObject::tr("%1 kbps").arg(bytes / 1'000);
+    if (bytes < 10'000'000'000)
+        return QObject::tr("%1 Mbps").arg(bytes / 1'000'000);
```
hebasto, line 784:
```suggestion
    if (bytes < 10'000)
        //: "Bit per second."
        return QObject::tr("%1 bps").arg(bytes);
    if (bytes < 10'000'000)
        //: "Kilobit per second."
        return QObject::tr("%1 kbps").arg(bytes / 1'000);
    if (bytes < 10'000'000'000)
        //: "Megabit per second."
        return QObject::tr("%1 Mbps").arg(bytes / 1'000'000);
```

src/qt/guiutil.cpp
hebasto: A translator [comment](https://github.com/bitcoin-core/gui/pull/305#discussion_r624665523) is the easiest way to let translators know what "bps" is.

src/qt/guiutil.cpp
hebasto: Oh, just noted that here are bytes not bits. In that case the correct contraction is upper case "B", not "b": Bps, kBps, MBps.

rebroad:
> What is the base period to calculate data-transfer rate?

In my first pull request it was from connection time to current time. Now it's from connection time to time of last send/recv.

src/qt/guiutil.cpp
rebroad: @hebasto actually, they are bits, but I ought to change the variable name to reflect that. Will do so, and thanks for the guidance on the commentation.

rebroad: > Oh, just noted that here are bytes not bits. In that case the correct contraction is upper case "B", not "b": Bps, kBps, MBps.

why is the k small but the M is big?

hebasto: https://en.wikipedia.org/wiki/Kilo-
https://en.wikipedia.org/wiki/Mega-

luke-jr:
Maybe have both and let the user decide which to hide/show?

src/qt/guiutil.cpp
rebroad: ah, ok. will change.

rebroad:
> Maybe have both and let the user decide which to hide/show?

where would you place the option?

luke-jr:
Just have 4 columns, total in, total out, rate in, rate out.

Other applications let you rightclick the header area to hide/show them, but even if we can't have that, users can resize them to ~0 width.

rebroad:
> Just have 4 columns, total in, total out, rate in, rate out.
> 
> Other applications let you rightclick the header area to hide/show them, but even if we can't have that, users can resize them to ~0 width.

It's a pain having to resize columns, and also they cannot be resized to zero. I'd rather wait until there's the ability to disable columns before creating a pull to add rate in addition to downloaded. There are already too many columns IMHO.

ShaMan239 (commented):
Tested ACK cf39e7d7b17e0affc1214cb2bb64de3f89093679

Compiled and Tested on Ubuntu 20.04

It makes more sense to add the rate of sending and receiving of data than the total as suggested by op.
But I have just one doubt
```
return GUIUtil::formatBps(rec->nodeStats.nSendBytes * 8 / (rec->nodeStats.nLastSend+1 - rec->nodeStats.nTimeConnected));
```

By the logic used by op, the displayed rate is the average rate of data transferred by the peer. But I personally would prefer that if the displayed rate was instantaneous and not average because it might happen that a node is inactive for quite some time, but the data they transferred earlier was so high, that the rate column is showing a significantly large rate of data transfer. 
And also:
Screenshot from 2021-07-30 20-56-41


For values larger than 1000 bps the data should be shown in kbps, but it is not happening as in the highlighted example.

rebroad:
> For values larger than 1000 bps the data should be shown in kbps, but it is not happening as in the highlighted example.

I agree. I'll upload an update shortly.

rebroad:
> By the logic used by op, the displayed rate is the average rate of data transferred by the peer. But I personally would prefer that if the displayed rate was instantaneous and not average because it might happen that a node is inactive for quite some time, but the data they transferred earlier was so high, that the rate column is showing a significantly large rate of data transfer.

Yes, this does happen, for example in the case where there's IBD (initial block download) which has much higher transfer rates and then skews the numbers once that completes. I have another commit, which deals with this, and allows the calculation to be reset at the end of IBD. Shall I include that with this pull, or perhaps raise another pull in the future to include that?

shaavan (commented):
Changes since my last review:
- The sent data is displayed in kbps when it is greater than 1000bps.
- `bits` variable has been changed from `uint64_t` to `float`.
- Modifies Sent and Recieve case in `peertablemodel.cpp` to avoid division by zero.

Changing bits variable from int to float has to lead to displaying sent and receive value with decimals. This makes the displayed data look a lot better compared to the previous version. Great Work, @rebroad!

I am adding some screenshots to emphasize the difference between the old commit and the latest commit.

|Old Commit|Latest Commit|
|---------------|--------------------|
|Screenshot from 2021-07-30 20-56-41|latest|

> Shall I include that with this pull, or perhaps raise another pull in the future to include that?

IMO, I think it would be better to add this as an additional commit to this PR. This is a significant change. Let’s not wait for the opening of another and then waiting for it to get merged.

shaavan (commented):
Changes since my last review:
- Correction has been made at line 809: `10'000` -> `100'000`

>I have another commit, which deals with this, and allows the calculation to be reset at the end of IBD. Shall I include that with this pull, or perhaps raise another pull in the future to include that?

Any updates on the above. Have you planned on adding it as another commit, or have you decided to do so in a future PR?

rebroad:
@shaavan I am still testing the functionality for resetting at the end of IBD - have you already created a method for detecting this? I'd be interested in having a look at your code if so.

PR #306, opened 2021-05-01
Revert "Peertableview alternating row colors"

rebroad:
Reverts bitcoin-core/gui#298

I think this has made the table more cluttered now, and perhaps this pull request could be re-merged in the future once it's done with making the changes optional.

### DIFF
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -923,9 +923,6 @@
              <property name="tabKeyNavigation">
               <bool>false</bool>
              </property>
-             <property name="alternatingRowColors">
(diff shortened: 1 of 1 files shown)


jonatack:
I agree.

hebasto:
> ... making the changes optional.

Done in #307.

jarolrod:
My preference would be for #307. Some people were for this change, some against. #307 is the democratic option for appeasing both sides.

Rspigler:
I am concept NACK this, without a plan for 307, which has now been closed

rebroad:
I actually quite like the stripes now :) acquired taste I guess....!

PR #308, opened 2021-05-01
Display progress of LoadBlockDB() on splash screen

rebroad:
On some machines the splash screen can be displaying the message "Loading block index..." for quite a few minutes, causing some users to panic thinking the software has crashed or frozen.

This pull request makes it clear that the loading is progressing by showing a percentage that updates per second (although limited to no more updates than per 1 percent so as to reduce logging).

Is every 1 second too often? Maybe 5 or 10 seconds is better. Also, is this the correct repository or should it be a pull request in http://github.com/bitcoin/bitcoin ?


rebroad:
moving to main repo

PR #309, opened 2021-05-01
Add access to the Peers tab from the network icon

### LINKED ISSUE #93: Network tray icon should open peers window, add button to disable network there
I doubt many users are looking for a quick way to disable the network:
Schermafbeelding 2020-09-16 om 13 56 57

More likely, at least from my experience, they want to see the list of peers.

I suggest linking this icon to the peers tab. And then add a button there to disable the network.

hebasto:
This PR add a small context menu to the network activity icon that provides an access to the Peers tab:

gui-network-icon

Closes #93.

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -43,6 +43,7 @@
 #include <QAction>
 #include <QApplication>
 #include <QComboBox>
+#include <QCursor>
 #include <QDateTime>
 #include <QDragEnterEvent>
 #include <QListWidget>
@@ -199,9 +200,6 @@ BitcoinGUI::BitcoinGUI(interfaces::Node& node, const PlatformStyle *_platformSty
     // Subscribe to notifications from core
     subscribeToCoreSignals();
 
-    connect(connectionsControl, &GUIUtil::ClickableLabel::clicked, [this] {
-        m_node.setNetworkActive(!m_node.getNetworkActive());
-    });
     connect(labelProxyIcon, &GUIUtil::ClickableLabel::clicked, [this] {
         openOptionsDialogWithTab(OptionsDialog::TAB_NETWORK);
     });
@@ -586,7 +584,10 @@ void BitcoinGUI::setClientModel(ClientModel *_clientModel, interfaces::BlockAndH
         createTrayIconMenu();
 
         // Keep up to date with client
-        updateNetworkState();
+        setNetworkActive(m_node.getNetworkActive());
+        connect(connectionsControl, &GUIUtil::ClickableLabel::clicked, [this] {
+            GUIUtil::PopupMenu(m_network_context_menu, QCursor::pos());
+        });
         connect(_clientModel, &ClientModel::numConnectionsChanged, this, &BitcoinGUI::setNumConnections);
         connect(_clientModel, &ClientModel::networkActiveChanged, this, &BitcoinGUI::setNetworkActive);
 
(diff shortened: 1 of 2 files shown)


src/qt/bitcoingui.cpp
```diff
@@ -933,9 +938,24 @@ void BitcoinGUI::setNumConnections(int count)
...
+        [this] {
+            rpcConsole->setTabFocus(RPCConsole::TabTypes::PEERS);
+            showDebugWindow();
+        });
+    m_network_context_menu->addAction(
+        network_active ?
+            //: A context menu item.
+            tr("Disable network activity") :
+            //: A context menu item. The network activity was disabled previously.
+            tr("Enable network activity again"),
```
kristapsk, line 957: Maybe not so important, but I think "again" is redundant here, could be just "Enable network activity".

hebasto:
Updated 1f208f4360df55f18afb89056eb4c18e10f9fd5f -> a7f5ea589fb7a6eddca3e060eade403b2dd6ff87 ([pr309.02](https://github.com/hebasto/gui/commits/pr309.02) -> [pr309.03](https://github.com/hebasto/gui/commits/pr309.03), [diff](https://github.com/hebasto/gui/compare/pr309.02..pr309.03)):

- addressed @kristapsk's [comment](https://github.com/bitcoin-core/gui/pull/309#discussion_r624580987)

src/qt/bitcoingui.cpp
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui/pull/309#issuecomment-830763475).

kristapsk (approved):
ACK a7f5ea589fb7a6eddca3e060eade403b2dd6ff87

jarolrod (commented):
tACK a7f5ea589fb7a6eddca3e060eade403b2dd6ff87

Tested on macOS 11.3 Qt 5.15.2, Also tested with `-disablewallet=1`.

This is ok to merge as-is. But, wanted to document how I think this should be changed around as part of a follow-up.

The 'Peers Tab' is for seeing your connected Peers and interacting with them, such as banning or disconnecting a peer.

The idea of "Disabling Network Activity" should be tied to the 'Network Traffic Tab' not the 'Peers Tab'. This means we could add a "Disable Network Activity" button to the Network Traffic tab and a 'Network Traffic' widget next to the 'Connection Count' widget. Pressing this 'Network Traffic' widget would open the 'Network Traffic' tab where the user has the option of disabling network activity.

hebasto:
@Sjors Do you mind looking into this PR?

Sjors (approved):
tACK a7f5ea589fb7a6eddca3e060eade403b2dd6ff87

src/qt/bitcoingui.cpp
```diff
@@ -586,7 +584,10 @@ void BitcoinGUI::setClientModel(ClientModel *_clientModel, interfaces::BlockAndH
         createTrayIconMenu();
 
         // Keep up to date with client
-        updateNetworkState();
+        setNetworkActive(m_node.getNetworkActive());
+        connect(connectionsControl, &GUIUtil::ClickableLabel::clicked, [this] {
+            m_network_context_menu->exec(QCursor::pos());
```
promag, line 589: `QMenu::exec` is blocking. If you want to keep this approach then you could declare a local `QMenu` instance instead - this way you can ditch the class member and the call to `.clear()`.

promag (commented):
Tested ACK a7f5ea589fb7a6eddca3e060eade403b2dd6ff87.

hebasto:
Updated a7f5ea589fb7a6eddca3e060eade403b2dd6ff87 -> d29ea72393ac1d9b32a6976062e9c9fb75876295 ([pr309.03](https://github.com/hebasto/gui/commits/pr309.03) -> [pr309.04](https://github.com/hebasto/gui/commits/pr309.04), [diff](https://github.com/hebasto/gui/compare/pr309.03..pr309.04)):

- addressed @promag's [comment](https://github.com/bitcoin-core/gui/pull/309#discussion_r626571342):
> `QMenu::exec` is blocking.

src/qt/bitcoingui.cpp
hebasto: @promag 

Thanks!

> `QMenu::exec` is blocking.

My bad! [Fixed](https://github.com/bitcoin-core/gui/pull/309#issuecomment-835154862).

> If you want to keep this approach then you could declare a local `QMenu` instance instead - this way you can ditch the class member and the call to `.clear()`.

How the lifetime of local `QMenu` instances could be managed in a proper way?

RandyMcMillan:
It would be great if a key combo was added to this functionality...
Maybe:

COMMAND + OPTION <w> or <n> would work.

or

Maybe COMMAND + OPTION <o> would work.

kristapsk:
> Adding a shortcut for enabling/disabling the network activity

I'm not even sure such shortcut is a good idea at all. Could accidentally hit that, the same that has happened to more with that icon without this PR.

promag (commented):
Code review ACK d29ea72393ac1d9b32a6976062e9c9fb75876295.

Not fond of the approach though. It seems it would be enough to update the menu item text on `BitcoinGUI::updateNetworkState()` instead of `QMenu::clear()` + `QMenu::addAction()` and the slot would just query the current network state.

It also seems more elegant to just call `BitcoinGUI::updateNetworkState()` in `BitcoinGUI::setClientModel` which already calls `m_node.getNetworkActive()`.

Worth noting that when the toggle action is triggered it calls `m_node.setNetworkActive()` which results in destroying that same action due to `QMenu::clear()`.

Don't mind my feedback, it's just another way of implementing the feature, thanks for adding the feature and addressing my previous comment.

Sjors:
re-ACK d29ea72393ac1d9b32a6976062e9c9fb75876295

PR #311, opened 2021-05-03
Peers Window rename 'Peer id' to 'Peer'

jarolrod:
Picking up https://github.com/bitcoin-core/gui/pull/290

**Original PR Description:**
- renames the peers tab column header from `Peer Id` to `Peer` to allow resizing the column more tightly (this will be particularly useful after #256) and does the same for the peer details area.

While here, we also add Qt translator comments for the Peer Table columns.

| ----------- | ----------- |
| Screen Shot 2021-05-03 at 1 23 05 AM | Screen Shot 2021-05-05 at 4 08 45 AM |

### DIFF
--- a/src/qt/peertablemodel.h
+++ b/src/qt/peertablemodel.h
@@ -74,7 +74,31 @@ public Q_SLOTS:
 
 private:
     interfaces::Node& m_node;
-    const QStringList columns{tr("Peer"), tr("Address"), tr("Type"), tr("Network"), tr("Ping"), tr("Sent"), tr("Received"), tr("User Agent")};
+    const QStringList columns{
+        /*: Title of Peers Table column which contains a
+            unique number used to identify a connection. */
+        tr("Peer"),
+        /*: Title of Peers Table column which contains the
+            IP/Onion/I2P address of the connected peer. */
+        tr("Address"),
+        /*: Title of Peers Table column which describes the type of
+            peer connection. The "type" describes why the connection exists. */
+        tr("Type"),
+        /*: Title of Peers Table column which states the network the peer
+            connected through. */
+        tr("Network"),
+        /*: Title of Peers Table column which indicates the current latency
+            of the connection with the peer. */
+        tr("Ping"),
+        /*: Title of Peers Table column which indicates the total amount of
+            network information we have sent to the peer. */
+        tr("Sent"),
+        /*: Title of Peers Table column which indicates the total amount of
+            network information we have received from the peer. */
+        tr("Received"),
+        /*: Title of Peers Table column which contains the peer's
+            User Agent string. */
+        tr("User Agent")};
     std::unique_ptr<PeerTablePriv> priv;
     QTimer *timer;
 };
--- a/src/qt/peertablemodel.h
+++ b/src/qt/peertablemodel.h
@@ -74,7 +74,7 @@ public Q_SLOTS:
 
 private:
     interfaces::Node& m_node;
-    const QStringList columns{tr("Peer Id"), tr("Address"), tr("Type"), tr("Network"), tr("Ping"), tr("Sent"), tr("Received"), tr("User Agent")};
+    const QStringList columns{tr("Peer"), tr("Address"), tr("Type"), tr("Network"), tr("Ping"), tr("Sent"), tr("Received"), tr("User Agent")};
     std::unique_ptr<PeerTablePriv> priv;
     QTimer *timer;
 };
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -1023,7 +1023,7 @@ void RPCConsole::updateDetailWidget()
     const auto stats = selected_peers.first().data(PeerTableModel::StatsRole).value<CNodeCombinedStats*>();
     // update the detail ui with latest node information
     QString peerAddrDetails(QString::fromStdString(stats->nodeStats.addrName) + " ");
-    peerAddrDetails += tr("(peer id: %1)").arg(QString::number(stats->nodeStats.nodeid));
+    peerAddrDetails += tr("(peer: %1)").arg(QString::number(stats->nodeStats.nodeid));
     if (!stats->nodeStats.addrLocal.empty())
         peerAddrDetails += "<br />" + tr("via %1").arg(QString::fromStdString(stats->nodeStats.addrLocal));
     ui->peerHeading->setText(peerAddrDetails);


hebasto:
@jarolrod 

Did you consider how it will look on a long-running node when one-digit id peer is next to, say, four-digit id one?

jarolrod:
@hebasto Actually, that's a good point. As a tabular figure, perhaps it makes more sense to leave as-is. Especially, if we enforce a monospace font at some point

jarolrod:
Update from 4ed95c0 -> 73a91c6

Changes: 
- Keep Peer # left-aligned as this is more appropriate for a tabular figure

hebasto:
> * renames the peers tab column header from `Peer Id` to `Peer` to allow resizing the column more tightly

If saving of the horizontal space is the motivation, leaving just "ID" or "id" is strictly better approach for the following reasons:
- the translated "Peer" could become much longer, e.g., the both variants in Ukrainian are "Вузол" or "З'єднання"
- "ID"/"id" is shorter, and often remains untranslated

jonatack:
> leaving just "ID" or "id" is strictly better approach

Hm, I disagree for the reasons I already described in the parent PR.

For laypeople (not devs or database people), the right word here is probably "number", "peer number" or just "peer", as "number" is self-understood. 

"ID" means "identification" and in common use, your ID card. "Show me some ID" is what the police officer or the security guard would ask you.

"Id" is a word from Freudian psychology IIRC.

Maybe a good solution is to have native speakers of each language decide on the best word for that language, and in cases where the English-as-a-first-language and English-as-a-second-language-or-as-dev-jargon differ, provide helpful translation hints in the translator comments.

RandyMcMillan:
№ - numero will work..

https://en.wikipedia.org/wiki/Numero_sign

https://github.com/RandyMcMillan/gui/tree/numero-column-name

https://github.com/RandyMcMillan/gui/commit/23dbd90e32127f72cd9a6e2dfec3909083df2e37#diff-3fcbd1d49f399ac9b4656011ea03e7ef80a855482d6de33855e6326ec065321fR77

Screen Shot 2021-05-11 at 12 28 46 AM

jonatack:
> № - numero will work..
> 
> https://en.wikipedia.org/wiki/Numero_sign

Interesting, it is a standard abbreviation in French, TIL it exists in English. I like it better than ID, not sure how well known it is and what the GUI policy is regarding acronyms and abbreviations.

hebasto:
Agree to use any neutral replacement for "id" string, even just "#" for English locale :)
https://en.wikipedia.org/wiki/Number_sign

jonatack:
These column headers are words, so for the English I would suggest using "Peer" and then describe alternatives ("This column means the peer number or ID") for translators in the tr comment.

hebasto:
It seem this PR needs translation comments only :)

promag:
Honest opinion here, could be empty. Otherwise ID or # LGTM.

RandyMcMillan:
In this old PR #135 "ID" looked ok...

jarolrod:
Updated from 73a91c63ec72e62ef76fbc857baff14b099a1358 -> b405f19c953d9cd5aa90b66001c31cc58230fb78 ([pr311.02](https://github.com/jarolrod/gui/commits/pr311.02), [pr311.03](https://github.com/jarolrod/gui/commits/pr311.03), [diff](https://github.com/jarolrod/gui/compare/master...jarolrod:pr311.03))

Changes:
- added translator comments for the columns of the Peers Table to address this [comment](https://github.com/bitcoin-core/gui/pull/311#issuecomment-839943546). Hopefully, this translator comment is enough to guide the translation of the term and can appease all sides on the debate of `ID` vs `Peer`

src/qt/peertablemodel.h
```diff
@@ -74,7 +74,33 @@ public Q_SLOTS:
 
 private:
     interfaces::Node& m_node;
-    const QStringList columns{tr("Peer Id"), tr("Address"), tr("Type"), tr("Network"), tr("Ping"), tr("Sent"), tr("Received"), tr("User Agent")};
+    const QStringList columns{
+        /*: Title of Peers Table column which contains a
+            unique number used to identify a connection. */
+        tr("Peer"),
+        /*: Title of Peers Table column which contains the
+            IP/Onion/I2P address we used to connect to the peer. */
```
hebasto, line 82: "we used to connect to the peer" -- looks like it is about outgoing connections only.

src/qt/peertablemodel.h
```diff
@@ -74,7 +74,33 @@ public Q_SLOTS:
...
+        tr("Address"),
+        /*: Title of Peers Table column which contains a value for
+            the type of connection we have established with the peer.
```
hebasto, line 85: "a value for the type" -- sounds unusual for me.

src/qt/peertablemodel.h
```diff
@@ -74,7 +74,33 @@ public Q_SLOTS:
...
+            Type is a term denoting what the connection is for. */
+        tr("Type"),
+        /*: Title of Peers Table column which contains a value for
+            the network the peer resides on. */
```
hebasto, line 89: "a value for the network" -- sounds unusual for me.

src/qt/peertablemodel.h
```diff
@@ -74,7 +74,33 @@ public Q_SLOTS:
...
+        tr("Network"),
+        /*: Title of Peers Table column which contains a value for
+            the latency of the connection established with the peer. */
+        tr("Ping"),
+        /*: Title of Peers Table column which contains a value for
+            the total amount of bytes we have sent to the peer. */
```
hebasto, line 95: "bytes" -- could the exact name of the unit be avoided? (two places)

src/qt/peertablemodel.h
```diff
@@ -74,7 +74,33 @@ public Q_SLOTS:
...
+        tr("Sent"),
+        /*: Title of Peers Table column which contains a value for
+            the total amount of bytes we have received from the peer. */
+        tr("Received"),
+        /*: Title of Peers Table column which contains the peer's
+            User Agent string. A peer can configure this value to
+            allow itself to be identifiable. */
```
hebasto, line 102: I assume there could be more reasons to use `-uacomment`, maybe avoid this particular pattern to make the comment more general?

hebasto (approved):
ACK b405f19c953d9cd5aa90b66001c31cc58230fb78, I agree on column title renaming, and as a translator I'm happy to see more translator comments.

I'm not an expert in the English Grammar, therefore, the following nits could be just ignored.

src/qt/peertablemodel.h
jarolrod: what do you think about this rephrasing:
 `Title of the Peers Table column which contains the IP/Onion/I2P address of the connected peer`

hebasto: lgtm

src/qt/peertablemodel.h
jonatack: agree that "we have established" sounds like outbound

"which describes the type of peer connection" ?

src/qt/peertablemodel.h
jonatack: "column for the network the peer connected through" ?￼

src/qt/peertablemodel.h
jonatack: "column for the peer address" ?

(reviewing from my phone so apologies if a bit out of context)

jarolrod:
updated from b405f19 -> 657b33e ([pr311.03](https://github.com/jarolrod/gui/tree/pr311.03), [pr311.04](https://github.com/jarolrod/gui/tree/pr311.04), [diff](https://github.com/jarolrod/gui/compare/pr311.03...jarolrod:pr311.04))

**Changes:**
Updated to address review comments on the translator comment contents, thanks @hebasto and @jonatack!

Here is a diff of how I've updated from [pr311.03](https://github.com/jarolrod/gui/tree/pr311.03):

``` diff
diff --git a/src/qt/peertablemodel.h b/src/qt/peertablemodel.h
index 268ae013e..3d195342f 100644
--- a/src/qt/peertablemodel.h
+++ b/src/qt/peertablemodel.h
@@ -79,27 +79,25 @@ private:
             unique number used to identify a connection. */
         tr("Peer"),
         /*: Title of Peers Table column which contains the
-            IP/Onion/I2P address we used to connect to the peer. */
+            IP/Onion/I2P address of the connected peer. */
         tr("Address"),
-        /*: Title of Peers Table column which contains a value for
-            the type of connection we have established with the peer.
-            Type is a term denoting what the connection is for. */
+        /*: Title of Peers Table column which describes the type of
+            peer connection. The "type" describes why the connection exists. */
         tr("Type"),
-        /*: Title of Peers Table column which contains a value for
-            the network the peer resides on. */
+        /*: Title of Peers Table column which states the network the peer
+            connected through. */
         tr("Network"),
-        /*: Title of Peers Table column which contains a value for
-            the latency of the connection established with the peer. */
+        /*: Title of Peers Table column which indicates the current latency
+            of the connection with the peer. */
         tr("Ping"),
-        /*: Title of Peers Table column which contains a value for
-            the total amount of bytes we have sent to the peer. */
+        /*: Title of Peers Table column which indicates the total amount of
+            network information we have sent to the peer. */
         tr("Sent"),
-        /*: Title of Peers Table column which contains a value for
-            the total amount of bytes we have received from the peer. */
+        /*: Title of Peers Table column which indicates the total amount of
+            network information we have received from the peer. */
         tr("Received"),
         /*: Title of Peers Table column which contains the peer's
-            User Agent string. A peer can configure this value to
-            allow itself to be identifiable. */
+            User Agent string. */
         tr("User Agent")};
     std::unique_ptr<PeerTablePriv> priv;
     QTimer *timer;
```

jonatack:
utACK 657b33ef2de77acd1061cdf4d1d64d0e086d75df

hebasto (approved):
re-ACK 657b33ef2de77acd1061cdf4d1d64d0e086d75df

PR #313, opened 2021-05-03
qt: Optimize string concatenation by default

hebasto:
From [Qt docs](https://doc.qt.io/qt-5/qstring.html#more-efficient-string-construction):
> ... multiple uses of the \[`QString`\] '+' operator usually means multiple memory allocations. When concatenating n substrings, where n > 2, there can be as many as n - 1 calls to the memory allocator.

With this PR
> ... the '+' will automatically be performed as the `QStringBuilder` '%' everywhere.

The change in the `src/Makefile.qt.include` file does not justify submitting this PR into the main repo, IMHO.


jarolrod (commented):
ACK b4e1480c02e845ab1df573d92acb814f62a172ea 🥃

This simplifies and reduces the mental burden of having to think about "am I concatenating a string in the most efficient way possible?"

I could not find any other occurrences of including `QStringBuilder` or occurrences of concatenating string with `%`.

hebasto:
@Talkless

Could you look into this PR?

src/qt/optionsmodel.cpp
```diff
@@ -244,7 +245,8 @@ static ProxySetting GetProxySetting(QSettings &settings, const QString &name)
 
 static void SetProxySetting(QSettings &settings, const QString &name, const ProxySetting &ip_port)
 {
-    settings.setValue(name, ip_port.ip + ":" + ip_port.port);
+    QString proxy = ip_port.ip + QLatin1Char(':') + ip_port.port;
+    settings.setValue(name, proxy);
```
Talkless, line 249: This still could be one-liner with wrapping second argument in `QString{}` temporary, or this is assumed to have better readability? If so, `const` could be added.

src/qt/peertablemodel.cpp
```diff
@@ -114,7 +114,7 @@ QVariant PeerTableModel::data(const QModelIndex &index, int role) const
             return (qint64)rec->nodeStats.nodeid;
         case Address:
             // prepend to peer address down-arrow symbol for inbound connection and up-arrow for outbound connection
-            return QString(rec->nodeStats.fInbound ? "↓ " : "↑ ") + QString::fromStdString(rec->nodeStats.addrName);
+            return QString::fromStdString((rec->nodeStats.fInbound ? "↓ " : "↑ ") + rec->nodeStats.addrName);
```
Talkless, line 117: What is motive for this change?

src/qt/transactiondesc.cpp
```diff
@@ -38,14 +40,19 @@ QString TransactionDesc::FormatTxStatus(const interfaces::WalletTx& wtx, const i
...
     {
         int nDepth = status.depth_in_main_chain;
-        if (nDepth < 0)
+        if (nDepth < 0) {
             return tr("conflicted with a transaction with %1 confirmations").arg(-nDepth);
-        else if (nDepth == 0)
-            return tr("0/unconfirmed, %1").arg((inMempool ? tr("in memory pool") : tr("not in memory pool"))) + (status.is_abandoned ? ", "+tr("abandoned") : "");
-        else if (nDepth < 6)
+        } else if (nDepth == 0) {
+            QString abandoned;
```
Talkless, line 46: Using ternary operator would allow marking it as const and avoiding re-initialization. Ir this is considered more readable?

Talkless (commented):
Concept ACK

hebasto:
@Talkless 
> It's true that QtCreator, and Qt itself is built with that flag, but we might want to discuss if we want that "implicit magic".

You are right! Discussion is a good mean to reach (rough) consensus about suggested changes.

>  There is some caveat with this kind of code:

You code snippet could be fixed by s/`auto`/`QString`/, and such caveats could be documented in https://github.com/bitcoin-core/bitcoin-devwiki/wiki/Developer-Notes-for-Qt-Code.

src/qt/peertablemodel.cpp
hebasto:
```
qt/peertablemodel.cpp:117:20: error: no viable conversion from returned value of type 'QStringBuilder<typename QConcatenable<QString>::type, typename QConcatenable<QString>::type>' (aka 'QStringBuilder<QString, QString>') to function return type 'QVariant'
            return QString(rec->nodeStats.fInbound ? "↓ " : "↑ ") + QString::fromStdString(rec->nodeStats.addrName);
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

hebasto:
Updated b4e1480c02e845ab1df573d92acb814f62a172ea -> a02c970eb001b456d74ddc30750fe8b55348ddac ([pr313.02](https://github.com/hebasto/gui/commits/pr313.02) -> [pr313.03](https://github.com/hebasto/gui/commits/pr313.03), [diff](https://github.com/hebasto/gui/compare/pr313.02..pr313.03)):

- addressed @Talkless's comments

src/qt/optionsmodel.cpp
hebasto: [Updated](https://github.com/bitcoin-core/gui/pull/313#issuecomment-841686529).

src/qt/transactiondesc.cpp
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui/pull/313#issuecomment-841686529).

Talkless:
> You code snippet could be fixed by s/`auto`/`QString`/

Yes, but that's the "danger", as one needs to know these kinda caveats. Yes, docs could help.

Talkless (approved):
utACK a02c970eb001b456d74ddc30750fe8b55348ddac, built successfully on Debian Sid with Qt 5.15.2, but did not check if any displayed strings are "wrong" after refactoring.

jarolrod (commented):
ACK a02c970eb001b456d74ddc30750fe8b55348ddac

Tested most of the strings on Qt 5.15.2 macOS 11.3

laanwj:
Code review ACK a02c970eb001b456d74ddc30750fe8b55348ddac

> This simplifies and reduces the mental burden of having to think about "am I concatenating a string in the most efficient way possible?"

I don't think there is any place in bitcoin where concatenating Qt strings is actually a bottleneck, but agree with this perspective.

> It's true that QtCreator, and Qt itself is built with that flag,

That's reassuring also in the sense that it probably won't just go away.

jonatack:
Approach ACK, this looks like a good idea indeed.

PR #317, opened 2021-05-05
Direction col peers

jarolrod:
Picking up #289 

This adds a `Direction column`, making the peers tab the same as the `Direction/Type` row in the peer details and the direction and type columns in our other user-facing peer connections table in `-netinfo`.

Users can now sort the peers table by direction. The default sort is set to inbound, then outbound.

This patch also makes the direction and type translations the same for the 3 places where they are used (the peers column, the peer details row, and its tooltip) and adds translator comments to link them together.

Changes:
- remove `QStringBuilder` and `%` concatenation as #313 will ensure the string concatenations here are efficient.

| ----------- | ----------- |
| Screen Shot 2021-05-05 at 3 51 09 AM | Screen Shot 2021-05-05 at 3 35 40 AM |


src/qt/peertablesortproxy.cpp
```diff
@@ -26,6 +26,8 @@ bool PeerTableSortProxy::lessThan(const QModelIndex& left_index, const QModelInd
         return left_stats.nodeid < right_stats.nodeid;
     case PeerTableModel::Address:
         return left_stats.addrName.compare(right_stats.addrName) < 0;
+    case PeerTableModel::Direction:
+        return left_stats.fInbound > right_stats.fInbound; // default sort Inbound, then Outbound
```
hebasto, line 30: What does mean "default sort"?

src/qt/guiutil.cpp
```diff
@@ -666,10 +666,12 @@ QString ConnectionTypeToQString(ConnectionType conn_type, bool prepend_direction
 {
     QString prefix;
     if (prepend_direction) {
+        //: Connection direction. Also used in CONNECTION_TYPE_DOC and PeerTableModel::data().
```
hebasto, line 669: This translator comment will be assigned to the first string only, i.e., "Inbound".

We should add a translator comment to one string at a time (here and other places).

jonatack:
Currently, direction isn't a sortable column, and it's potentially confusing to users to display both connection direction and type in the details area but only the the connection type without the direction (and so no type at all for inbounds) in the main table.  It was the reason I reopened the parent PR after an issue was filed in bitcoin/bitcoin where this came up.

jarolrod:
updated from a44eec9 -> 5d7463a ([pr317.01](https://github.com/jarolrod/gui/tree/pr317.01), [pr317.02](https://github.com/jarolrod/gui/tree/pr317.02), [diff](https://github.com/jarolrod/gui/compare/pr317.01...jarolrod:pr317.02))

**Changes:**
- drop the last three commits (f4bde8c4067dd94fd53bf043934af8bc9a9b3ec6, 8090db4bbb47ec787918ca1e7f4a4a818b94fff5, a44eec94805dc523b8c7de638bb68f555cc63873) inherited from picking up #289 in order to make the PR focused on adding a `Direction` column. The work done in those commits will be picked up in another PR.
- Now based on the last commit of #311 so that we are consistent with adding the new `Direction` column and the translator comment for it
- Address https://github.com/bitcoin-core/gui/pull/317#discussion_r627172107 by removing the (possibly confusing) comment
- Fix translator comment for the `Direction` column in `PeerTableModel::data`

src/qt/peertablesortproxy.cpp
jarolrod: addressed in 5d7463a

src/qt/guiutil.cpp
jarolrod: Will address in a follow-up PR which focuses on improving these translations. For now, the commit referenced here has been dropped to make the PR focused on just adding a new `Direction` column. See: https://github.com/bitcoin-core/gui/pull/317#issuecomment-848432166

src/qt/peertablesortproxy.cpp
```diff
@@ -26,6 +26,8 @@ bool PeerTableSortProxy::lessThan(const QModelIndex& left_index, const QModelInd
         return left_stats.nodeid < right_stats.nodeid;
     case PeerTableModel::Address:
         return left_stats.addrName.compare(right_stats.addrName) < 0;
+    case PeerTableModel::Direction:
+        return left_stats.fInbound > right_stats.fInbound;
```
jonatack, line 30: if you retouch, the comment at https://github.com/bitcoin-core/gui/pull/289/commits/678b57227e5c163cfbe77444d2c8b39c357678c2#diff-dafd60d44b87b3de8fc005415f63356de4fdd1fa772d1c1131a2c8b58c8b1d9aR30 seemed helpful to describe why `>` rather than `<` operator is used here, unlike the other columns

jonatack (commented):
Light code review ACK 5d7463aec2c3a786af6aa724c417af16f8a787b6

src/qt/peertablesortproxy.cpp
jonatack: The sort direction when you first click to a column.

jarolrod:
updated from 5d7463a -> 0183415 ([pr317.02](https://github.com/jarolrod/gui/tree/pr317.02), [pr317.03](https://github.com/jarolrod/gui/tree/pr317.03))

**Changes:**
- rebased over master due to conflict with #313 
- Addressed @jonatacks [suggestion](https://github.com/bitcoin-core/gui/pull/317#discussion_r639688568). For the purposes of clarity, reintroduce comment about ordering

src/qt/peertablesortproxy.cpp
jarolrod: Thanks for the review, Addressed in 0183415

jarolrod:
updated from 0183415 -> 6971e79 ([pr317.03](https://github.com/jarolrod/gui/tree/pr317.03) -> [pr317.04](https://github.com/jarolrod/gui/tree/pr317.04), [diff](https://github.com/jarolrod/gui/compare/pr317.03...jarolrod:pr317.04))

**Changes:**
- rebase over conflict with #311 (drop commit from 311 this was based on)

jonatack:
Tested ACK 6971e790c32253ecddb6108e796afd3892ced7fe

jonatack:
@hebasto mind having a look into this? Would be nice to have it in v22.

luke-jr:
This still takes up an annoying chunk of horizontal size. #363 ?

jonatack:
> This still takes up an annoying chunk of horizontal size. #363 ?

Recently merged PRs gained us a lot of horizontal space so I thought it might no longer be an obstacle. 

The issue with arrows is they would not resolve the discrepancy between the connection direction and type, which are displayed in full in the details area ("Outbound Full Relay") but lack the direction in the main table ("Full Relay").  This causes confusion, e.g. for a bitcoin dev in a bitcoin/bitcoin issue who thought it was a bug.

rebroad:
NACK - I prefer the existing functionality. It could be possible to have a 4-click sort method on the existing column though if you wanted to choose to sort by the direction.

jonatack:
Given the opposition to having a direction column, I wonder if the connection type column should be removed.  Direction and type of connection go together and displaying only half of the two has been seen by experienced users as a bug.  I think that's that wrong direction, but it is what reviewers seems to want.

jonatack:
> NACK - I prefer the existing functionality.

The lack of ability to sort by direction doesn't seem like a functionality, but ok, I'll reluctantly open a PR to remove the type column and too bad for sorting by direction.

wodry:
> Users can now sort the peers table by direction.

With the up/down arrow before the address in the same column, you can already sort by direction. The column is sorted (or you could also say "grouped" like in #13537 where the arrow prefix was introduced) by direction first (since the arrow is the first character), then by address (the following characters). You only cannot sort by address regardless of the direction, but I wonder what the use case could be for that.

jonatack:
Note that #372 (which reverts the addition of the "type" column) or this PR ought to be merged for v22.0 to avoid confusing users with a half-implemented feature that may look like a bug.

jarolrod:
@luke-jr 
> This still takes up an annoying chunk of horizontal size. #363 ?

#256 now allows a user to customize the widths of the table columns. It does not take up an annoying chunk of horizontal size if the user has resized the columns to their preference.

@rebroad 
> NACK - I prefer the existing functionality. 

The existing functionality of *no* Direction column? 🤔

ShaMan239 (commented):
tACK 6971e790c32253ecddb6108e796afd3892ced7fe

Compiled and tested successfully on Ubuntu 20.04

Displaying Connection Direction would help clarify some confusion regarding connection direction, as was suggested in #289. I like the approach of this PR over #363, because for the following reasons:
- The horizontal arrows do not have a strong connotation of IN and OUT as compared to vertical arrows.
- Without mentioning the purpose of the left and right arrows in the column header, it creates more confusion to a new user than clearing it.
- Finally, the aesthetics of the arrow in #363 does not match with the rest of the table.

I would, however, like to make a suggestion. Instead of removing the arrow, they should be moved to the Direction Column. And also‘OUTBOUND’ and ‘INBOUND’ should be replaced by ‘OUT’ and ‘IN” as it is done [here](https://github.com/bitcoin-core/gui/pull/363#pullrequestreview-686734381) (under PR alternative diff header in the table) as it would convey the information correctly at the same time being aesthetically more pleasing.

hebasto:
@promag
> No strong opinion, just think that with arrows it was more compact and clean.

Arrows "up" and "down" have no meaning about the connection direction.

hebasto:
@wodry 
> You only cannot sort by address regardless of the direction, but I wonder what the use case could be for that.

To check if one's node has more than one peers from a particular subnet.

hebasto:
@rebroad 
> NACK - I prefer the existing functionality. It could be possible to have a 4-click sort method on the existing column though if you wanted to choose to sort by the direction.

What is a "4-click sort method"?

hebasto (approved):
ACK 6971e790c32253ecddb6108e796afd3892ced7fe, tested on Linux Mint 20.2 (Qt 5.12.8).

This change is an improvement for the following reasons:
- both "Direction" and "Type" columns are coherent with peer details
- the "Address" column is free from independent data
- it is possible to sort by address and by direction independently
-  "up" and "down" arrows with no implicit or ubiquitous semantics of "inbound" and "outbound" have been replaced with the texts that are clear and understandable for everyone, including new users

I do not consider taking horizontal space by the new "Direction" column as a blocker for approving this PR.

A screenshot from my system:

Screenshot from 2021-08-08 21-41-12

It looks really great!

wodry:
I like #401 very much, please see my suggestion B.3 there.

An empty connection type I find confusing. If the user sees (like I did with 22.0rc2), that always only all inbound connections have an empty type, the user thinks, OK I guess it's not detectable or something for inbound. But it leaves uncertainty is just confusing.

RandyMcMillan:
Here is an example of a simple arrow based direction column for comparison.

Using that much visual real estate for direction seems impractical.

https://github.com/bitcoincore-dev/gui/pull/1

()

hebasto:
@RandyMcMillan 
> Here is an example of a simple arrow based direction column for comparison.

What does mean "↑" in the column with unnamed header title?

RandyMcMillan:
one possible option is to put a unicode earth symbol in the column title. 

🜨

This is both prudent on visual real-estate and universally recognized. 

arrow up (as in upload) means outbound data to most people these days. 

arrow down (as in download) means inbound data these days.

hebasto:
@RandyMcMillan 
> arrow up (as in upload) means outbound data to most people these days.
> 
> arrow down (as in download) means inbound data these days.

Upload/download are about a direction of _sending_ data.
Inbound/outbound are about a connection _initiator_.

Let's don't add more mess :)

wodry:
@RandyMcMillan -1 for the unicode symbol. I have never seen this (cross?) symbol, it would have no meaning for me in the column header.

RandyMcMillan:
I was using the analogy to express an idea and possible approach to visual communication with the node operator.
I apologize if something is getting lost in translation.
I agree with you - a mess it is.

RandyMcMillan:
I suspect the long term solution is to leverage existing symbolic/unicode libraries to communicate with the node operator.
This is provide a lot of information about the node "at a glance" and minimize the need to translate.
Using a symbolic library has the added benefit of being prudent in its use of gui real-estate.
This will also affect on-going Android mobile GUI design. 

Reference: https://google.github.io/material-design-icons/

Explaining fundamental GUI designs and concepts is fun.  :) (not really)

wodry:
Please see my updated https://github.com/bitcoin-core/gui/issues/401#issuecomment-894964337 in the brainstorming issue, with a small commit and a screenshot. Please comment what you think of it.

RandyMcMillan:
Aa I mentioned in the comment below - drawing inspiration from the established iconography of git is going to go a long way.
In fact reusing the established icons for some gui elements may avoid reinventing the wheel on this...

https://github.com/bitcoin-core/gui/issues/401#issuecomment-895557567

REF: https://github.com/primer/octicons/tree/main/icons

src/qt/peertablemodel.cpp
```diff
@@ -113,8 +113,13 @@ QVariant PeerTableModel::data(const QModelIndex &index, int role) const
         case NetNodeId:
             return (qint64)rec->nodeStats.nodeid;
         case Address:
-            // prepend to peer address down-arrow symbol for inbound connection and up-arrow for outbound connection
-            return QString::fromStdString((rec->nodeStats.fInbound ? "↓ " : "↑ ") + rec->nodeStats.addrName);
+            return QString::fromStdString(rec->nodeStats.addrName);
+        case Direction:
+            return QString(rec->nodeStats.fInbound ?
+                               //: An Inbound Connection from a Peer.
+                               tr("Inbound") :
```
rebroad, line 120: why not continue using the up and down arrows?

hebasto: @rebroad 

How could the "down arrow" could mean a "connection initiated by a peer" for a new user?

PR #275, opened 2021-04-09
Support runtime appearance adjustment on macOS

hebasto:
On macOS switching appearance (Light -> Dark or Dark -> Light) when Bitcoin Core is running makes the GUI pretty unusable.
This bug is especially important when a user chose the "Auto" mode to adjust appearance automatically.

This PR fixes Bitcoin Core behavior.

This is an alternative to #268.

PR #319, opened 2021-05-05
Paste button in Open URI dialog

kristapsk:
Picking up https://github.com/bitcoin/bitcoin/pull/17955, with some review comments addressed.

### DIFF
--- a/src/qt/forms/openuridialog.ui
+++ b/src/qt/forms/openuridialog.ui
@@ -30,6 +30,27 @@
        </property>
       </widget>
      </item>
+     <item>
+      <widget class="QToolButton" name="pasteButton">
+       <property name="toolTip">
+        <string extracomment="Tooltip text for button that allows you to paste an address that is in your clipboard.">Paste address from clipboard</string>
+       </property>
+       <property name="text">
+        <string/>
+       </property>
+       <property name="icon">
+        <iconset resource="../bitcoin.qrc">
+         <normaloff>:/icons/editpaste</normaloff>:/icons/editpaste
+        </iconset>
+       </property>
+       <property name="iconSize">
+        <size>
+         <width>22</width>
+         <height>22</height>
+        </size>
+       </property>
+      </widget>
+     </item>
     </layout>
    </item>
    <item>
--- a/src/qt/openuridialog.cpp
+++ b/src/qt/openuridialog.cpp
@@ -6,15 +6,20 @@
 #include <qt/forms/ui_openuridialog.h>
 
 #include <qt/guiutil.h>
+#include <qt/platformstyle.h>
 #include <qt/sendcoinsrecipient.h>
 
+#include <QAbstractButton>
+#include <QLineEdit>
 #include <QUrl>
 
-OpenURIDialog::OpenURIDialog(QWidget *parent) :
-    QDialog(parent, GUIUtil::dialog_flags),
-    ui(new Ui::OpenURIDialog)
+OpenURIDialog::OpenURIDialog(const PlatformStyle* platformStyle, QWidget* parent) : QDialog(parent, GUIUtil::dialog_flags),
+                                                                                    ui(new Ui::OpenURIDialog),
+                                                                                    m_platform_style(platformStyle)
 {
     ui->setupUi(this);
+    ui->pasteButton->setIcon(m_platform_style->SingleColorIcon(":/icons/editpaste"));
+    QObject::connect(ui->pasteButton, &QAbstractButton::clicked, ui->uriEdit, &QLineEdit::paste);
 
     GUIUtil::handleCloseWindowShortcut(this);
 }
@@ -32,11 +37,19 @@ QString OpenURIDialog::getURI()
 void OpenURIDialog::accept()
 {
     SendCoinsRecipient rcp;
-    if(GUIUtil::parseBitcoinURI(getURI(), &rcp))
-    {
+    if (GUIUtil::parseBitcoinURI(getURI(), &rcp)) {
         /* Only accept value URIs */
         QDialog::accept();
     } else {
         ui->uriEdit->setValid(false);
     }
+}
+
+void OpenURIDialog::changeEvent(QEvent* e)
+{
+    if (e->type() == QEvent::PaletteChange) {
+        ui->pasteButton->setIcon(m_platform_style->SingleColorIcon(":/icons/editpaste"));
+    }
+
+    QDialog::changeEvent(e);
 }
--- a/src/qt/openuridialog.h
+++ b/src/qt/openuridialog.h
@@ -7,6 +7,8 @@
 
 #include <QDialog>
 
+class PlatformStyle;
+
 namespace Ui {
     class OpenURIDialog;
 }
@@ -16,16 +18,19 @@ class OpenURIDialog : public QDialog
     Q_OBJECT
 
 public:
-    explicit OpenURIDialog(QWidget *parent);
+    explicit OpenURIDialog(const PlatformStyle* platformStyle, QWidget* parent);
     ~OpenURIDialog();
 
     QString getURI();
 
 protected Q_SLOTS:
     void accept() override;
+    void changeEvent(QEvent* e) override;
 
 private:
-    Ui::OpenURIDialog *ui;
+    Ui::OpenURIDialog* ui;
+
+    const PlatformStyle* m_platform_style;
 };
 
 #endif // BITCOIN_QT_OPENURIDIALOG_H
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -872,7 +872,7 @@ void BitcoinGUI::showHelpMessageClicked()
 #ifdef ENABLE_WALLET
 void BitcoinGUI::openClicked()
 {
-    OpenURIDialog dlg(this);
+    OpenURIDialog dlg(platformStyle, this);
     if(dlg.exec())
     {
         Q_EMIT receivedURI(dlg.getURI());


src/qt/forms/openuridialog.ui
```diff
@@ -30,6 +30,23 @@
...
+       </property>
+       <property name="text">
+        <string/>
+       </property>
+       <property name="icon">
+        <iconset resource="../bitcoin.qrc">
+         <normaloff>:/icons/editpaste</normaloff>:/icons/editpaste</iconset>
+        </property>
+        <property name="shortcut">
+         <string>Alt+P</string>
```
jarolrod, line 46: not sure how I feel about this shortcut. How intuitive/practical is this?

jarolrod (commented):
Concept ACK

While this looks ok on Linux:

Screen Shot 2021-05-05 at 10 14 42 PM

macOS behaves differently when appropriate `maximumSize` property with values for `width` and `height` are not set on Qt > 5.9, iirc correctly from my testing this was not a problem when we were using Qt 5.9.8:

Screen Shot 2021-05-05 at 1 16 12 PM

You'll want to play around with `width` and `height` until you get something that looks good.

src/qt/forms/openuridialog.ui
hebasto: https://github.com/bitcoin/bitcoin/pull/17955#discussion_r368223931
https://github.com/bitcoin/bitcoin/pull/17955#discussion_r368906725

src/qt/openuridialog.h
```diff
@@ -7,6 +7,8 @@
 
 #include <QDialog>
 
+#include <qt/platformstyle.h>
```
hebasto, line 10: https://github.com/bitcoin/bitcoin/pull/17955/files#r368907744:
> Just forward declare `PlatformStyle`.

src/qt/openuridialog.h
```diff
@@ -16,7 +18,7 @@ class OpenURIDialog : public QDialog
     Q_OBJECT
 
 public:
-    explicit OpenURIDialog(QWidget *parent);
+    explicit OpenURIDialog(const PlatformStyle* platformStyle, QWidget *parent);
```
hebasto, line 21: style nit:
```suggestion
    explicit OpenURIDialog(const PlatformStyle* platformStyle, QWidget* parent);
```

src/qt/openuridialog.cpp
```diff
@@ -8,13 +8,16 @@
 #include <qt/guiutil.h>
 #include <qt/sendcoinsrecipient.h>
 
+#include <QClipboard>
 #include <QUrl>
 
-OpenURIDialog::OpenURIDialog(QWidget *parent) :
+OpenURIDialog::OpenURIDialog(const PlatformStyle *platformStyle, QWidget *parent) :
```
hebasto, line 14: style nit:
```suggestion
OpenURIDialog::OpenURIDialog(const PlatformStyle* platformStyle, QWidget* parent) :
```

hebasto (commented):
Concept ACK.

The icon is set both in the `*.ui` and in the `*.cpp` files. Is this required?

kristapsk:
> You'll want to play around with `width` and `height` until you get something that looks good.

I don't have a Mac where to test that.

src/qt/openuridialog.h
kristapsk: You mean forward declare it here, and move this include to openuridialog.cpp? Otherwise there is "invalid use of incomplete type" compiler error.

hebasto: > You mean forward declare it here, and move this include to openuridialog.cpp?

Yeap.

hebasto:
Btw, the following patch
```diff
--- a/src/qt/forms/openuridialog.ui
+++ b/src/qt/forms/openuridialog.ui
@@ -81,7 +81,9 @@
    <header>qt/qvalidatedlineedit.h</header>
   </customwidget>
  </customwidgets>
- <resources/>
+ <resources>
+  <include location="../bitcoin.qrc"/>
+ </resources>
  <connections>
   <connection>
    <sender>buttonBox</sender>
```
makes the added icon available in Qt Designer.

hebasto:
> > You'll want to play around with `width` and `height` until you get something that looks good.
> 
> I don't have a Mac where to test that.

This issue could be fixed by replacing `QPushButton` with `QToolButton`.
Also suggesting to use 22x22 icon size for consistency with other cases of this icon usage.

In total:
```diff
diff --git a/src/qt/forms/openuridialog.ui b/src/qt/forms/openuridialog.ui
index 981d9255e..15c0a440e 100644
--- a/src/qt/forms/openuridialog.ui
+++ b/src/qt/forms/openuridialog.ui
@@ -31,7 +31,7 @@
       </widget>
      </item>
      <item>
-      <widget class="QPushButton" name="pasteButton">
+      <widget class="QToolButton" name="pasteButton">
        <property name="toolTip">
         <string>Paste address from clipboard</string>
        </property>
@@ -41,10 +41,16 @@
        <property name="icon">
         <iconset resource="../bitcoin.qrc">
          <normaloff>:/icons/editpaste</normaloff>:/icons/editpaste</iconset>
-        </property>
-        <property name="shortcut">
-         <string>Alt+P</string>
-        </property>
+       </property>
+       <property name="iconSize">
+        <size>
+         <width>22</width>
+         <height>22</height>
+        </size>
+       </property>
+       <property name="shortcut">
+        <string>Alt+P</string>
+       </property>
       </widget>
      </item>
     </layout>
diff --git a/src/qt/openuridialog.cpp b/src/qt/openuridialog.cpp
index ef4678a6a..131be0f4f 100644
--- a/src/qt/openuridialog.cpp
+++ b/src/qt/openuridialog.cpp
@@ -8,7 +8,9 @@
 #include <qt/guiutil.h>
 #include <qt/sendcoinsrecipient.h>
 
+#include <QAbstractButton>
 #include <QClipboard>
+#include <QLineEdit>
 #include <QUrl>
 
 OpenURIDialog::OpenURIDialog(const PlatformStyle* platformStyle, QWidget* parent) :
@@ -16,8 +18,7 @@ OpenURIDialog::OpenURIDialog(const PlatformStyle* platformStyle, QWidget* parent
     ui(new Ui::OpenURIDialog)
 {
     ui->setupUi(this);
-    ui->pasteButton->setIcon(platformStyle->SingleColorIcon(":/icons/editpaste"));
-    QObject::connect(ui->pasteButton, &QPushButton::clicked, ui->uriEdit, &QLineEdit::paste);
+    QObject::connect(ui->pasteButton, &QAbstractButton::clicked, ui->uriEdit, &QLineEdit::paste);
 
     GUIUtil::handleCloseWindowShortcut(this);
 }
```

On macOS Big Sur 11.3.1 (20E241):

Screenshot from 2021-05-09 19-23-38

RandyMcMillan:
It may be useful to truncate the send value so that the user still has to manually input a send value.
This may help prompt the user to verify the target address? To minimize any hijacking schemes?

I am also wondering if there is a way to block access to this dialogue from AppleScript events. May be a security issue?

src/qt/forms/openuridialog.ui
promag: NACK the shortcut, see #17955 review.

jarolrod:
@hebasto In regards to https://github.com/bitcoin-core/gui/pull/319#issuecomment-835838387

I like the approach, except the part where we remove the following line: 
``` diff
- ui->pasteButton->setIcon(platformStyle->SingleColorIcon(":/icons/editpaste"));
```

Without the above line and using:

``` diff
- <resources/>
+ <resources>
+  <include location="../bitcoin.qrc"/>
+ </resources>
```

The icon will not be colorized white when using macOS dark mode
Screen Shot 2021-05-11 at 4 33 03 PM


If we keep the line, the icon will be colorized white when starting in macOS dark mode
Screen Shot 2021-05-11 at 4 28 07 PM

@kristapsk 

I would suggest to change the shortcut to `Ctrl+v`

``` diff
        <property name="shortcut">
-         <string>Alt+P</string>
+         <string>Ctrl+v</string>
        </property>
```

hebasto:
> I would suggest to change the shortcut to `Ctrl+v`

Why? Add any shortcut to the button, if `Ctrl+v` is the default shortcut for `QLineEdit`-based widgets?
It works right now, on master :)

jarolrod:
> Why add any shortcut to the button, if `Ctrl+v` is the default shortcut for `QLineEdit`-based widgets?
> It works right now, on master :)

right 🤦. Let's remove the shortcut on the `pasteButton`

kristapsk:
Addressed review comments above, this is ready for another review.

kristapsk:
@RandyMcMillan Your comment is out of scope of this PR, this just adds visual button to existing paste functionality.

hebasto:
After #275 is merged, we should be sure that every new added widget supports runtime appearance adjustment on macOS. That is not the case now:

Screenshot from 2021-06-06 13-06-15

Also there is no need to adjust copyright year here, because we have a dedicated [script](https://github.com/bitcoin-core/gui/tree/master/contrib/devtools#copyright_headerpy) which is ran once per year.

src/qt/openuridialog.cpp
```diff
@@ -1,20 +1,26 @@
...
 
 #include <qt/openuridialog.h>
 #include <qt/forms/ui_openuridialog.h>
 
 #include <qt/guiutil.h>
+#include <qt/platformstyle.h>
 #include <qt/sendcoinsrecipient.h>
 
+#include <QAbstractButton>
+#include <QClipboard>
```
luke-jr, line 13: Is this needed?

kristapsk:
Addressed comments by @hebasto and @luke-jr.

kristapsk:
Modified `changeEvent` to match #366.

src/qt/forms/openuridialog.ui
```diff
@@ -30,6 +30,27 @@
        </property>
       </widget>
      </item>
+     <item>
+      <widget class="QToolButton" name="pasteButton">
+       <property name="toolTip">
+        <string>Paste address from clipboard</string>
```
jarolrod, line 36: we can add translator comments by filling in the `extracomment` field:
```suggestion
        <string extracomment="Tooltip text for button that allows you to paste an address that is in your clipboard.">Paste address from clipboard</string>
```

jarolrod (commented):
in regards to https://github.com/bitcoin-core/gui/pull/319#issuecomment-872039063:
> Modified changeEvent to match #366.

This doesn't include https://github.com/bitcoin-core/gui/pull/275. Can you rebase to include this. This allows for easier testing and might be a cause for the CI failure.

ShaMan239 (commented):
Concept ACK
Successfully compiled and tested on Ubuntu 20.04

This PR adds a button to the **openbitcoinuri** dialog box that allows the pasting of URI in the QLineEdit box. This is done by adding a button to the **openbitcoinuri** dialog box, an object of the **platformStyle** class, which allows setting a custom icon on the button. Finally, the pasting of URI is connected with pressing of this button, resulting in given the button its functionality.
I am adding a screenshot to show the correct working of PR.
image1


As already mentioned by @jarolrod, OP should fix the CI formatting. Also, OP should rebase this PR on the current master head to allow the proper test of the UI changing on macOS

src/qt/openuridialog.cpp
```diff
@@ -6,15 +6,21 @@
...
 #include <qt/guiutil.h>
+#include <qt/platformstyle.h>
 #include <qt/sendcoinsrecipient.h>
 
+#include <QAbstractButton>
+#include <QLineEdit>
 #include <QUrl>
 
-OpenURIDialog::OpenURIDialog(QWidget *parent) :
+OpenURIDialog::OpenURIDialog(const PlatformStyle* platformStyle, QWidget* parent) :
```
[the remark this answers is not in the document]
hebasto: @kristapsk 

https://github.com/bitcoin/bitcoin/tree/master/contrib/devtools#clang-format-diffpy

Just run `alias clang-format-diff='git diff -U0 HEAD~1.. | ./contrib/devtools/clang-format-diff.py -p1 -i -v'` after each commit, then `git commit --all --amend` to apply suggested changes.

kristapsk:
Addressed additional review comments, this ready for another review.

shaavan (approved):
tACK dbde055
Changes since my last review:
- The formatting of the code has been fixed
- The PR is rebased over the current master

I retested this PR, and the paste button is working perfectly.

Though it might just be because of the screenshot software difference, but the paste button looks slightly different since I last reviewed this PR.
|Old Commit|Latest Commit|
|----------------|-------------------|
|image1 | prnew|

Btw if it is a deliberate change, the new button looks a lot better!

jarolrod (commented):
ACK dbde055

Tested on macOS 12 and on Ubuntu. Confirmed that this works as intended. Also confirmed that this responds to macOS theme switching.

| Start Dark | Switch to Light |
|---------------|---------------------|
| Screen Shot 2021-11-02 at 1 05 29 AM | Screen Shot 2021-11-02 at 1 05 37 AM |

promag (commented):
Tested ACK dbde0558ce73db4c901dbaa2eddc634701fa1d0d.

PR #320, opened 2021-05-06
WIP qt: mempool stats chart 

RandyMcMillan:
add graphical mempool display to Node window

### DIFF
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -22,6 +22,7 @@ QT_FORMS_UI = \
   qt/forms/helpmessagedialog.ui \
   qt/forms/intro.ui \
   qt/forms/modaloverlay.ui \
+  qt/forms/mempoolstats.ui \
   qt/forms/openuridialog.ui \
   qt/forms/optionsdialog.ui \
   qt/forms/overviewpage.ui \
@@ -54,6 +55,7 @@ QT_MOC_CPP = \
   qt/moc_intro.cpp \
   qt/moc_macdockiconhandler.cpp \
   qt/moc_macnotificationhandler.cpp \
+  qt/moc_mempoolstats.cpp \
   qt/moc_modaloverlay.cpp \
   qt/moc_notificator.cpp \
   qt/moc_openuridialog.cpp \
@@ -126,6 +128,7 @@ BITCOIN_QT_H = \
   qt/macdockiconhandler.h \
   qt/macnotificationhandler.h \
   qt/macos_appnap.h \
+  qt/mempoolstats.h \
   qt/modaloverlay.h \
   qt/networkstyle.h \
   qt/notificator.h \
@@ -228,6 +231,7 @@ BITCOIN_QT_BASE_CPP = \
   qt/csvmodelwriter.cpp \
   qt/guiutil.cpp \
   qt/intro.cpp \
+  qt/mempoolstats.cpp \
   qt/modaloverlay.cpp \
   qt/networkstyle.cpp \
   qt/notificator.cpp \
--- a/src/interfaces/node.h
+++ b/src/interfaces/node.h
@@ -49,6 +49,24 @@ struct BlockAndHeaderTipInfo
     double verification_progress;
 };
 
+class mempool_feeinfo {
+public:
+    uint64_t total_size;
+    uint64_t total_fee;
+    uint64_t tx_count;
+    CAmount fee_from;
+    CAmount fee_to;
+
+    //TODO: remove
+    // added for storing and loading a mempool set during development to avoid waiting hours for collecting enought samples
+    SERIALIZE_METHODS(mempool_feeinfo, obj)
+    {
+        READWRITE(obj.total_size, obj.total_fee, obj.tx_count, obj.fee_from, obj.fee_to);
+    }
+};
+
+typedef std::vector<mempool_feeinfo> mempool_feehistogram;
+
 //! Top-level interface for a bitcoin node (bitcoind process).
 class Node
 {
@@ -121,6 +139,9 @@ public:
 
     //! Get mempool dynamic usage.
     virtual size_t getMempoolDynamicUsage() = 0;
+
+    //! Get mempool fee histogram
+    virtual mempool_feehistogram getMempoolFeeHistogram() = 0;
 
     //! Get header tip height and time.
     virtual bool getHeaderTip(int& height, int64_t& block_time) = 0;
--- a/src/node/interfaces.cpp
+++ b/src/node/interfaces.cpp
@@ -174,6 +174,58 @@ public:
     int64_t getTotalBytesSent() override { return m_context->connman ? m_context->connman->GetTotalBytesSent() : 0; }
     size_t getMempoolSize() override { return m_context->mempool ? m_context->mempool->size() : 0; }
     size_t getMempoolDynamicUsage() override { return m_context->mempool ? m_context->mempool->DynamicMemoryUsage() : 0; }
+    interfaces::mempool_feehistogram getMempoolFeeHistogram() override {
+         /* TODO: define log scale formular for dynamically creating the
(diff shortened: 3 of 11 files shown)


RandyMcMillan:
There is a lot if interest in #108
This is a rebase of original PR - it has grown stale.

Sjors:
Thanks for picking this up. Looks like it broke a few CI instances, probably because of the compiler warnings (that are treated as errors).

jarolrod (commented):
Concept ACK

I don't know if @jonasschnelli has completely given up on his original PR yet.

This is still very much a work in progress. So, it should be labeled as such and made into a draft. 

When you push changes, it would be nice if you could write a small post explaining what has changed between updates. This makes it easier for people to follow along.

rebroad:
Has anyone worked out how to make the Y axis show memory usage rather than tx count? (ideally in a way that it shows a flat line once we reach the maxmempool).

Sjors:
I would prefer MB over count too if it's not too difficult. I might take a look a later to see how...

Can you split this PR a bit? E.g. one commit for the `src/interfaces` changes, one for `mempoolstats.cpp` (which probably should mostly be moved out of the GUI code), one for the UI change, etc.

prayank23:
Concept ACK

> Has anyone worked out how to make the Y axis show memory usage rather than tx count? (ideally in a way that it shows a flat line once we reach the maxmempool).

Yes There should be a way to select Tx count or vSize

Not sure if others will agree with it or maybe we can do this in follow up PRs, I created a rough design of mempool tab:

image

Left: History for mempool over a time period. User can select the time frame.

Right: Current fee rate distribution in mempool. I have used image from https://btc.bitaps.com (An option to switch between Tx count and vSize can be added)

User can enter a transaction id in the text edit widget to see where it is in the mempool (similar to https://mempool.observer):

Can be highlighted in the 'fee distribution' graph with dotted lines. I have used dotted lines and "glasses" sticker:

image

0xB10C:
I agree that insights into a nodes mempool are really interesting and useful. I too understand that users want to use their own full node and not rely on external services. However, I'm not sure if the Bitcoin Core GUI is the right place for it. 

From my experience, different types of users want different stats, charts, chart axis and so forth... I think having good interfaces to access this data and then visualizing it, for example, on a web page, scales way better and requires much less maintenance on the Bitcoin Core side. I'm leaning towards a **Concept NACK** for now. 

Disclaimer: _I run such a site visualizing the mempool of my node, but I don't think this influences my reasoning here. I have no incentive to NACK here. I don't make money off site visitors. The hosting is payed by myself and donations (sometimes listed as sponsors) and I'm not compensated for my time maintaining it._

hebasto:
@0xB10C 
> I agree that insights into a nodes mempool are really interesting and useful. I too understand that users want to use their own full node and not rely on external services. However, I'm not sure if the Bitcoin Core GUI is the right place for it.
> 
> From my experience, different types of users want different stats, charts, chart axis and so forth... I think having good interfaces to access this data and then visualizing it, for example, on a web page, scales way better and requires much less maintenance on the Bitcoin Core side. I'm leaning towards a **Concept NACK** for now.
> 
> Disclaimer: _I run such a site visualizing the mempool of my node, but I don't think this influences my reasoning here. I have no incentive to NACK here. I don't make money off site visitors. The hosting is payed by myself and donations (sometimes listed as sponsors) and I'm not compensated for my time maintaining it._

Could we limit this feature to some set of data (texts, tables, charts) that just could give to a pro-user useful info to manually set a feerate for his transaction?

Rspigler:
I am a strong Concept ACK for @prayank23 's design. Improving Core's GUI and widening the user base is important.

Sjors:
Personally I'm happy to review and test this, and it's something I would use. Mempool weather is something I check before and during every transaction. Although websites can indeed iterate faster, it's nice to have it in the node. The Hoenicke design that this is based on hasn't changed in many years, so I'm also not terribly worried about maintenance.

And if we correctly split the data collection (node & rpc) from the visualisation (gui), then there may be other use cases too.

Rspigler:
:disappointed: Reason for close? Mark up for grabs?

prayank23:
@0xB10C 

> However, I'm not sure if the Bitcoin Core GUI is the right place for it.

Mempool related charts and options are a basic thing which should have been there in Bitcoin Core GUI few years back. Sad that we won't see it in v22.0 but still not too late to start working on it for future release.

> From my experience, different types of users want different stats, charts, chart axis and so forth.

Agree. But that is true for lot of things in Bitcoin Core GUI. We should add atleast basic things that everyone finds useful.

> I think having good interfaces to access this data and then visualizing it, for example, on a web page, scales way better and requires much less maintenance on the Bitcoin Core side. 

Yes, website works better for charts. However, both can be used. One or two basic charts and options in Core GUI and other detailed mempool stats on different websites.

Is https://mempool.observer open source? Can users run it locally and use their own node? That would be helpful

Rspigler:
> users want to use their own full node and not rely on external services

This.

rebroad:
> I would prefer MB over count too if it's not too difficult. I might take a look a later to see how...
> 
> Can you split this PR a bit? E.g. one commit for the `src/interfaces` changes, one for `mempoolstats.cpp` (which probably should mostly be moved out of the GUI code), one for the UI change, etc.
> 
> Also for a followup: little dots to indicate your wallet unconfirmed transactions :-)

I've managed to do this... but I cheated a little bit as there is memory used in addition to the TXs that fluctuates over time, so if you want a nice flat line when your node starts pruning TXs from the mempool, then you won't get that unless you cheat :) (which I have). I will upload my patches soon.

kristapsk:
Concept ACK

>> However, I'm not sure if the Bitcoin Core GUI is the right place for it.
>
> Mempool related charts and options are a basic thing which should have been there in Bitcoin Core GUI few years back. Sad that we won't see it in v22.0 but still not too late to start working on it for future release.

I agree with @prayank23 here, if you use Bitcoin Core as a wallet (and I don't see a reason running GUI if you don't), this is probably the most important metric to decide what transaction fee to use for transaction.

promag:
Concept ACK.

@0xB10C I think some basic mempool viz is reasonable. Will review.

promag:
Concept ACK.

@0xB10C I think some basic mempool viz is reasonable. Will review.

0xB10C:
Thanks everybody for the feedback and discussion. I've reconsidered my position on this. I agree that some basic mempool weather in the GUI makes sense.

Rspigler:
People who have (Concept) ACKd this should look at #108

prayank23:
@RandyMcMillan Do you want to reopen this? I can work on it if you have bandwidth issues

prayank23:
Why do I see a different chart and nothing on right with `Detail Stub` written?


<details>
  <summary>Screenshot</summary>

image

</details>


1. If I stretch that part to right, this is how it looks:

<details>
  <summary>Screenshot</summary>

image

</details>

Same thing on Ubuntu and Windows.

2. Chart in https://github.com/bitcoin/bitcoin/pull/8550 looks better IMO:

<details>
  <summary>Screenshot</summary>

image

</details>

3. Or maybe we can just draw a simple bar chart with https://github.com/bitcoin/bitcoin/pull/21422 and https://doc.qt.io/qt-5/qtcharts-barchart-example.html

I had used the results of `getmempoolinfo` to create a similar chart using Python Flask: https://github.com/bitcoin/bitcoin/pull/21422#issuecomment-899258623

<details>
  <summary>Screenshot</summary>

image

</details>

So which chart would look better and everyone agrees to add in debug window?

1. https://github.com/bitcoin-core/gui/pull/108 👍 
2. https://github.com/bitcoin/bitcoin/pull/8550 🚀 
3. https://github.com/bitcoin/bitcoin/pull/21422 and https://doc.qt.io/qt-5/qtcharts-barchart-example.html ❤️

RandyMcMillan:
rebase

PR #322, opened 2021-05-08
Improve Overview tab look on Windows

### LINKED ISSUE #303: Remove indented/raised borders elements on Windows
Somewhere between the last GUI update and current master 4cfe6c37d956be58e6228f50833d25b84aafe55a the Windows version of the GUI has these indented / raised border designs applied to them which has resulted in a very visually cluttered GUI. I think this style is only visually bad on Windows as other recent screenshots of master from macOS / Linux don't seem to be effected.  Below is an example from the overview section:

Group 265

It has also made the sections more condensed as I assume each widget has a size constraint and these borders are taking up extra space. It would be much more visually appealing to have this reverted back to the old style (like in the before above).

hebasto:
This PR replaces raised frames with plane ones on the Overview tab, and improves its look on Windows.

Fixes #303.

### DIFF
--- a/src/qt/forms/overviewpage.ui
+++ b/src/qt/forms/overviewpage.ui
@@ -40,10 +40,7 @@
        <item>
         <widget class="QFrame" name="frame">
          <property name="frameShape">
-          <enum>QFrame::StyledPanel</enum>
-         </property>
-         <property name="frameShadow">
-          <enum>QFrame::Raised</enum>
+          <enum>QFrame::Panel</enum>
          </property>
          <layout class="QVBoxLayout" name="verticalLayout_4">
           <item>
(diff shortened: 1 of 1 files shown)


Bosch-0:
Tested this change and the borders on the overview tab indentation is removed but now they have black borders (previously had no border element). 

image

I only showed an example of the overview page in that previous issue. This indentation effect seems to be present GUI wide not just the overview page.

hebasto:
> Tested this change and the borders on the overview tab indentation is removed but now they have black borders (previously had no border element).

Right. Now it looks the same as on other platforms. If it looks bad, why do not change it for all platforms?

> I only showed an example of the overview page in that previous issue. This indentation effect seems to be present GUI wide not just the overview page.

You are right. As a non-designer I'm not able to make conclusions about nicety of a look. But I'm able to (and do) code what were suggested by designers :)

hebasto:
Closed in favor of https://github.com/bitcoin/bitcoin/pull/22133.

PR #324, opened 2021-05-11
Peers tab hide horizontal scroll bars

RandyMcMillan:


RandyMcMillan:
The horizontal scrollbars block click access to rows in the table view and detail view.
The vertical scrollbars are useful. The horz. bars are not very useful (IMO).

### Before:



### After: 

Screen Shot 2021-05-11 at 7 28 39 PM

src/qt/rpcconsole.cpp
```diff
@@ -616,6 +616,7 @@ void RPCConsole::setClientModel(ClientModel *model, int bestblock_height, int64_
         ui->peerWidget->setColumnWidth(PeerTableModel::Subversion, SUBVERSION_COLUMN_WIDTH);
         ui->peerWidget->setColumnWidth(PeerTableModel::Ping, PING_COLUMN_WIDTH);
         ui->peerWidget->horizontalHeader()->setStretchLastSection(true);
+        ui->peerWidget->horizontalScrollBar()->setStyleSheet("QScrollBar {height:0px;}");
```
promag, line 619: Have you tried `ui->peerWidget->setHorizontalScrollBarPolicy(Qt::ScrollBarAlwaysOff)`?

promag (commented):
I don't see a reason to hide them, especially considering that the default policy is `Qt::ScrollBarAsNeeded`.

src/qt/rpcconsole.cpp
jarolrod: It's a NACK from me on removing the horizontal scroll bar from the Peers Table. The user can be left in a state where it is inconvenient to scroll right to see the rest of the content. This is a UX downgrade.

Note: if there's consensus for this change (for some reason), you missed disabling the horizontal scroll bar on the banned peers table

Screen Shot 2021-05-12 at 8 37 04 PM

src/qt/rpcconsole.cpp
```diff
@@ -658,6 +659,7 @@ void RPCConsole::setClientModel(ClientModel *model, int bestblock_height, int64_
         ui->blocksDir->setText(model->blocksDir());
         ui->startupTime->setText(model->formatClientStartupTime());
         ui->networkName->setText(QString::fromStdString(Params().NetworkIDString()));
+        ui->scrollArea->horizontalScrollBar()->setStyleSheet("QScrollBar {height:0px;}");
```
jarolrod, line 662: This should be ok to do since #293 was merged. Instead of changing the stylesheet, you should implement this suggestion: https://github.com/bitcoin-core/gui/pull/324/files#r630627901 

You should change this line as so:

```diff
-   ui->scrollArea->horizontalScrollBar()->setStyleSheet("QScrollBar {height:0px;}");
+   ui->scrollArea->setHorizontalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
```

src/qt/rpcconsole.cpp
```diff
@@ -616,6 +616,7 @@ void RPCConsole::setClientModel(ClientModel *model, int bestblock_height, int64_
         ui->peerWidget->setColumnWidth(PeerTableModel::Subversion, SUBVERSION_COLUMN_WIDTH);
         ui->peerWidget->setColumnWidth(PeerTableModel::Ping, PING_COLUMN_WIDTH);
         ui->peerWidget->horizontalHeader()->setStretchLastSection(true);
+        ui->scrollArea->setHorizontalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
```
hebasto, line 619: How does this change differ from the one in the line 663?

hebasto (commented):
Agree with @promag 
> I don't see a reason to hide them, especially considering that the default policy is `Qt::ScrollBarAsNeeded`.

and with @jarolrod 
> It's a NACK from me on removing the horizontal scroll bar from the Peers Table. The user can be left in a state where it is inconvenient to scroll right to see the rest of the content. This is a UX downgrade.

src/qt/rpcconsole.cpp
RandyMcMillan: I just didn't see that option in the docs.
A simple one-liner is obviously better. :)

RandyMcMillan:
How does/should the user toggle/hide the detail widget when the table data fills the view?

Screen Shot 2021-05-19 at 4 08 21 PM 2
Screen Shot 2021-05-19 at 4 08 55 PM 2

Maybe assigning a key command to hide/toggle the detail view would work? \<ESC\> key?
But I believe it is optimal to allow for a mouse click event as well as a key command event.

PR #325, opened 2021-05-12
Align numbers in the "Peer Id" column to the right

hebasto:
On master (6b49d88a5dda97fdbabe363fd7d3c4f1ce29082a):
Screenshot from 2021-05-12 21-53-52

With this PR:
Screenshot from 2021-05-12 21-48-19

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -37,6 +37,8 @@
 #include <QDateTime>
 #include <QFont>
 #include <QKeyEvent>
+#include <QLatin1String>
+#include <QLocale>
 #include <QMenu>
 #include <QMessageBox>
 #include <QScreen>
@@ -44,9 +46,10 @@
 #include <QSettings>
 #include <QString>
 #include <QStringList>
+#include <QStyledItemDelegate>
 #include <QTime>
 #include <QTimer>
-
+#include <QVariant>
 
 const int CONSOLE_HISTORY = 50;
 const int INITIAL_TRAFFIC_GRAPH_MINS = 30;
@@ -128,6 +131,20 @@ public:
     }
 };
 
+class PeerIdViewDelegate : public QStyledItemDelegate
+{
+    Q_OBJECT
+public:
+    explicit PeerIdViewDelegate(QObject* parent = nullptr)
+        : QStyledItemDelegate(parent) {}
+
+    QString displayText(const QVariant& value, const QLocale& locale) const override
+    {
+        // Additional spaces should visually separate right-aligned content
+        // from the next column to the right.
+        return value.toString() + QLatin1String("   ");
+    }
+};
 
 #include <qt/rpcconsole.moc>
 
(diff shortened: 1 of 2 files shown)


src/qt/rpcconsole.cpp
```diff
@@ -634,6 +651,7 @@ void RPCConsole::setClientModel(ClientModel *model, int bestblock_height, int64_
         ui->peerWidget->setColumnWidth(PeerTableModel::Subversion, SUBVERSION_COLUMN_WIDTH);
         ui->peerWidget->setColumnWidth(PeerTableModel::Ping, PING_COLUMN_WIDTH);
         ui->peerWidget->horizontalHeader()->setStretchLastSection(true);
+        ui->peerWidget->setItemDelegateForColumn(PeerTableModel::NetNodeId, new PeerIdViewDelegate(this));
```
jarolrod, line 654: I don't have a strong opinion on this. I think with and without the `PeerIdViewDelegate` both look fine:

| With Delegate     | Without Delegate |
| ---------------- | ------------------ |
| Screen Shot 2021-05-12 at 9 46 31 PM | Screen Shot 2021-05-12 at 10 09 38 PM |

jarolrod (commented):
Concept ACK

jarolrod (commented):
ACK 69b8b5d72c47d42a9e69e6666af89606057be15b

Having the `Peer Id` numbers right-aligned is a UX improvement in my opinion.

Through testing and resizing columns, it's possible for the `Peer id` and `Address` columns to look a bit cluttered. 

This PR's approach to prevent this clutter is to use a [QStyledItemDelegate](https://doc.qt.io/qt-5/qstyleditemdelegate.html), named `PeerIdViewDelegate`, which gets around this by *drawing* the value with a few spaces after the number. 

It's an important distinction that these spaces are being *drawn* and not *appended* to the `ID` value. We could instead just *append* these spaces to the following line directly and not use a delegate:
https://github.com/bitcoin-core/gui/blob/b295395664bd37e26d168c329f238237b34aef8c/src/qt/peertablemodel.cpp#L114

But, that would mess up cases where we access the `NetNodeId` value such as here:
https://github.com/bitcoin-core/gui/blob/b295395664bd37e26d168c329f238237b34aef8c/src/qt/rpcconsole.cpp#L1155

If other reviewers think the `clutter` is a non-issue, or the added complexity of a `QStyledItemDelegate` is not worth it, then this PR can just be comprised of this line:

https://github.com/bitcoin-core/gui/blob/69b8b5d72c47d42a9e69e6666af89606057be15b/src/qt/peertablemodel.cpp#L135

src/qt/rpcconsole.cpp
jonatack: Thanks for bringing this patch to my attention @jarolrod and for this helpful comparison. I prefer with the added separation between the right-aligned peer number and the left-aligned address (which is why I originally proposed center-aligned to have the spacing as well and aligned under the header that is also center-aligned, but the version proposed in this PR seems good too).

jonatack:
Concept ACK. I wanted to change this column away from left-alignment as well, so thank you for proposing this.

jonatack (commented):
Tested rebased to master ("Peer" header, this PR could be renamed) and with the arrows removed from the neighboring address column to test the case of them being replaced by a dedicated Direction column.

Center-aligned:

Screenshot from 2021-05-28 11-51-04

Right-aligned per this patch:

Screenshot from 2021-05-28 11-56-23

I think this is an improvement, though maybe a bit more separation from the Address column might be good and I do prefer center-aligned for overall readability and balance in the design.
 
ACK 69b8b5d72c47d42a9e69e6666af89606057be15b happy to re-ack with more separation or center-aligned

kristapsk (approved):
ACK 69b8b5d72c47d42a9e69e6666af89606057be15b

hebasto:
@jonatack 
> happy to re-ack with more separation...

Let's collect some feedback with current one :)

> ... or center-aligned

I'm really opposite against center-aligned ids, as it ends with such a weird layout:

Screenshot from 2021-06-05 15-14-05

PR #217, opened 2021-02-22
qt: Make warning label look clickable

jarolrod:
The warning icon on the overview page indicates that there is something important the user should know about, but a user may not be aware that they can click it because, on `master`, the warning label does not look clickable. As detailed in issue #23, the reason to make it look clickable is that it if they "had a more clickable-appearance (borders or beveled button edges) it could help users more quickly understand what they are being alerted to." 

This PR removes the `flat` property from both `QPushButton`'s to make them look like a button, and therefore clickable. Furthermore, it updates the `Maximum Width` to `45` to fix the small hit-box issue outlined in issue #215.

Below are screenshots showing how the warning icon looks under `master` and this `PR`:

**macOS 11.1: Qt 5.15** 
| Master        | PR               |
| ----------- | ----------- |
|  Screen Shot 2021-02-22 at 5 00 40 PM | Screen Shot 2021-02-22 at 3 08 40 PM |

**Ubuntu 20.04: Qt 5.12**

| Master        | PR               |
| ----------- | ----------- |
| Screen Shot 2021-02-22 at 4 57 32 PM |   Screen Shot 2021-02-22 at 4 12 54 PM |

Closes #23 
Closes #215

PR #329, opened 2021-05-14
Make console buttons look clickable

jarolrod:
On master, for macOS, the console buttons' hitboxes are quite small. This makes clicking on the button with your mouse a little more tedious than it should be. The Issue is related to recent versions of Qt (>5.9.8) not playing so nice on macOS when there are "incorrect" `width` and `height` values set for a `QPushButton` (here is another example: https://github.com/bitcoin-core/gui/pull/319#pullrequestreview-652907740). 

This fixes this small hitbox issue by converting the buttons from `QPushButton` to `QToolButton`, which in turn makes the buttons look explicitly clickable. This approach was chosen as it helps us avoid having to play around with `width` and `height` values until we find values that play nice with macOS and look good on Linux & Windows. Also, `QToolButton` is an appropriate class for these buttons. 

Per [Qt Docs](https://doc.qt.io/qt-5/qtoolbutton.html#details): 
> A tool button is a special button that provides quick-access to specific commands or options. As opposed to a normal command button, a tool button usually doesn't show a text label, but shows an icon instead.

Since we are changing the type of the buttons, we need to change the respective actions connection logic in `rpcconsole`. Instead of plugging in `QToolButton`, we abstract it to the base class: `QAbstractButton`.

per [Qt Dev Notes](https://github.com/bitcoin-core/bitcoin-devwiki/wiki/Developer-Notes-for-Qt-Code#inherited-signals-and-slot)
> Use base class functions as this makes the code more general, e.g., use QAbstractButton::clicked instead of QPushButton::clicked.

While here, we also update the size of the icons to `22x22` to be consistent with other tool buttons.

**macOS: Master vs PR:**

| ----------- | ----------- |
| master-ss-macos | pr-ss-macos |

**Linux: Master vs PR:**

| ----------- | ----------- |
| master-ss-linux | pr-ss-linux |

### DIFF
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -470,13 +470,7 @@
           </spacer>
          </item>
          <item>
-          <widget class="QPushButton" name="fontSmallerButton">
-           <property name="maximumSize">
-            <size>
-             <width>24</width>
-             <height>24</height>
-            </size>
-           </property>
+          <widget class="QToolButton" name="fontSmallerButton">
            <property name="toolTip">
             <string>Decrease font size</string>
            </property>
@@ -489,26 +483,14 @@
            </property>
            <property name="iconSize">
             <size>
-             <width>24</width>
-             <height>16</height>
+             <width>22</width>
+             <height>22</height>
             </size>
-           </property>
-           <property name="autoDefault">
-            <bool>false</bool>
-           </property>
-           <property name="flat">
-            <bool>true</bool>
            </property>
           </widget>
          </item>
          <item>
-          <widget class="QPushButton" name="fontBiggerButton">
-           <property name="maximumSize">
-            <size>
-             <width>24</width>
-             <height>24</height>
-            </size>
-           </property>
+          <widget class="QToolButton" name="fontBiggerButton">
            <property name="toolTip">
             <string>Increase font size</string>
            </property>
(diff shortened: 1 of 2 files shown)


hebasto:
Concept ACK on making clickable elements explicit (see #217).

> On master, for macOS, the console buttons' hitboxes are quite small

How do your changes increase their sizes?

jarolrod:
@hebasto 

> How do your changes increase their sizes?

It seems a better name for this PR would be "Make Console Buttons Look Clickable" as that is what it really does for all systems. The PR was motivated by the small hitbox bug on macOS, this is fixed by making the buttons a `QToolButton` instead of finding appropriate `width` and `height` values for a`QPushButton`.

jarolrod:
updated from 05808da -> 8b419b5:

Changes:
- Change naming of commit and pr to better represent what the PR is doing

hebasto (approved):
ACK 8b419b5163d41c0caef7b9fa6d3008b73917e8ee, tested on Linux Mint 20.1 (Qt 5.12.8).

promag:
Not a NACK, personally I think they are fine as is.

promag (commented):
Tested ACK 8b419b5163d41c0caef7b9fa6d3008b73917e8ee on macOS Big Sur M1, this drops only relevant usages to `flat` buttons.

nit, could drop the following:
https://github.com/bitcoin-core/gui/blob/aedf71dcc54f56ba8ba93274add2f40ed26cafd4/src/qt/forms/psbtoperationsdialog.ui#L94-L96

promag:
Played around and with
```diff
          </item>
          <item>
           <widget class="QPushButton" name="fontSmallerButton">
-           <property name="maximumSize">
-            <size>
-             <width>24</width>
-             <height>24</height>
-            </size>
-           </property>
            <property name="toolTip">
             <string>Decrease font size</string>
            </property>
@@ -490,7 +484,7 @@
            <property name="iconSize">
             <size>
              <width>24</width>
-             <height>16</height>
+             <height>24</height>
             </size>
            </property>
            <property name="autoDefault">
```
I get this
Screenshot 2021-06-01 at 00 30 31
and when clicking
Screenshot 2021-06-01 at 00 30 50
So it seems that `maximumSize` is screwing the button layout.

PR #337, opened 2021-05-17
test: Use Regex Search in Apptests

jarolrod:
This PR refactors our GUI `apptests` so that it uses regex search to find values in our console/qtextedit output regardless if it is in `plaintext`, `html`, or `markdown`. 

This introduces a new function `FindInConsole` which uses [QRegularExpression](https://doc.qt.io/qt-5/qregularexpression.html) to search the output of the console. The function must be provided with a [perl compatible regex](https://www.debuggex.com/cheatsheet/regex/pcre) pattern which wants to match a single group. The function then returns the matched group. If no match is found, an empty `QString` is returned.

We then use this new function in `TestRpcCommand` to find the current `chain` value instead of reading with univalue. 

This approach can apply to a wider variety of testing scenarios as we can reuse this function to search for values when the console output is exported in a different format than `plaintext`. As an example, A follow up PR will add tests for console resizing and needs to look for the size in `html` tags after exporting the console text with `toHtml()`.

### DIFF
--- a/src/qt/test/apptests.cpp
+++ b/src/qt/test/apptests.cpp
@@ -21,6 +21,7 @@
 
 #include <QAction>
 #include <QLineEdit>
+#include <QRegularExpression>
 #include <QScopedPointer>
 #include <QSignalSpy>
 #include <QTest>
@@ -30,6 +31,13 @@
 #include <QtTest/QtTestGui>
 
 namespace {
+//! Regex find a string group inside of the console output
+QString FindInConsole(const QString& output, const QString& pattern)
+{
+    const QRegularExpression re(pattern);
+    return re.match(output).captured(1);
+}
+
 //! Call getblockchaininfo RPC and check first field of JSON output.
 void TestRpcCommand(RPCConsole* console)
 {
--- a/src/qt/test/apptests.cpp
+++ b/src/qt/test/apptests.cpp
@@ -12,7 +12,6 @@
 #include <qt/rpcconsole.h>
 #include <shutdown.h>
 #include <test/util/setup_common.h>
-#include <univalue.h>
 #include <validation.h>
 
 #if defined(HAVE_CONFIG_H)
@@ -24,6 +23,7 @@
 #include <QRegularExpression>
 #include <QScopedPointer>
 #include <QSignalSpy>
+#include <QString>
 #include <QTest>
 #include <QTextEdit>
 #include <QtGlobal>
@@ -49,10 +49,9 @@ void TestRpcCommand(RPCConsole* console)
     QTest::keyClick(lineEdit, Qt::Key_Return);
     QVERIFY(mw_spy.wait(1000));
     QCOMPARE(mw_spy.count(), 4);
-    QString output = messagesWidget->toPlainText();
-    UniValue value;
-    value.read(output.right(output.size() - output.lastIndexOf(QChar::ObjectReplacementCharacter) - 1).toStdString());
-    QCOMPARE(value["chain"].get_str(), std::string("regtest"));
+    const QString output = messagesWidget->toPlainText();
+    const QString pattern = QStringLiteral("\"chain\": \"(\\w+)\"");
+    QCOMPARE(FindInConsole(output, pattern), QString("regtest"));
 }
 } // namespace
 


jarolrod:
updated from d32b704 -> d342630

Changes:
- Address [CI failure](https://cirrus-ci.com/task/6137803319803904?logs=ci#L4339) by explicitly making the "regtest" string a `QString`
  - Note: Using `QStringLiteral` would be more efficient since it avoids constructing a `QString` and copying data to it. But we are not in a loop, the performance benefit is minimal, and just `QString` is cleaner.
- While here, include missing `QString` header

src/qt/test/apptests.cpp
```diff
@@ -22,6 +22,7 @@
 #include <QAction>
 #include <QEventLoop>
 #include <QLineEdit>
+#include <QRegularExpression>
```
hebasto, line 24: https://github.com/bitcoin/bitcoin/pull/17213#issuecomment-546684680

jarolrod:
updated from d342630 -> 1a63f44 ([pr337.01](https://github.com/jarolrod/gui/commits/pr337.01) -> [pr337.02](https://github.com/jarolrod/gui/commits/pr337.02))

Changes:
- Rebase due to conflict with #335

src/qt/test/apptests.cpp
hebasto: I verified that Qt itself uses `QRegularExpression` header, at least for version we use in depends.

So, I'm ok with it.

src/qt/test/apptests.cpp
```diff
@@ -41,10 +49,9 @@ void TestRpcCommand(RPCConsole* console)
     QTest::keyClick(lineEdit, Qt::Key_Return);
     QVERIFY(mw_spy.wait(1000));
     QCOMPARE(mw_spy.count(), 2);
-    QString output = messagesWidget->toPlainText();
-    UniValue value;
-    value.read(output.right(output.size() - output.lastIndexOf(QChar::ObjectReplacementCharacter) - 1).toStdString());
-    QCOMPARE(value["chain"].get_str(), std::string("regtest"));
+    const QString output = messagesWidget->toPlainText();
+    const QString pattern = "\"chain\": \"(\\w+)\"";
```
hebasto, line 53: nit:
```suggestion
    const QString pattern = QStringLiteral("\"chain\": \"(\\w+)\"");
```

hebasto (commented):
Approach ACK 1a63f44609d1c7a4de0f5c4727e0d4cd3d7448ad.

jarolrod:
updated from 1a63f44 -> 9bde381 ([pr337.02](https://github.com/jarolrod/gui/tree/pr337.02), [pr337.03](https://github.com/jarolrod/gui/tree/pr337.03), [diff](https://github.com/jarolrod/gui/compare/pr337.02...jarolrod:pr337.03))

**Changes:**
- Address @hebasto's suggestions ([1](https://github.com/bitcoin-core/gui/pull/337#discussion_r637455152), [2](https://github.com/bitcoin-core/gui/pull/337#discussion_r637455353))

src/qt/test/apptests.cpp
```diff
@@ -21,15 +20,24 @@
...
 #include <QTextEdit>
 #include <QtGlobal>
 #include <QtTest/QtTestWidgets>
 #include <QtTest/QtTestGui>
 
 namespace {
+//! Regex find a string group inside of the console output
+QString FindInConsole(const QString& output, const QString& pattern)
+{
+    QRegularExpression re(pattern);
```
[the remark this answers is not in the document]
jarolrod: Addressed in 9bde381

src/qt/test/apptests.cpp
jarolrod: Addressed in 9bde381

jarolrod:
Updated from 9bde381 -> 6969b2b ([pr337.03](https://github.com/jarolrod/gui/tree/pr337.03) -> [pr337.04](https://github.com/jarolrod/gui/tree/pr337.04))

**Changes:**
- address conflict with #123

hebasto (approved):
ACK 6969b2bb98a2f44e1b51c905db92ec2e28345078

ShaMan239 (commented):
ACK 6969b2bb98a2f44e1b51c905db92ec2e28345078
Compiled the PR properly on Ubuntu 20.04

The PR introduces regex search functionality to apptest.cpp file by replacing univalue based search, which requires the qmessagewidget to be output in plaintext to work properly, with a more general Qregularexpression based search which doesn’t explicitly require a plaincheck to function.

Tested if the new apptest is running properly by running `./src/qt/test/test_bitcoin-qt` and from there obtained: `PASS   : AppTests::appTests()`
For the sake of completion, the final result obtained from running the `test_bitcoin-qt` file was:
`Totals: 3 passed, 0 failed, 0 skipped, 0 blacklisted, 96ms`

This PR is an improvement over the master, as now the search string need not have to be a plaintext, which increases the number of use cases of the search action. An example of an additional use case is if we want to output the qmessagewidget in HTML to examine the values of the current HTML elements.

PR #338, opened 2021-05-17
qt: peers-tab resizeColumnsToContents

RandyMcMillan:


RandyMcMillan:
Related to #256 

This PR handles resizing column widths programmatically.

RandyMcMillan:
Fields that commonly need to be resized to view are resized based on content length.

Screen Shot 2021-05-17 at 2 35 09 AM

src/qt/bantablemodel.cpp
```diff
@@ -120,6 +122,8 @@ QVariant BanTableModel::data(const QModelIndex &index, int role) const
     const auto column = static_cast<ColumnIndex>(index.column());
     if (role == Qt::DisplayRole) {
         switch (column) {
+        case Bumper:
+            return QString::fromStdString("");
```
jarolrod, line 126: why does this need to be `QString::fromStdString("")`

src/qt/peertablemodel.cpp
```diff
@@ -140,10 +142,12 @@ QVariant PeerTableModel::data(const QModelIndex &index, int role) const
         case Ping:
         case Sent:
         case Received:
-            return QVariant(Qt::AlignRight | Qt::AlignVCenter);
-        case Subversion:
             return {};
-        } // no default case, so the compiler can warn about missing cases
+        case Subversion:
+            return QVariant(Qt::AlignLeft | Qt::AlignVCenter);
```
jarolrod, line 147: isn't this unrelated to what you want to do?

src/qt/peertablesortproxy.cpp
```diff
@@ -38,6 +38,8 @@ bool PeerTableSortProxy::lessThan(const QModelIndex& left_index, const QModelInd
         return left_stats.nRecvBytes < right_stats.nRecvBytes;
     case PeerTableModel::Subversion:
         return left_stats.cleanSubVer.compare(right_stats.cleanSubVer) < 0;
+    case PeerTableModel::Bumper:
+        return left_stats.nodeid < right_stats.nodeid;
```
jarolrod, line 42: Why this sorting scheme?

jarolrod (commented):
I'm failing to understand what this is trying to address, can you clarify a bit? 🤔

promag:
Please describe the goal, what is trying to improve and how it relates to #256. Looks like #256 conflicts with `QHeaderView::ResizeToContents`?

src/qt/peertablesortproxy.cpp
RandyMcMillan: Sorting by ```nodeid``` was simply to have sorting functionality for that column as opposed to no sorting functionality.

RandyMcMillan:
##### Goal of this PR/Discussion
- Ideally formatting the initial column widths can reduce the number of click events the user needs to perform to be able to view the data presented.
- Reducing click events where ever possible is ideal for normal users & tremendously valuable for scenarios where the user doesn't interact with this software primarily with mouse/touch methods.

##### Motivations for \/gui fork as far as I can tell...
- It is rooted in improving UI/UX ([User Experience](https://en.wikipedia.org/wiki/User_experience))
- REVIEW [ISO 9241-210:2019(en)](https://www.iso.org/obp/ui/#iso:std:iso:9241:-210:ed-2:v1:en)
- It is really impractical to require /gui PRs have to restate basic GUI/UI/UX motivations every time.

hebasto:
Concept NACK.

There are some concerns about `QHeaderView::ResizeToContents`.

1) It increases complexity in such a way that could lead, at some scale, to deteriorating the GUI responsiveness
2) From Qt [docs](https://doc.qt.io/qt-5/qheaderview.html#ResizeMode-enum):
> The size cannot be changed by the user...

Usually, in this project, we strive do not make decisions instead of users in such questions :)

3) Some users could feel uncomfortable with possible continuous reflowing of the peers table. I'm among of them.

jarolrod:
Concept NACK

This unnecessarily restricts the user-configurability of the GUI. Testing this PR, I am no longer able to resize the tabs on my own to my liking. Per Qt docs, this is because of the use of [QHeaderView::ResizeToContents](https://doc.qt.io/qt-5/qheaderview.html#ResizeMode-enum). 

In general, we should *ONLY* restrict user freedom when there is good reason and motivation to do so. Tables are meant to be resized to one's liking; not everyone wants to see the `User Agent` string, and some may not want to see the full `Address` string (this is one reason why we have #276). 

Users of the GUI *do want* to customize the sizing of various aspects of the interface and have that saved and restored. This is a precedent that has been set by the direction GUI development has taken, See #165, #194, #205, and #256.

rebroad:
Can we make it do this only until the user tries to manually resize a column?

PR #343, opened 2021-05-22
Improve the GUI responsiveness when progress dialogs are used

hebasto:
[`QProgressDialog`](https://doc.qt.io/qt-5/qprogressdialog.html) estimates the time the operation will take (based on time for steps), and only shows itself if that estimate is beyond [`minimumDuration`](https://doc.qt.io/qt-5/qprogressdialog.html#minimumDuration-prop).

The default `minimumDuration` value is [4 seconds](https://doc.qt.io/qt-5/qprogressdialog.html#details), and it could make users think that the GUI is frozen.

This PR sets `minimumDuration` to zero for all progress dialogs, that affects ones in the `WalletControllerActivity` class.

### DIFF
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -818,9 +818,12 @@ void PolishProgressDialog(QProgressDialog* dialog)
     // Workaround for macOS-only Qt bug; see: QTBUG-65750, QTBUG-70357.
     const int margin = TextWidth(dialog->fontMetrics(), ("X"));
     dialog->resize(dialog->width() + 2 * margin, dialog->height());
-#else
-    Q_UNUSED(dialog);
 #endif
+    // QProgressDialog estimates the time the operation will take (based on time
+    // for steps), and only shows itself if that estimate is beyond minimumDuration.
+    // The default minimumDuration value is 4 seconds, and it could make users
+    // think that the GUI is frozen.
+    dialog->setMinimumDuration(0);
 }
 
 int TextWidth(const QFontMetrics& fm, const QString& text)
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -818,7 +818,6 @@ void PolishProgressDialog(QProgressDialog* dialog)
     // Workaround for macOS-only Qt bug; see: QTBUG-65750, QTBUG-70357.
     const int margin = TextWidth(dialog->fontMetrics(), ("X"));
     dialog->resize(dialog->width() + 2 * margin, dialog->height());
-    dialog->show();
 #else
     Q_UNUSED(dialog);
 #endif
--- a/src/qt/walletcontroller.cpp
+++ b/src/qt/walletcontroller.cpp
@@ -207,6 +207,9 @@ void WalletControllerActivity::showProgressDialog(const QString& label_text)
     m_progress_dialog->setCancelButton(nullptr);
     m_progress_dialog->setWindowModality(Qt::ApplicationModal);
     GUIUtil::PolishProgressDialog(m_progress_dialog);
+    // The setValue call forces QProgressDialog to start the internal duration estimation.
+    // See details in https://bugreports.qt.io/browse/QTBUG-47042.
+    m_progress_dialog->setValue(0);
 }
 
 void WalletControllerActivity::destroyProgressDialog()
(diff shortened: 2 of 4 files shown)


src/qt/walletcontroller.cpp
```diff
@@ -207,6 +207,9 @@ void WalletControllerActivity::showProgressDialog(const QString& label_text)
     m_progress_dialog->setCancelButton(nullptr);
     m_progress_dialog->setWindowModality(Qt::ApplicationModal);
     GUIUtil::PolishProgressDialog(m_progress_dialog);
+    // The setValue call forces QProgressDialog to start the internal duration estimation.
+    // See details in https://bugreports.qt.io/browse/QTBUG-47042.
```
ryanofsky, line 211: In commit "qt: Improve GUI responsiveness" (4935ac583bbdc289dd31a1caae3d711edef742b6)

It could be useful to also add this comment other places we're relying on this behavior (or maybe dedup the code more generally):

https://github.com/bitcoin/bitcoin/blob/860093401840d7aad7b439aeba0d1598933bc9c6/src/qt/bitcoingui.cpp#L1380
https://github.com/bitcoin/bitcoin/blob/860093401840d7aad7b439aeba0d1598933bc9c6/src/qt/walletview.cpp#L335

ryanofsky (approved):
Code review ACK 4935ac583bbdc289dd31a1caae3d711edef742b6. I'm not very familiar with this API but all the changes and explanations make sense and are very clear, and this seems like it should be an improvement.

src/qt/guiutil.cpp
```diff
@@ -818,10 +818,12 @@ void PolishProgressDialog(QProgressDialog* dialog)
...
     dialog->resize(dialog->width() + 2 * margin, dialog->height());
-    dialog->show();
-#else
-    Q_UNUSED(dialog);
 #endif
+    // QProgressDialog estimates the time the operation will take (based on time
+    // for steps), and only shows itself if that estimate is beyond minimumDuration.
+    // The default minimumDuration value is 4 seconds, and it could make users
+    // think that the GUI is frozen.
+    dialog->setMinimumDuration(0);
```
promag, line 826: Why not just `show()`?

promag (commented):
Concept ACK.

src/qt/guiutil.cpp
hebasto: From [QTBUG-47042](https://bugreports.qt.io/browse/QTBUG-47042?focusedCommentId=301657&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-301657):
> `QProgressDialog` is _designed_ to show itself automatically, based on an internal estimate for the duration of the operation and the `minimumDuration` property. You _never_ call `show()` or `exec()` on it manually. You're also not supposed to keep it around when it's not used. In 5.4, the only way to start the internal duration estimation was to call `setValue(0)`. But we noticed that many people didn't call `setValue(0)`, but just handed external progress information on to the progress dialog, and whether or not that progress information started at 0 or, say, 1, would determine whether the dialog would show itself at all, or not. This was thought to be a bit fragile, so we changed it so that the start time of the duration estimate was the dialog construction time unless `setValue(0)` (or, more correctly, `setValue(minimum())` was called, leading to the dialog showing itself `minimumDuration` after its construction, even if it was constructed not for immediate use (which the dialog cannot distinguish from a really long-running operation).

src/qt/walletcontroller.cpp
hebasto: @ryanofsky

Thank you for your review.

I agree with you that there is a possibility to dedup the code more generally. As such a change must be refactoring, I intentionally leaved it for a follow up, if there are no objections from you.

A comment added here because this dialog is not actually a "progress" one, it is a busy indicator without meaningful values. But it requires to call `setValue(0)` to start [internal timer](https://github.com/bitcoin-core/gui/pull/343#discussion_r640163408).

In other cases calling `setValue()` is natural, as dialogs are real "progress" ones, and the value shows the progress.

promag: Right, the progress here is [0,0].

promag (commented):
Code review ACK 4935ac583bbdc289dd31a1caae3d711edef742b6.

jarolrod (commented):
ACK 4935ac583bbdc289dd31a1caae3d711edef742b6

Code review ACK and light tested ACK of progress bars. Patch looks good.

PR #332, opened 2021-05-15
Replace disambiguation strings with translator comments

hebasto:
Since https://github.com/bitcoin/bitcoin/pull/21694 is merged, translator comments is the right way to pass context to translators.

This PR fixes changes were made:
- in #220 before https://github.com/bitcoin/bitcoin/pull/21694
- in https://github.com/bitcoin/bitcoin/pull/21694 on testing purpose
- in #125

Closes #288.

PR #346, opened 2021-05-27
English translations update

hebasto:
Update for Transifex.

After changing translator comments in #332 this update will show if Transifex triggers strings to be re-translated.

### DIFF
--- a/src/qt/bitcoinstrings.cpp
+++ b/src/qt/bitcoinstrings.cpp
@@ -75,12 +75,6 @@ QT_TRANSLATE_NOOP("bitcoin-core", ""
 "Prune: last wallet synchronisation goes beyond pruned data. You need to -"
 "reindex (download the whole blockchain again in case of pruned node)"),
 QT_TRANSLATE_NOOP("bitcoin-core", ""
-"SQLiteDatabase: Failed to prepare the statement to fetch sqlite wallet "
-"schema version: %s"),
-QT_TRANSLATE_NOOP("bitcoin-core", ""
-"SQLiteDatabase: Failed to prepare the statement to fetch the application id: "
-"%s"),
-QT_TRANSLATE_NOOP("bitcoin-core", ""
 "SQLiteDatabase: Unknown sqlite wallet schema version %d. Only version %d is "
 "supported"),
 QT_TRANSLATE_NOOP("bitcoin-core", ""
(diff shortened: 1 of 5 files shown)


src/qt/bitcoinstrings.cpp
```diff
@@ -189,7 +183,7 @@ QT_TRANSLATE_NOOP("bitcoin-core", "Invalid netmask specified in -whitelist: '%s'
 QT_TRANSLATE_NOOP("bitcoin-core", "Loading P2P addresses…"),
 QT_TRANSLATE_NOOP("bitcoin-core", "Loading banlist…"),
 QT_TRANSLATE_NOOP("bitcoin-core", "Loading block index…"),
-QT_TRANSLATE_NOOP("bitcoin-core", "Loading wallet…"),
+QT_TRANSLATE_NOOP("bitcoin-core", "Loading wallet..."),
```
laanwj, line 186: Strange, why is this changing back

laanwj: Looks like bitcoin/bitcoin#20773 accidentally reverted this

src/qt/bitcoinstrings.cpp
```diff
@@ -75,12 +75,6 @@ QT_TRANSLATE_NOOP("bitcoin-core", ""
 "Prune: last wallet synchronisation goes beyond pruned data. You need to -"
 "reindex (download the whole blockchain again in case of pruned node)"),
 QT_TRANSLATE_NOOP("bitcoin-core", ""
-"SQLiteDatabase: Failed to prepare the statement to fetch sqlite wallet "
```
laanwj, line 78: Nice, this definitely doesn't need to be translated.

src/qt/bitcoinstrings.cpp
hebasto: yeah...

don't throw a stone at me, but a linter could help

laanwj: Please no…

hebasto: Fixed in the recent push.

laanwj:
ACK df4c81fda490e556775f6006f3d4f7d54f61073b
`make translate` gives me an empty diff

jarolrod (commented):
ACK df4c81fda490e556775f6006f3d4f7d54f61073b

Also get an empty diff after running `make -C src translate`.

hebasto:
#### The results of this testing PR.

In case of changing of translator comments, Transifex:
- does _not_ change the "Translated" status of the translated string
- drops the "Reviewed" status of the translated string

PR #354, opened 2021-06-03
Refactor open date range to use std::optional

promag:
Use `std::nullopt` for open date range instead of `TransactionFilterProxy::MIN_DATE` and `TransactionFilterProxy::MAX_DATE`.

### DIFF
--- a/src/qt/transactionfilterproxy.cpp
+++ b/src/qt/transactionfilterproxy.cpp
@@ -9,15 +9,8 @@
 
 #include <cstdlib>
 
-// Earliest date that can be represented (far in the past)
-const QDateTime TransactionFilterProxy::MIN_DATE = QDateTime::fromTime_t(0);
-// Last date that can be represented (far in the future)
-const QDateTime TransactionFilterProxy::MAX_DATE = QDateTime::fromTime_t(0xFFFFFFFF);
-
 TransactionFilterProxy::TransactionFilterProxy(QObject *parent) :
     QSortFilterProxyModel(parent),
-    dateFrom(MIN_DATE),
-    dateTo(MAX_DATE),
     m_search_string(),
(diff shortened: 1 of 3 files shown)


monstro805 (approved):

monstro805 (approved):

hebasto:
Concept ACK

hebasto (approved):
ACK a7837ba08bc10c9aaca6f28d9c2be9b8430323f5, I have reviewed the code and it looks OK, I agree it can be merged.

hebasto:
nit: include `<optional>` to the `src/qt/transactionview.cpp` as `std::nullopt` are use there.

hebasto (approved):
re-ACK 4830f4912a538600e9e9133442e9f8d929006630, only missed header included since my [previous](https://github.com/bitcoin-core/gui/pull/354#pullrequestreview-682108182) review.

src/qt/transactionfilterproxy.cpp
```diff
@@ -46,8 +39,8 @@ bool TransactionFilterProxy::filterAcceptsRow(int sourceRow, const QModelIndex &
         return false;
 
     QDateTime datetime = index.data(TransactionTableModel::DateRole).toDateTime();
-    if (datetime < dateFrom || datetime > dateTo)
-        return false;
+    if (dateFrom && datetime < *dateFrom) return false;
```
Talkless, line 42: `std::optional` does support comparison without dereferencing (see (26) in https://en.cppreference.com/w/cpp/utility/optional/operator_cmp) but your variant is better as `operator<` would repeat `if (opt)` check you already (rightly) did.

Talkless (approved):
tACK 4830f4912a538600e9e9133442e9f8d929006630, tested on Debian Sid, filtering seems to work as expected.

PR #355, opened 2021-06-03
qt: Avoid ambiguous interpreting of placeholders

hebasto:
Transifex.com has special meaning for `%1%`, `%2%` etc substrings, so it could confuse translators.

For example, it renders the source "Unknown. Syncing Headers (%1, %2%)…" as the following:

Screenshot from 2021-06-03 23-44-49

This PR prevents such unintentional behavior.


luke-jr:
Is that some kind of error indication?

What if you make it `%2%%`?

hebasto:
@luke-jr 

> Is that some kind of error indication?

It looks like Transifex considers `%...%` as a kind of a placeholder, but our [custom placeholder](https://docs.transifex.com/translation-checks/setting-translation-checks#checking-for-custom-variables) list is empty.

> What if you make it `%2%%`?

I don't know for sure, but it leaves the `%2%` substring untouched.

luke-jr:
Populating custom placeholders seems to have fixed Transifex's behaviour in this regard.

hebasto:
> Populating custom placeholders seems to have fixed Transifex's behaviour in this regard.

Right. Closing.

PR #330, opened 2021-05-15
Allow prompt icon to be colorized

jarolrod:
Opening the console on macOS, while in dark mode, the console prompt icon will not be colorized white like other icons. This applies the `platformStyle` to the icon so that It can be colorized white.

While here, refactor the `promptIcon` widget from a `QPushButton` to `QLabel`; which is more appropriate, per [Qt Docs](https://doc.qt.io/qt-5/qlabel.html#details):
> QLabel is used for displaying text or an image. No user interaction functionality is provided. 

| Master        | PR               |
| ----------- | ----------- |
| Screen Shot 2021-05-14 at 11 46 33 PM | Screen Shot 2021-05-14 at 11 45 41 PM |

PR #358, opened 2021-06-07
make GUIUtil::ThemedLabel UI form friendly

jarolrod:
This allows the `GUIUtil::ThemedLabel` class able to be used in UI form files. This is accomplished by introducing a form file friendly constructor which only takes in a `QWidget* parent`.

You cannot pass in the `platform_style` when initializing a `GUIUtil::ThemedLabel` in a UI form file. This introduces a `setPlatformStyle` function to be able to pass in the `platform_style` at a later point.

Additionally; because we are not initializing the class with a `platform_style`, we need to implement a check in `updateThemedPixmap` so that it only executes when the `m_platform_style` has been set. A segfault would occur without this conditional as a `ChangeEvent` occurs on application start

This allows this class to be used in the form file for #330 for an elegant solution.

### DIFF
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -793,6 +793,11 @@ qreal calculateIdealFontSize(int width, const QString& text, QFont font, qreal m
     return font_size;
 }
 
+ThemedLabel::ThemedLabel(QWidget* parent)
+{
+    QLabel{parent};
+}
+
 ThemedLabel::ThemedLabel(const PlatformStyle* platform_style, QWidget* parent)
     : QLabel{parent}, m_platform_style{platform_style}
 {
@@ -807,6 +812,11 @@ void ThemedLabel::setThemedPixmap(const QString& image_filename, int width, int 
     updateThemedPixmap();
 }
 
+void ThemedLabel::setPlatformStyle(const PlatformStyle* platform_style)
+{
+    m_platform_style = platform_style;
+    assert(m_platform_style);
+}
 void ThemedLabel::changeEvent(QEvent* e)
 {
 #ifdef Q_OS_MACOS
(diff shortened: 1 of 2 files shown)


src/qt/guiutil.h
```diff
@@ -237,14 +237,16 @@ namespace GUIUtil
...
         explicit ThemedLabel(const PlatformStyle* platform_style, QWidget* parent = nullptr);
         void setThemedPixmap(const QString& image_filename, int width, int height);
+        void setPlatformStyle(const PlatformStyle* platform_style);
 
     protected:
         void changeEvent(QEvent* e) override;
 
     private:
-        const PlatformStyle* m_platform_style;
+        const PlatformStyle* m_platform_style = nullptr;
```
luke-jr, line 249: `{nullptr}`

luke-jr (changes requested):

src/qt/guiutil.cpp
```diff
@@ -793,6 +793,11 @@ qreal calculateIdealFontSize(int width, const QString& text, QFont font, qreal m
     return font_size;
 }
 
+ThemedLabel::ThemedLabel(QWidget* parent)
+{
+    QLabel{parent};
+}
+
```
hebasto, line 800:
```suggestion
    : QLabel{parent} {}
```

src/qt/guiutil.cpp
```diff
@@ -807,6 +812,11 @@ void ThemedLabel::setThemedPixmap(const QString& image_filename, int width, int
     updateThemedPixmap();
 }
 
+void ThemedLabel::setPlatformStyle(const PlatformStyle* platform_style)
+{
+    m_platform_style = platform_style;
+    assert(m_platform_style);
+}
 void ThemedLabel::changeEvent(QEvent* e)
```
hebasto, line 820: style nit: Add an empty line as a separator?

hebasto (commented):
@jarolrod 

I think we should adopt your suggestion on IRC to move the `ThemedLabel` into its own module for the following reasons:
- all of so called ["promoted widgets"](https://doc.qt.io/qt-5/designer-using-custom-widgets.html#promoting-widgets) in our repo have their own modules
- see https://github.com/bitcoin-core/gui/pull/330/files#r650460316

Also, it seams reasonable to me to drop this pr, and review both commits in #330, because without actual usage of a promoted widget in a `*.ui` file it is not easy to test it.

jarolrod:
This was meant to be a part of https://github.com/bitcoin-core/gui/pull/330. A simpler approach was chosen for that PR and this is no longer worth the effort as the issue has been fixed. 

Note: If there is ever a reason to make the `ThemedLabel` class UI friendly, the approach should be to do so in its own class file; moving it away from `GUIUtil`

PR #361, opened 2021-06-11
Fix gui segfault caused by bitcoin/bitcoin#22216

ryanofsky:
Reported by Hennadii Stepanov https://github.com/bitcoin/bitcoin/pull/22216#issuecomment-859790682

Fixes bitcoin/bitcoin#22227

### DIFF
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -490,6 +490,7 @@ int GuiMain(int argc, char* argv[])
 
     /// 2. Parse command-line options. We do this after qt in order to show an error if there are problems parsing these
     // Command-line options take precedence:
+    node_context.args = &gArgs;
     SetupServerArgs(gArgs);
     SetupUIArgs(gArgs);
     std::string error;


hebasto:
Why tests and CI did not catch it?

ryanofsky:
Tests would not catch it because tests initialize context differently than the gui app.

I didn't catch it because I only tested the gui with #22216 and #22219 together. They were initially part of the same PR and #22219 also fixes this bug.

hebasto (approved):
ACK d7f3b1af21daa93165b3726fee059a55b7727280, tested on Linux Mint 20.1 (Qt 5.12.8).

UPDATE: tested on Windows 10 Pro (20H2, build 19042.1052).

jarolrod (commented):
ACK d7f3b1af21daa93165b3726fee059a55b7727280

Tested on macOS 11.3, Qt 5.15.2

**master:**
```
xyz@xyzs-MBP bitcoin % ./src/qt/bitcoin-qt -testnet
dbus[52399]: Dynamic session lookup supported but failed: launchd did not provide a socket path, verify that org.freedesktop.dbus-session.plist is loaded!
Assertion failed: ("node.args" && check), function operator(), file wallet/init.cpp, line 127.
zsh: abort      ./src/qt/bitcoin-qt -testnet
```

**pr:**
Runs and works fine upon manual testing

Rspigler:
> Tests would not catch it because tests initialize context differently than the gui app.

Is there a way to improve tests so that they initialize in the same way?

ryanofsky:
> Is there a way to improve tests so that they initialize in the same way?

One way would be to make the python tests invoke bitcoin-qt.

PR #366, opened 2021-06-22
Dark Mode fixes/portability

luke-jr:
The changes to support macOS "Dark Mode" are valid for any platform, and should work so long as Qt implements the PaletteChange event. (Worst case, we're no worse off with trying.)

Additionally, we shouldn't block the parent classes from implementing event handlers. Who knows what side effects that could have.

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -1173,14 +1173,12 @@ void BitcoinGUI::message(const QString& title, QString message, unsigned int sty
 
 void BitcoinGUI::changeEvent(QEvent *e)
 {
-#ifdef Q_OS_MACOS
     if (e->type() == QEvent::PaletteChange) {
         overviewAction->setIcon(platformStyle->SingleColorIcon(QStringLiteral(":/icons/overview")));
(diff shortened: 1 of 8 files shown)


hebasto:
> ... we shouldn't block the parent classes from implementing event handlers. Who knows what side effects that could have.

Looks reasonable.

saadbitcoin:
OKY

jonatack:
Concept ACK. Lightly tested, debug build is clean.  I have my system set up (on Debian testing) so the GUI follows my system settings, and this change doesn't affect it; everything looks the same for me.

saadbitcoin:
GOOD

hebasto (approved):
ACK 9d5bf6bf01af40a9684f1b1f06a8df4aaf36b8f3, tested on Linux Mint 20.1 (Qt 5.12.8) with the [`qt5ct`](https://packages.ubuntu.com/focal/qt5ct) package installed.

kristapsk (approved):
ACK 9d5bf6bf01af40a9684f1b1f06a8df4aaf36b8f3. Tested on Gentoo Linux with Xfce4 and Qt 5.15.2, does not break anything on my computer.

PR #368, opened 2021-06-24
Bugfix: When restoring table columns, still set their minimum column width and stretch on last section

luke-jr:
~~Regression introduced in #205~~

Correction: Undefined behaviour currently makes this a non-issue, but probably better to be safe and do it correctly.

### DIFF
--- a/src/qt/receivecoinsdialog.cpp
+++ b/src/qt/receivecoinsdialog.cpp
@@ -69,14 +69,14 @@ ReceiveCoinsDialog::ReceiveCoinsDialog(const PlatformStyle *_platformStyle, QWid
     tableView->setAlternatingRowColors(true);
     tableView->setSelectionBehavior(QAbstractItemView::SelectRows);
     tableView->setSelectionMode(QAbstractItemView::ContiguousSelection);
+    tableView->horizontalHeader()->setMinimumSectionSize(MINIMUM_COLUMN_WIDTH);
+    tableView->horizontalHeader()->setStretchLastSection(true);
 
     QSettings settings;
     if (!tableView->horizontalHeader()->restoreState(settings.value("RecentRequestsViewHeaderState").toByteArray())) {
         tableView->setColumnWidth(RecentRequestsTableModel::Date, DATE_COLUMN_WIDTH);
         tableView->setColumnWidth(RecentRequestsTableModel::Label, LABEL_COLUMN_WIDTH);
         tableView->setColumnWidth(RecentRequestsTableModel::Amount, AMOUNT_MINIMUM_COLUMN_WIDTH);
-        tableView->horizontalHeader()->setMinimumSectionSize(MINIMUM_COLUMN_WIDTH);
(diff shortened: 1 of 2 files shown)


saadbitcoin:
GOOD

hebasto:
> When restoring table columns, still set their minimum column width and stretch on last section

Are `minimumSectionSize` and `stretchLastSection` parts of the saved state?

> Regression introduced in #205

How is it observable by a user?

jarolrod (commented):
> Are minimumSectionSize and stretchLastSection parts of the saved state?

yes, they both are. See: https://code.qt.io/cgit/qt/qt.git/tree/src/gui/itemviews/qheaderview.cpp#n3556
- [minimumSectionSize](https://code.qt.io/cgit/qt/qt.git/tree/src/gui/itemviews/qheaderview.cpp#n3579)
- [stretchLastSection](https://code.qt.io/cgit/qt/qt.git/tree/src/gui/itemviews/qheaderview.cpp#n3574)

> Regression introduced in #205

While I think this change is ok, is it possible for a user to actually manipulate the header view in a way where they can run into an issue here?

hebasto:
@luke-jr 

The [`minimumSectionSize`](https://doc.qt.io/qt-5/qheaderview.html#minimumSectionSize-prop) and [`stretchLastSection`](https://doc.qt.io/qt-5/qheaderview.html#stretchLastSection-prop) properties of the [`QHeaderView`](https://doc.qt.io/qt-5/qheaderview.html) class are saved with [`QHeaderView::saveState`](https://doc.qt.io/qt-5/qheaderview.html#saveState) and restored with [`QHeaderView::restoreState`](https://doc.qt.io/qt-5/qheaderview.html#restoreState).

> Regression introduced in #205

1. From the PR description it is not clear what is actually a bug in the current master branch. Could you elaborate please?

2. What is the reason to set `minimumSectionSize` and `stretchLastSection` _before_ restoring the header state, when during this restoring (if it succeeds, of course) the values of these properties will be overwritten?

hebasto:
> I guess this is more about not relying on undefined behaviour rather than a bugfix.
> 
> Leaving the assignment prior to restoring sizes, so that users who actually go to the trouble to hack their setting are respected.

Why only are `setMinimumSectionSize` and `setStretchLastSection` considered? And not `setColumnWidth` methods?

hebasto:
This discussion has been inactive for a long time.

Should it be closed?

RandyMcMillan:
Let's leave this open - In previous work I added a "bumper" column to the tableview - this helped address some of the issues. The bumper column was an addon to other PRs but not a stand alone PR - I will put together a PR that focuses on why this is useful.

hebasto:
@luke-jr 

> This discussion has been inactive for a long time.
> 
> Should it be closed?

Instead of reacting with the "thumbs down" emoji, maybe just answer questions from the [comments](https://github.com/bitcoin-core/gui/pull/368#issuecomment-916192090)?

PR #372, opened 2021-06-26
Remove "Type" column from peers window

jonatack:
This PR removes the "Type" column from the peers main window that was added in #179, as it is confusing to users and does not make sense to have a Type column without its corresponding Direction column immediately preceding it to indicate Inbound or Outbound, per our current connection type naming (one of the connection types is "Inbound" and is not ever displayed currently).

Three attempts to add the Direction column in #179 (original version), #289, and #317 have seen opposition from some reviewers. 

Therefore, before the v22.0 release, remove the Type column until such time as there is rough consensus to add the Direction Inbound/Outbound column with it (or better yet, merge #317).

### DIFF
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -669,12 +669,9 @@ QString NetworkToQString(Network net)
     assert(false);
 }
 
-QString ConnectionTypeToQString(ConnectionType conn_type, bool prepend_direction)
+QString ConnectionTypeToQString(ConnectionType conn_type)
 {
-    QString prefix;
-    if (prepend_direction) {
-        prefix = (conn_type == ConnectionType::INBOUND) ? QObject::tr("Inbound") : QObject::tr("Outbound") + " ";
-    }
+    QString prefix{conn_type == ConnectionType::INBOUND ? QObject::tr("Inbound") : QObject::tr("Outbound") + " "};
     switch (conn_type) {
     case ConnectionType::INBOUND: return prefix;
     case ConnectionType::OUTBOUND_FULL_RELAY: return prefix + QObject::tr("Full Relay");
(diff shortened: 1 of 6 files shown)


jonatack:
See #289 and #317 for more context.

jarolrod (commented):
Indeed, it is a sub-optimal situation. My opinion is that there is still value in maintaining the `Type` column. I am mainly concerned about removing this, and then going through opposition to reintroduce it next release cycle.

While we may not have a separate and sortable `Direction` column, we do still have arrows prepended before the Address. This has been the default behavior for three years now (https://github.com/bitcoin-core/gui/commit/4132ad3bfa5b5ead3f4a592d819d0097881b4827) so user's know to look here for the direction of the selected peer. While the situation is not optimal; this means the `Type` information does have orientation context in regards to `Direction`.

jonatack:
It's too confusing for users with the current halfway implementation. One type (Inbound) isn't even displayed, and the empty type looks like a bug.

At least this PR and its predecessors can be pointed to when more user bug reports arrive, if this isn't fixed before the release: "Look, we tried 3 times." So this had to be proposed.

But it would be better to resolve it before the release, one way or the other.

prayank23:
> But it would be better to resolve it before the release, one way or the other

We should add directions column

ShaMan239 (commented):
Tested on Ubuntu 20.04

The PR claims to remove the type column from the PR table and is successful in doing so. Along with the type column, the PR is also removing the direction arrows preceding the node's address, which makes sense to do.

Though I understand the approach of op to solve this problem, I don’t think it is optimal. As @jarolrod already mentioned, removing the type column will inevitably result in a debate over bringing it back. The optimal approach will be merging pr317, or something similar to 317 if there is a debate there, as suggested by op.

hebasto:
Closing, as #317 has been merged.

jonatack:
> Closing, as #317 has been merged.

Thanks!

Issue #160, opened 2020-12-22
Selected peer in the "Peers" tab gets unselected if any previous peer disconnects


hebasto:
On master(209974d379e5c2f32f9f691c992a2fb58be2abf2) the selected peer in the "Peers" tab gets unselected if any previous peer (in the upper rows) disconnects.

**Step 1** -- connected three peers, and "peerC" has been selected:
![Screenshot from 2020-12-22 15-04-19](https://user-images.githubusercontent.com/32963518/102891836-92ca8c00-4467-11eb-957c-bf6f25ed917c.png)

**Step 2** -- new connected peers ("peerD") do not hurt:
![Screenshot from 2020-12-22 15-04-30](https://user-images.githubusercontent.com/32963518/102891955-ca393880-4467-11eb-954a-b5a7d75b5e71.png)

**Step 3** -- but disconnection of any peer from upper rows (e.g., "peerB") causes the loss of peer selection:
![Screenshot from 2020-12-22 15-05-28](https://user-images.githubusercontent.com/32963518/102892156-2439fe00-4468-11eb-913e-100e096cb0d9.png)

thantkyaw1:
Bug

jarolrod:
This bug no longer operates in the way described. I have not investigated fully, but I believe that #18 fixed this bug as it is currently described. There is *still* a bug related to peer selection that exists and it is fixed by #164.

**New Bug:**
on `master`, if you select a peer, and a peer above it disconnects or is disconnected, the selection will **slip** to the next available peer.

| Select Peer  6   | Disconnect Peer 0  |
| --------------- | ------------------- |
| ![Screen Shot 2021-06-06 at 9 17 11 PM](https://user-images.githubusercontent.com/23396902/120947119-be3aed80-c70c-11eb-99e1-ad9a739df5be.png) | ![Screen Shot 2021-06-06 at 9 17 48 PM](https://user-images.githubusercontent.com/23396902/120947129-c430ce80-c70c-11eb-9c5a-8b4f32a07d7f.png) |

As shown in the screenshots above, we start by selecting `Peer 6: 149.248.3.82:38333`. `Peer 0` is then disconnected, our selection then slips to `Peer 7: 88.99.101.67:38333`.

**#164 Fixes New Bug:**
So while the bug has changed, I can confirm that #164 fixes this.


| Select Peer  9   | Disconnect Peer 1  |
| --------------- | ------------------- |
[...]

PR #374, opened 2021-06-28
Fix #367

### LINKED ISSUE #367: Peer table (node window) does not remain sorted on update
image

This issue was introduced somewhere between 672870ab7ba51b1fc40f263a1c71ebcf165b1229 and c8571486364d6e9ca8c86bd1c81e230ca64f8904

rebroad:
Attempting to fix issue #367

### DIFF
--- a/src/qt/peertablemodel.cpp
+++ b/src/qt/peertablemodel.cpp
@@ -154,30 +154,7 @@ void PeerTableModel::refresh()
         new_peers_data.append(stats);
     }
 
-    // Handle peer addition or removal as suggested in Qt Docs. See:
-    // - https://doc.qt.io/qt-5/model-view-programming.html#inserting-and-removing-rows
-    // - https://doc.qt.io/qt-5/model-view-programming.html#resizable-models
-    // We take advantage of the fact that the std::vector returned
-    // by interfaces::Node::getNodesStats is sorted by nodeid.
-    for (int i = 0; i < m_peers_data.size();) {
-        if (i < new_peers_data.size() && m_peers_data.at(i).nodeStats.nodeid == new_peers_data.at(i).nodeStats.nodeid) {
-            ++i;
-            continue;
-        }
-        // A peer has been removed from the table.
-        beginRemoveRows(QModelIndex(), i, i);
-        m_peers_data.erase(m_peers_data.begin() + i);
-        endRemoveRows();
-    }
-
-    if (m_peers_data.size() < new_peers_data.size()) {
-        // Some peers have been added to the end of the table.
-        beginInsertRows(QModelIndex(), m_peers_data.size(), new_peers_data.size() - 1);
-        m_peers_data.swap(new_peers_data);
-        endInsertRows();
-    } else {
-        m_peers_data.swap(new_peers_data);
-    }
-
-    Q_EMIT changed();
+    Q_EMIT layoutAboutToBeChanged();
+    m_peers_data.swap(new_peers_data);
+    Q_EMIT layoutChanged();
 }
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -684,7 +684,7 @@ void RPCConsole::setClientModel(ClientModel *model, int bestblock_height, int64_
 
         // peer table signal handling - update peer details when selecting new node
         connect(ui->peerWidget->selectionModel(), &QItemSelectionModel::selectionChanged, this, &RPCConsole::updateDetailWidget);
-        connect(model->getPeerTableModel(), &PeerTableModel::changed, this, &RPCConsole::updateDetailWidget);
+        connect(model->getPeerTableModel(), &PeerTableModel::layoutChanged, this, &RPCConsole::updateDetailWidget);
 
         // set up ban table
         ui->banlistWidget->setModel(model->getBanTableModel());


MarcoFalke:
Please explain the change and remove the unscoped GitHub attribute. The unscoped link will mess with all monotree repos, not just this one.

rebroad:
@MarcoFalke what is an unscoped github attribute please?

MarcoFalke:
`#367` is unscoped. At the very least it should say `bitcoin-core/gui#374`.

rebroad:
> `#367` is unscoped. At the very least it should say `bitcoin-core/gui#374`.

A pull/issue always defaults to the same project that it's in - it's only necessary to add the project if it's a different project to that which the pull/issue is in.

rebroad:
How do I change this to a draft pull request? I think there may be a chance of Segmentation fault in this code.

MarcoFalke:
> A pull/issue always defaults to the same project that it's in - it's only necessary to add the project if it's a different project to that which the pull/issue is in.

This is a monotree project. Meaning that different project will have the same tree and share the (merge) commits. So without a scope it won't be trivially possible to recover the metadata.

fanquake:
> A pull/issue always defaults to the same project that it's in - it's only necessary to add the project if it's a different project to that which the pull/issue is in.

Rather than arguing the point, how about just updating the link. As mentioned, having an unscoped link will mess with our merge process / resultant commit messages.

> How do I change this to a draft pull request?

Click the button that says "Convert to draft"

draft

You do realise you could read the GitHub documentation, or use any search engine to also find the answer to that question? Rather than posting another comment here?

I'd suggest you spend some time reading this projects developer docs and contributing guide. Opening a PR with no title, no explanation of the change, a useless commit message, and no commit body is not the standard that is expected here. I'm also almost certain you are aware of that, given you've been around the project for many years now.

hebasto (commented):
Tested 0d41edb8b3ce6923821985a633b49c3be002a05b on Linux Mint 20.1 (Qt 5.12.8) -- works as intended.

The change is correct (UPDATE: see https://github.com/bitcoin-core/gui/pull/374#issuecomment-869662723), and Qt [docs](https://doc.qt.io/qt-5/model-view-programming.html#resizable-models) clearly states:
> For more complex changes to the model's structure, perhaps involving internal reorganization, _sorting of data_ or any other structural change, it is necessary to perform the following sequence:
>
>  - Emit the layoutAboutToBeChanged() signal
>  - Update internal data which represents the structure of the model.
>  - Update persistent indexes using changePersistentIndexList()
>  - Emit the layoutChanged() signal.

UPDATE 2: see https://github.com/bitcoin-core/gui/pull/374#issuecomment-869744938

Also this signal seems redundant now:
```diff
--- a/src/qt/peertablemodel.h
+++ b/src/qt/peertablemodel.h
@@ -73,9 +73,6 @@ public:
 public Q_SLOTS:
     void refresh();
 
-Q_SIGNALS:
-    void changed();
-
 private:
     //! Internal peer data structure.
     QList<CNodeCombinedStats> m_peers_data{};
```

Hope that aforementioned comments will be addressed as well.

hebasto:
OTOH, this change re-introduces #191 bug.

rebroad:
> OTOH, this change re-introduces #191 bug.

Out of the two bugs #191 or #367 - I'd rather have former than the latter.

MarcoFalke:
> I'm trying to understand the issue. I can't fix something when I'm not aware of the problem.

There are two issues with the commit message:

* The hashtag is missing the repo namespace, so when the commit gets pushed to another repo it can't be resolved properly.
* The commit message is lacking any sort of explanation.

You can read our contributing guide to learn on how to create patches:  https://github.com/bitcoin-core/gui/blob/master/CONTRIBUTING.md#committing-patches . It also links to a guide on how to write commit messages.

hebasto:
I still don't understand why [this](https://doc.qt.io/qt-5/qsortfilterproxymodel.html#details) does not work:
> By default, the model dynamically re-sorts and re-filters data whenever the original model changes.

And we have `dynamicSortFilter == true` for sure.

rebroad:
> * The hashtag is missing the repo namespace, so when the commit gets pushed to another repo it can't be resolved properly.

Ah, I understand now - you're implying that it will also get pushed to https://github.com/bitcoin/bitcoin, yes?

> * The commit message is lacking any sort of explanation.

"Fixes #367" is an explanation. so I would disagree with your statement, but if it gets pushed to another repository then I can understand how it could point people to the wrong information. I'm not really understanding why we have two repositories with duplicate code though. Is there an explanation for this somewhere?

> You can read our contributing guide to learn on how to create patches: https://github.com/bitcoin-core/gui/blob/master/CONTRIBUTING.md#committing-patches . It also links to a guide on how to write commit messages.

Thanks. I shall read it now.

hebasto:
My previous [comment](https://github.com/bitcoin-core/gui/pull/374#pullrequestreview-693918856) about using of `layoutAboutToBeChanged` and `layoutChanged` signals is not correct because it should be applied to cases when no proxy models are used.

I believe that correct fix of #367 is suggested in #375.

hebasto:
Does 71ab42d00e12c3bbdf3ca588da6408ec209a10d8 actually revert ecbd91153875c8cdd5b92b840afc116f65e457fb?

src/qt/peertablemodel.cpp
```diff
@@ -154,30 +154,7 @@ void PeerTableModel::refresh()
         new_peers_data.append(stats);
     }
 
-    // Handle peer addition or removal as suggested in Qt Docs. See:
```
jarolrod, line 157: removing this reintroduces #160 as described here: https://github.com/bitcoin-core/gui/issues/160#issuecomment-855515389

jarolrod (commented):
NACK

This is not the way to fix the mentioned issue. This also reintroduces two bugs that we had fixed (#160 and #191)

hebasto:
Closing due to the #375 is merged.

PR #164, opened 2020-12-24
Handle peer addition/removal in a right way

hebasto:
This PR makes `PeerTableModel` handle a peer addition/removal in a right way. See:
- https://doc.qt.io/qt-5/model-view-programming.html#inserting-and-removing-rows
- https://doc.qt.io/qt-5/model-view-programming.html#resizable-models

Fixes #160.

Fixes #191.

PR #375, opened 2021-06-28
Emit dataChanged signal to dynamically re-sort Peers table

### LINKED ISSUE #367: Peer table (node window) does not remain sorted on update

hebasto:
[By default](https://doc.qt.io/qt-5/qsortfilterproxymodel.html#details), the `PeerTableSortProxy`
> dynamically re-sorts ... data whenever the original model changes.

That is not the case on master (8cdf91735f2bdc55577d84a9915f5920ce23b00a) as in ecbd91153875c8cdd5b92b840afc116f65e457fb (#164) no signals are emitted to notify about model changes.

This PR uses a dedicated [`dataChanged`](https://doc.qt.io/qt-5/qabstractitemmodel.html#dataChanged) signal.

Fixes #367.

An alternative to #374.

### DIFF
--- a/src/qt/peertablemodel.cpp
+++ b/src/qt/peertablemodel.cpp
@@ -179,5 +179,7 @@ void PeerTableModel::refresh()
         m_peers_data.swap(new_peers_data);
     }
 
-    Q_EMIT changed();
+    const auto top_left = index(0, 0);
+    const auto bottom_right = index(rowCount() - 1, columnCount() - 1);
+    Q_EMIT dataChanged(top_left, bottom_right);
 }
--- a/src/qt/peertablemodel.h
+++ b/src/qt/peertablemodel.h
@@ -73,9 +73,6 @@ public:
 public Q_SLOTS:
     void refresh();
 
-Q_SIGNALS:
-    void changed();
-
 private:
     //! Internal peer data structure.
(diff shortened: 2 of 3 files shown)


hebasto:
@rebroad  Do you mind testing this PR?

jarolrod (commented):
ACK 986bf78d7e8fd9b69841ecb0decaff840efe9cff

Tested on `Arch Linux`, `Ubuntu`, `macOS 11.3`, and `Windows 10`. The usage of `dataChanged` here is correct and solves the mentioned issue.

The following screenshots will compare the behavior of `master` and this `PR`. Note that the screenshots will sort by `Received`. 

On the `master` screenshots you will see the ordering get out of whack because it is not dynamically sorting the peers as the data changes.

On the `PR` screenshots, you will see that the ordering is correct because we now allow to dynamically sort the peers as the data changes.

## Ubuntu
| Master | PR |
| ----------- | --- | 
| master | pr |

## macOS
| Master | PR |
| ----------- | --- | 
| master | pr |

## Windows 10
| Master | PR |
| ----------- | --- | 
| master | pr |

PR #381, opened 2021-07-14
refactor: Make BitcoinCore class reusable

hebasto:
This PR makes the `BitcoinCore` class reusable, i.e., it can be used by the widget-based GUI or by the [QML-based](https://github.com/bitcoin-core/gui-qml/tree/main/src/qml) one, and it makes the divergence between these two repos minimal.

The small benefit to the current branch is more structured code.

Actually, this PR is ported from https://github.com/bitcoin-core/gui-qml/pull/10.
The example of the re-using of the `BitcoinCore` class is https://github.com/bitcoin-core/gui-qml/pull/11.

### DIFF
--- a/build_msvc/libbitcoin_qt/libbitcoin_qt.vcxproj
+++ b/build_msvc/libbitcoin_qt/libbitcoin_qt.vcxproj
@@ -24,6 +24,7 @@
     <ClCompile Include="..\..\src\qt\csvmodelwriter.cpp" />
     <ClCompile Include="..\..\src\qt\editaddressdialog.cpp" />
     <ClCompile Include="..\..\src\qt\guiutil.cpp" />
+    <ClCompile Include="..\..\src\qt\initexecutor.cpp" />
     <ClCompile Include="..\..\src\qt\intro.cpp" />
     <ClCompile Include="..\..\src\qt\modaloverlay.cpp" />
     <ClCompile Include="..\..\src\qt\networkstyle.cpp" />
@@ -78,6 +79,7 @@
     <ClCompile Include="$(GeneratedFilesOutDir)\moc\moc_csvmodelwriter.cpp" />
     <ClCompile Include="$(GeneratedFilesOutDir)\moc\moc_editaddressdialog.cpp" />
     <ClCompile Include="$(GeneratedFilesOutDir)\moc\moc_guiutil.cpp" />
+    <ClCompile Include="$(GeneratedFilesOutDir)\moc\moc_initexecutor.cpp" />
     <ClCompile Include="$(GeneratedFilesOutDir)\moc\moc_intro.cpp" />
     <ClCompile Include="$(GeneratedFilesOutDir)\moc\moc_modaloverlay.cpp" />
     <ClCompile Include="$(GeneratedFilesOutDir)\moc\moc_networkstyle.cpp" />
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -7,12 +7,19 @@
 #endif
 
 #include <qt/bitcoin.h>
-#include <qt/bitcoingui.h>
 
 #include <chainparams.h>
+#include <init.h>
+#include <interfaces/handler.h>
+#include <interfaces/node.h>
+#include <node/context.h>
+#include <node/ui_interface.h>
+#include <noui.h>
+#include <qt/bitcoingui.h>
 #include <qt/clientmodel.h>
 #include <qt/guiconstants.h>
 #include <qt/guiutil.h>
+#include <qt/initexecutor.h>
 #include <qt/intro.h>
 #include <qt/networkstyle.h>
 #include <qt/optionsmodel.h>
@@ -20,24 +27,17 @@
 #include <qt/splashscreen.h>
 #include <qt/utilitydialog.h>
 #include <qt/winshutdownmonitor.h>
+#include <uint256.h>
+#include <util/system.h>
+#include <util/threadnames.h>
+#include <util/translation.h>
+#include <validation.h>
 
 #ifdef ENABLE_WALLET
 #include <qt/paymentserver.h>
 #include <qt/walletcontroller.h>
 #include <qt/walletmodel.h>
 #endif // ENABLE_WALLET
-
-#include <init.h>
-#include <interfaces/handler.h>
-#include <interfaces/node.h>
-#include <node/context.h>
-#include <node/ui_interface.h>
-#include <noui.h>
-#include <uint256.h>
-#include <util/system.h>
-#include <util/threadnames.h>
-#include <util/translation.h>
-#include <validation.h>
 
 #include <boost/signals2/connection.hpp>
 #include <memory>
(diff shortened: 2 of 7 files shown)


src/qt/bitcoincore.cpp
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) 2014-2021 The Bitcoin Core developers
```
hebasto, line 1: The initial code was contributed by @laanwj back in [2014](https://github.com/bitcoin/bitcoin/pull/3493).

hebasto:
Begging my fellow colleagues for review :
- @laanwj as an author of the initial code
- @promag as an expert in Qt code
- @ryanofsky as an expert in multi-threading

ryanofsky:
This seems good. Also maybe if `BitcoinCore` class is moving to it's own file, it could be renamed to something that better describes what it does like `InitExecutor` (similar to `RPCExecutor`) or `InitActivity` (like `CreateWalletActivity` and `OpenWalletActivity`) or just `InitThread` `InitTask`.

(Longer term, I think these asynchronous executor / activity classes are mostly unnecessary and could be replaced with an [AsyncUpdate](https://github.com/bitcoin/bitcoin/issues/16874#issuecomment-632424392) or similar utility, but that's beyond the scope of this PR.)

hebasto:
@ryanofsky 
> This seems good. Also maybe if `BitcoinCore` class is moving to it's own file, it could be renamed to something that better describes what it does like `InitExecutor` (similar to `RPCExecutor`) or `InitActivity` (like `CreateWalletActivity` and `OpenWalletActivity`) or just `InitThread` `InitTask`.

Many thanks for feedback. Which name could also describe that this class is also used to provide a shutdown activity?

src/qt/bitcoin.cpp
```diff
@@ -291,22 +301,14 @@ bool BitcoinApplication::baseInitialize()
 
 void BitcoinApplication::startThread()
 {
-    if(coreThread)
-        return;
-    coreThread = new QThread(this);
```
ryanofsky, line 296: In commit "qt: Add BitcoinCore::m_thread member" (6b769d9e9e1c73cef37826e3c842fe9b65737456)

The coreThread member doesn't actually do anything after this change. Would be less confusing to squash the next commit into this one and delete coreThread entirely instead of keeping it half-used.

ryanofsky (approved):
Code review ACK 7ec552676c66488fe00fb503d02ec4a389a715b7

> Many thanks for feedback. Which name could also describe that this class is also used to provide a shutdown activity?

Personally I like `InitExecutor`. Shutdown just seems like part of initialization. Maybe `StartupExecutor`? :shrug: I do think pretty much any name would be more descriptive than a class or file called bitcoin core. But feel free to keep the current name or choose any name that seems good to you.

hebasto:
Updated 6c2969da1dd164c6d7acfeb45b55ed95d530c76e -> 0c7667f522f7da8155b7e959e550868210dbf033 ([pr381.02](https://github.com/hebasto/gui/commits/pr381.02) -> [pr381.03](https://github.com/hebasto/gui/commits/pr381.03), [diff](https://github.com/hebasto/gui/compare/pr381.02..pr381.03)).

@ryanofsky

Thank you for your review! Your comments are addressed.

build_msvc/libbitcoin_qt/libbitcoin_qt.vcxproj
```diff
@@ -14,6 +14,7 @@
     <ClCompile Include="..\..\src\qt\bitcoin.cpp" />
     <ClCompile Include="..\..\src\qt\bitcoinaddressvalidator.cpp" />
     <ClCompile Include="..\..\src\qt\bitcoinamountfield.cpp" />
+    <ClCompile Include="..\..\src\qt\bitcoincore.cpp" />
```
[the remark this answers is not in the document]
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui/pull/381#issuecomment-880033334).

src/qt/bitcoin.cpp
```diff
@@ -291,22 +301,14 @@ bool BitcoinApplication::baseInitialize()
...
-    BitcoinCore *executor = new BitcoinCore(node());
-    executor->moveToThread(coreThread);
+    m_core = std::make_unique<BitcoinCore>(node());
```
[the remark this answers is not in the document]
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui/pull/381#issuecomment-880033334).

src/qt/bitcoin.cpp
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui/pull/381#issuecomment-880033334).

src/qt/bitcoin.cpp
```diff
@@ -158,6 +158,16 @@ void DebugMessageHandler(QtMsgType type, const QMessageLogContext& context, cons
...
     QObject(), m_node(node)
 {
+    this->moveToThread(&m_thread);
+    m_thread.start();
+}
+
+BitcoinCore::~BitcoinCore()
+{
+    qDebug() << __func__ << ": Stopping thread";
+    m_thread.quit();
```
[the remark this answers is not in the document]
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui/pull/381#issuecomment-880033334).

src/qt/bitcoin.h
```diff
@@ -112,7 +116,7 @@ public Q_SLOTS:
     void windowShown(BitcoinGUI* window);
 
 private:
-    QThread *coreThread;
+    std::unique_ptr<BitcoinCore> m_executor;
```
ryanofsky: In commit "qt: Add BitcoinCore::m_thread member" (efacacba98aa6d9d0794f62032eb353ced1e0791)

Not important, but this could use `std::optional` instead of `std::unique_ptr` to avoid a dynamic allocation

ryanofsky (approved):
Code review ACK 0c7667f522f7da8155b7e959e550868210dbf033. Thanks! Just renaming and other suggested changes since last review

hebasto:
Updated 0c7667f522f7da8155b7e959e550868210dbf033 -> 8169fc4e73a87331e02502fc24e293831765c8b1 ([pr381.03](https://github.com/hebasto/gui/commits/pr381.03) -> [pr381.04](https://github.com/hebasto/gui/commits/pr381.04), [diff](https://github.com/hebasto/gui/compare/pr381.03..pr381.04)):

- addressed @ryanofsky's [comment](https://github.com/bitcoin-core/gui/pull/381#discussion_r669835948)

src/qt/bitcoin.h
hebasto: Thank! [Updated](https://github.com/bitcoin-core/gui/pull/381#issuecomment-880133880).

saadbitcoin:
How to mine bitcoin core

src/qt/bitcoin.h
saadbitcoin: very good

saadbitcoin:
very good

saadbitcoin:
i am need to downlod

RandyMcMillan:
MacOS 11.4

Remote Desktop Picture July 16, 2021 at 8 50 16 PM EDT

laanwj:
> I do think pretty much any name would be more descriptive than a class or file called bitcoin core. 

Originally I called the class "BitcoinCore", not because the software is called like that (not sure if it even did at the time?), but because it wraps an instance of the core (=non-gui) functionality, to be executed in its own thread. But like `bitcoind`'s main thread, it does nothing but initialization and shutdown.

(I don't mind it being renamed, but it was not as silly asa "let's call a class the same name as the software" at the time)

ryanofsky:
Yeah I figured "core" was a synonym for "init". Just makes sense to make the old name more descriptive when the class is moved and more widely exposed now.

ryanofsky (approved):
Code review ACK 8169fc4e73a87331e02502fc24e293831765c8b1. Only change is switching from `m_executor` from pointer to optional type (thanks for update!)

laanwj:
> Yeah I figured "core" was a synonym for "init". Just makes sense to make the old name more descriptive when the class is moved and more widely exposed now.

Well not "init", more like "root model object". But it's okay, I'm fine with naming it after what it is now doing.

ACK 8169fc4e73a87331e02502fc24e293831765c8b1

promag (commented):
Code review ACK 8169fc4e73.

Issue #262, opened 2021-03-29
Long(V3) Onion address is not wrapped in peer details


ghost:
In the upper right corner of the screenshot you can see that the long(V3) Onion address is not wrapped, but cut.
Also, the colon before the port is missing, which is confusing.

When you enlarge the detail window, you can see that wrapping automatically works, but only at word-breakpoints, e.g. space or colon.

![kein_doppelpunkt](https://user-images.githubusercontent.com/8447873/112802904-07b6ee80-9073-11eb-979a-8e5ce374e1cb.png)

Copying the v3 Onion address and port with the mouse works fine, though. The complete address and port is copied.

hebasto:
Will a contraction of presented address like `fbuf3g…a27cad.onion:8333` work for you?

wodry:
I would like it better, if the address would be longer than 28 characters (which is half of the onion V3 length 56) it would get a linefeed wrap after the 28th character. I think this would fit well into the window, and there would be no information loss.

But the whole address should be copy-able.

I did not find another way to copy the address, aside of marking it in the details tab with the mouse (this was also the use case made me stumbling upon this issue).

Alternatively, if right-cklicking the address in the main peer window column, there could be a context menu "copy to clipboard". I would like that most.

hebasto:
@wodry 

Do you mind looking into #264?

wodry:
> Do you mind looking into #264?

Have done, thank you very much. If this copy-menu is there, the detail window regarding the node address is not so important anymore, and I could live with a contracted address display like `fbuf3g…a27cad.onion:8333` there.

hebasto:
@wodry
> ... contracted address display like `fbuf3g…a27cad.onion:8333` there.

Done in #276.

jarolrod:
@wodry  can this can be closed since #276 was merged?

> I could live with a contracted address display like fbuf3g…a27cad.onion:8333 there.

wodry:
The use case I stumbled upon this issue was that I wanted to copy a peer address, and it was confusing to try to copy it from the details header where it was not wrapped.

#276 (shorten address display with '...' in the middle) is not regarding this issue.

This issue would be solved with a copy entry in the addresses context menu. That was intended with #264, which was closed and then picked up in #318, which is still open, not merged. The "copy address" address context menu entry is missing in current master:
[...]

PR #384, opened 2021-07-20
add copy subnet action for banned peer

ShaMan239:
This PR adds a Copy Subnet context menu action to the Banned Peers Table.

This feature is quite helpful if a node using GUI might want to alert its peer about a particular malicious user. So it can copy that user’s subnet and broadcast it to its peers so they can ban it instantly using the setban command in the console.

| ----------- | ----------- |
| Screenshot_from_2021-07-21_00-01-331 | Screenshot_from_2021-07-20_23-59-271 |

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -706,6 +706,11 @@ void RPCConsole::setClientModel(ClientModel *model, int bestblock_height, int64_
 
         // create ban table context menu
         banTableContextMenu = new QMenu(this);
+        //: Context menu action to copy the subnet of a banned peer
+        banTableContextMenu->addAction(tr("Copy subnet"), [this] {
+            GUIUtil::copyEntryData(ui->banlistWidget, BanTableModel::Address, Qt::DisplayRole);
+        });
+        banTableContextMenu->addSeparator();
         banTableContextMenu->addAction(tr("&Unban"), this, &RPCConsole::unbanSelectedNode);
         connect(ui->banlistWidget, &QTableView::customContextMenuRequested, this, &RPCConsole::showBanTableContextMenu);
 


hebasto:
Concept ACK.

> This PR adds a Copy Subnet context menu action to the Banned Peers Table.

But on the screenshot a menu item is called "Copy Address". Which variant is correct?

src/qt/rpcconsole.cpp
```diff
@@ -706,6 +706,11 @@ void RPCConsole::setClientModel(ClientModel *model, int bestblock_height, int64_
 
         // create ban table context menu
         banTableContextMenu = new QMenu(this);
+        //: Context menu action to copy the subnet of a banned peer
+        banTableContextMenu->addAction(tr("Copy subnet"), [this] {
```
jarolrod, line 710: Since #362 we are now defaulting to adding mnemonic shortcuts for the context menus. This means that you should add one here

Below is how you would do so here:
```suggestion
        banTableContextMenu->addAction(tr("&Copy subnet"), [this] {
```

jarolrod (commented):
Concept ACK

Looking good, one note: https://github.com/bitcoin-core/gui/pull/384/files#r673396366

hebasto:
@ShaMan239

And warm welcome as a new contributor :fireworks:

ShaMan239:
> @ShaMan239
> 
> And warm welcome as a new contributor fireworks

Thank You very much @hebasto

src/qt/rpcconsole.cpp
ShaMan239: this is addressed in d518d67

kristapsk:
Concept ACK. I think "Copy address" should be added to connected peers list's context menu too, both for consistency and usefulness.

RandyMcMillan:
Concept ACK 

I will be curious to see how "Copy subnet" translates.
"Copy address" or simply "Copy" will be ok if there is a translation issue.

I lean toward simply "Copy".

It would be great to link a key command to the action as well - such as <control> + c maybe?

jarolrod:
@RandyMcMillan 
> I will be curious to see how "Copy subnet" translates.

That's a valid concern. Maybe we can add additional context to the translator comment stating that the subnet is a combination of its `IP address` and it's `Netmask`. Additionally we can add that if there is no clear translation in the target language the translation of `IP address` can be used.

hebasto:
> @RandyMcMillan
> 
> > I will be curious to see how "Copy subnet" translates.
> 
> That's a valid concern. Maybe we can add additional context to the translator comment stating that the subnet is a combination of its `IP address` and it's `Netmask`. Additionally we can add that if there is no clear translation in the target language the translation of `IP address` can be used.

If the translator comment will be touched, please, end it with a full stop.
Rationale: it is presented to translators in Transifex.com web-interface, and they usually do not look into the source code. The absence of a full stop could make an impression that Transifex.com just trimmed out a part of a comment.

ShaMan239:
updated from 7c21de2 -> 32b4a54 ([pr384.02](https://github.com/ShaMan239/gui/tree/pr384.02) -> [pr384.03](https://github.com/ShaMan239/gui/tree/pr384.03))
changes: addressed https://github.com/bitcoin-core/gui/pull/384#issuecomment-884352943 and https://github.com/bitcoin-core/gui/pull/384#issuecomment-884357068

jarolrod (commented):
tACK 32b4a54429278dcfb6be1fbaa10c2b91e73a9a1a

Tested on macOS 11.3. Banned a peer, then was able to successfully copy the Peer's subnet by using the new `Copy subnet` context menu action.

The new translator comment looks good to me. I think the additional context behind it's meaning and a possible alternative will help with any translator confusion.

RandyMcMillan:
tACK 32b4a54

Tested on macOS 10.15.7

psancheti110:
Tested 32b4a54429278dcfb6be1fbaa10c2b91e73a9a1a

I tried to build the PR on `macOS 11.4`. It delivers what is projected in the attached screenshots. The code patch looks neat and consistent. Congratulations on that 🎉 

Although I had two doubts about the functionality and the use-case mentioned for the added feature. 

1) The functionality implemented with the patch does not seem a new feature, instead just another way to copy the address. The Qt widget `banlistWidget` by default supports keyboard shortcuts like `Ctrl+C` on Linux/Windows and `⌘+C` on macOS. You could look at this comment for a better understanding of what I am trying to say: https://github.com/bitcoin-core/gui/issues/262#issuecomment-887350423

2) The use-case mentioned states about a node using GUI who might want to alert its peer about a particular malicious user. Would other nodes really welcome/interest such messages about a fellow peer, where they know nothing else about the user but its IP Address and assume it to be malicious by **Trusting** some other user?

In case I understood the use-case wrongly, would appreciate it if you'd correct me. 

Happy Coding!

ShaMan239:
@psancheti110

Thank you for addressing these doubts. Let me try to answer them one by one:

1. Though we can directly copy addresses using keyboard shortcuts, this PR makes it more accessible for those who need to use the mouse.
2. Here is an example of when this is useful:
    - Sometimes, there is a list of peers known to spy that is passed around. This makes it more convenient to add to the list.
    - You have two GUI nodes and banned a peer on one node. You can copy the subnet from GUI Node A, take it to GUI Node B, and ban the peer there as well.

psancheti110:
Hello @ShaMan239 

Appreciate you clearing my doubts and concerns 🙌🏻

1) Agreed. It can be a good addition for people who want to use a mouse for a change. 

I'd like to suggest something maybe you could add to make the use-case mentioned even better. 

- Currently, in the banned peers' list, it is not possible to select multiple addresses and copy them together. For now one has to copy individual addresses one by one.

- You can also checkout **rpcconsole**. Try running `listbanned` and that shall return a list of all banned peers in JSON format, which can be copied with just one click of the mouse. Attaching a screenshot for your reference. 
   > rpcconsole : `listbanned`

Screenshot 2021-08-02 at 3 50 41 PM (2)

hebasto (commented):
> I think "Copy address" should be added to connected peers list's context menu too, both for consistency and usefulness.

Yes, I guess it is an improvement in terms of accessibility (that is why I concept ACKed this pr).

OTOH, such changes open a door for prs that suggest to add a "Copy some_another_fancy_column_content" into the context menus. It could ends with ugly bloated menus.

---

I don't think the use cases discussed above are a good practice for the following reasons:
- for an attacker it is easy and cheap to switch to another IP address in another subnet, and it seems unfeasible to track attacker manually
- aggressive banning (I believe it is the case when we use `setban` RPC call) increases the risk of splitting the network

jarolrod:
> Should "XXX" in "Copy XXX" menu action match a header title of the column?

I think that in the scope of this PR it should. While the combination of IP and Netmask is called subnet in other parts of the code, the column header here is 'IP/Netmask` and not `Subnet`. @hebasto is completely correct here. Having different names for the same data here can be confusing for both users and translators. 

For now lets keep the same name as in the header. If there is valid motivation to rename this header to `subnet`, then a follow up PR can do so and update the name of this action from `Copy IP/Netmask` to `Copy subnet`.

For now lets rename this action to `Copy IP/Netmask`

ShaMan239:
updated from  32b4a54 ->  2da5d46 ([pr384.03](https://github.com/ShaMan239/gui/tree/pr384.03) -> [pr384.04](https://github.com/ShaMan239/gui/tree/pr384.04))
changes: addressed the comments [1](https://github.com/bitcoin-core/gui/pull/384#discussion_r682461249) and [2](https://github.com/bitcoin-core/gui/pull/384#issuecomment-902406496) by @hebasto and @jarolrod 
- Updated Name of Action from `copy subnet` to `copy IP/Netmask`
- Updated Translator's Comments accordingly

src/qt/rpcconsole.cpp
```diff
@@ -706,6 +706,14 @@ void RPCConsole::setClientModel(ClientModel *model, int bestblock_height, int64_
 
         // create ban table context menu
         banTableContextMenu = new QMenu(this);
+        /*: Context menu action to copy the IP/Netmask of a banned peer.
+            IP/Netmask is the combination of a peer's IP address and it's Netmask.
+            For IP Address see: https://en.wikipedia.org/wiki/Subnetwork
```
jarolrod, line 711: This is similar to what was done in #332, see: https://github.com/bitcoin-core/gui/pull/332/files#diff-fe41db46679f280131a7b0a04d39383dd4d0ab623cec8de8ddc74f79b163dabfR285

But, is this an appropriate link for more information on `IP Address`. What about: https://en.wikipedia.org/wiki/IP_address

ShaMan239:
Updated from 2da5d46 to a52f72f
Addressed [this](https://github.com/bitcoin-core/gui/pull/384#discussion_r692998313) comment by @jarolrod
- Updated translator comment to now direct towards the correct link with the information regarding the meaning of IP Address.

jarolrod:
tACK a52f72f

tested on Ubuntu 20.04 and macOS 12. PR works as intended. Translator comments look good to me

src/qt/rpcconsole.cpp
```diff
@@ -706,6 +706,14 @@ void RPCConsole::setClientModel(ClientModel *model, int bestblock_height, int64_
 
         // create ban table context menu
         banTableContextMenu = new QMenu(this);
+        /*: Context menu action to copy the IP/Netmask of a banned peer.
+            IP/Netmask is the combination of a peer's IP address and it's Netmask.
+            For IP Address see: https://en.wikipedia.org/wiki/IP_address
+            For NetMask see: https://wiki.teltonika-networks.com/view/What_is_a_Netmask%3F */
```
hebasto, line 712: I'd avoid mentioning any links to proprietary sites in the Bitcoin Core codebase.

Maybe, just skip last two lines of the comment?

jarolrod: wikipedia or the other website? we already link to wikipedia articles to help with translations: https://github.com/bitcoin-core/gui/blob/d3203a99d886177eee9d1f9cd8411e215118a4e6/src/qt/addressbookpage.cpp#L285

hebasto: At least, drop `https://wiki.teltonika-networks.com`, please.

PR #385, opened 2021-07-22
Always log context function of QtDebugMsg messages

promag:
Use `QMessageLogContext::function` consistently for all `qDebug` calls instead of explicitly logging `__func__` in some places.

### DIFF
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -147,9 +147,8 @@ static void initTranslations(QTranslator &qtTranslatorBase, QTranslator &qtTrans
 /* qDebug() message handler --> debug.log */
 void DebugMessageHandler(QtMsgType type, const QMessageLogContext& context, const QString &msg)
 {
-    Q_UNUSED(context);
     if (type == QtDebugMsg) {
-        LogPrint(BCLog::QT, "GUI: %s\n", msg.toStdString());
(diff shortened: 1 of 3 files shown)


hebasto:
The diff of last lines of logs for master and this pr with `-debug=qt -nologtimestamps`:
```diff
 [shutoff] Shutdown: done
-[shutoff] GUI: shutdown : Shutdown finished
-[main] GUI: ~InitExecutor : Stopping thread
-[main] GUI: ~InitExecutor : Stopped thread
+[shutoff] GUI: void InitExecutor::shutdown(): Shutdown finished
+[main] GUI: virtual InitExecutor::~InitExecutor(): Stopping thread
+[main] GUI: virtual InitExecutor::~InitExecutor(): Stopped thread
```

The output is more verbose.

What are benefits of this change?

Why `InitExecutor::~InitExecutor()` destructor is `virtual`?

hebasto:
Update messages in `qWarning()` calls?

hebasto:
How is this feature combined with `-logsourcelocations` command-line option?

promag:
> What are benefits of this change?

The idea is to clean up a bit (remove __func__) and make logging consistent.

> Why `InitExecutor::~InitExecutor()` destructor is `virtual`?

Because base class has a virtual destructor?

> How is this feature combined with `-logsourcelocations` command-line option?

Didn't test, will check.

PR #391, opened 2021-07-29
Add cancel button to configuration options popup

shaavan:
This PR renames the **OK** button to **Continue** and adds a **Cancel** button to the configuration options pop-up.

This feature will give the user an option to abort opening the configuration file if they want to. This is an essential helpful feature that was missing in the master branch.

In some windows managers such as Windows I3. The exit button at the top right corner is missing. So this feature becomes crucial there. And even when the exit button is there, it doesn't prevent the opening of the configuration file even when pressed.

Additionally, it will always be possible to close using Keyboard Shortcut. This PR helps accessibility for those who need to use a mouse.

<table>
  <tr>
   <td>Master
   </td>
   <td>PR
   </td>
  </tr>
  <tr>
   <td>

Cancel-conf master(1)

   </td>
   <td>

Screenshot from 2021-09-07 20-15-28

   </td>
  </tr>
</table>

### DIFF
--- a/src/qt/optionsdialog.cpp
+++ b/src/qt/optionsdialog.cpp
@@ -295,10 +295,22 @@ void OptionsDialog::on_resetButton_clicked()
 
 void OptionsDialog::on_openBitcoinConfButton_clicked()
 {
-    /* explain the purpose of the config file */
-    QMessageBox::information(this, tr("Configuration options"),
-        tr("The configuration file is used to specify advanced user options which override GUI settings. "
-           "Additionally, any command-line options will override this configuration file."));
+    QMessageBox config_msgbox(this);
+    config_msgbox.setIcon(QMessageBox::Information);
+    //: Window title text of pop-up box that allows opening up of configuration file.
+    config_msgbox.setWindowTitle(tr("Configuration options"));
+    /*: Explanatory text about the priority order of instructions considered by client.
+        The order from high to low being: command-line, configuration file, GUI settings. */
+    config_msgbox.setText(tr("The configuration file is used to specify advanced user options which override GUI settings. "
+                             "Additionally, any command-line options will override this configuration file."));
+
+    QPushButton* open_button = config_msgbox.addButton(tr("Continue"), QMessageBox::ActionRole);
+    config_msgbox.addButton(tr("Cancel"), QMessageBox::RejectRole);
+    open_button->setDefault(true);
+
+    config_msgbox.exec();
+
+    if (config_msgbox.clickedButton() != open_button) return;
 
     /* show an error if there was some problem opening the file */
     if (!GUIUtil::openBitcoinConf())


prayank23:
Concept ACK. Can remove extra `QMessageBox::Ok`

src/qt/optionsdialog.cpp
```diff
@@ -293,12 +293,21 @@ void OptionsDialog::on_resetButton_clicked()
...
 void OptionsDialog::on_openBitcoinConfButton_clicked()
 {
     /* explain the purpose of the config file */
-    QMessageBox::information(this, tr("Configuration options"),
-        tr("The configuration file is used to specify advanced user options which override GUI settings. "
-           "Additionally, any command-line options will override this configuration file."));
+    int ret = QMessageBox::information(this, tr("Configuration options"),
+                                       tr("The configuration file is used to specify advanced user options which override GUI settings. "
+                                          "Additionally, any command-line options will override this configuration file."),
+                                       QMessageBox::Cancel | QMessageBox::Ok, QMessageBox::Ok);
```
prayank23, line 303:
```suggestion
                                       QMessageBox::Cancel | QMessageBox::Ok);
```

ShaMan239: That extra `QMessageBox::Ok` specifies selecting the Ok button by default, and I think it is better to keep it as it is.

psancheti110:
I agree with the concept proposed @ShaMan239 conceptually, which aborts the opening of the config file in case of keystroke by mistake on the tab. 

There are a few points I believe should be addressed here about this window. 

-  The text that appears on the screen on clicking **_Open Configuration File_** is a message, meant for the user describing what a config file is. It is a pretty essential message for **new users** to Bitcoin and are trying to get hands-on experience with GUI. But because of the UI Symbol present in the dialogue, it appears more like a **warning** than **message**. This needs to be addressed.  

- I would like to put forward another approach that might enhance the feature. There are users/contributors who are very much familiar with Bitcoin GUI and hence are aware of what a configuration file is. Getting a message every time they try to open the configuration file can get irritating at some point.  
     
     Adding a **_Do not show this message again_** option will resort to it. This will allow the user to allow the UI to show this message or NOT to show this message based on their preference. If a user checks this option and proceeds, they will not be shown this message ever again unless they change the setting back manually. I believe this will serve both the purpose of why this message was placed in the window in the first place (to provide an important message to new users) and providing existing experienced users with an option to turn it off forever. 

Please let me know your views on the suggested addition. If it looks good, we could work on it together :)

jarolrod (commented):
Since the logic here is now of close | open, perhaps we could change the ok button into an open button. But the icon associated with the open button isn't great.:

alt


Neither is the Close button icon as pointed out by @psancheti110:
>  But because of the UI Symbol present in the dialogue, it appears more like a warning than message. This needs to be addressed.

To address this let's just not use any icons.

psancheti110:
@jarolrod, 

> To address this let's just not use any icons. 

I like this idea better. That might clear the confusion and depict it like a message as it is supposed to be.

Also would appreciate it if you could provide your views on the **Do not show this message again** I have mentioned above?

src/qt/optionsdialog.cpp
```diff
@@ -293,12 +293,21 @@ void OptionsDialog::on_resetButton_clicked()
     }
 }
 
+
```
hebasto, line 296: Why this extra line?

src/qt/optionsdialog.cpp
```diff
@@ -293,12 +293,21 @@ void OptionsDialog::on_resetButton_clicked()
...
 void OptionsDialog::on_openBitcoinConfButton_clicked()
 {
     /* explain the purpose of the config file */
-    QMessageBox::information(this, tr("Configuration options"),
-        tr("The configuration file is used to specify advanced user options which override GUI settings. "
-           "Additionally, any command-line options will override this configuration file."));
+    int ret = QMessageBox::information(this, tr("Configuration options"),
```
hebasto, line 300: Could be `const`?

src/qt/optionsdialog.cpp
```diff
@@ -293,12 +293,21 @@ void OptionsDialog::on_resetButton_clicked()
...
+                                       tr("The configuration file is used to specify advanced user options which override GUI settings. "
```
hebasto, line 301: While these lines are touched, maybe add a commit with translator comments?

src/qt/optionsdialog.cpp
```diff
@@ -293,12 +293,21 @@ void OptionsDialog::on_resetButton_clicked()
...
+                                          "Additionally, any command-line options will override this configuration file."),
+                                       QMessageBox::Cancel | QMessageBox::Ok, QMessageBox::Ok);
+
+    switch (ret) {
+    case QMessageBox::Cancel:
+        return;
+    case QMessageBox::Ok:
+        break;
+    }
```
hebasto, line 310: Is it simpler
```suggestion
    if (ret == QMessageBox::Cancel) return;
```
?

hebasto (commented):
Approach ACK f9ff47c6ed77cb6502f42fc37c7aa71ca92af476

> This feature will give the user an option to abort opening the configuration file if they want to.

Agree.

> In some windows managers such as Windows I3. The exit button at the top right corner is missing. So this feature becomes crucial there. And even when the exit button is there, it does not allow you to cancel opening the configuration file.
> 
> Additionally, it will always be possible to close using Keyboard Shortcut. This PR helps accessibility for those who need to use a mouse.

I didn't quite get it. On master I have no means to abort a configuration file opening at all:
- no Cancel button
- closing window with mouse just continues the file opening
- closing window with Esc just continues the file opening

hebasto:
In the PR description

127555187-db217eb1-e4c0-43eb-91fd-1ef04cba3f9d

there is a "Close" button. Is it a correct screenshot?

src/qt/optionsdialog.cpp
hebasto: Or even better:
```
    if (ret != QMessageBox::Ok) return;
```

The latter variant fixes a bug which also is present in master: when the dialog is opened and the program is being shutdown (e.g., via the `Exit` action of the systray icon menu) a configuration file gets opened.

ShaMan239:
Updated from f9ff47c6ed77cb6502f42fc37c7aa71ca92af476 to 25bee9251374fa78d19077f2fb5fa0bfc5d651a2 ([pr391.01](https://github.com/ShaMan239/gui/tree/pr391.01) -> [pr391.03](https://github.com/ShaMan239/gui/tree/pr391.03))

Addressed comments of @JarolRod, @PSancheti, and @Hebasto
Changes made are:
1. Removed Icons for Buttons
2. Changed ordering of button to now match the sequence of open and cancel button of the parent option dialog box
3. Fixed Clang formatting

src/qt/optionsdialog.cpp
ShaMan239: > Why this extra line?

Sorry, that was a blunder on my part. Fixed in 25bee9251374fa78d19077f2fb5fa0bfc5d651a2

src/qt/optionsdialog.cpp
ShaMan239: Changed the declaration style in 25bee9251374fa78d19077f2fb5fa0bfc5d651a2. Now instead of `int ret` now `QMessageBox config_msgbox`, is used.

src/qt/optionsdialog.cpp
ShaMan239: Added translator comments in 25bee9251374fa78d19077f2fb5fa0bfc5d651a2

src/qt/optionsdialog.cpp
ShaMan239: > if (ret != QMessageBox::Ok) return;

That's a great suggestion. Updated `switch/case` to this format, in 25bee9251374fa78d19077f2fb5fa0bfc5d651a2

ShaMan239:
> there is a "Close" button. Is it a correct screenshot?

The PR description was erroneous and is now updated.

hebasto:
```
qt/optionsdialog.cpp: In member function ‘void OptionsDialog::on_openBitcoinConfButton_clicked()’:
qt/optionsdialog.cpp:308:18: error: unused variable ‘close_button’ [-Werror=unused-variable]
     QPushButton* close_button = config_msgbox.addButton(tr("Close"), QMessageBox::RejectRole);
                  ^~~~~~~~~~~~
```

ShaMan239:
Updated from 25bee92 to 13930ea ([pr391.03](https://github.com/ShaMan239/gui/tree/pr391.03) to [pr391.04](https://github.com/ShaMan239/gui/tree/pr391.04))

Addressed [this](https://github.com/bitcoin-core/gui/pull/391#issuecomment-896842514) comment by @hebasto 

Changes made:
- Removed the added **Close** button to be declared as a separate variable, which corrected the `unused variable warning.

src/qt/optionsdialog.cpp
```diff
@@ -295,10 +295,22 @@ void OptionsDialog::on_resetButton_clicked()
...
 {
-    /* explain the purpose of the config file */
-    QMessageBox::information(this, tr("Configuration options"),
-        tr("The configuration file is used to specify advanced user options which override GUI settings. "
-           "Additionally, any command-line options will override this configuration file."));
+    QMessageBox config_msgbox;
+    config_msgbox.setIcon(QMessageBox::Information);
+    //: Option file to give custom user configuration instructions.
+    config_msgbox.setWindowTitle(tr("Configuration options"));
+    /*: The instructions written in this file will overwrite settings of Graphical User Interface.
```
jarolrod, line 302: The translator comment is not very effective in helping me translate the corresponding string. Maybe re-think how it could be better.

A good exercise is to really put yourself in the shoes of someone tasked to translate this. What information/context would you need to be able to effectively translate this string. Then use that to craft your translator comment.

ShaMan239:
Updated from 13930ea to bb17cd8 ([pr391.04](https://github.com/ShaMan239/gui/tree/pr391.04) -> [pr391.05](https://github.com/ShaMan239/gui/tree/pr391.05))
changes: addressed [this](https://github.com/bitcoin-core/gui/pull/391#discussion_r692685054) change suggested by @jarolrod
- Updated translator comments.

ShaMan239:
Updated from bb17cd8 to ddec692 ([pr391.04](https://github.com/ShaMan239/gui/tree/pr391.04) -> [pr391.05](https://github.com/ShaMan239/gui/tree/pr391.05))
changes: addressed [this](https://github.com/bitcoin-core/gui/pull/391#discussion_r692685054) and [this](https://github.com/bitcoin-core/gui/pull/391#discussion_r687033508) changes suggested by @jarolrod
- Updated translator comments.
- Removed unnecessary comment about action role.

Thanks, @jarolrod  for the suggestions.

ShaMan239:
Updated from ddec692 to 4a0b791
The first force commit was pushed by mistake, and the second one is a correction to it.
For the reviewers: **There are no changes made by the above two commits**

jarolrod:
tACK  4a0b791

This window looks good now without the confusing icons. PR functions as intended.

Screenshot of current window:
pr391

hebasto:
@GBKS @bosch-0 

Could you look into this PR, please?

What, in your designer's opinion, is the best text on:
1) the button which continues the procedure and actually opens the `bitcoin.conf` file
2) the button which aborts the procedure of opening the `bitcoin.conf`, and returns the GUI to the previous state

?

GBKS:
How about "Continue" and "Cancel"?

Just as a side note, on my Mac, the pop-up does not have a title, so I don't see the text "Configuration options" shown in the screenshots above.

src/qt/optionsdialog.cpp
```diff
@@ -295,10 +295,22 @@ void OptionsDialog::on_resetButton_clicked()
...
 {
-    /* explain the purpose of the config file */
-    QMessageBox::information(this, tr("Configuration options"),
-        tr("The configuration file is used to specify advanced user options which override GUI settings. "
-           "Additionally, any command-line options will override this configuration file."));
+    QMessageBox config_msgbox;
+    config_msgbox.setIcon(QMessageBox::Information);
+    //: Window title text of pop-up box, which allows opening up configuration file.
+    config_msgbox.setWindowTitle(tr("Configuration options"));
+    /*: Self explanatory text about the priority order of instructions considered by client.
```
jarolrod, line 302:
```suggestion
    /*: Explanatory text about the priority order of instructions considered by client.
```

src/qt/optionsdialog.cpp
```diff
@@ -295,10 +295,22 @@ void OptionsDialog::on_resetButton_clicked()
...
+        the order from high to low being: command-line -> configuration file -> GUI settings. */
```
jarolrod, line 303: You cannot use the arrow in the translator comments; they show up as something else, see the output of `make -C src translate`:

```
<extracomment>Self explanatory text about the priority order of instructions considered by client. the order from high to low being: command-line -&gt; configuration file -&gt; GUI settings.</extracomment>
```

as you can see, the arrow's show up as `-&gt;`

```suggestion
        The order from high to low being: command-line, configuration file, GUI settings. */
```

jarolrod (commented):
@ShaMan239 we should adopt @GBKS suggestion as a designer, so I retract my ack. Please update the button text according to his suggestions. Below are some other suggestions related to the translator comments.

@GBKS thanks for taking the time to review a GUI PR. What you've noticed about the missing window title of a QMessageBox on macOS is the expected behavior, see: [QMessageBox::setWindowTitle](https://doc.qt.io/qt-5/qmessagebox.html#setWindowTitle)

> Sets the title of the message box to title. On macOS, the window title is ignored (as required by the macOS Guidelines).

ShaMan239:
Updated from 4a0b791 to 52aec5dd2338a3af627a6897dd5efe59a928bd10 ([pr391.05](https://github.com/ShaMan239/gui/tree/pr391.05) -> [pr391.06](https://github.com/ShaMan239/gui/tree/pr391.06))
Addressed @GBKS and @jarolrod comments. Thanks, @GBKS, for taking the time to review this PR and giving your valuable suggestions. Thanks, @jarolrod, for suggesting the appropriate changes in the translator's comments that I had missed earlier.

Changes:
- Updated buttons names from “Open” and “Close” to “Continue” and “Cancel.”
- Updated translator comments as per @jarolrod suggestions.
- Updated PR description accordingly

src/qt/optionsdialog.cpp
```diff
@@ -295,10 +295,22 @@ void OptionsDialog::on_resetButton_clicked()
...
+    QMessageBox config_msgbox;
+    config_msgbox.setIcon(QMessageBox::Information);
+    //: Option file to give custom user configuration instructions.
+    config_msgbox.setWindowTitle(tr("Configuration options"));
+    /*: The instructions written in this file will overwrite settings of Graphical User Interface.
+        This instructions will themselves be overwritten by instructions given through command-line.*/
+    config_msgbox.setText(tr("The configuration file is used to specify advanced user options which override GUI settings. "
+                             "Additionally, any command-line options will override this configuration file."));
+
+    QPushButton* open_button = config_msgbox.addButton(tr("Open"), QMessageBox::ActionRole); //setting ActionRole to fix ordering issue of the buttons
```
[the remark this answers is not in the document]
ShaMan239: Addressed in ddec692.

src/qt/optionsdialog.cpp
```diff
@@ -295,10 +295,22 @@ void OptionsDialog::on_resetButton_clicked()
 
 void OptionsDialog::on_openBitcoinConfButton_clicked()
 {
-    /* explain the purpose of the config file */
-    QMessageBox::information(this, tr("Configuration options"),
-        tr("The configuration file is used to specify advanced user options which override GUI settings. "
-           "Additionally, any command-line options will override this configuration file."));
+    QMessageBox config_msgbox;
+    config_msgbox.setIcon(QMessageBox::Information);
+    //: Option file to give custom user configuration instructions.
```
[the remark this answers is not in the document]
ShaMan239: Addressed in ddec692.

src/qt/optionsdialog.cpp
ShaMan239: Addressed in ddec692.

src/qt/optionsdialog.cpp
```diff
@@ -295,10 +295,22 @@ void OptionsDialog::on_resetButton_clicked()
 
 void OptionsDialog::on_openBitcoinConfButton_clicked()
 {
-    /* explain the purpose of the config file */
-    QMessageBox::information(this, tr("Configuration options"),
-        tr("The configuration file is used to specify advanced user options which override GUI settings. "
-           "Additionally, any command-line options will override this configuration file."));
+    QMessageBox config_msgbox;
+    config_msgbox.setIcon(QMessageBox::Information);
+    //: Window title text of pop-up box, which allows opening up configuration file.
```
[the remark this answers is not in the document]
ShaMan239: Addressed in 52aec5d.

src/qt/optionsdialog.cpp
ShaMan239: Addressed in 52aec5d.

src/qt/optionsdialog.cpp
ShaMan239: Addressed in 52aec5d.

jarolrod: It is not specific to the combination of characters that to us mean an arrow: `->`. What's happening is that the `>` symbol gets converted to `&gt`.

prayank23 (approved):
utACK https://github.com/bitcoin-core/gui/pull/391/commits/52aec5dd2338a3af627a6897dd5efe59a928bd10

src/qt/optionsdialog.cpp
```diff
@@ -295,10 +295,22 @@ void OptionsDialog::on_resetButton_clicked()
 
 void OptionsDialog::on_openBitcoinConfButton_clicked()
 {
-    /* explain the purpose of the config file */
-    QMessageBox::information(this, tr("Configuration options"),
-        tr("The configuration file is used to specify advanced user options which override GUI settings. "
-           "Additionally, any command-line options will override this configuration file."));
+    QMessageBox config_msgbox;
```
hebasto, line 298: To preserve behavior
```suggestion
    QMessageBox config_msgbox(this);
```

https://doc.qt.io/qt-5/qdialog.html:
> A dialog is always a top-level widget, but if it has a parent, its default location is centered on top of the parent's top-level widget (if it is not top-level itself).

src/qt/optionsdialog.cpp
```diff
@@ -295,10 +295,22 @@ void OptionsDialog::on_resetButton_clicked()
...
+    config_msgbox.setIcon(QMessageBox::Information);
+    //: Window title text of pop-up box that allows you to open up your configuration file.
```
hebasto, line 300: nit:

Translator comments are kind of documentation. Could they be more formal, without "you" and "your"?

src/qt/optionsdialog.cpp
```diff
@@ -295,10 +295,22 @@ void OptionsDialog::on_resetButton_clicked()
...
+    config_msgbox.setWindowTitle(tr("Configuration options"));
+    /*: Explanatory text about the priority order of instructions considered by client.
+        The order from high to low being: command-line, configuration file, GUI settings. */
+    config_msgbox.setText(tr("The configuration file is used to specify advanced user options which override GUI settings. "
+                             "Additionally, any command-line options will override this configuration file."));
+
+    QPushButton* open_button = config_msgbox.addButton(tr("Continue"), QMessageBox::ActionRole);
```
hebasto, line 307: Not sure if `QMessageBox::ActionRole` is a good choice.

[`QMessageBox::ActionRole`](https://doc.qt.io/qt-5/qmessagebox.html#ButtonRole-enum): 
> Clicking the button causes _changes to the elements within the dialog_.

Maybe `QMessageBox::AcceptRole` or `QMessageBox::YesRole`?

In case of `QMessageBox::AcceptRole` the button positions are evaluated according to the system-dependant styles.
And s/`open_button->setDefault(true);`/`config_msgbox.setDefaultButton(open_button);`/ is required.

shaavan: Both `QMessageBox::AcceptRole` and `QMessageBox::YesRole` are leading to changing the ordering of `continue` and `cancel` buttons. 
I am trying to keep ordering buttons in the `Configuration Options` window and `Options window` same to maintain consistency but both `AcceptRole` and `YesRole` mess that up.
Screenshot from 2021-09-10 18-47-46
I guess there must be some way of using one of these roles without disrupting the ordering, but I cannot find it yet.

shaavan:
Updated from 52aec5d to 0b869df1c913839855148d728514c76ba7664092 ([pr391.06](https://github.com/shaavan/gui/tree/pr391.06) -> [pr391.07](https://github.com/shaavan/gui/tree/pr391.07))
Addressed [1](https://github.com/bitcoin-core/gui/pull/391#discussion_r705467338) and [2](https://github.com/bitcoin-core/gui/pull/391#discussion_r705469034) comments by @hebasto

Changes made:
- Rewritten Translators’ comment to now be not addressed to a second person
- `QMessageBox config_msgbox` -> `QMessageBox config_msgbox(this)` to preserve behavior.

src/qt/optionsdialog.cpp
shaavan: Addressed in  0b869df

src/qt/optionsdialog.cpp
shaavan: Addressed in  0b869df

hebasto (approved):
ACK 0b869df1c913839855148d728514c76ba7664092, tested on Linux Mint 20.2 (Qt 5.12.8):

Screenshot from 2021-09-12 18-27-00

prayank23 (approved):
tACK https://github.com/bitcoin-core/gui/pull/391/commits/0b869df1c913839855148d728514c76ba7664092

Tried on Windows 10

image

`Continue` and `Cancel` works as expected.

shaavan:
Thank you very much to everyone who had given their valuable time to review this PR.

PR #29, opened 2020-07-10
refactor: Optimize signal-slot connections logic

hebasto:
This PR:
- implements an idea from https://github.com/bitcoin/bitcoin/pull/17937#issuecomment-575991765
- simplifies `WalletFrame` class interface
- as a side effect, removes `bitcoingui` -> `walletframe` -> `bitcoingui` circular dependency
- is an alternative to https://github.com/bitcoin/bitcoin/pull/17500

PR #393, opened 2021-08-03
Fix regression in "Encrypt Wallet" menu item

### LINKED ISSUE #392: Control changes from clickable to unclickable upon canceling.
Load a wallet, "Settings" > "Encrypt Wallet..." can be clicked. Click it, modal opens, cancel modal, now "Settings" > "Encrypt Wallet..." cannot be clicked ("disabled grey").

**Expected behavior**

The control being disabled from the beginning.

**Actual behavior**

Control started clickable and idempotence not happening until after having clicked the control.

**To reproduce**

Open Bitcoin GUI, load the wallet (if not loaded). Close Bitcoin GUI. Open it again so that it loads the wallet automatically. Reproduce. After closing the wallet, the issue doesn't happen anymore by reloading the wallet, until closing Bitcoin GUI and launching it again.

**System information**

This is v22.0rc1 built from sources, but I can check master and previous tags as necessary. I'd say the 0.21.1 tag is free from this behavior.

This is Debian Buster (10) right on top of the amd64 metal.

Qt reports 5.11.3, Plasma is 5.14.5 (I think).

I'd be bisecting this soon.

hebasto:
Fix #392.

Adding a new item to the `m_wallet_selector` must follow the establishment of a connection between the `WalletView::encryptionStatusChanged` signal and the `BitcoinGUI::updateWalletStatus` slot.

This was a regression introduced in https://github.com/bitcoin/bitcoin/commit/20e2e24e90d782219e853ef0676ac66dc6a9de6a (#29).

---

An _encrypted_ wallet being auto-loaded at the GUI startup:
- on master (eaf09bda4ab21f79f89822d2c6fa3d7a3ce57b0d)

Screenshot from 2021-08-03 22-38-49

- with this PR

Screenshot from 2021-08-03 22-34-58

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -682,8 +682,6 @@ void BitcoinGUI::addWallet(WalletModel* walletModel)
         m_wallet_selector_label_action->setVisible(true);
         m_wallet_selector_action->setVisible(true);
     }
-    const QString display_name = walletModel->getDisplayName();
-    m_wallet_selector->addItem(display_name, QVariant::fromValue(walletModel));
 
     connect(wallet_view, &WalletView::outOfSyncWarningClicked, this, &BitcoinGUI::showModalOverlay);
     connect(wallet_view, &WalletView::transactionClicked, this, &BitcoinGUI::gotoHistoryPage);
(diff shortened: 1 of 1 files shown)


hebasto:
cc @katesalazar @jarolrod @meshcollider @achow101 @promag

achow101:
ACK d54d94959869b0c363939163b99ba0475751dcb6

Tested that this fixes the issue.

For those wondering why this fixes the issue, `m_wallet_selector->addItem` ends up calling a lot of setup functions and emits several signals. One of those signals is `encryptionStatusChanged`. However previously this signal was not connected to its slot so when that signal was emitted, it did not have the intended effect of greying out the Encrypt Wallet option. By moving `addItem` to after all of the signals have been connected, the signal will be correctly handled.

jarolrod (commented):
ACK d54d94959869b0c363939163b99ba0475751dcb6

Tested on macOS 12 and Ubuntu 20.4.02

The related code is a bit entangled, but I have ensured that this is correct in its implementation and is a good way to fix the issue. Below are screenshots comparing master and this PR:

| Master.       |  PR.             |
| ----------- | ----------- |
| master | pr |

This was tested by starting with an already encrypted wallet. On the master branch, we can notice the issue described in #392. On the PR branch, we can see that this is fixed.

hebasto:

PR #396, opened 2021-08-06
Ensure external signer option remains disabled without signers

### LINKED ISSUE #395: External Signer can be enabled even when external signer script is not set
The `External Signer` option when creating a wallet is supposed to be disabled while the external signer script has not been set in settings. There is a way to enable this option even though the script has not been set, then if you were to create an External Signer wallet the gui will crash because the external signer script was never set:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  GetExternalSigner: restart bitcoind with -signer=<cmd>
Aborted (core dumped)
```

**Steps to reproduce:**

No special preparation. Just required that the external signer script has not be set in settings. This is to ensure that proper functionality is that the `External Signer` script option is supposed to be disabled.

1) Go to create a new wallet
2) Notice that the `External Signer` option is disabled, as it is supposed to be
3) Select, then de-select `Encrypt Wallet`
4) The `External Signer` option is now enabled, when it is not supposed to be
5) Select the now enabled `External Signer` option and create wallet
6) GUI will crash

achow101:
When no external signers are available, the option to enable external signers should always be disabled. However the encrypt wallet checkbox can erroneously re-enable the external signer checkbox. To avoid this, CreateWalletDialog now stores whether signers were available during setSigners so that future calls to external_signer_checkbox->setEnabled can account for whether signers are available.

Fixes #395

### DIFF
--- a/src/qt/createwalletdialog.cpp
+++ b/src/qt/createwalletdialog.cpp
@@ -32,7 +32,7 @@ CreateWalletDialog::CreateWalletDialog(QWidget* parent) :
         // set to true, enable it when isEncryptWalletChecked is false.
         ui->disable_privkeys_checkbox->setEnabled(!checked);
 #ifdef ENABLE_EXTERNAL_SIGNER
-        ui->external_signer_checkbox->setEnabled(!checked);
+        ui->external_signer_checkbox->setEnabled(m_has_signers && !checked);
 #endif
         // When the disable_privkeys_checkbox is disabled, uncheck it.
         if (!ui->disable_privkeys_checkbox->isEnabled()) {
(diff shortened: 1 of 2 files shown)


achow101:
Needs backport to 22.0

src/qt/createwalletdialog.cpp
```diff
@@ -32,7 +32,7 @@ CreateWalletDialog::CreateWalletDialog(QWidget* parent) :
         // set to true, enable it when isEncryptWalletChecked is false.
         ui->disable_privkeys_checkbox->setEnabled(!checked);
 #ifdef ENABLE_EXTERNAL_SIGNER
-        ui->external_signer_checkbox->setEnabled(!checked);
+        ui->external_signer_checkbox->setEnabled(m_has_signers && !checked);
```
hebasto, line 35: Maybe
```suggestion
        if (m_has_signers) {
            ui->external_signer_checkbox->setEnabled(!checked);
        }
```
to make the same code structure as it is in `CreateWalletDialog::setSigners` for better readability?

hebasto (approved):
ACK a9b9ca82daefc77ee3c884d3f250460d7cf734a5, tested on Linux Mint 20.2 (Qt 5.12.8).

Sjors (approved):
tACK a9b9ca82daefc77ee3c884d3f250460d7cf734a5

jarolrod (commented):
ACK a9b9ca82daefc77ee3c884d3f250460d7cf734a5

Extensively tested on Ubuntu 20.04. Thank you for the speedy fix!

hebasto:

PR #398, opened 2021-08-07
refactor: Pass WalletModel object to the WalletView constructor 

hebasto:
An instance of the `WalletView` class without the `walletModel` data member being set is invalid. So, it is better to set it in the constructor.

Establishing one more `WalletView` class's invariant in constructor:
- allows to drop all of checks of the`walletModel` in member functions
- makes reasoning about the code that uses instances of the `WalletView` class  easier

Possible follow ups could extend this approach to other classes, e.g., `OverviewPage`, `TransactionView`, `ReceiveCoinsDialog`, `SendCoinsDialog`, `AddressBookPage`.

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -676,7 +676,7 @@ void BitcoinGUI::addWallet(WalletModel* walletModel)
 {
     if (!walletFrame) return;
 
-    WalletView* wallet_view = new WalletView(platformStyle, walletFrame);
+    WalletView* wallet_view = new WalletView(walletModel, platformStyle, walletFrame);
     if (!walletFrame->addWallet(walletModel, wallet_view)) return;
 
     rpcConsole->addWallet(walletModel);
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -677,7 +677,7 @@ void BitcoinGUI::addWallet(WalletModel* walletModel)
     if (!walletFrame) return;
 
     WalletView* wallet_view = new WalletView(walletModel, platformStyle, walletFrame);
-    if (!walletFrame->addWallet(walletModel, wallet_view)) return;
+    if (!walletFrame->addView(wallet_view)) return;
 
     rpcConsole->addWallet(walletModel);
     if (m_wallet_selector->count() == 0) {
--- a/src/qt/walletview.cpp
+++ b/src/qt/walletview.cpp
@@ -30,19 +30,24 @@
 #include <QPushButton>
 #include <QVBoxLayout>
 
-WalletView::WalletView(const PlatformStyle *_platformStyle, QWidget *parent):
-    QStackedWidget(parent),
-    clientModel(nullptr),
-    walletModel(nullptr),
-    platformStyle(_platformStyle)
+WalletView::WalletView(WalletModel* wallet_model, const PlatformStyle* _platformStyle, QWidget* parent)
+    : QStackedWidget(parent),
+      clientModel(nullptr),
+      walletModel(wallet_model),
+      platformStyle(_platformStyle)
 {
+    assert(walletModel);
+
     // Create tabs
     overviewPage = new OverviewPage(platformStyle);
+    overviewPage->setWalletModel(walletModel);
 
     transactionsPage = new QWidget(this);
     QVBoxLayout *vbox = new QVBoxLayout();
     QHBoxLayout *hbox_buttons = new QHBoxLayout();
     transactionView = new TransactionView(platformStyle, this);
+    transactionView->setModel(walletModel);
+
     vbox->addWidget(transactionView);
     QPushButton *exportButton = new QPushButton(tr("&Export"), this);
     exportButton->setToolTip(tr("Export the data in the current tab to a file"));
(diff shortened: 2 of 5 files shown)


Talkless:
Concept ACK

src/qt/walletview.cpp
```diff
@@ -100,49 +127,15 @@ void WalletView::setClientModel(ClientModel *_clientModel)
...
-    {
-        // Receive and pass through messages from wallet model
-        connect(_walletModel, &WalletModel::message, this, &WalletView::message);
-
-        // Handle changes in encryption status
-        connect(_walletModel, &WalletModel::encryptionStatusChanged, this, &WalletView::encryptionStatusChanged);
-        updateEncryptionStatus();
-
-        // update HD status
-        Q_EMIT hdEnabledStatusChanged();
```
Talkless, line 128: I don't see this signal being emmited in new code. It's because initialization in the constructor, as there's no "change" per se?

src/qt/walletview.h
```diff
@@ -35,26 +35,26 @@ class WalletView : public QStackedWidget
...
         The client model represents the part of the core that communicates with the P2P network, and is wallet-agnostic.
     */
     void setClientModel(ClientModel *clientModel);
-    WalletModel *getWalletModel() { return walletModel; }
-    /** Set the wallet model.
-        The wallet model represents a bitcoin wallet, and offers access to the list of transactions, address book and sending
-        functionality.
-    */
-    void setWalletModel(WalletModel *walletModel);
+    WalletModel* getWalletModel() const { return walletModel; }
```
Talkless, line 45: Maybe worth adding `noexcept` and `[[nodiscard]]` too? Or at least just `noexcept`, as it's not "heavyweight" operation to ask for "nodsicardness", but still...

Talkless: Also forgot to add, `WalletModel *walletModel;` pointer itself now can be `const`.

hebasto: > Also forgot to add, `WalletModel *walletModel;` pointer itself now can be `const`.

```
In file included from qt/moc_walletview.cpp:9:
./qt/walletview.h:45:50: error: cannot initialize return object of type 'WalletModel *' with an lvalue of type 'const WalletModel *const'
    WalletModel* getWalletModel() const { return walletModel; }
                                                 ^~~~~~~~~~~
1 error generated
```

Talkless (approved):
utACK bb56486a170aacb355f4a973f0cd40ab3918a0cd

src/qt/walletview.h
Talkless: Not `const WalletModel *` but `WalletModel *const walletModel`. Pointer itself `const`, not the object pointed to.

hebasto:
@Talkless 

> I don't see this signal being emmited in new code. It's because initialization in the constructor, as there's no "change" per se?

#403 could make things simpler.

So labeling this PR as a draft for now, and begging to review #403 first.

src/qt/bitcoingui.cpp
```diff
@@ -676,7 +676,7 @@ void BitcoinGUI::addWallet(WalletModel* walletModel)
 {
     if (!walletFrame) return;
 
-    WalletView* wallet_view = new WalletView(platformStyle, walletFrame);
+    WalletView* wallet_view = new WalletView(walletModel, platformStyle, walletFrame);
     if (!walletFrame->addWallet(walletModel, wallet_view)) return;
```
promag, line 680: Could a commit to simplify like
```diff
-bool WalletFrame::addWallet(WalletModel* walletModel, WalletView* walletView)
+bool WalletFrame::addView(WalletView* walletView)
```

hebasto:
Updated 92ddc02a16a74e10f24190929f05e2dcf2b55871 -> d319c4dae9ed7d59d71b926e677707fce4194d0c ([pr398.02](https://github.com/hebasto/gui/commits/pr398.02) -> [pr398.03](https://github.com/hebasto/gui/commits/pr398.03), [diff](https://github.com/hebasto/gui/compare/pr398.02..pr398.03)):

- addressed @promag's [comment](https://github.com/bitcoin-core/gui/pull/398#discussion_r701421756)

src/qt/bitcoingui.cpp
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui/pull/398#issuecomment-912071917).

ShaMan239 (commented):
Code review ACK d319c4dae9ed7d59d71b926e677707fce4194d0c

These changes seem to simplify the code a great bit. Thanks for this amazing work!

I have compiled some notes regarding the changes made in each of the commits in this PR. However, @hebasto has very clearly mentioned the details in the commit message of each commit. I still think my notes might be of some help to fellow reviewers.

NOTES:
- **qt, refactor: Pass WalletModel object to WalletView constructor**
    - The process of creating `walletmodel` has been moved under the constructor of the `WalletView` class.
    - The redundant function of creating a `walletmodel` has been removed.
    - `walletview.h` file has been appropriately modified to account for all the changes made in the `walletview.cpp` file.
- **qt, refactor: Drop redundant checks of walletModel**
    - As the name suggests, all the redundant checks of `walletModel` in the `walletview.cpp` file have been removed, as the `walletmodel` is now constructed during the construction of the `walletview` object itself.
- **qt, refactor: Declare getWalletModel with const and noexcept qualifiers** 
    - This is done to ensure that the function is not changing a member variable’s value. And the given function will not throw an exception (since the `walletmodel` is generated with the constructor itself).
- **qt, refactor: Replace WalletFrame::addWallet with WalletFrame::addView**
    - Modified the `addWallet` function such that it now doesn’t need to take `walletmodel` as an argument. And also removed the need to check for the existence of a `walletmodel` inside this function.

If some of my observation is erroneous or lacking thorough observation, please do inform me. 
<p>

promag (commented):
Code review ACK d319c4dae9ed7d59d71b926e677707fce4194d0c.

hebasto:
@Talkless @jarolrod 

Do you mind having another look into this PR?

jarolrod (commented):
ACK d319c4dae9ed7d59d71b926e677707fce4194d0c

I did light testing with different wallets and switching between them. I also ran the qt tests locally. Performed identical testing steps on macOS and on Ubuntu.

This is a nice simplification of the WalletModel/WalletView logic. It ties up nicely at the end with the last commit where we now only need to pass the WalletView object; not both a WalletModel and WalletView object.

@ShaMan239 your notes on the PR are excellent and I looked at them before beginning my review. Thanks for posting them!

PR #399, opened 2021-08-07
Fix "Load PSBT" functionality when no wallet loaded

### LINKED ISSUE #232: "Load PSBT from file..." does not work when no wallet is loaded
Linux Mint 20.1 (Qt 5.12.8), the master (ad89812c70a9b88be642fceffb970859f6fdcba2):
Screenshot from 2021-03-01 19-18-59

opens no dialog.

From [IRC](http://www.erisian.com.au/bitcoin-core-dev/log-2021-03-01.html#l-192):
```
<hebasto> achow101: does "File" -> "Load PSBT from {file|clipboard}" make any sense when no wallet is loaded?
<achow101> hebasto: yes, for finalize and sending
```

psancheti110:
This PR provides a fix to the issue mentioned in #232. 

Currently, the **_Load PSBT_** functionality works well in case a wallet is loaded but does nothing when a wallet isn't loaded. 

If a function cannot work without a wallet being loaded, it is disabled by default (It is unclickable as shown in the image). 

For e.g. One cannot `Close Wallet` or `Backup Wallet` or `Sign Messages` without a wallet being loaded. And hence they are disabled. But if you notice, `Load PSBT` options are not disabled by default even when a wallet isn't loaded.

>  Screenshot 2021-08-07 at 11 46 30 PM

As mentioned by hebasto in the issue description : 

``` 
<hebasto> achow101: does "File" -> "Load PSBT from {file|clipboard}" make any sense when no wallet is loaded?
<achow101> hebasto: yes, for finalize and sending
```

This means **_Load PSBT_**  should be working just as similar whether wallets are being loaded or not. 

After making the required changes to the code, The **_Load PSBT_** works as expected even with no wallet loaded and the PSBT is finalized. 

| Master | PR | 
|-------------|---------------|
| Hnet com-image (1) | Hnet com-image (2) | 

Close #232

### DIFF
--- a/src/qt/psbtoperationsdialog.cpp
+++ b/src/qt/psbtoperationsdialog.cpp
@@ -47,18 +47,22 @@ void PSBTOperationsDialog::openWithPSBT(PartiallySignedTransaction psbtx)
 {
     m_transaction_data = psbtx;
 
-    bool complete;
-    size_t n_could_sign;
-    FinalizePSBT(psbtx);  // Make sure all existing signatures are fully combined before checking for completeness.
-    TransactionError err = m_wallet_model->wallet().fillPSBT(SIGHASH_ALL, false /* sign */, true /* bip32derivs */, &n_could_sign, m_transaction_data, complete);
-    if (err != TransactionError::OK) {
-        showStatus(tr("Failed to load transaction: %1")
-            .arg(QString::fromStdString(TransactionErrorString(err).translated)), StatusLevel::ERR);
-        return;
+    bool complete = FinalizePSBT(psbtx); // Make sure all existing signatures are fully combined before checking for completeness.
+    if (m_wallet_model) {
+        size_t n_could_sign;
+        TransactionError err = m_wallet_model->wallet().fillPSBT(SIGHASH_ALL, false /* sign */, true /* bip32derivs */, &n_could_sign, m_transaction_data, complete);
+        if (err != TransactionError::OK) {
+            showStatus(tr("Failed to load transaction: %1")
+                           .arg(QString::fromStdString(TransactionErrorString(err).translated)),
+                       StatusLevel::ERR);
+            return;
+        }
+        m_ui->signTransactionButton->setEnabled(!complete && !m_wallet_model->wallet().privateKeysDisabled() && n_could_sign > 0);
+    } else {
+        m_ui->signTransactionButton->setEnabled(false);
     }
 
     m_ui->broadcastTransactionButton->setEnabled(complete);
-    m_ui->signTransactionButton->setEnabled(!complete && !m_wallet_model->wallet().privateKeysDisabled() && n_could_sign > 0);
 
     updateTransactionDisplay();
 }
@@ -133,7 +137,7 @@ void PSBTOperationsDialog::saveTransaction() {
         }
         CTxDestination address;
         ExtractDestination(out.scriptPubKey, address);
-        QString amount = BitcoinUnits::format(m_wallet_model->getOptionsModel()->getDisplayUnit(), out.nValue);
+        QString amount = BitcoinUnits::format(m_client_model->getOptionsModel()->getDisplayUnit(), out.nValue);
         QString address_str = QString::fromStdString(EncodeDestination(address));
         filename_suggestion.append(address_str + "-" + amount);
         first = false;
@@ -224,6 +228,10 @@ void PSBTOperationsDialog::showStatus(const QString &msg, StatusLevel level) {
 }
 
 size_t PSBTOperationsDialog::couldSignInputs(const PartiallySignedTransaction &psbtx) {
+    if (!m_wallet_model) {
+        return 0;
+    }
+
     size_t n_signed;
     bool complete;
     TransactionError err = m_wallet_model->wallet().fillPSBT(SIGHASH_ALL, false /* sign */, false /* bip32derivs */, &n_signed, m_transaction_data, complete);
@@ -246,7 +254,10 @@ void PSBTOperationsDialog::showTransactionStatus(const PartiallySignedTransactio
         case PSBTRole::SIGNER: {
             QString need_sig_text = tr("Transaction still needs signature(s).");
             StatusLevel level = StatusLevel::INFO;
-            if (m_wallet_model->wallet().privateKeysDisabled()) {
+            if (!m_wallet_model) {
+                need_sig_text += " " + tr("(But no wallet is loaded.)");
+                level = StatusLevel::WARN;
+            } else if (m_wallet_model->wallet().privateKeysDisabled()) {
                 need_sig_text += " " + tr("(But this wallet cannot sign transactions.)");
                 level = StatusLevel::WARN;
             } else if (n_could_sign < 1) {
--- a/src/qt/walletframe.cpp
+++ b/src/qt/walletframe.cpp
@@ -4,12 +4,18 @@
 
 #include <qt/walletframe.h>
 
+#include <node/ui_interface.h>
+#include <psbt.h>
+#include <qt/guiutil.h>
 #include <qt/overviewpage.h>
+#include <qt/psbtoperationsdialog.h>
 #include <qt/walletmodel.h>
 #include <qt/walletview.h>
 
 #include <cassert>
 
+#include <QApplication>
+#include <QClipboard>
 #include <QGroupBox>
 #include <QHBoxLayout>
 #include <QLabel>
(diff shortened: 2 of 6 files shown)


hebasto:
Concept ACK.

Disclaimer: I am mentoring @psancheti110 for the https://summerofbitcoin.org.

src/qt/psbtoperationsdialog.cpp
```diff
@@ -237,7 +244,6 @@ size_t PSBTOperationsDialog::couldSignInputs(const PartiallySignedTransaction &p
 void PSBTOperationsDialog::showTransactionStatus(const PartiallySignedTransaction &psbtx) {
     PSBTAnalysis analysis = AnalyzePSBT(psbtx);
     size_t n_could_sign = couldSignInputs(psbtx);
-
```
achow101, line 240: nit: Unnecessary whitespace change.

src/qt/psbtoperationsdialog.cpp
```diff
@@ -46,20 +46,23 @@ PSBTOperationsDialog::~PSBTOperationsDialog()
...
-    FinalizePSBT(psbtx);  // Make sure all existing signatures are fully combined before checking for completeness.
-    TransactionError err = m_wallet_model->wallet().fillPSBT(SIGHASH_ALL, false /* sign */, true /* bip32derivs */, &n_could_sign, m_transaction_data, complete);
-    if (err != TransactionError::OK) {
-        showStatus(tr("Failed to load transaction: %1")
-            .arg(QString::fromStdString(TransactionErrorString(err).translated)), StatusLevel::ERR);
-        return;
+    if (m_wallet_model) {
+        bool complete;
+        size_t n_could_sign;
+        FinalizePSBT(psbtx); // Make sure all existing signatures are fully combined before checking for completeness.
```
achow101, line 52: `FinalizePSBT` is not a wallet function, it should always be done. It also returns a bool indicating whether the transaction is completely signed, so this could be

    bool complete = FinalizePSBT(psbt);

src/qt/psbtoperationsdialog.cpp
```diff
@@ -46,20 +46,23 @@ PSBTOperationsDialog::~PSBTOperationsDialog()
...
+            return;
+        }
+        m_ui->broadcastTransactionButton->setEnabled(complete);
+        m_ui->signTransactionButton->setEnabled(!complete && !m_wallet_model->wallet().privateKeysDisabled() && n_could_sign > 0);
+    } else {
+        m_ui->broadcastTransactionButton->setEnabled(true);
+        m_ui->signTransactionButton->setEnabled(false);
     }
-
-    m_ui->broadcastTransactionButton->setEnabled(complete);
```
achow101, line 60: If `complete` is set by `FinalizePSBT` as I suggested above, then this line can be unchanged and the `broadcastTransactionButton` lines can be removed from the if-else blocks.

src/qt/psbtoperationsdialog.cpp
```diff
@@ -46,20 +46,23 @@ PSBTOperationsDialog::~PSBTOperationsDialog()
...
+        if (err != TransactionError::OK) {
+            showStatus(tr("Failed to load transaction: %1")
+                           .arg(QString::fromStdString(TransactionErrorString(err).translated)),
+                       StatusLevel::ERR);
+            return;
+        }
+        m_ui->broadcastTransactionButton->setEnabled(complete);
+        m_ui->signTransactionButton->setEnabled(!complete && !m_wallet_model->wallet().privateKeysDisabled() && n_could_sign > 0);
+    } else {
+        m_ui->broadcastTransactionButton->setEnabled(true);
```
achow101, line 63: It's not always correct for the broadcast button to be enabled since the user could provide a PSBT that is incomplete. By always doing `FinalizePSBT` as I have suggested above, then we will know whether the PSBT is complete by the time we get here.

achow101 (commented):
Concept ACK

src/qt/walletframe.cpp
```diff
@@ -4,17 +4,23 @@
...
 #include <QVBoxLayout>
+#include <cassert>
+#include <qt/guiutil.h>
+#include <qt/overviewpage.h>
+#include <qt/walletmodel.h>
+#include <qt/walletview.h>
+
+#include <node/ui_interface.h>
+#include <psbt.h>
+#include <qt/psbtoperationsdialog.h>
```
hebasto, line 22: Please keep the grouping of headers unchanged -- (1) our own headers, (2) C++ headers, (3) Qt headers.

src/qt/walletframe.cpp
```diff
@@ -4,17 +4,23 @@
...
+
```
hebasto, line 23: Why this line added?

psancheti110:
PR Updated from a77d04fb1a4e788b51424e8fb41122a0568e2756 -> cb7b222cb2ccd1b7d92f7123557097559b0c16a1. Click [_here_](https://github.com/bitcoin-core/gui/compare/a77d04fb1a4e788b51424e8fb41122a0568e2756..cb7b222cb2ccd1b7d92f7123557097559b0c16a1) to check diff.

Addressed changes suggested by @achow101 in https://github.com/bitcoin-core/gui/pull/399#pullrequestreview-725704910.

src/qt/walletframe.cpp
```diff
@@ -184,10 +190,38 @@ void WalletFrame::gotoVerifyMessageTab(QString addr)
...
+        bool invalid;
+        data = DecodeBase64(raw, &invalid);
+        if (invalid) {
+            Q_EMIT message(tr("Error"), tr("Unable to decode PSBT from clipboard (invalid base64)"), CClientUIInterface::MSG_ERROR);
+            return;
+        }
+    } else {
+        QString filename = GUIUtil::getOpenFileName(this,
+                                                    tr("Load Transaction Data"), QString(),
+                                                    tr("Partially Signed Transaction (*.psbt)"), nullptr);
```
hebasto, line 205: I understand these indentations were suggested by `clang-format-diff.py`, but to keep verifying easy with `diff --color-moved=dimmed-zebra`, I'd suggest to keep the same indentations as they were in `walletview.cpp`.

hebasto (commented):
Approach ACK cb7b222cb2ccd1b7d92f7123557097559b0c16a1.

After moving PSBT dialog code from `walletview.cpp` into `waletframe.cpp`, some `include`s could be removed:
```diff
--- a/src/qt/walletview.cpp
+++ b/src/qt/walletview.cpp
@@ -8,7 +8,6 @@
 #include <qt/askpassphrasedialog.h>
 #include <qt/clientmodel.h>
 #include <qt/guiutil.h>
-#include <qt/psbtoperationsdialog.h>
 #include <qt/optionsmodel.h>
 #include <qt/overviewpage.h>
 #include <qt/platformstyle.h>
@@ -21,13 +20,10 @@
 
 #include <interfaces/node.h>
 #include <node/ui_interface.h>
-#include <psbt.h>
 #include <util/strencodings.h>
 
 #include <QAction>
 #include <QActionGroup>
-#include <QApplication>
-#include <QClipboard>
 #include <QFileDialog>
 #include <QHBoxLayout>
 #include <QProgressDialog>
```

psancheti110:
PR Updated from cb7b222cb2ccd1b7d92f7123557097559b0c16a1 -> e5e00910ae5fc3d07e1a4ed5aff5b23a26166495. Click [_here_](https://github.com/bitcoin-core/gui/compare/cb7b222cb2ccd1b7d92f7123557097559b0c16a1..e5e00910ae5fc3d07e1a4ed5aff5b23a26166495) to check diff.

Addressed changes suggested by @hebasto in https://github.com/bitcoin-core/gui/pull/399#pullrequestreview-726541283

achow101:
ACK e5e00910ae5fc3d07e1a4ed5aff5b23a26166495

hebasto (commented):
Tested e5e00910ae5fc3d07e1a4ed5aff5b23a26166495.

I think the `PSBTOperationsDialog::showTransactionStatus` should be modified to correctly handle unsigned PSBTs:
```diff
--- a/src/qt/psbtoperationsdialog.cpp
+++ b/src/qt/psbtoperationsdialog.cpp
@@ -254,7 +254,10 @@ void PSBTOperationsDialog::showTransactionStatus(const PartiallySignedTransactio
         case PSBTRole::SIGNER: {
             QString need_sig_text = tr("Transaction still needs signature(s).");
             StatusLevel level = StatusLevel::INFO;
-            if (m_wallet_model->wallet().privateKeysDisabled()) {
+            if (!m_wallet_model) {
+                need_sig_text += " " + tr("(But no wallet is loaded.)");
+                level = StatusLevel::WARN;
+            } else if (m_wallet_model->wallet().privateKeysDisabled()) {
                 need_sig_text += " " + tr("(But this wallet cannot sign transactions.)");
                 level = StatusLevel::WARN;
             } else if (n_could_sign < 1) {
```

psancheti110:
PR Updated from e5e00910ae5fc3d07e1a4ed5aff5b23a26166495 -> 0237d95323dec7c65b7223c314afdf63aab6b11b. Click [_here_](https://github.com/psancheti110/gui/compare/PR399.03..PR399.04) to check diff.

Addressed changes suggested by @hebasto in https://github.com/bitcoin-core/gui/pull/399#pullrequestreview-727158375

hebasto (approved):
ACK 0237d95323dec7c65b7223c314afdf63aab6b11b, tested on Linux Mint 20.2 (Qt 5.12.8).

No wallet is loaded when testing: `bitcoin-qt -nowallet`.

**A.** The loading of an unsigned PSBT being prepared with the `walletcreatefundedpsbt` RPC command:

Screenshot from 2021-08-11 14-06-59

**B.** The loading of a signed PSBT which is the PSBT mentioned above being processed with the `walletprocesspsbt` RPC command:

Screenshot from 2021-08-11 14-07-33

achow101:
re-ACK 0237d95323dec7c65b7223c314afdf63aab6b11b

psancheti110:
**_A big thanks to all the contributors who took out time to test and review my PR. 
Appreciate each and every comment made here by fellow mates!_**

psancheti110:
**_A big thanks to all the contributors who took out time to test and review my PR. 
Appreciate each and every comment made here by fellow mates_**  🙌🏻

PR #400, opened 2021-08-08
Honor inbound/outbound arrow prefix when sorting peer address column

### LINKED ISSUE #397: Peer address column sorting does not honor the inbound/outbound arrow prefix
See this peer address column:
screenshot

**Expected behavior**
Peer number 196 is sorted below peer number 424, because it is sorted by the first character (arrow) first.

**Actual behavior**
Peer number 196 is sorted below peer number 110. The first (arrow) character seems not to be relevant for sorting?

**To reproduce**
Run with both Onion and I2P network enabled and watch the peer table.

**System information**
22.0rc2
Debian Buster amd_64

wodry:
Fixes #397

~This re-establishes a functionality that was lost I guess since 0.21.~

Peer address column sort with current master:
Peers 27 and 51 are sorted without honoring the connection direction.
address_sort_without_prefix

Peer address column sort with this PR:
Peers 16 and 22 are sorted honoring the connection direction.
address_sort_with_prefix

I would be grateful for implementation improvement suggestions.

### DIFF
--- a/src/qt/peertablesortproxy.cpp
+++ b/src/qt/peertablesortproxy.cpp
@@ -21,11 +21,15 @@ bool PeerTableSortProxy::lessThan(const QModelIndex& left_index, const QModelInd
     const CNodeStats left_stats = Assert(sourceModel()->data(left_index, PeerTableModel::StatsRole).value<CNodeCombinedStats*>())->nodeStats;
     const CNodeStats right_stats = Assert(sourceModel()->data(right_index, PeerTableModel::StatsRole).value<CNodeCombinedStats*>())->nodeStats;
 
+    // Add inbound/outbound arrow prefix to include it in sorting Address
+    const std::string left_arrowAddrName = (left_stats.fInbound ? "↓ " : "↑ ") + left_stats.addrName;
+    const std::string right_arrowAddrName = (right_stats.fInbound ? "↓ " : "↑ ") + right_stats.addrName;
+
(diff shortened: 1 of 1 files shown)


hebasto:
> This re-establishes a functionality that was lost I guess since 0.21.

This is not true: https://github.com/bitcoin-core/gui/blob/4c097f9669a28420da74b159a4d61e509da80d33/src/qt/peertablemodel.cpp#L26-L27

wodry:
OK thanks, then my guess was wrong and i did really not stumble upon this issue by chance. I thought it might have been lost with the new sortProxy.

ShaMan239 (commented):
Tested on Ubuntu 20.04

The sorting of addresses in Master is purely lexicographical, irrespective of the direction of peer. This PR adds the functionality of taking the direction of peers as a prior consideration while sorting the addresses. The PR still sorts addresses in lexicographical order for the same direction peers.
I was able to compile and test the PR on Ubuntu 20.04 successfully.


Though I like the approach, since we want #317, it is not worth merging this PR because PR #317 would revert these changes.

jarolrod (commented):
The issue here is that the meaning of arrows are ambiguous, I am personally concept 0 on this in favor of #317. These PR's are essentially competing as #317 is just going to remove the arrows.

hebasto:
Closing, as #317 has been merged.

PR #403, opened 2021-08-11
refactor: Make paths to update Encryption and HD wallet statuses simpler

hebasto:
This PR makes signal-slot paths to reach `setHDStatus` and `setEncryptionStatus` functions shorter and easier to reason about them.

Required to simplify #398 (see https://github.com/bitcoin-core/gui/pull/398#discussion_r686094883).

---

**Note for reviewers.** Please verify that "Encrypt Wallet..." menu item, and the following icons

DeepinScreenshot_select-area_20210811202120

and updated properly in each and every possible scenario.

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -692,7 +692,6 @@ void BitcoinGUI::addWallet(WalletModel* walletModel)
     });
     connect(wallet_view, &WalletView::encryptionStatusChanged, this, &BitcoinGUI::updateWalletStatus);
     connect(wallet_view, &WalletView::incomingTransaction, this, &BitcoinGUI::incomingTransaction);
-    connect(wallet_view, &WalletView::hdEnabledStatusChanged, this, &BitcoinGUI::updateWalletStatus);
     connect(this, &BitcoinGUI::setPrivacy, wallet_view, &WalletView::setPrivacy);
     wallet_view->setPrivacy(isPrivacyModeActivated());
     const QString display_name = walletModel->getDisplayName();
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -110,6 +110,7 @@ BitcoinGUI::BitcoinGUI(interfaces::Node& node, const PlatformStyle *_platformSty
             connect(activity, &CreateWalletActivity::finished, activity, &QObject::deleteLater);
             activity->create();
         });
+        connect(walletFrame, &WalletFrame::currentWalletSet, [this] { updateWalletStatus(); });
         setCentralWidget(walletFrame);
     } else
 #endif // ENABLE_WALLET
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -1337,9 +1337,8 @@ void BitcoinGUI::setEncryptionStatus(int status)
 
 void BitcoinGUI::updateWalletStatus()
 {
-    if (!walletFrame) {
-        return;
-    }
+    assert(walletFrame);
+
     WalletView * const walletView = walletFrame->currentWalletView();
     if (!walletView) {
         return;
--- a/src/qt/walletframe.cpp
+++ b/src/qt/walletframe.cpp
@@ -103,7 +103,8 @@ void WalletFrame::setCurrentWallet(WalletModel* wallet_model)
     walletView->updateGeometry();
 
     walletStack->setCurrentWidget(walletView);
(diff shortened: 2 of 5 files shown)


src/qt/walletframe.cpp
```diff
@@ -103,7 +103,8 @@ void WalletFrame::setCurrentWallet(WalletModel* wallet_model)
     walletView->updateGeometry();
 
     walletStack->setCurrentWidget(walletView);
-    walletView->updateEncryptionStatus();
```
ryanofsky, line 106: In commit "qt: Add WalletFrame::currentWalletSet signal" (c0f84d9d4354e128b16524c14b6c2dadb891a1b6)

Note: At was unclear to me at first if this commit was changing behavior. But it does not seem to change anything since the previous `updateEncryptionStatus` just called `updateWalletStatus` indirectly.

ryanofsky (approved):
Code review ACK ba9a309a4a8f7c07510e4a033fb477d14a115923. All these cleanups do look good and seem to simplify things without changing behavior.

hebasto:
Rebased ba9a309a4a8f7c07510e4a033fb477d14a115923 -> b8aa84b1a116599a6dd3b9ddb4e6c178a6688b1b ([pr403.01](https://github.com/hebasto/gui/commits/pr403.01) -> [pr403.02](https://github.com/hebasto/gui/commits/pr403.02)) due to the conflict with #399.

src/qt/bitcoingui.cpp
```diff
@@ -113,6 +113,7 @@ BitcoinGUI::BitcoinGUI(interfaces::Node& node, const PlatformStyle *_platformSty
         connect(walletFrame, &WalletFrame::message, [this](const QString& title, const QString& message, unsigned int style) {
             this->message(title, message, style);
         });
+        connect(walletFrame, &WalletFrame::currentWalletSet, [this] { updateWalletStatus(); });
```
Talkless, line 116: Do we really need to wrap that call with lambda? I see that `updateWalletStatus` is already a slot, or am I missing something?

Talkless (changes requested):

src/qt/bitcoingui.cpp
hebasto: No, we don't.

The small benefit is that my editor could easy find this call place of `updateWalletStatus()` function member. It fails to find it by pointer `&BitcoinGUI::updateWalletStatus`.

Are there any drawbacks of such an approach?

ryanofsky (approved):
Code review ACK b8aa84b1a116599a6dd3b9ddb4e6c178a6688b1b. Only change since last review was rebase

src/qt/bitcoingui.cpp
fanquake: The drawback is writing more verbose / unnecessary code, for no reason other than to appease your editor. If it's been questioned during review, it's just as likely someone will open a PR to remove it later on, citing the same reasoning, that, it's confusing / unnecessary. We really shouldn't be bloating our codebase just because your editor doesn't work.

hebasto: I strictly prefer [`QObject::connect(const QObject *sender, PointerToMemberFunction signal, Functor functor)`](https://doc.qt.io/qt-5/qobject.html#connect-4) overload over [`QObject::connect(const QObject *sender, PointerToMemberFunction signal, const QObject *receiver, PointerToMemberFunction method, Qt::ConnectionType type)`](https://doc.qt.io/qt-5/qobject.html#connect-3) one for the following reasons:

- the former states used signal-slot parameters explicitly, i.e., here it is clear that we use no signal parameters
- the former is less verbose, compare `[this] { updateWalletStatus(); }` vs `this, &BitcoinGUI::updateWalletStatus`

> We really shouldn't be bloating our codebase just because your editor doesn't work.

Absolutely agree. But this is not that case, no?

Btw, this--very helpful syntax--is already actively used in the code syntax. E.g. https://github.com/bitcoin/bitcoin/pull/14123 and other prs by @promag.

fanquake: Thanks for providing the actual reasoning; this should have been the first comment. I don't really have an opinion on the code, but was always going to leave the response above, regardless of the change, because the justification "it's better for my editor", is not a good one.

hebasto: I apologize for that.

fanquake: No worries. No need for an apology. Feel free to mark as resolved and merge etc.

PR #404, opened 2021-08-12
Fix various edge case bugs in QValidatedLineEdit

luke-jr:
1. Use a CSS selector to avoid changing the background colour of the tooltip.
2. Re-check validity of input when we first set the validator (probably a no-op in practice).
3. Check validity of input when it is set programmatically via `setText` (eg, via the address book). Probably no-op in practice UNTIL merging https://github.com/bitcoin/bitcoin/pull/15987 or any other PR that adds a warning for valid addresses.

Moved from https://github.com/bitcoin/bitcoin/pull/18133 (just concept ACKs)

### DIFF
--- a/src/qt/qvalidatedlineedit.cpp
+++ b/src/qt/qvalidatedlineedit.cpp
@@ -106,6 +106,7 @@ void QValidatedLineEdit::checkValidity()
 void QValidatedLineEdit::setCheckValidator(const QValidator *v)
 {
     checkValidator = v;
+    checkValidity();
 }
 
 bool QValidatedLineEdit::isValid()
--- a/src/qt/qvalidatedlineedit.cpp
+++ b/src/qt/qvalidatedlineedit.cpp
@@ -15,6 +15,12 @@ QValidatedLineEdit::QValidatedLineEdit(QWidget *parent) :
     connect(this, SIGNAL(textChanged(QString)), this, SLOT(markValid()));
 }
 
+void QValidatedLineEdit::setText(const QString& text)
+{
+    QLineEdit::setText(text);
+    checkValidity();
+}
+
 void QValidatedLineEdit::setValid(bool _valid)
 {
     if(_valid == this->valid)
--- a/src/qt/qvalidatedlineedit.cpp
+++ b/src/qt/qvalidatedlineedit.cpp
@@ -28,7 +28,7 @@ void QValidatedLineEdit::setValid(bool _valid)
     }
     else
     {
-        setStyleSheet(STYLE_INVALID);
+        setStyleSheet("QValidatedLineEdit { " STYLE_INVALID "}");
     }
     this->valid = _valid;
 }
--- a/src/qt/qvalidatedlineedit.h
+++ b/src/qt/qvalidatedlineedit.h
@@ -29,6 +29,7 @@ private:
     const QValidator *checkValidator;
 
 public Q_SLOTS:
+    void setText(const QString&);
     void setValid(bool valid);
     void setEnabled(bool enabled);
 


ShaMan239 (commented):
I looked through the PR and the changes that are made. I have got some doubts that I want to ask the op regarding some changes made here:
- The default color property is now enclosed in a CSS selector to avoid changing the tooltip's background color. **But was that a problem in the first place?**
- If enclosing the color property has its benefits, shouldn’t a similar thing be done in line 33 ` setStyleSheet(""); ` to maintain consistency?
- The validity of text is checked after setting **checkValidator** in the **setCheckValidator** function. Why is there a need to check the validity here?

It would be helpful if op could clarify these doubts. Thank You.

luke-jr:
>The default color property is now enclosed in a CSS selector to avoid changing the tooltip's background color. But was that a problem in the first place?

Yes. Without it, the tooltip bg colour is changed too. That is incorrect behaviour.

>If enclosing the color property has its benefits, shouldn’t a similar thing be done in line 33 setStyleSheet(""); to maintain consistency?

No, that doesn't make sense... Do you know CSS? :/

>The validity of text is checked after setting checkValidator in the setCheckValidator function. Why is there a need to check the validity here?

It's the logically correct thing to do to ensure a consistent state.

ShaMan239:
> Yes. Without it, the tooltip bg color is changed too. That is incorrect behavior.

That makes sense. And I would agree with it.

> No, that doesn't make sense... Do you know CSS? :/

As a matter of fact, I do know CSS. I wrongly interpreted the changes introduced here, and that was a genuine mistake on my part.

Let me go through some further testings to check if I am not missing something. In the meantime, @luke-jr, it would be great if you could post some screenshots that clearly express the differences introduced in this PR.

src/qt/qvalidatedlineedit.cpp
```diff
@@ -106,6 +106,7 @@ void QValidatedLineEdit::checkValidity()
 void QValidatedLineEdit::setCheckValidator(const QValidator *v)
 {
     checkValidator = v;
+    checkValidity();
```
promag, line 115: b1a544be109d336c0b53722e3f8b51687972c94e

What bug is this change fixing? Or is this just making the code more correct?

src/qt/qvalidatedlineedit.cpp
```diff
@@ -15,6 +15,12 @@ QValidatedLineEdit::QValidatedLineEdit(QWidget *parent) :
     connect(this, SIGNAL(textChanged(QString)), this, SLOT(markValid()));
 }
 
+void QValidatedLineEdit::setText(const QString& text)
+{
+    QLineEdit::setText(text);
```
promag, line 20: aeb18b665c616c3326671b4c7e9d6421306564f0

Maybe add another connection? See
https://github.com/bitcoin-core/gui/blob/6490a3ef6c4c6525d169818d18a99050f780c5b7/src/qt/qvalidatedlineedit.cpp#L15

promag (commented):
Concept ACK fixing and improving this custom widget. Does anyone know why https://doc.qt.io/qt-5/qlineedit.html#setValidator is not used?

src/qt/qvalidatedlineedit.cpp
jarolrod: bump

this is also not evident to me

jarolrod (commented):
@luke-jr, just following up on this comment: https://github.com/bitcoin/bitcoin/pull/15987#issuecomment-933027946

Why would https://github.com/bitcoin/bitcoin/pull/15987 depend on this?

src/qt/qvalidatedlineedit.cpp
luke-jr: > Re-check validity of input when we first set the validator (probably a no-op in practice).

src/qt/qvalidatedlineedit.cpp
luke-jr: That's less correct

luke-jr:
Because without this, validators don't work correctly.  bitcoin/bitcoin#15987 uses a validator.

Additionally, inline marking of bech32 error detection (https://github.com/bitcoin-core/gui/issues/527) also depends on this.

src/qt/qvalidatedlineedit.cpp
Sjors: Maybe add "Probably a no-op in practice." to the commit message.

Sjors (approved):
This PR builds on rather ancient commits (from 2017; in case anyone wonders why building fails over missing OpenSSL), but I tested a rebased version.

tACK aeb18b665c616c3326671b4c7e9d6421306564f0

Before 2385b508d5f2db118513c3e0b343d2309cdfdcd8:
before

After:
after

hebasto (approved):
ACK aeb18b665c616c3326671b4c7e9d6421306564f0, tested on Linux Mint 20.3 (Qt 5.12.8).

PR #406, opened 2021-08-21
Handle new added plurals in bitcoin_en.ts

hebasto:
This step was missed. See [`translation_process.md`](https://github.com/bitcoin/bitcoin/blob/master/doc/translation_process.md#handling-plurals-in-source-files).

Transifex is smart enough to handle `%n` in strings as a plural form. Therefore, only two English strings are broken in the GUI on master (and in 22.0).

### DIFF
--- a/src/qt/locale/bitcoin_en.ts
+++ b/src/qt/locale/bitcoin_en.ts
@@ -749,8 +749,8 @@ Signing is only possible with addresses of the type &apos;legacy&apos;.</source>
         <source>%n active connection(s) to Bitcoin network.</source>
         <extracomment>A substring of the tooltip.</extracomment>
         <translation type="unfinished">
-            <numerusform></numerusform>
-            <numerusform></numerusform>
+            <numerusform>%n active connection to Bitcoin network.</numerusform>
+            <numerusform>%n active connections to Bitcoin network.</numerusform>
         </translation>
     </message>
     <message>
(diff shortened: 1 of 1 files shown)


ShaMan239 (commented):
Concept ACK

This PR adds the plural translation for the two strings that use plurals but hitherto didn’t had a plural translation. Adding screenshots to compare the difference between the master and the PR.

Screenshots (taken from QT Linguistic):
| Master   |      PR   |
|----------|:-------------:|
| Screenshot from 2021-08-23 11-40-05| Screenshot from 2021-08-23 11-41-28 |
| Screenshot from 2021-08-23 11-40-32| Screenshot from 2021-08-23 11-42-00|

laanwj:
I don't understand, `bitcoin_en.ts` is auto-generated by `make translate`, right?

hebasto:
> I don't understand, `bitcoin_en.ts` is auto-generated by `make translate`, right?

Yes. But `make translate` do not create plural forms (and, I guess, never did it).

laanwj:
>  Yes. But make translate does not create plural forms (and, I guess, never did it).

Ok, so `bitcoin_en.ts` is *updated*, not recreated. This is slightly unexpected to me but makes sense.

ACK 2b3d8f3dde383a53f29b7e7ee53ea364d4ef8938

hebasto:

PR #407, opened 2021-08-23
bugfix: initialize IsHDEnabled correctly.

Saibato:
the result of `IsHDEnabled()` was initialized with true.

But in case of no keys or a blank hd wallet the iterator would be skipped
and not set to false.
What had resulted in a wrong return and subsequent false hd and watch only icon display
in gui when reloading a wallet after closing.

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -1364,9 +1364,13 @@ CAmount CWallet::GetDebit(const CTransaction& tx, const isminefilter& filter) co
 bool CWallet::IsHDEnabled() const
 {
     // All Active ScriptPubKeyMans must be HD for this to be true
-    bool result = true;
+    bool result = false;
     for (const auto& spk_man : GetActiveScriptPubKeyMans()) {
-        result &= spk_man->IsHDEnabled();
+        if (spk_man->IsHDEnabled()) {
+            result = true;
+        } else {
+            return false;
+        }
     }
     return result;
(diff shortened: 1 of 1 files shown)


src/wallet/wallet.cpp
```diff
@@ -1364,9 +1364,13 @@ CAmount CWallet::GetDebit(const CTransaction& tx, const isminefilter& filter) co
...
+    bool result = false;
     for (const auto& spk_man : GetActiveScriptPubKeyMans()) {
-        result &= spk_man->IsHDEnabled();
+        if (spk_man->IsHDEnabled()) {
+            result = true;
+        } else {
+            return false;
+        }
     }
     return result;
```
MarcoFalke, line 1375: easier would be:

```cpp
for ...
  if ! enabled: return false;
return true;

MarcoFalke:
This is the wrong repo. To modify wallet code, you will need to submit to the main repo.

jarolrod:
#403 is a fix for this

Saibato:
> This is the wrong repo. To modify wallet code, you will need to submit to the main repo.

fwics it effects only the gui, but you obviously correct, i will file it too under main,

src/wallet/wallet.cpp
Saibato: not quite, since we need to return false of the init value in case the for loop is never entered. afaics.

PR #410, opened 2021-08-23
Uppercase 'bitcoin:' in QR code URI

benthecarman:
The folks at BTCPayServer did research on this earlier this year and it seems we have reached enough wallets supporting completely uppercase URIs for the QR code

https://github.com/btcpayserver/btcpayserver/issues/2110

### DIFF
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -195,7 +195,7 @@ QString formatBitcoinURI(const SendCoinsRecipient &info)
 {
     bool bech_32 = info.address.startsWith(QString::fromStdString(Params().Bech32HRP() + "1"));
 
-    QString ret = QString("bitcoin:%1").arg(bech_32 ? info.address.toUpper() : info.address);
+    QString ret = QString("bitcoin:%1").arg(info.address);
     int paramCount = 0;
 
     if (info.amount)
@@ -216,6 +216,10 @@ QString formatBitcoinURI(const SendCoinsRecipient &info)
         QString msg(QUrl::toPercentEncoding(info.message));
         ret += QString("%1message=%2").arg(paramCount == 0 ? "?" : "&").arg(msg);
         paramCount++;
+    }
+
+    if (bech_32 && paramCount == 0) {
+        ret = ret.toUpper();
     }
 
     return ret;


ben-kaufman (approved):
ACK ae7bc4dc4014ffca0d68f53cdc9b27ed0aae52f0

kristapsk:
It only makes sense to do that if final `paramCount` is 0, as BIP21 parameters are case sensitive. And, apart for QR code encoding, BIP173 says that lower case is preferred. See https://github.com/JoinMarket-Org/joinmarket-clientserver/pull/972, how I propose to do it for JoinMarket GUI.

benthecarman:
Used @kristapsk suggestion

kristapsk (approved):
ACK b6704cf7e1b005d4122c032485c9b5bba1359da8

ShaMan239 (commented):
Tested on Ubuntu 20.04.

This PR allows the capitalization of the schema of a “bitcoin:” based URI. This is done only when all the parameters of URI are not prespecified.
The motivation of this PR is that QR codes for all uppercase URIs are much simpler and hence easier to scan, as discussed [here](https://bitcoinops.org/en/bech32-sending-support/#creating-more-efficient-qr-codes-with-bech32-addresses).

Tested this PR by creating a payment receiving request and copying the URI of that request.
The URI with no specified info works as expected. The schema is non-capitalized on master while it is capitalized on PR.

**Copying a URI with no prespecified info:** 
|Master|PR|
|-----|----|
|**bitcoin:TB1Q7A0G8SFXRFKCGY77PPDTS964ZHAHP7KMRFF9VH**|**BITCOIN:TB1Q7A0G8SFXRFKCGY77PPDTS964ZHAHP7KMRFF9VH**|

While tested okay with URI of no-predefined info. I observed some weird behavior while copying a URI with a specified label.

**Copying a URI with a “test_label”:**

|Master|PR|
|-----|----|
|**bitcoin:TB1QFT0PFF87246F607S895CUSYEHJNGWUZLN69A3F?label=test_label**|**bitcoin:tb1qft0pff87246f607s895cusyehjngwuzln69a3f?label=test_label**|

When I specified the request label, the PR lowercased the address part of the URI. I don’t know if this is a deliberate behavior by the PR or a bug. If this is indeed an intentional behavior, then the PR description should address and explain this. Otherwise, OP should look into this issue.

benthecarman:
> When I specified the request label, the PR lowercased the address part of the URI. I don’t know if this is a deliberate behavior by the PR or a bug. If this is indeed an intentional behavior, then the PR description should address and explain this. Otherwise, OP should look into this issue.

This is intentional and shouldn't have an effect on users. Since the address is case insensitive it doesn't change anything.

ShaMan239:
> This is intentional and shouldn't have an effect on users. Since the address is case insensitive it doesn't change anything.

Makes sense. But I would personally prefer to avoid adding any unnecessary changes, that are changing the behavior with no actual benefits.

benthecarman:
There are benefits, the code is much cleaner this way and the URI makes more sense to have a lowercase bech32 addr when we aren't going full caps since this is the default defined by the BIP

ShaMan239:
Now you do mention it, I think these benefits are worth the change.

tACK b6704cf7e1b005d4122c032485c9b5bba1359da8

hebasto:
Tested b6704cf7e1b005d4122c032485c9b5bba1359da8.

- **`tb1q4dcd7mdjdlfgdhjhz8rg5gw2834udp2u0n9jjs`:**

| master | PR |
|---|---|
| Screenshot from 2021-09-03 14-00-06 | Screenshot from 2021-09-03 13-53-28 |

I see no benefits for encoding: both QR codes have the same 29x29 size.

- **`tb1qnle64twzzjae9mrryy02z30hqaca5747rsanxt` + `amount=0.42000000`:**

| master | PR |
|---|---|
| Screenshot from 2021-09-03 14-00-28 | Screenshot from 2021-09-03 13-54-30 |

Both QR codes have the same 33x33 size. Is this behavior change justified (at least, it is not mentioned in the PR description)?

benthecarman:
> I see no benefits for encoding: both QR codes have the same 29x29 size.

Huh, does qrencode not support mixed modes? I don't see anything about it on the repo https://github.com/fukuchi/libqrencode

benthecarman:
> There are benefits, the code is much cleaner this way...
> I guess, this statement is wrong.

This was inrespect to @ShaMan239's comment about keeping the address uppercase even when everything else is lowercase, but still doing all uppercase when possible

hebasto:
> > There are benefits, the code is much cleaner this way...
> > I guess, this statement is wrong.
> 
> This was inrespect to @ShaMan239's comment about keeping the address uppercase even when everything else is lowercase, but still doing all uppercase when possible

I beg pardon for misunderstanding on my side. I was thinking about QR code, not the code.
Sorry again.

hebasto:
> > I see no benefits for encoding: both QR codes have the same 29x29 size.
> 
> Huh, does qrencode not support mixed modes? I don't see anything about it on the repo https://github.com/fukuchi/libqrencode

It does. If an URI were 47 symbols length, this PR makes the GUI generate a QR code version 2 (25x25).

But the shortest URI with bech32 address is 50 characters long.

benthecarman:
> It does. If an URI were 47 or less symbols length, this PR makes the GUI generate a QR code version 2 (25x25).
> But the shortest URI with bech32 address is 50 characters long.

hmm so is it even worth it to do caps at all?

hebasto:
> > It does. If an URI were 47 or less symbols length, this PR makes the GUI generate a QR code version 2 (25x25).
> > But the shortest URI with bech32 address is 50 characters long.
> 
> hmm so is it even worth it to do caps at all?

For the one address type—P2WPKH—no, not worth it. Other address types need some research work.

RandyMcMillan:
tested b6704cf7e1b005d4122c032485c9b5bba1359da8

on MacOS 10.15.7

Screen Shot 2021-11-22 at 3 39 49 PM

achow101:
I don't see the benefit. With regtest taproot addresses (so the longest addresses we may make), the QR codes have identical 29x29 dimensions:
Screenshot_20211206_235940
Screenshot_20211207_000612

hebasto:
Closing this for now. Feel free to re-open if it seems required.

Issue #78, opened 2020-08-28
Switch mainnet / testnet / regtest from GUI


Sjors:
The only ways to start QT with a different network is to launch it from the command line or edit `bitcoin.conf`. It would be nice if it can be done from the settings screen (a restart is fine).

michaelfolkson:
Concept ACK. Wait for Signet to be merged first? So there would be mainnet, testnet, regtest and signet options in the GUI?

prayank23:
> The only ways to start QT with a different network is to launch it from the command line or edit `bitcoin.conf`. It would be nice if it can be done from the settings screen (a restart is fine).

In Windows if you search for "bitcoin" you get two options for mainnet and testnet which are links to shortcuts that have below targets set in their properties and same thing for electrum wallet:

"C:\Program Files\Bitcoin\bitcoin-qt.exe"
"C:\Program Files\Bitcoin\bitcoin-qt.exe" -testnet

![mainnet-testnet-shortcuts](https://i.imgur.com/jeV5Edb.png)

Shortcuts can be created in Ubuntu as well.

jonasschnelli:
Concept ACK.
Since the different nets can run in parallel, would it be possible to open – say – regtest form mainnet (new application launch instead of restart)?

Sjors:
@prayank23 that's pretty cool. If that can be done on macOS too that's nice. That also lets you run in parallel, which as @jonasschnelli points out is useful.

RandyMcMillan:
In reference to color selection:

https://github.com/bitcoin-core/gui/issues/89#issuecomment-695752844

If multiple networks can be easily ran in parallel - color blindness should be considered and an effort should be made so the UIs are distinguishable by people that may not be able to tell a difference between orange and green (or whatever colors are ultimately used).

jonatack:
> If multiple networks can be easily ran in parallel - color blindness should be considered and an effort should be made so the UIs are distinguishable by people that may not be able to tell a difference between orange and green (or whatever colors are ultimately used).

Regtest/testnet/mainnet GUIs can be run at the same time, just like bitcoind.

On Debian at least, the splash/loading and main screens display "[testnet]" or "[test]". It might be good to have it also displayed in the node windows. 

I'm not sure if that is sufficient for color-blind people; it's good to raise the question.

Bosch-0:
[...]

PR #336, opened 2021-05-16
Do not exit and re-enter main event loop during shutdown

hebasto:
On master (1ef34ee25ed34b2b092f15bf3dca5c0508092829) during shutdown `QApplication` exits the main event loop, then re-enter again.

This PR streamlines shutdown process by removing the need to interrupt the main event loop, that is required for #59.

Also, blocking [`QDialog::exec()`](https://doc.qt.io/qt-5/qdialog.html#exec) calls are replaced with safer [`QDialog::show()`](https://doc.qt.io/qt-5/qwidget.html#show), except for `SendConfirmationDialog` as that change is not trivial (marked as TODO).

The [`QDialog::open()`](https://doc.qt.io/qt-5/qdialog.html#open) was not used because the actual modality mode (application modal or window modal) of a dialog depends on whether it has a parent.

This PR does not change behavior, and all touched dialogs are still application modal.
As a follow up, a design research could suggest to make some dialogs window modal.

NOTE for reviewers: quitting app while a dialog is open (e.g., via systray icon menu) must work fine.

PR #414, opened 2021-08-31
Toggle chain from menu (using Settings)

Sjors:
Implements #78

It's currently not easy for GUI users to switch to a test chain like signet. E.g on macOS this requires using the command line, since Spotlight doesn't let you add an argument like `-signet`.

This PR makes it easier by adding a menu option to restart with a different chain. ~It does this by adding a `.chain_gui` file. Unfortunately we can't use `settings.json`, because this file is network specific. This file is ignored by `bitcoind`.~ It does this by adding `chain` to the mainnet QSettings.

Schermafbeelding 2021-09-01 om 13 01 37
(it's now under Settings, not File)

### DIFF
--- a/doc/release-notes.md
+++ b/doc/release-notes.md
@@ -190,6 +190,10 @@ GUI changes
 - UTXOs which are locked via the GUI are now stored persistently in the
   wallet database, so are not lost on node shutdown or crash. (#23065)
 
+- Toggle chain from mainnet to testnet, signet or regtest. The GUI now always sets `-chain`,
+  so any occurrences of `-testnet`, `-signet` or `-regtest` in `bitcoin.conf` should
+  be replace with `chain=test`, `chain=signet` or `chain=regtest` respectively. (bitcoin-core/gui#414)
+
 - The Bech32 checkbox has been replaced with a dropdown for all address types, including the new Bech32m (BIP-350) standard for Taproot enabled wallets.
 
 Low-level changes
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -564,6 +564,10 @@ int GuiMain(int argc, char* argv[])
     // - QSettings() will use the new application name after this, resulting in network-specific settings
     // - Needs to be done before createOptionsModel
 
+    QSettings settings;
+    std::string chain = settings.value("chain", "main").toString().toStdString();
+    gArgs.SoftSetArg("-chain", chain);
+
     // Check for chain settings (Params() calls are only valid after this clause)
     try {
         SelectParams(gArgs.GetChainName());
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -454,6 +454,46 @@ void BitcoinGUI::createMenuBar()
     appMenuBar = menuBar();
 #endif
 
+    // Add chain selection submenu
+    QAction* chain_selection_action{nullptr};
+    QMenu* chain_selection_menu{nullptr};
+    chain_selection_action = new QAction(tr("&Switch chain"), this);
+    chain_selection_action->setStatusTip(tr("Restart application using a different (test) network"));
+    chain_selection_menu = new QMenu(this);
+
+    connect(chain_selection_menu, &QMenu::aboutToShow, [this, chain_selection_menu] {
+        chain_selection_menu->clear();
+        const std::vector<std::pair<QString, QString>> chains = {{"main", "&Bitcoin"}, {"test", "&Testnet"}, {"regtest", "&Regtest"}, {"signet", "&Signet"}};
+        for (auto chain : chains) {
+            const bool is_current = gArgs.GetChainName() == chain.first.toStdString();
+            QAction* action = chain_selection_menu->addAction(chain.second);
+            action->setCheckable(true);
+            action->setChecked(is_current);
+            action->setEnabled(!is_current);
+
+            connect(action, &QAction::triggered, [this, chain] {
+                //: Switch to the mainnet, testnet, signet or regtest chain.
+                QMessageBox::StandardButton btn_ret_val = QMessageBox::question(this, tr("Switch chain"),
+                    //: Switching between mainnet, testnet, signet or regtest chain requires a restart.
+                    tr("Client restart required to switch chain.\n\nClient will be shut down. Do you want to proceed?"),
+                    QMessageBox::Yes | QMessageBox::Cancel, QMessageBox::Cancel);
+
+                if (btn_ret_val == QMessageBox::Cancel) return;
+
+                // QSettings are stored seperately for each network. Switch application name
+                // to mainnet before storing the selected chain.
+                QScopedPointer<const NetworkStyle> networkStyle(NetworkStyle::instantiate("main"));
+                assert(!networkStyle.isNull());
+                QApplication::setApplicationName(networkStyle->getAppName());
+
+                QSettings settings;
+                settings.setValue("chain", chain.first);
+                Q_EMIT quitRequested();
+            });
+        }
+    });
+    chain_selection_action->setMenu(chain_selection_menu);
+
     // Configure the menus
     QMenu *file = appMenuBar->addMenu(tr("&File"));
     if(walletFrame)
@@ -483,6 +523,7 @@ void BitcoinGUI::createMenuBar()
         settings->addSeparator();
     }
     settings->addAction(optionsAction);
+    settings->addAction(chain_selection_action);
 
     QMenu* window_menu = appMenuBar->addMenu(tr("&Window"));
 


ryanofsky:
I think you could just use QSettings for this, the same way the GUI datadir preference and all other other GUI-only settings are stored in QSettings. Even #15936 which consolidates bitcoin-qt and bitcoind settings in settings.json still stores GUI-only settings in QSettings.

Sjors:
If we're keeping some QSettings around anyway, I might as well use that. With `bitcoind` it's trivial to just specify the network as a parameter.

Sjors:
I switched to using QSettings. Similar to `settings.json` these are also network dependent. The GUI does this by changing its name, which it does during the bootstrap process right after deciding which chain its on. This PR uses the same trick in reverse, namely changing the application name back to mainnet right before updating the setting and quitting.

ShaMan239 (commented):
Concept ACK

This PR adds the functionality to select a network chain directly from the GUI, making it easier to toggle between the chains. This is achieved by adding a context menu option to set the -chain suboption to the selected chain type and closing the GUI. So that when next time GUI will start, the value of -chain suboption will by default be set to the last selected network by the user. This value, as I observed, can always be overridden while opening the bitcoin GUI through the command line.

Tested the PR on Ubuntu 20.04. The PR works as described by OP. I am adding the screenshot of the added context menu options.

Screenshot from 2021-09-01 19-42-53

This PR also has a byproduct benefit that is not discussed till now. For a person like me who mainly works on a signet network, I do not have to explicitly create an alias or select the signet chain each time I want to open GUI. So that’s great!

I have a few suggestions I want to put forward that might further improve the PR:
- Before line 459:
```
QMessageBox::StandardButton btnRetVal = QMessageBox::question(this, tr("Confirm chain"),
``` 
Add translator’s comment for text in the StandardButton.
- Add mnemonic shortcuts for the added context menu options.
- It would be great if the GUI could automatically start after toggling the chain instead of simply closing the app.

Sjors:
@ShaMan239 thanks for the review! I added a translator hint and keyboard shortcuts (I can't test those on macOS, new screenshot is welcome). Unfortunately afaik Bitcoin Core can't restart itself. This is a problem with some settings changes too.

shaavan (commented):
The Translator’s comments look good.
Regarding the mnemonics shortcuts, there are some collisions between them.

Screenshot from 2021-09-01 23-11-52

`Load PSBT from &clipboard` <-> `Switch &chain`

~~`&Backup wallet` <-> `&Bitcoin`~~
> Unfortunately, afaik Bitcoin Core can't restart itself.

Well, that’s a bummer. I was just wondering if this feature could be added and would that even be worth adding.

Sjors:
Maybe I misunderstand how these shortcuts work... Wouldn't "c" take you to the chain submenu and then "b" to Bitcoin inside that menu? Or all the shortcuts global?

src/qt/bitcoingui.cpp
```diff
@@ -441,6 +441,42 @@ void BitcoinGUI::createActions()
...
+    m_chain_selection_menu = new QMenu(this);
+
+    connect(m_chain_selection_menu, &QMenu::aboutToShow, [this] {
+        m_chain_selection_menu->clear();
+        const std::vector<std::pair<QString, QString>> chains = {{"main", "&Bitcoin"}, {"test", "&Testnet"}, {"regtest", "&Regtest"}, {"signet", "&Signet"}};
+        for (auto chain : chains) {
+            QAction* action = m_chain_selection_menu->addAction(chain.second);
+
+            if (gArgs.GetChainName() == chain.first.toStdString()) {
+              action->setEnabled(false);
```
promag, line 455: nit, fix indentation heer and below.

promag: nit, could do `action->setEnabled(gArgs.GetChainName() != chain.first.toStdString())`.

src/qt/bitcoingui.cpp
```diff
@@ -441,6 +441,42 @@ void BitcoinGUI::createActions()
...
+            }
+
+            connect(action, &QAction::triggered, [this, chain] {
+              //: Switch to the mainnet, testnet, signet or regtest chain.
+              QMessageBox::StandardButton btnRetVal = QMessageBox::question(this, tr("Switch chain"),
+                  tr("Client restart required to switch chain.") + "<br><br>" + tr("Client will be shut down. Do you want to proceed?"),
+                  QMessageBox::Yes | QMessageBox::Cancel, QMessageBox::Cancel);
+
+              if(btnRetVal == QMessageBox::Cancel) return;
```
promag, line 464: nit, space after `if`.

promag (commented):
Approach ACK, makes sense to have this separate from other options.

promag (commented):
Tested ACK 79af0d657bc080bfd78b70a968c50f654c1aa329.

Also added the check on the current chain to better indicate why the current is disabled.
image

with the following
```diff
+            const bool is_current = gArgs.GetChainName() == chain.first.toStdString();
             QAction* action = m_chain_selection_menu->addAction(chain.second);
-
-            if (gArgs.GetChainName() == chain.first.toStdString()) {
-              action->setEnabled(false);
-            }
+            action->setCheckable(true);
+            action->setChecked(is_current);
+            action->setEnabled(!is_current);
```

promag:
Now I can't pick a different chain from the command line, for instance `src/qt/bitcoin-qt -regtest` gives `Error: Invalid combination of -regtest, -signet, -testnet and -chain. Can use at most one.`.

Sjors:
@promag thanks, addressed comments and added the checkmark. Though on my machine I can't see the checkmark (QT 6.1.2 via homebrew).

Once you use the GUI to select a chain, there's no way back :-) The chain is a special case with so many weird rules, I'd rather not mess with:

https://github.com/bitcoin-core/gui/blob/f4e12fd50c23875f4b5f272c94449eb81de43d5d/src/util/settings.cpp#L129-L143

promag (commented):
Code review ACK 246d0658b41b232151567356694c8580f32a800e.

src/qt/bitcoin.cpp
```diff
@@ -559,6 +559,10 @@ int GuiMain(int argc, char* argv[])
     // - QSettings() will use the new application name after this, resulting in network-specific settings
     // - Needs to be done before createOptionsModel
 
+    QSettings settings;
+    const std::string chain = settings.value("chain", "").toString().toStdString();
+    if (chain != "") gArgs.SoftSetArg("-chain", chain);
```
hebasto, line 564:
```suggestion
    if (!chain.empty()) gArgs.SoftSetArg("-chain", chain);
```

src/qt/bitcoingui.cpp
```diff
@@ -441,6 +441,42 @@ void BitcoinGUI::createActions()
...
+        for (auto chain : chains) {
+            const bool is_current = gArgs.GetChainName() == chain.first.toStdString();
+            QAction* action = m_chain_selection_menu->addAction(chain.second);
+            action->setCheckable(true);
+            action->setChecked(is_current);
+            action->setEnabled(!is_current);
+
+            connect(action, &QAction::triggered, [this, chain] {
+                //: Switch to the mainnet, testnet, signet or regtest chain.
+                QMessageBox::StandardButton btnRetVal = QMessageBox::question(this, tr("Switch chain"),
```
hebasto, line 460: Could follow [our naming convention](https://github.com/bitcoin/bitcoin/blob/master/doc/developer-notes.md#coding-style-c) with `btnRetVal`?

src/qt/bitcoingui.cpp
```diff
@@ -441,6 +441,42 @@ void BitcoinGUI::createActions()
...
+                    tr("Client restart required to switch chain.") + "<br><br>" + tr("Client will be shut down. Do you want to proceed?"),
```
hebasto, line 461: Translators will much appreciate if both strings belong to the same context:
```suggestion
                    tr("Client restart required to switch chain.\n\nClient will be shut down. Do you want to proceed?"),
```

hebasto (commented):
Concept ACK.

1. Still having [this error](https://github.com/bitcoin-core/gui/pull/414#issuecomment-912456779) on Linux:
> Now I can't pick a different chain from the command line, for instance `src/qt/bitcoin-qt -regtest` gives `Error: Invalid combination of -regtest, -signet, -testnet and -chain. Can use at most one.`.

2. It seems possible to move all new code from `BitcoinGUI::createActions()` to `BitcoinGUI::createMenuBar()` and use the local variable `chain_selection_action` instead of the data member `m_chain_selection_action`.

3. Why not add translator comments to all of the added `tr()` arguments?

shaavan (commented):
Changes since my last review:
- The checkmark is added along with the options to emphasize which chain is currently selected. I think it makes sense to add a checkmark, so I agree with this change
Adding screenshots of the visual change before and after updating PR.

|Before Update|After Update|
|------|-----|
|Screenshot from 2021-09-01 23-11-52|pr1|

> Maybe I misunderstand how these shortcuts work... Wouldn't "c" take you to the chain submenu and then "b" to Bitcoin inside that menu? Or all the shortcuts global?

I just checked if the shortcuts are global. They are not, and so I have updated my comment accordingly.
But there is still the collision of mnemonics between `Load PSBT from &clipboard` and `Switch &chain` as they are part of the same menu. So I would recommend updating the mnemonics shortcut to `&S`

```diff
diff --git a/src/qt/bitcoingui.cpp b/src/qt/bitcoingui.cpp
index 39a58f4d1..5aedbbf43 100644
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -441,7 +441,7 @@ void BitcoinGUI::createActions()
     }
 #endif // ENABLE_WALLET
 
-    m_chain_selection_action = new QAction(tr("Switch &chain"), this);
+    m_chain_selection_action = new QAction(tr("&Switch chain"), this);
     m_chain_selection_action->setStatusTip(tr("Restart application using a different (test) network"));
     m_chain_selection_menu = new QMenu(this);
```

src/qt/bitcoingui.cpp
Sjors: Done

Sjors:
> Maybe always add the `chain` value to settings at the first run the new code, and notify users about new behavior?

Done, and added a release note.

I moved the menu stuff to `BitcoinGUI::createMenuBar()`

> Why not add translator comments to all of the added tr() arguments?

Done (I think).

I switched the shortcut to `Switch &chain`

shaavan (approved):
tACK bc72b19 

Updates since my last review:

- Release notes are added to inform users about the change in behavior in the way of using chain variables.
- **GUI** will open on the **mainnet** when the user specifies no specific chain, either through **Command-Line** or **GUI**.
- The scope of the `chain_selction_menu` variable has been changed from global to local.
`btnRetVal` renamed to `btn_ret_val` according to the naming convention.
- The mnemonic shortcut of the switch chain option has been changed from `&c` to `&S`, removing shortcut collision.

I tested the updated PR on Ubuntu 20.04. The PR works perfectly. Also, I followed the release notes word by word while testing. They are correct too. The collision between mnemonic shortcuts has been resolved, and the new mnemonic shortcut is working perfectly. 

I am adding the screenshot of the latest commit.

PR

Thanks for this amazing work, @Sjors!

Sjors:
@kallewoof @ajtowns you might enjoy this PR

luke-jr:
Concept NACK. I think there should be a very firm line between Bitcoin and test networks. Something like this can too easily lead to loss because someone forgot they switched back to mainnet.

kallewoof:
Slightly agree with @luke-jr but I can see why it would be nice to have. Just not sure how often you switch between networks like that.

Maybe this should be a default-disabled feature that developers can turn on when compiling.

achow101:
Concept NACK, agree with @luke-jr

jarolrod:
After further consideration, I agree with @luke-jr. While this is a nice-to-have, without including several new UI/UX changes to prevent a user from footgunning; there's a possibility that this can do harm.

To be clear, I'm concept ACK on the general Idea, I think there's room for a feature like this. But, this should be done with careful attention and perhaps safeguarded, as mentioned by @kallewoof, behind an options menu-setting.

Sjors:
> Something like this can too easily lead to loss because someone forgot they switched back to mainnet.

How? The addresses are different and incompatible. It seems to me that losing funds would require deception, not just forgetting something.

I'm happy to move this into the options dialog, but I don't see how that helps in the deception scenario. For that it might be more useful to add a stern warning.

> Maybe this should be a default-disabled feature that developers can turn on when compiling.

The purpose of this PR is to make it practical for GUI users to use signet and testnet. Anyone who can self-compile can also just open a terminal and enter `/Applications/Bitcoin\ Core.app/Contents/MacOS/Bitcoin-Qt -signet`.

Being able to use signet and testnet lets users practice with (multisig) wallets without risking real funds. So you lose some safety in one place and gain it in another.

Sjors:
Also note that in Windows it's already possible to launch with a different network: https://github.com/bitcoin-core/gui/issues/78#issuecomment-682673772

ryanofsky:
This seems harmless, but why is not possible to have different icons on macos like we can have on windows or linux? Is this an operating system limitation? An installer limitation? Also why is it better to shut down the mainnet instance when you want to start a testnet instance? Would you want to shut down your main wallet while experimenting and trying something in your test wallet?

Mainnet and testnet just being separate apps with separate icons that can be started and stopped independently would seem like the ideal thing. How does Microsoft Office work on macos? Can you open Word and Powerpoint at the same time, or is there a menu where you switch from Word mode to Powerpoint mode?

doc/release-notes.md
```diff
@@ -131,6 +131,10 @@ GUI changes
 - UTXOs which are locked via the GUI are now stored persistently in the
   wallet database, so are not lost on node shutdown or crash. (#23065)
 
+- Toggle chain from mainnet to testnet, signet or regtest. The GUI now always sets `-chain`,
+  so any occurrences of `-testnet`, `-signet` or `-regtest` in `bitcoin.conf` should
+  be replace with `chain=test`, `chain=signet` or `chain=regtest` respectively. (#414)
```
Sjors, line 136: Should I prefix the GitHub issue with `bitcoin-core/gui#414`?

hebasto: I guess so.

Sjors: Ok, done for for release notes. Not sure how to test the end result.

dlarchikov:
@Sjors thx for PR. Please add support of custom private rpc for regtest.

RandyMcMillan:
Receiving this alert on MacOS 

Screen Shot 2022-02-08 at 12 44 02 AM

```
Process:               Bitcoin-Qt [46840]
Path:                  /Users/USER/*/Bitcoin-Qt.app/Contents/MacOS/Bitcoin-Qt
Identifier:            org.bitcoinfoundation.Bitcoin-Qt
Version:               22.99.0 (22.99.0)
Code Type:             X86-64 (Native)
Parent Process:        bash [79038]
Responsible:           Terminal [425]
User ID:               501

Date/Time:             2022-02-08 00:52:00.534 -0500
OS Version:            Mac OS X 10.15.7 (19H1615)
Report Version:        12
Anonymous UUID:        50406######################9E77F6

Sleep/Wake UUID:       E9EA##########################199E

Time Awake Since Boot: 96000 seconds
Time Since Wake:       95000 seconds

System Integrity Protection: disabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Application Specific Information:
abort() called

```

Screen Shot 2022-02-08 at 12 55 14 AM

RandyMcMillan:
Running Cirrus on the rebased branch:
Cirrus: https://cirrus-ci.com/build/6466679780671488

RandyMcMillan:
It seems that Switch chain belongs in the Settings menu:
https://github.com/RandyMcMillan/gui/commit/9b1456cf59e4ee41a6a9ab4985b4716f231163fc
https://cirrus-ci.com/build/4848117060206592

Screen Shot 2022-02-08 at 2 30 11 AM

RandyMcMillan:
On a side note:

It could be argued that "FIle" is incorrect since mostly wallet operations are listed under this menu.
The wallet related functions could be grouped together - with the other functions in a sub menu group. 

Screen Shot 2022-02-08 at 2 38 13 AM

Sjors:
@RandyMcMillan thanks for testing! That crash is worrying. I'm getting this too now some of the time. Inspired by #336 I changed `QApplication::quit();` to `Q_EMIT quitRequested();`.

I also moved the menu to Settings. And rebased.

@dlarchikov:
> Please add support of custom private rpc for regtest.

I'm not sure what you mean by this.

RandyMcMillan:
reACK Approach https://github.com/bitcoin-core/gui/commit/df3ad9a68ff84c01f62bfb49766da890f4be2c82

Tested on MacOS Arm64 - shutting down cleanly.

I am going to do a bit more testing running multiple networks concurrently (see issue #544) as well as pointing -datadir to internal/external storage - signet_challenge - etc...

src/qt/bitcoingui.cpp
```diff
@@ -454,6 +454,46 @@ void BitcoinGUI::createMenuBar()
...
+    // Add chain selection submenu
+    QAction* chain_selection_action{nullptr};
+    QMenu* chain_selection_menu{nullptr};
+    chain_selection_action = new QAction(tr("&Switch chain"), this);
+    chain_selection_action->setStatusTip(tr("Restart application using a different (test) network"));
+    chain_selection_menu = new QMenu(this);
+
+    connect(chain_selection_menu, &QMenu::aboutToShow, [this, chain_selection_menu] {
+        chain_selection_menu->clear();
+        const std::vector<std::pair<QString, QString>> chains = {{"main", "&Bitcoin"}, {"test", "&Testnet"}, {"regtest", "&Regtest"}, {"signet", "&Signet"}};
```
ryanofsky, line 466: In commit "gui: toggle chain via menu" (12431e0d95157b4df18912b38735acdc927c9762)

Interesting the text is is "Bitcoin" instead of "Mainnet". I think that it is good because it implies the other networks are not the Bitcoin network. Another way to do this while being a little more consistent could be to make the choices "Mainnet (Bitcoin)" / "Testnet" / "Regtest" / "Signet"

ryanofsky (approved):
I have reservations about this feature, and also think the implementation could be improved, but code review ACK 12431e0d95157b4df18912b38735acdc927c9762 since I don't think there are problems with the code.

Code review comments below suggest ways to improve the implementation, and I also agree with luke-jr (https://github.com/bitcoin-core/gui/pull/414#issuecomment-923255521) and kalle and achow and jarolrod that in general this change could result in confusion and be a net negative. I mostly just don't think it makes sense from a user perspective that you are forced to shut down your connection and wallets for one network to connect to a different network, instead of just opening different networks in different windows. There's no problem doing this on Linux or Windows (IIUC from https://github.com/bitcoin-core/gui/issues/78#issuecomment-682673772) because Linux and Windows packages can provide separate icons for launching testnet and signet instances. Ideally the mac package would just provide separate icons as well so we wouldn't need this less flexible and potentially confusing toggle.

I don't see a technical reason why providing separate icons shouldn't be possible on mac. This guide about creating DMGs https://intmaker.com/deploy-to-macos/ has a section "Level 3: additional applications" about adding helper applications alongside the main application. In this case, helper applications could be `Bitcoin-Qt (testnet)` and `Bitcoin-Qt (signet)` scripts that call the main Bitcoin-Qt application with appropriate command line arguments. https://superuser.com/a/16777 shows how to write applescripts that invoke applications with command lines and https://apple.stackexchange.com/a/407885 shows how to package scripts as applications with icons.

Another approach would be to replace the "Switch Chain" toggle menu with an "Open Chain" menu, which would just open new windows connecting to other networks, instead of closing down the whole application and restarting. The implementation could still use QSettings to remember what windows are open so relaunching bitcoin will reopen the same windows (Or if every window was manually closed by the user, reopen the last closed window).

Sjors:
@ryanofsky thanks for all the comments, I'm a bit behind on things, but will get to them.

src/qt/bitcoin.cpp
```diff
@@ -564,6 +564,10 @@ int GuiMain(int argc, char* argv[])
     // - QSettings() will use the new application name after this, resulting in network-specific settings
     // - Needs to be done before createOptionsModel
 
+    QSettings settings;
+    std::string chain = settings.value("chain", "main").toString().toStdString();
+    gArgs.SoftSetArg("-chain", chain);
```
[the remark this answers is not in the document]
ryanofsky: re: https://github.com/bitcoin-core/gui/pull/414#discussion_r815245955

> It might be good to show a message on startup the first time bitcoin is started after switching networks. [...]

Replying to myself, maybe a better  startup message would be: "Bitcoin-Qt configuration has changed and will now connect to the {Testnet} network instead of the {Mainnet} network. {Mainnet} wallets will not be loaded while connected to the {Testnet} network. Also {Testnet} coins are separate from actual bitcoins, and are not supposed to have any value." Dialog could include a checkbox "Do not show this message again," and buttons "Continue connecting to {Testnet} network" / "Exit and switch back to {Mainnet network}" for a safe escape.

(And again I really think ideally none of these dialogs would be necessary and new networks would just open in new windows without closing existing ones. But showing clear messages could at least avoid pitfalls if only one window can be open at a time)

ryanofsky (approved):
re: https://github.com/bitcoin-core/gui/pull/414#issuecomment-1064384414

> @ryanofsky thanks for all the comments, I'm a bit behind on things, but will get to them.

:+1: 

Even though I suggested a lot of changes, I think this PR is ok in its current form. The other reviewers who objected to this PR previously should speak up if they still have issues with it, or if they think any changes would make it better. Sjors has already put a good amount of work into this, and some of the changes I suggested would require even more effort. So it'd be good to sync up and make sure effort is not wasted.

src/qt/bitcoingui.cpp
RandyMcMillan: I noted this as well - IMO "Bitcoin" is a better way to go - "Mainnet" may actually confuse new users at this point.

RandyMcMillan:
Concept reACK

If a custom signet is detected - It would be great to add a truncated "finger print" of the active signet_challenge - I have experimented with adding a signet challenge to the window title and menu and the string is simply too long - but a "finger print" will work in the window title. Maybe use a finger print with 8 chars?

examples:
1. BitcoinCore - [signet] [stuvwxyz]
or
2. BitcoinCore - [signet] [...stuvwxyz]
or
3. BitcoinCore - [signet] [abcd...wxyz]

Just an idea...but if this was added - option 3 seems best IMO.

luke-jr:
Concept re-NACK.

How about for macOS only, a menu item that simply launches a separate instance on a test network?

hebasto:

> But if anyone else wants to do it sooner, don't hesitate!

Will be labeled "Up for grabs".

PR #416, opened 2021-09-02
Add RPC setting

Sjors:
RPC access is disabled by default for the GUI.

With the proliferation of third party desktop applications that use the Bitcoin Core RPC (e.g. Specter Desktop, Sparrow and Wasabi), this PR makes them slight easier to configure. It's no longer required to find and edit `bitcoin.conf` to add `server=1` to it.

Schermafbeelding 2021-09-02 om 14 25 58

### DIFF
--- a/src/qt/forms/optionsdialog.ui
+++ b/src/qt/forms/optionsdialog.ui
@@ -33,7 +33,7 @@
           <string>Automatically start %1 after logging in to the system.</string>
          </property>
          <property name="text">
-          <string>&amp;Start %1 on system login</string>
+          <string>Start %1 on system &amp;login</string>
          </property>
         </widget>
        </item>
@@ -173,12 +173,22 @@
            <property name="sizeHint" stdset="0">
             <size>
              <width>40</width>
-             <height>20</height>
+             <height>40</height>
             </size>
            </property>
           </spacer>
          </item>
         </layout>
+       </item>
+       <item>
+        <widget class="QCheckBox" name="enableServer">
+         <property name="toolTip">
+          <string extracomment="Tooltip text for Options window setting that enables the RPC server.">This allows you or a third party tool to communicate with the node through command-line and JSON-RPC commands.</string>
+         </property>
+         <property name="text">
+          <string extracomment="An Options window setting to enable the RPC server.">Enable RPC &amp;server</string>
+         </property>
+        </widget>
        </item>
        <item>
         <spacer name="verticalSpacer_Main">
--- a/src/qt/optionsdialog.cpp
+++ b/src/qt/optionsdialog.cpp
@@ -210,6 +210,7 @@ void OptionsDialog::setModel(OptionsModel *_model)
     connect(ui->spendZeroConfChange, &QCheckBox::clicked, this, &OptionsDialog::showRestartWarning);
     /* Network */
     connect(ui->allowIncoming, &QCheckBox::clicked, this, &OptionsDialog::showRestartWarning);
+    connect(ui->enableServer, &QCheckBox::clicked, this, &OptionsDialog::showRestartWarning);
     connect(ui->connectSocks, &QCheckBox::clicked, this, &OptionsDialog::showRestartWarning);
     connect(ui->connectSocksTor, &QCheckBox::clicked, this, &OptionsDialog::showRestartWarning);
     /* Display */
@@ -246,6 +247,7 @@ void OptionsDialog::setMapper()
     mapper->addMapping(ui->mapPortUpnp, OptionsModel::MapPortUPnP);
     mapper->addMapping(ui->mapPortNatpmp, OptionsModel::MapPortNatpmp);
     mapper->addMapping(ui->allowIncoming, OptionsModel::Listen);
+    mapper->addMapping(ui->enableServer, OptionsModel::Server);
 
     mapper->addMapping(ui->connectSocks, OptionsModel::ProxyUse);
     mapper->addMapping(ui->proxyIp, OptionsModel::ProxyIP);
--- a/src/qt/optionsmodel.cpp
+++ b/src/qt/optionsmodel.cpp
@@ -149,6 +149,13 @@ void OptionsModel::Init(bool resetSettings)
     if (!gArgs.SoftSetBoolArg("-listen", settings.value("fListen").toBool()))
         addOverriddenOption("-listen");
 
+    if (!settings.contains("server")) {
+        settings.setValue("server", false);
+    }
+    if (!gArgs.SoftSetBoolArg("-server", settings.value("server").toBool())) {
+        addOverriddenOption("-server");
+    }
+
     if (!settings.contains("fUseProxy"))
         settings.setValue("fUseProxy", false);
     if (!settings.contains("addrProxy"))
@@ -363,6 +370,8 @@ QVariant OptionsModel::data(const QModelIndex & index, int role) const
             return settings.value("nThreadsScriptVerif");
         case Listen:
             return settings.value("fListen");
+        case Server:
+            return settings.value("server");
         default:
             return QVariant();
         }
@@ -525,6 +534,12 @@ bool OptionsModel::setData(const QModelIndex & index, const QVariant & value, in
         case Listen:
             if (settings.value("fListen") != value) {
                 settings.setValue("fListen", value);
+                setRestartRequired(true);
+            }
+            break;
+        case Server:
+            if (settings.value("server") != value) {
+                settings.setValue("server", value);
                 setRestartRequired(true);
             }
             break;
--- a/src/qt/optionsmodel.h
+++ b/src/qt/optionsmodel.h
@@ -69,6 +69,7 @@ public:
         ExternalSignerPath,     // QString
         SpendZeroConfChange,    // bool
         Listen,                 // bool
+        Server,                 // bool
         OptionIDRowCount,
     };
 


jarolrod:
Strong Concept ACK!
Didn't realize the solution to this was seemingly so simple.

hebasto:
Concept ACK.

jonatack:
Concept ACK

promag (approved):
Tested ACK 6afe80253fae12265e3744bc35a796670489ab81 on macOS.

Could have a release note or maybe just tag.

kristapsk:
If there is already `server=1` present in `bitcoin.conf`, checkbox will be unchecked by default, I don't think it's the right behaviour.

hebasto:
Concept ACK.

hebasto:
The "Network" tab contains settings that are related to the Bitcoin network. I don't think this is a right place for a private RPC interface setting. Maybe the "Main" tab? Or a new one?

Does the node restart is really required? Could we just call `StartRPC();` / `InterruptRPC(); StopRPC();`?

---

@kristapsk 
> If there is already `server=1` present in `bitcoin.conf`, checkbox will be unchecked by default, I don't think it's the right behaviour.

In such a case you'll see a note: _"Options set in this dialog are overridden by the command line or in the configuration file: -server=1"._

This behavior is well established and documented in [`bitcoin-conf.md`](https://github.com/bitcoin/bitcoin/blob/master/doc/bitcoin-conf.md).

ShaMan239 (commented):
Concept ACK
This PR adds the functionality to set `server=1` from GUI, eliminating editing the `bitcoin.conf` file and setting the server’s ‘value' there. This is done by mapping the state of the newly added `enableServer` button with the value of the server option.

I have tested this PR on **Ubuntu 20.04** **(using Qt version 5.12.8)**. The PR seems to work as expected. But I have observed some behavior that I would like to discuss.

- I am being warned to restart the app, to apply changes each time I switch the state of Checkbox. Shouldn’t a correct behavior be warned only when I change the state from the last saved state? Like: being warned when changing state from 1 -> 0. But the warning disappears when I go from 1 -> 0 -> 1.
- When I select the box, the warning message is “Client restart required to activate changes.” When I deselect it, the message is, “This change would require a client restart.” Is there some reason behind having two different sentences implying the same thing instead of displaying the same message in both cases?

| Selected the Checkbox | Disselected the Checkbox |
| ------------------------------ | ---------------------------------- |
|Screenshot from 2021-09-03 18-37-50|Screenshot from 2021-09-03 18-37-53|
- The message “This change would require a client restart” disappears after around 9.5 seconds. While the message “Client restart required to activate changes” doesn’t seem to disappear. I think this inconsistency should be looked into.

I hope my observations would be a help to the OP to improve upon the PR.
<p>

src/qt/forms/optionsdialog.ui
```diff
@@ -319,6 +319,16 @@
          </property>
         </widget>
        </item>
+       <item>
+        <widget class="QCheckBox" name="enableServer">
+         <property name="toolTip">
+          <string>Accept command line and JSON-RPC commands.</string>
```
jarolrod, line 325: We can add a translator comment by using the `extracomment` field:
```suggestion
          <string extracomment="Tooltip text for Options window setting that enables the RPC server. This allows you to communicate with your node through command line and JSON-RPC commands.">Accept command line and JSON-RPC commands.</string>
```

src/qt/forms/optionsdialog.ui
```diff
@@ -319,6 +319,16 @@
...
+         </property>
+         <property name="text">
+          <string>Enable RPC server</string>
```
jarolrod, line 328: Let's add a mnemonic shortcut as well as a translator comment:
```suggestion
          <string extracomment="An Options window setting to enable the RPC server.">Enable &amp;RPC server</string>
```

jarolrod: I agree with @hebasto [comment](https://github.com/bitcoin-core/gui/pull/416#issuecomment-912438215). This is probably a better fit under the main tab as it is not related to the P2P network. As we collect more settings, this could be a good candidate to move into a new "Advanced" or "Developer" tab.

jarolrod (commented):
Approach ACK

Tested on macOS 12, Ubuntu 20.04, and Windows 10. Here is what I tested:
 - macOS
   - Compiled natively, toggled settings on and off, connected with specter wallet, ran some commands through `bitcoin-cli`.
   - Built and deployed a `dmg`. Installed the bundled app and confirmed that feature works by connecting to specter wallet.
 - Ubuntu 20.04
   - Compiled natively, toggled settings on and off, ran commands through `bitcoin-cli`.
 - Windows 10
   - Cross-compiled, toggled settings on and off, connected with specter wallet, ran commands through `bitcoin-cli`.
   - Built and deployed a setup `.exe` file to install Bitcoin Core, connected to specter wallet.

**Steps forward**
Here are some actionable steps I believe we should take:
- Add translator comments to the new strings
- Add a mnemonic shortcut for this new setting
- Move under the `Main` tab

**Precompiled Binary Users**
When considering users who will be grabbing precompiled binaries; there is a subset of users that won't actually have the ability to interact through the `bitcoin-cli` interface, macOS users who installed bitcoin through the `.dmg` file we provide, and Windows users through the `.exe` file. For these user's, and without extra effort on their part, the functionality of these feature is limited.

Because this is a function of what we provide with the `.dmg`/`.exe`, I don't see this as a big issue. A user who installed Bitcoin through the `.dmg` file is probably a user who is not too concerned with interacting with bitcoin through `bitcoin-cli`. 

If such users were in fact in need of this functionality, they can always grab the respective `zip` or `tar.gz` files and utilize the `bitcoin-cli` executable.

This still expands the feature set for both users. I'm not aware of a way to enable `-server` mode for the `.dmg`/`.exe` users previous to this PR. Additionally, for both users, the killer feature is still available, which is the ability of third-party apps to connect to bitcoin core through the RPC interface.

src/qt/forms/optionsdialog.ui
Sjors: Added the translator comment. Also using some of that text for the tooltip itself.

Sjors:
I moved it to the Main tab.

@hebasto:
> Is the node restart really required? Could we just call `StartRPC();` / `InterruptRPC(); StopRPC();`?

The RPC startup is more convoluted than that it seems, see `AppInitServers` in `init.cpp`. Maybe better for a followup to remove the need to restart.

@ShaMan239:
> I am being warned to restart the app, to apply changes each time I switch the state of Checkbox. Shouldn’t a correct behavior be warned only when I change the state from the last saved state? Like: being warned when changing state from 1 -> 0. But the warning disappears when I go from 1 -> 0 -> 1.

That would be nice, but it's a problem for all settings. In order to fix that we'd have to track whether _any_ setting changed. Because you might change setting A and B, both of which need a restart, and then restore setting A. Right now we just trigger this warning when _one_ setting changes.

>  Is there some reason behind having two different sentences implying the same thing instead of displaying the same message in both cases?

No, but it seems unrelated this change.

@jarolrod:

> When considering users who will be grabbing precompiled binaries; there is a subset of users that won't actually have the ability to interact through the bitcoin-cli interface

Third party applications that use the RPC don't need bitcoin-cli for that.

> I'm not aware of a way to enable -server mode for the .dmg/.exe users previous to this PR.

You can, by editing bitcoin.conf, but this PR makes it easier.

jarolrod:
Shouldn't the UI element react and be aware of the scenario where the server is already enabled through bitcoin.conf. or do we want to keep this mutually exclusive? As in enabling it through the GUI and enabling it through bitcoin.conf are two separate events. Or enabling it at all, whether through the GUI RPC setting or through bitcoin.conf, is a universal event.

It's surprising behavior for someone to have enabled it through their bitcoin.conf, then their GUI settings implies that server mode is not on because this new RPC setting is not checked.

src/qt/forms/optionsdialog.ui
```diff
@@ -173,13 +173,23 @@
...
             </size>
            </property>
           </spacer>
          </item>
         </layout>
        </item>
+       <item>
+        <widget class="QCheckBox" name="enableServer">
+         <property name="toolTip">
+          <string  extracomment="Tooltip text for Options window setting that enables the RPC server.">This allows you or a third party tool to communicate with the node through command line and JSON-RPC commands.</string>
```
jarolrod, line 186: just jotting down a nit, only apply if you get a good reason to retouch. 
```suggestion
          <string  extracomment="Tooltip text for Options window setting that enables the RPC server.">This allows you or a third party tool to communicate with the node through command-line and JSON-RPC commands.</string>
```

Sjors:
> Shouldn't the UI element react and be aware of the scenario where the server is already enabled through bitcoin.conf

It is existing (bad) behaviour that the GUI settings are overriden by bitcoin.conf / arguments and that this is confusing. It's better to discuss that in a separate issue.

hebasto:
@jarolrod 
> It's surprising behavior for someone to have enabled it through their bitcoin.conf, then their GUI settings implies that server mode is not on because this new RPC setting is not checked.

See https://github.com/bitcoin-core/gui/pull/416#issuecomment-912438215

shaavan (approved):
tACK 2c420467a980487752ae2b0c9aa89819b027ab70

Changes since my last review:

- The RPC console option has been moved under the main tab.
- The height of the object, `horizontalSpacer_Main_Threads`, just above the newly added RPC console option, has been changed from `20` to `40`. I am assuming this is done to create enough white space between these two objects.

Tested on Ubuntu 20.04. The changes work correctly and behaving as expected. Adding screenshots to emphasize the difference between master and PR.

|Master|PR|
|----|----|
|Master|PR|

> That would be nice, but it's a problem for all settings. In order to fix that we'd have to track whether any setting changed. Because you might change setting A and B, both of which need a restart, and then restore setting A. Right now we just trigger this warning when one setting changes.

That makes sense. It will unnecessarily make the code complicated. And the additional benefit doesn’t seem to justify the complication introduced. 

In the end, I want to say, thanks for this amazing work, @Sjors!

promag (commented):
ACK 2c420467a980487752ae2b0c9aa89819b027ab70, only moved option to main tab since last review.

src/qt/forms/optionsdialog.ui
```diff
@@ -173,13 +173,23 @@
...
+         </property>
+         <property name="text">
+          <string extracomment="An Options window setting to enable the RPC server.">Enable &amp;RPC server</string>
```
hebasto, line 189: Tested 2c420467a980487752ae2b0c9aa89819b027ab70 on Linux Mint 20.2 (Qt 5.12.8).

The `Alt-R` shortcut is ambiguous with this PR:
```
$ git grep -n -e '&amp;R' src/qt/forms/optionsdialog.ui
src/qt/forms/optionsdialog.ui:189:          <string extracomment="An Options window setting to enable the RPC server.">Enable &amp;RPC server</string>
src/qt/forms/optionsdialog.ui:944:          <string>&amp;Reset Options</string>
```

To solve it, suggesting the following patch:
```diff
--- a/src/qt/forms/optionsdialog.ui
+++ b/src/qt/forms/optionsdialog.ui
@@ -33,7 +33,7 @@
           <string>Automatically start %1 after logging in to the system.</string>
          </property>
          <property name="text">
-          <string>&amp;Start %1 on system login</string>
+          <string>Start %1 on system &amp;login</string>
          </property>
         </widget>
        </item>
@@ -186,7 +186,7 @@
           <string  extracomment="Tooltip text for Options window setting that enables the RPC server.">This allows you or a third party tool to communicate with the node through command line and JSON-RPC commands.</string>
          </property>
          <property name="text">
-          <string extracomment="An Options window setting to enable the RPC server.">Enable &amp;RPC server</string>
+          <string extracomment="An Options window setting to enable the RPC server.">Enable RPC &amp;server</string>
          </property>
         </widget>
        </item>
```

src/qt/forms/optionsdialog.ui
hebasto, line 186: nit: extra space:
```suggestion
          <string extracomment="Tooltip text for Options window setting that enables the RPC server.">This allows you or a third party tool to communicate with the node through command line and JSON-RPC commands.</string>
```

hebasto (commented):
Approach ACK 2c420467a980487752ae2b0c9aa89819b027ab70.

src/qt/forms/optionsdialog.ui
Sjors: Is there documentation somewhere that explains how global / local these keyboard shortcuts are? I can't test it on Mac.

hebasto: Not exactly what you ask for, but https://doc.qt.io/qt-5/qshortcut.html#activatedAmbiguously

I think the scope of shortcuts is limited by a window (not a widget) which keeps the focus.

Sjors:
Addressed feedback.

hebasto (approved):
ACK bd5c826a9630f41255497e0c9a0f1872b5ab78d5, tested on Linux Mint 20.2 (Qt 5.12.8):

Screenshot from 2021-09-16 13-30-29

shaavan (approved):
reACK bd5c826a9630f41255497e0c9a0f1872b5ab78d5

Changes since my last commit:
- Some Changes to mnemonic shortcuts to avoid collisions:
   - the mnemonic shortcut of text of `bitcoinAtStartup` QCheckbox, has been changed from `&amp;Start` to `&amp;login`
   - mnemonic shortcut of the text of added `enableServer` QCheckbox has changed from `&amp;RPC`to `&amp;server`

promag (commented):
Code review ACK bd5c826a9630f41255497e0c9a0f1872b5ab78d5. Just minor fixes to the .ui form since last review.

jonatack (commented):
Post-merge tested ACK bd5c826a9630f41255497e0c9a0f1872b5ab78d5

A practical feature.

Improved state tracking WRT the "need to restart" messages (or even better, avoiding the need to restart) could be nice follow-ups.

PR #417, opened 2021-09-02
Ditch wallet model juggling

promag:
Instantiate `WalletModel` in the main thread - avoids calling`setParent` and `moveToThread`.

Also ensures loading (fetching transactions, addresses, etc) still occurs in the background thread.

### DIFF
--- a/src/qt/walletcontroller.cpp
+++ b/src/qt/walletcontroller.cpp
@@ -125,19 +125,10 @@ WalletModel* WalletController::getOrCreateWallet(std::unique_ptr<interfaces::Wal
         }
     }
 
-    // Instantiate model and register it.
-    WalletModel* wallet_model = new WalletModel(std::move(wallet), m_client_model, m_platform_style,
-                                                nullptr /* required for the following moveToThread() call */);
-
-    // Move WalletModel object to the thread that created the WalletController
-    // object (GUI main thread), instead of the current thread, which could be
-    // an outside wallet thread or RPC thread sending a LoadWallet notification.
-    // This ensures queued signals sent to the WalletModel object will be
-    // handled on the GUI event loop.
-    wallet_model->moveToThread(thread());
-    // setParent(parent) must be called in the thread which created the parent object. More details in #18948.
-    QMetaObject::invokeMethod(this, [wallet_model, this] {
-        wallet_model->setParent(this);
+    WalletModel* wallet_model;
+    // Instantiate model in GUI main thread.
+    QMetaObject::invokeMethod(this, [&] {
+        wallet_model = new WalletModel(std::move(wallet), m_client_model, m_platform_style, this);
     }, GUIUtil::blockingGUIThreadConnection());
 
     m_wallets.push_back(wallet_model);
--- a/src/qt/walletcontroller.cpp
+++ b/src/qt/walletcontroller.cpp
@@ -131,6 +131,8 @@ WalletModel* WalletController::getOrCreateWallet(std::unique_ptr<interfaces::Wal
         wallet_model = new WalletModel(std::move(wallet), m_client_model, m_platform_style, this);
     }, GUIUtil::blockingGUIThreadConnection());
 
+    wallet_model->preload();
+
     m_wallets.push_back(wallet_model);
 
     // WalletModel::startPollBalance needs to be called in a thread managed by
--- a/src/qt/addresstablemodel.cpp
+++ b/src/qt/addresstablemodel.cpp
@@ -70,12 +70,13 @@ class AddressTablePriv
 {
 public:
     QList<AddressTableEntry> cachedAddressTable;
-    AddressTableModel *parent;
+    AddressTableModel* const parent;
+    const bool pk_hash_only;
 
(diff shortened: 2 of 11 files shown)


hebasto:
> There is a change in behavior because now _everything_ that is loaded in `WalletModel` instantiation _happens in the GUI thread_.

Shouldn't we move in the opposite direction?

promag:
@hebasto at the moment loading happens on a separate thread but it loads everything, which is not ideal. For now, I think I'll move the loading code from constructors to separate functions so that only QObject instantiations happen on the GUI thread. But later we need to lazy load wallet data.

promag:
@hebasto added fa425d10753fb1a0ce331343b4548a68fcb0b954 to address the above comment, also updated OP.

promag:
Rebased.

hebasto:
> Depends on [bitcoin/bitcoin#22868](https://github.com/bitcoin/bitcoin/pull/22868) because otherwise, a deadlock would occur during the wallet loading.

bitcoin/bitcoin#22868 has been merged. Rebase?

src/qt/walletcontroller.cpp
```diff
@@ -130,6 +130,8 @@ WalletModel* WalletController::getOrCreateWallet(std::unique_ptr<interfaces::Wal
         wallet_model = new WalletModel(std::move(wallet), m_client_model, m_platform_style, this);
     }, GUIUtil::blockingGUIThreadConnection());
 
+    wallet_model->preload();
```
promag, line 133: In a follow-up, some preload will be replaced by lazy loading.

Gemk83 (approved):

hebasto:
> This change takes advantage of `GUIUtil::ObjectInvoke` to avoid workarounds around how Qt handles object instantiation and threads.

`GUIUtil::ObjectInvoke` seems a bit outdated :)

hebasto:
cc @ryanofsky

promag:
> The first commit "qt: Ditch wallet model juggling" ([f3e7047](https://github.com/bitcoin-core/gui/commit/f3e70479257770bc9f61de694db8704140139548)) is more confusing, because it seems like the new code is doing the same thing as the old code, just making slightly different function calls to do the same thing. This is probably ok, but it would be helpful if commit message said whether this is just supposed to be a local code cleanup, or a bigger change that affects other code. Also that commit seems to delete comments which I think are helpful about _why_ particular threads are assigned. The new comment says what thread is assigned but not why.

It doesn't do the same thing as before since `WalletModel` constructor is executed in the GUI thread whereas before it was executed in the wallet controller background thread or on the RPC handling thread. This change makes loading big wallets block the GUI. But this blocking issue is "fixed" in the following commit.

In terms of Qt calls, it avoids `moveToThread` and `setParent`, since now `WalletModel` is instanced in the GUI thread with the right parent.

> IIUC, the second commit "qt: Move wallet preloading out of model constructors" ([e2ffc98](https://github.com/bitcoin-core/gui/commit/e2ffc98e73df68d0de9755d2ebbb26daa17c92fc)) seems like a good change, moving wallet "preloading" tasks off of the GUI thread into timer threads or notification callback threads. But I'm not 100% sure this is what it is doing. Would be helpful if the commit description said what loading behavior was before the change and how it is different afterward.

So, the `WalletModel` constructor instantiated the remaining models (like `TransactionTableModel`) and those models would load everything from `interfaces::Wallet`. This happened on the wallet controller background thread or the RPC handling thread, depending on how the wallet was loaded. So, the `WalletModel` constructor could block for a while for big wallets - that was the reason for not instantiating the model in the GUI thread.

This commit splits instancing models and loading data. Instancing models still happens on the GUI thread while loading data happens on a separate thread. This "fixes" the blocking issue introduced in the 1st commit.

---

Note that, even though `preload` happens on a background thread, we can have a blocked GUI because underneath loading locks some mutexes that other parts of the GUI could also try to lock. For instance, `WalletImpl::getWalletTxs` locks `cs_wallet`, and the more transactions the worse.

Now that we have the loading split from the instancing, we can improve the loading for big wallets. For instance:
 - load transactions only when the user opens the transactions list, same for addresses, etc
 - adopt a lazy loading approach
 - chunked loading where locks don't get locked for so long.
 - read-write locks (?)

ryanofsky (approved):
Thanks for the explanations. This seems like a pretty good change that just took me a long time to understand. I'd describe the change as "Stop loading wallet data inside the WalletModel constructor, so the WalletModel can be constructed more straightforwardly on the GUI thread without the need for setParent and moveToThread calls, but still not blocking the GUI thread"

Code review ACK e2ffc98e73df68d0de9755d2ebbb26daa17c92fc, but I left a lot of suggestions to clarify, and really would like to see the order of commits switched so wallet data keeps being loaded on the same thread and does not switch to a different thread and then back again.

promag:
@ryanofsky cool, I'll address your comments, makes sense to reorder commits.

hebasto:
Closing for now due to lack of progress. Let us know if this should be reopened.

Labeling "Up for grabs" as well.

PR #419, opened 2021-09-07
Add missing tooltips to options menu settings

jarolrod:
This adds missing tooltips to the text of the `Size of database cache` and the `Number of script verification threads` settings. 

All settings in the Options window will now have appropriate tooltip texts.

### DIFF
--- a/src/qt/forms/optionsdialog.ui
+++ b/src/qt/forms/optionsdialog.ui
@@ -104,6 +104,9 @@
         <layout class="QHBoxLayout" name="horizontalLayout_2_Main">
          <item>
           <widget class="QLabel" name="databaseCacheLabel">
+           <property name="toolTip">
+            <string extracomment="Tooltip text for Options window setting that sets the size of the database cache. Explains the corresponding effects of increasing/decreasing this value.">Maximum database cache size. A larger cache can contribute to faster sync, after which the benefit is less pronounced for most use cases. Lowering the cache size will reduce memory usage. Unused mempool memory is shared for this cache.</string>
+           </property>
            <property name="text">
             <string>Size of &amp;database cache</string>
            </property>
@@ -147,6 +150,9 @@
         <layout class="QHBoxLayout" name="horizontalLayout_Main_VerifyLabel">
          <item>
           <widget class="QLabel" name="threadsScriptVerifLabel">
+           <property name="toolTip">
+            <string extracomment="Tooltip text for Options window setting that sets the number of script verification threads. Explains that negative values mean to leave these many cores free to the system.">Set the number of script verification threads. Negative values correspond to the number of cores you want to leave free to the system.</string>
+           </property>
            <property name="text">
             <string>Number of script &amp;verification threads</string>
            </property>


ShaMan239 (approved):
tACK b669974a42c6e6e09e6402d8d108a783c63119d9

The newly added tooltip messages, as well as extracomments, look great. If I imagine myself in the shoes of the translator, I would be able to translate these tooltip messages in my own language effectively.
Great Work, @jarolrod!
Tested Successfully on Ubuntu 20.04. Adding screenshots of the newly added tooltip messages by this PR.
Screenshot from 2021-09-07 23-33-29
Screenshot from 2021-09-07 23-33-40

kristapsk (approved):
utACK b669974a42c6e6e09e6402d8d108a783c63119d9

src/qt/forms/optionsdialog.ui
```diff
@@ -104,6 +104,9 @@
         <layout class="QHBoxLayout" name="horizontalLayout_2_Main">
          <item>
           <widget class="QLabel" name="databaseCacheLabel">
+           <property name="toolTip">
+            <string extracomment="Tooltip text for Options window setting that sets the size of the database cache. Explains the corresponding effects of increasing/decreasing this value.">Limit the database cache size to a certain value. A larger cache size can contribute to faster sync times. Lowering the cache size will reduce memory usage.</string>
```
hebasto, line 108: Could this tooltip be aligned with https://github.com/bitcoin-core/gui/blob/ee1db7b6dc4e8a3458a7cfb5ead445c56fb75c51/src/init.cpp#L388 and mention "unused mempool memory"?

hebasto (commented):
Concept ACK.

src/qt/forms/optionsdialog.ui
jonatack: Maybe something like this.
```suggestion
            <string extracomment="Tooltip text for Options window setting that sets the size of the database cache. Explains the corresponding effects of increasing/decreasing this value.">Maximum database cache size. A larger cache size can contribute to faster sync. Lowering the cache size will reduce memory usage. Unused mempool memory is shared for this cache.</string>
```

A sync can be faster, a sync time can be lower/higher, so maybe `s/faster sync time/faster sync/`

(Aside, ISTM the "Unused mempool memory is shared for this cache" text in `src/init.cpp` is not very clear, or at least not as precise as it could be.)

jonatack (commented):
Tested ACK b669974a42c6e6e09e6402d8d108a783c63119d9

modulo suggestion below

src/qt/forms/optionsdialog.ui
jonatack: In `doc/reduce-memory.md`, this may be worth appending:

"faster synce, after which the benefit is less pronounced for most use cases, unless fast validation of blocks is important, such as for mining."

jarolrod:
Updated from b669974 -> 9bd168b ([pr419.01](https://github.com/jarolrod/gui/tree/pr419.01) -> [pr419.02](https://github.com/jarolrod/gui/tree/pr419.02))

Changes:
- Addressed @jonatack suggestions: [1](https://github.com/bitcoin-core/gui/pull/419#discussion_r706874149). [2](https://github.com/bitcoin-core/gui/pull/419#discussion_r706874616)
- Dropped the word 'size' from "A larger cache size ...", as to me this reads better.

src/qt/forms/optionsdialog.ui
jarolrod: Thanks for the excellent review @jonatack, I've adopted your suggestions in 9bd168b

hebasto (approved):
ACK 9bd168bf5457c6fd9770769547d8757bf14813b0, tested on Linux Mint 20.2 (Qt 5.12.8).

jonatack:
ACK 9bd168bf5457c6fd9770769547d8757bf14813b0 tested on Debian 5.10.46-4 (2021-08-03)

PR #318, opened 2021-05-05
Add `Copy address` Peers Tab Context Menu Action

jarolrod:
Picking up #264 

This adds a `Copy Address` context menu action to the `Peers Tab`.

Based on the first commit of PR #317 so that we can use `Qt::DisplayRole` in the `copyEntryData` function.

| Master        | PR               |
| ----------- | ----------- |
|  Screen Shot 2021-05-05 at 4 51 11 AM | Screen Shot 2021-05-05 at 4 49 15 AM |

PR #420, opened 2021-09-12
Ensure translator comments end in full stop

jarolrod:
This is a follow-up to #318 which addresses this [nit](https://github.com/bitcoin-core/gui/pull/318#discussion_r706856893) by addressing it globally.

This ensures that all GUI translator comments end in a full stop. If a comment does not end in a full stop, a translator may think  that the rest of the comment is being cut off.

While here, add a colon to the word "see" for any comments touched which point to look at a link.

### DIFF
--- a/src/qt/addressbookpage.cpp
+++ b/src/qt/addressbookpage.cpp
@@ -282,7 +282,7 @@ void AddressBookPage::on_exportButton_clicked()
     QString filename = GUIUtil::getSaveFileName(this,
         tr("Export Address List"), QString(),
         /*: Expanded name of the CSV file format.
-            See https://en.wikipedia.org/wiki/Comma-separated_values */
+            See: https://en.wikipedia.org/wiki/Comma-separated_values. */
         tr("Comma separated file") + QLatin1String(" (*.csv)"), nullptr);
 
     if (filename.isNull())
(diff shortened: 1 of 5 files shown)


hebasto (approved):
ACK 5cc783f5f32ffd550d1b298fc2e9cf6c0439f9fe, I have reviewed the code and it looks OK, I agree it can be merged.

jonatack (commented):
Light ACK.

I'm not sure about appending a full stop to URLs, but I don't know what the translator UX is like.

hebasto:
> I'm not sure about appending a full stop to URLs, but I don't know what the translator UX is like.

I think it's ok as the Transifex recognizes links:

Screenshot from 2021-09-13 17-59-37

shaavan:
Code Review ACK 5cc783f5f32ffd550d1b298fc2e9cf6c0439f9fe

The changes made here look good. Also, I checked and couldn't find any translator comment in `src/qt` that does not end in full stop and is not addressed here.

This is a small but necessary change to maintain the consistency of translators' comments. Great work, @jarolrod!

shaavan (approved):
Code Review ACK 5cc783f5f32ffd550d1b298fc2e9cf6c0439f9fe

The changes made here look good. Also, I checked and couldn't find any translator comment in `src/qt` that does not end in full stop and is not addressed here.

This is a small but necessary change to maintain the consistency of translators' comments. Great work, @jarolrod!

Issue #413, opened 2021-08-31
Checkboxes for onlynet combinations


prayank23:
### Is your feature request related to a problem? Please describe
Using `onlynet` and deciding outbound connection types should be easy for GUI users

### Describe the solution you'd like
![image](https://user-images.githubusercontent.com/13405205/131547261-4aa9a201-0dcf-4dc1-8f5e-fc720ff1a1fd.png)

### Additional context
There are few issues with `onlynet` and they should be fixed with https://github.com/bitcoin/bitcoin/pull/22834

Possible combinations are:

```
ipv4
ipv6
onion
i2p

onion-ipv4
onion-ipv6
i2p-ipv4
i2p-ipv6
i2p-onion
ipv6-ipv4

onion-ipv6-ipv4
i2p-ipv6-ipv4
i2p-onion-ipv4
i2p-onion-ipv6
```
Ignoring `i2p-onion-ipv6-ipv4` because it will be same as not using onlynet. So default can be all options checked if there is nothing mentioned in `bitcoin.conf`.

jonatack:
In terms of number of combinations, there may be Bitcoin Core support for additional BIP155 networks in the future.￼

Rspigler:
Strong Concept ACK.  More settings easily exposed to the GUI the better.

`onion` and `i2p` should be unclickable unless installed and setup properly (and maybe also point to documentation so the user can know how).

It would be more complicated, but having an `Inbound Connections` options would be great as well

PR #421, opened 2021-09-14
Remove proxy checkboxes from options

prayank23:
Never used these checkboxes. No response on IRC or SE so I am assuming its not useful: https://bitcoin.stackexchange.com/questions/109636/understanding-network-options-in-bitcoin-core-gui

Will get some extra space and need to add checkboxes for `onlynet` next: https://github.com/bitcoin-core/gui/issues/413


jarolrod (commented):
NACK

The use cases for these check boxes is specifying where to use a proxy and then setting a different proxy for usage with tor. 

You can't go and start removing code because "you don't use it"; especially when you don't provide any justification and don't fully understand what it's doing.

prayank23:
> The use cases for these check boxes is specifying where to use a proxy and then setting a different proxy for usage with tor.

Can you share docs or steps how is this options being used or helpful?

hebasto:
https://bitcoin.stackexchange.com/a/109651/113152

prayank23:
@hebasto Thanks for answering. Will close this PR as I think removal isn't required although I think some changes or discussion is required because:

1. I2P is added which uses 127.0.0.1:7656 as proxy
2. Few PRs are open in bitcoin/bitcoin which change behaviour of proxy and onlynet. It was discussed that proxy is just a proxy and should remain same for all networks if it can be used.

PR #423, opened 2021-09-15
Make initTranslations a member function of BitcoinApplication class

jarolrod:
Moves the `initTranslations` function from a static function to a member function of the `BitcoinApplication` class. This is for correctness and to clarify that the `QApplication` being worked on is our `BitcoinApplication` object.

### DIFF
--- a/src/qt/bitcoin.h
+++ b/src/qt/bitcoin.h
@@ -17,6 +17,7 @@
 #include <optional>
 
 #include <QApplication>
+#include <QTranslator>
 
 class BitcoinGUI;
 class ClientModel;
@@ -47,6 +48,8 @@ public:
     void createOptionsModel(bool resetSettings);
     /// Initialize prune setting
     void InitPruneSetting(int64_t prune_MiB);
+    /** Set up translations */
+    void initTranslations(QTranslator &qtTranslatorBase, QTranslator &qtTranslator, QTranslator &translatorBase, QTranslator &translator);
     /// Create main window
     void createWindow(const NetworkStyle *networkStyle);
     /// Create splash screen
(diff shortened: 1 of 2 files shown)


shaavan (commented):
Concept ACK

`removeTranslator` and `installTranlator` function removes/installs translation files to **“this”** application. (Source [removeTranslator](https://doc.qt.io/qt-5/qcoreapplication.html#removeTranslator), [installTranslator](https://doc.qt.io/qt-5/qcoreapplication.html#installTranslator) )
It makes sense for a function that deals with these two functions to be part of the BitcoinApplication class, then be a standalone static function.

hebasto:
> This is for correctness...

[`QCoreApplication::installTranslator`](https://doc.qt.io/qt-5/qcoreapplication.html#installTranslator) is a `static` function member of the `QCoreApplication` class.

And I cannot get the point of moving it into the non-static function member of the `BitcoinApplication` class when we actually need no access to any data member of the latter.

jarolrod:
the mentioned installTranslator is a function that we use within *our* initTranslator function. I think this makes things clearer, but this is extremely low-priority. 

Feel free to close if not wanted 🥃

src/qt/bitcoin.h
```diff
@@ -47,6 +48,8 @@ class BitcoinApplication: public QApplication
     void createOptionsModel(bool resetSettings);
     /// Initialize prune setting
     void InitPruneSetting(int64_t prune_MiB);
+    /** Set up translations */
+    void initTranslations(QTranslator &qtTranslatorBase, QTranslator &qtTranslator, QTranslator &translatorBase, QTranslator &translator);
```
laanwj, line 52: What about making it a static member function? I slightly like the idea of grouping it that way, but making it a normal member function is not needed)
(Sure, closing is also fine with me)

PR #430, opened 2021-09-24
Improvements to the open up transaction in third-party link action

### LINKED ISSUE #431: Separator can be omitted when adding third-party link actions to the transaction table
[#4092](https://github.com/bitcoin/bitcoin/pull/4092) added the ability to add third-party links to block explorers and the functionality to show a selected translation in that block explorer. As stated in a review in that PR, it is important that we separate these third-party links from all of the other context-menu actions.
> "Nit: please add a menu separator between the 'normal' actions and the 'third party' actions."

Since the merge of that PR, there has been a bug where it's possible that this separator does not get added, and the third-party link actions merge in with our "normal" actions.

Because the addition of the separator is conditional on being at 0, if the first string passed is empty; then the separator will never get added. Now that we've iterated past 0, any subsequent actions will get merged in with our normal actions instead of being visually separated into its distinct section.

**To Replicate:**
1. Pass in arguments to the `Options->Display->Third party transaction URLs` menu setting that would result in the first string being empty. Example: `| | https://mempool.space/tx/%s | https://blockstream.info/tx/%s`.
2. Restart, navigate to Transaction window and right-click on a transaction.
3. Notice that any third-party links you had passed are now actions that have merged in with "normal" actions.

jarolrod:
[#4092](https://github.com/bitcoin/bitcoin/pull/4092) introduced the ability to open up a transaction in a block explorer. This improves the related code by simplifying the addition and connection of the action through an [overloaded](https://doc.qt.io/archives/qt-5.9/qmenu.html#addAction-5) `addAction` function and prepends action description text to the host, "Show in". The reason to add this text is to make it clear what the action does. It also creates a clearer mental correlation between a user doing the work to add the 3rd-party tx link and this new menu action popping up.

This updates the setting text so that "third-party" is hyphenated. It should be hyphenated because it is being used as a modifier of both "URL" and "transaction URLs".

Additionally, this fixes #431 by ensuring that the seperator will be added before creating action.

Screenshots of visual changes:

**Context menu actions**
|   master   |   pr   |
|--------------|--------|
| 3pt-master | 3pt-pr |

**Setting text**
(tooltip text containing usage of "third-party" is also properly hyphenated)
|   master   |   pr   |
|--------------|--------|
| unnamed | pr-hyphenate |

### DIFF
--- a/src/qt/transactionview.cpp
+++ b/src/qt/transactionview.cpp
@@ -222,14 +222,17 @@ void TransactionView::setModel(WalletModel *_model)
         {
             // Add third party transaction URLs to context menu
             QStringList listUrls = GUIUtil::SplitSkipEmptyParts(_model->getOptionsModel()->getThirdPartyTxUrls(), "|");
+            bool actions_created = false;
             for (int i = 0; i < listUrls.size(); ++i)
             {
                 QString url = listUrls[i].trimmed();
                 QString host = QUrl(url, QUrl::StrictMode).host();
                 if (!host.isEmpty())
                 {
-                    if (i == 0)
+                    if (!actions_created) {
                         contextMenu->addSeparator();
+                        actions_created = true;
+                    }
                     /*: Transactions table context menu action to show the
                         selected transaction in a third-party block explorer.
                         %1 is a stand-in argument for the URL of the explorer. */
--- a/src/qt/transactionview.cpp
+++ b/src/qt/transactionview.cpp
@@ -228,11 +228,9 @@ void TransactionView::setModel(WalletModel *_model)
                 QString host = QUrl(url, QUrl::StrictMode).host();
                 if (!host.isEmpty())
                 {
-                    QAction *thirdPartyTxUrlAction = new QAction(host, this); // use host as menu item label
                     if (i == 0)
                         contextMenu->addSeparator();
-                    contextMenu->addAction(thirdPartyTxUrlAction);
-                    connect(thirdPartyTxUrlAction, &QAction::triggered, [this, url] { openThirdPartyTxUrl(url); });
+                    contextMenu->addAction(host, [this, url] { openThirdPartyTxUrl(url); });
                 }
             }
         }
--- a/src/qt/transactionview.cpp
+++ b/src/qt/transactionview.cpp
@@ -230,7 +230,10 @@ void TransactionView::setModel(WalletModel *_model)
                 {
                     if (i == 0)
                         contextMenu->addSeparator();
-                    contextMenu->addAction(host, [this, url] { openThirdPartyTxUrl(url); });
+                    /*: Transactions table context menu action to show the
+                        selected transaction in a third-party block explorer.
+                        %1 is a stand-in argument for the URL of the explorer. */
+                    contextMenu->addAction(tr("Show in %1").arg(host), [this, url] { openThirdPartyTxUrl(url); });
                 }
             }
         }
--- a/src/qt/forms/optionsdialog.ui
+++ b/src/qt/forms/optionsdialog.ui
@@ -729,10 +729,10 @@
          <item>
           <widget class="QLabel" name="thirdPartyTxUrlsLabel">
            <property name="toolTip">
-            <string>Third party URLs (e.g. a block explorer) that appear in the transactions tab as context menu items. %s in the URL is replaced by transaction hash. Multiple URLs are separated by vertical bar |.</string>
+            <string>Third-party URLs (e.g. a block explorer) that appear in the transactions tab as context menu items. %s in the URL is replaced by transaction hash. Multiple URLs are separated by vertical bar |.</string>
            </property>
            <property name="text">
-            <string>&amp;Third party transaction URLs</string>
+            <string>&amp;Third-party transaction URLs</string>
            </property>
            <property name="buddy">
             <cstring>thirdPartyTxUrls</cstring>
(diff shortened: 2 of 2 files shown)


jarolrod:
Updated from a70a98075a0d258d41c1310553a2337538a1d80c -> 8177578b296adb82fb8ab6c64cd76b832ebcc132 ([pr430.01](https://github.com/jarolrod/gui/tree/pr430.01) -> [pr430.02](https://github.com/jarolrod/gui/tree/pr430.02), [diff](https://github.com/jarolrod/gui/compare/pr430.01..pr430.02))

Changes: include new commit to fix #431

src/qt/transactionview.cpp
```diff
@@ -222,17 +222,21 @@ void TransactionView::setModel(WalletModel *_model)
...
-                    if (i == 0)
+                    if (!actions_created) {
                         contextMenu->addSeparator();
-                    contextMenu->addAction(thirdPartyTxUrlAction);
-                    connect(thirdPartyTxUrlAction, &QAction::triggered, [this, url] { openThirdPartyTxUrl(url); });
+                        actions_created = true;
+                    }
+                    /*: Transactions table context menu action to show the
+                        selected transaction in a third-party block explorer.
+                        %1 is a stand-in argument for the URL of the explorer. */
```
shaavan, line 238: Regarding the Translator's comments. The beginning of the comment looks a little confusing. How about...

```suggestion
                    /*: The context menu action of the Transaction table to show the
                        selected transaction in a third-party block explorer.
                        %1 is a stand-in argument for the URL of the explorer. */
```

shaavan (commented):
Concept ACK

This PR introduced the overloaded **addAction** function, which simplifies the code while keeping the functionality intact. The introduced translator’s comments also look good. I have a suggestion regarding them that I will come to in a minute. Finally, the added **actions_created** boolean allows the code to be more understandable without comments while keeping the same functionality.

src/qt/transactionview.cpp
jarolrod: I think this is better as is, will leave this unresolved for now.

jarolrod:
Updated from 8177578b296adb82fb8ab6c64cd76b832ebcc132 -> 2ccde2f9329d2685563b09ff0830f0d5916f57d5 ([pr430.02](https://github.com/jarolrod/gui/tree/pr430.02) -> [pr430.03](https://github.com/jarolrod/gui/tree/pr430.03), [diff](https://github.com/jarolrod/gui/compare/pr430.02..pr430.03))

Changes: included one more commit to address the proper hyphenation of "third-party".

shaavan (commented):
The new commit looks good. I have tested the PR on Ubuntu 20.04 (using Qt version 5.12.8) with the recent commit, and the changes are reflected perfectly in the GUI.
|Master|PR|
|---------|----|
|master|pr|

stratospher (commented):
tested ACK 2ccde2f. 
In agreement with the PR's objectives. The PR was tested and verified to be consistent with the proposed changes:

1. Adding the description "Show in"  - gave a clearer idea of the menu action.
2. Hyphenating to “third-party”
3. Fixing  [#431](https://github.com/bitcoin-core/gui/issues/431) so that the separator between third-party link actions and normal actions are in different sections even when the first string in a Third party transaction URL is empty. ( Example from [#431](https://github.com/bitcoin-core/gui/issues/431) : `| | https://mempool.space/tx/%s | https://blockstream.info/tx/%s`)

Notes:
- [x] 9980f4a
addAction() adds openThirdPartyTxUrl() to the menu's list of actions and returns it directly. This simplifies the code as compared to the previous version where connect() was used.
- [x] a70a980
Adds the description "Show in" to get a clear picture that we want to show the transaction in a third party transaction URL.
- [x] 8177578
Makes a separate section for third-party links
Uses a boolean variable to identify the first non-empty string and then adds the separator instead of arbitrarily adding the separator after the first string in listUrls[].
- [x] 2ccde2f
Makes “Third party” -> “Third-party”

promag (commented):
Code view ACK 2ccde2f9329d2685563b09ff0830f0d5916f57d5.

hebasto (commented):
Tested 2ccde2f9329d2685563b09ff0830f0d5916f57d5 on Linux Mint 20.2 (Qt 5.12.8) -- #431 is fixed.

An idea for a follow up -- introduce a warning message about possibility to break privacy while using external links.

@prayank23 What do you think about such a warning?

hebasto (approved):
ACK 2ccde2f9329d2685563b09ff0830f0d5916f57d5

prayank23:
> An idea for a follow up -- introduce a warning message about possibility to break privacy while using external links.

Yes we can inform the user about privacy issues because using such block explorers when you are running your own node doesn't make sense. Can add one line about using open source explorers with Bitcoin Core so URL will be localhost.

promag:
The user might want to share the link with someone?

prayank23:
> The user might want to share the link with someone?

Makes sense. Warning can be avoided in that case or sharing transaction id without explorer links is also good enough as the other person can check it in the explorer they prefer to use.

I was testing different things in GUI today to see if attackers can get IP for users and a fun experiment to share problems with users can be:

1. Create a DNS token: https://canarytokens.org/generate
2. Save `https://randomchars.canarytokens.com/tx/%s` in Bitcoin Core settings
3. Right click on any transaction
4. Check your email for IP address

image

Why would someone use a random explorer? Attackers can use several ways to convince newbies with features that are not available in other explorers. Example: https://bitcointalk.org/index.php?topic=2100392.0

PR #435, opened 2021-09-27
Move third-party tx URL setting from Display to Wallet options tab

jarolrod:
Out of all the settings in the Display tab, the `Third-party transaction URLs` is the odd one out. All other settings here affect the way the GUI is displayed. This action affects what context menu actions are available when wallet functionality is enabled, and we have a transaction. As such, it is a better fit for this to be in the Wallet tab.

Not a part of #430 as it is not a direct improvement of the related code. Instead, it is a different opinion on the position of this setting.

Screenshots of visual changes:
|  master  |  pr  |
|-------------|------|
| master-setting-placement | pr-setting-placement |

Based on the last commit of #430 to not conflict.

### DIFF
--- a/src/qt/forms/optionsdialog.ui
+++ b/src/qt/forms/optionsdialog.ui
@@ -189,7 +189,7 @@
        <item>
         <widget class="QCheckBox" name="enableServer">
          <property name="toolTip">
-          <string extracomment="Tooltip text for Options window setting that enables the RPC server.">This allows you or a third party tool to communicate with the node through command-line and JSON-RPC commands.</string>
+          <string extracomment="Tooltip text for Options window setting that enables the RPC server.">This allows you or a third-party tool to communicate with the node through command-line and JSON-RPC commands.</string>
          </property>
          <property name="text">
           <string extracomment="An Options window setting to enable the RPC server.">Enable RPC &amp;server</string>
@@ -225,6 +225,33 @@
           <string extracomment="An Options window setting to set subtracting the fee from a sending amount as default.">Subtract &amp;fee from amount by default</string>
          </property>
         </widget>
+       </item>
+       <item>
+        <layout class="QHBoxLayout" name="horizontalLayout_3_Display">
+         <item>
+          <widget class="QLabel" name="thirdPartyTxUrlsLabel">
+           <property name="toolTip">
+            <string>Third-party URLs (e.g. a block explorer) that appear in the transactions tab as context menu items. "%s" in the URL is replaced by the transaction hash. Multiple URLs are separated by a vertical bar |.</string>
+           </property>
+           <property name="text">
+            <string>&amp;Third-party transaction URLs</string>
+           </property>
+           <property name="buddy">
+            <cstring>thirdPartyTxUrls</cstring>
+           </property>
+          </widget>
+         </item>
+         <item>
+          <widget class="QLineEdit" name="thirdPartyTxUrls">
+           <property name="toolTip">
+            <string>Third-party URLs (e.g. a block explorer) that appear in the transactions tab as context menu items. "%s" in the URL is replaced by the transaction hash. Multiple URLs are separated by a vertical bar |.</string>
+           </property>
+           <property name="placeholderText">
+            <string notr="true">https://example.com/tx/%s</string>
+           </property>
+          </widget>
+         </item>
+        </layout>
        </item>
        <item>
         <widget class="QGroupBox" name="groupBox">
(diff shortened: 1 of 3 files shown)


promag (commented):
Concept ACK, agree wallet tab is a better place for this setting.

Code review ACK 870393fd357f7fc2fedc3a3877b43f1daa736a01. Verified 2nd commit is move-only.

You can ditch https://github.com/bitcoin-core/gui/blob/efa227f5df5f5a9669dec5f1d574cf22d3c0903f/src/qt/optionsdialog.cpp#L91-L92

now that it's under wallet tab.

shaavan (commented):
While I can understand the logic behind moving it under the wallet section, I don’t think the reason is strong enough to make sense of the change. Let me explain why I feel so:

1. **Logical Reason:**
    - Though this option might be an odd one out under the **Display** tab. Still, the purpose of this option is to “Display” the transaction in third-party transaction explorer. To put it under the Wallet section doesn’t seem natural from a non-techie’s perspective.
2. **Pragmatic Reason:**
    - This option has been under the **Display** tab for quite a while now (in fact, since it was introduced). To move it under the **Wallet** tab would introduce unnecessary confusion for a general user.

However, I do have some suggestions to add upon for this PR, which might help form a solid point for this PR, assisting this change in being added.

1. **Rename the option's name:** 
To be honest, when I first read the options name, I was not immediately clear about its purpose, and I had to do some further research to understand it. This is a sign of a bad UX design. Rename it to something like “Third Party Transaction explorer” or something else (I am sure you can think of better names of the option than me, :D) and explain clearly in the tooltip what this option means.
This will have two advantages:
    - First, it would better UX for this option.
    - Second, this would come as an upgraded version of the earlier setting rather than a mere transfer, which would make it seem logical to be moved under the Wallet Tab.

2. **Giving this option its own subsection:**
This setting is too different from other options under the **Wallet** Tab. To put it under the common subsection doesn’t seem to do justice with this setting. A subsection of its own, with a heading that clearly states this setting’s purpose and its relationship to **Wallet** in general, will significantly increase the understandability of this setting as well as make it seems obvious to be put under the **Wallet** section.

These were my 2 cents on this PR. I hope these suggestions might help you make this PR better!

hebasto:
Concept ACK.

hebasto (commented):
bfb3037acafc29827e0f2469a6f4c0ddaceab73a 

"Our usage of "third-party" should be hyphenated
as it is being used as a modifier of both "URL"
and "transaction URLs".

Sorry, but English is poor to get it. Could you elaborate it?

hebasto:
Please rebase :)

src/qt/forms/optionsdialog.ui
```diff
@@ -216,6 +216,33 @@
          </property>
         </widget>
        </item>
+       <item>
+        <layout class="QHBoxLayout" name="horizontalLayout_3_Display">
+         <item>
+          <widget class="QLabel" name="thirdPartyTxUrlsLabel">
+           <property name="toolTip">
+            <string>Third-party URLs (e.g. a block explorer) that appear in the transactions tab as context menu items. %s in the URL is replaced by transaction hash. Multiple URLs are separated by vertical bar |.</string>
```
jonatack, line 224: (change in both the tooltips if taken)
```suggestion
            <string>Third-party URLs (e.g. a block explorer) that appear in the transactions tab as context menu items. %s in the URL is replaced by the transaction hash. Multiple URLs are separated by a vertical bar |.</string>
```
(Feel free to ignore: I wonder if `%s` would be better as `"%s"` to help it not appear to be a bug like `"Dear {$customer_name}"`)

jonatack (commented):
Tested ACK. 

"Third-party" is for an adjective that modifies a subject/noun and "third party" for a subject/noun. So this seems correct. Info: https://www.merriam-webster.com/dictionary/third-party.

I wonder if it would be marginally better UI to group the checkboxes together and the user input boxes together (tx urls and script path).

src/qt/forms/optionsdialog.ui
jonatack: Looks like while doing this you could add a missing hyphen to "third party tool" in `src/qt/forms/optionsdialog.ui:192`.

stratospher (commented):
Concept ACK. In agreement with Third-party transaction URLs being a better fit for the wallet tab.

We’ll also need to make the following changes:

1. In optionsdialog.cpp 

- Move [L217](https://github.com/bitcoin-core/gui/blob/870393fd357f7fc2fedc3a3877b43f1daa736a01/src/qt/optionsdialog.cpp#L217)(from display section) to [L211](https://github.com/bitcoin-core/gui/blob/870393fd357f7fc2fedc3a3877b43f1daa736a01/src/qt/optionsdialog.cpp#L210)(wallet section).
- Move [L270](https://github.com/bitcoin-core/gui/blob/870393fd357f7fc2fedc3a3877b43f1daa736a01/src/qt/optionsdialog.cpp#L270)(from display section) to [L239](https://github.com/bitcoin-core/gui/blob/870393fd357f7fc2fedc3a3877b43f1daa736a01/src/qt/optionsdialog.cpp#L239)(wallet section)

2. In optionsmodel.cpp - Since Third-party transaction URLs come into picture only if ENABLE_WALLET is true,

- Move [L78](https://github.com/bitcoin-core/gui/blob/870393fd357f7fc2fedc3a3877b43f1daa736a01/src/qt/optionsmodel.cpp#L78) to [L115](https://github.com/bitcoin-core/gui/blob/870393fd357f7fc2fedc3a3877b43f1daa736a01/src/qt/optionsmodel.cpp#L115) 
- Move [L348](https://github.com/bitcoin-core/gui/blob/870393fd357f7fc2fedc3a3877b43f1daa736a01/src/qt/optionsmodel.cpp#L348) to [L344](https://github.com/bitcoin-core/gui/blob/870393fd357f7fc2fedc3a3877b43f1daa736a01/src/qt/optionsmodel.cpp#L344)
- Move [L478](https://github.com/bitcoin-core/gui/blob/870393fd357f7fc2fedc3a3877b43f1daa736a01/src/qt/optionsmodel.cpp#L478) to [L473](https://github.com/bitcoin-core/gui/blob/870393fd357f7fc2fedc3a3877b43f1daa736a01/src/qt/optionsmodel.cpp#L473)

unnamed

luke-jr:
Concept NACK, I think the Display tab makes more sense since it's about the GUI not the wallet itself.

Additionally, it may be useful outside wallet stuff when (eg) https://github.com/bitcoin/bitcoin/pull/9849 finally gets merged

PR #436, opened 2021-09-27
Include vout when copying transaction ID from coin selection

### LINKED ISSUE #432: Add ability to copy vout to clipboard from "Coin Selection" dialogue
While testing https://github.com/bitcoin/bitcoin/pull/23065 noticed that in context menu of "Coin Selection" dialogue one can copy to clipboard address, label, amount, transaction ID, but not vout, which is also not displayed anywhere. Both txid and vout are needed for `lockunspent` RPC (I wanted to copy them from GUI, then run RPC command and then check do GUI updates the state).

I propose adding another context menu action that copies either vout or txid:vout to the clipboard.

meshcollider:
Fixes #432

I think it makes sense to just add the vout to the existing function because I can't imagine a situation where a user in the coin selection dialog would want just the transaction ID rather than the specific outpoint, and they can just delete it from the end anyway.

### DIFF
--- a/src/qt/coincontroldialog.cpp
+++ b/src/qt/coincontroldialog.cpp
@@ -55,7 +55,7 @@ CoinControlDialog::CoinControlDialog(CCoinControl& coin_control, WalletModel* _m
     contextMenu->addAction(tr("&Copy address"), this, &CoinControlDialog::copyAddress);
     contextMenu->addAction(tr("Copy &label"), this, &CoinControlDialog::copyLabel);
     contextMenu->addAction(tr("Copy &amount"), this, &CoinControlDialog::copyAmount);
-    copyTransactionHashAction = contextMenu->addAction(tr("Copy transaction &ID"), this, &CoinControlDialog::copyTransactionHash);
+    m_copy_transaction_outpoint_action = contextMenu->addAction(tr("Copy transaction &ID and output index"), this, &CoinControlDialog::copyTransactionOutpoint);
     contextMenu->addSeparator();
     lockAction = contextMenu->addAction(tr("L&ock unspent"), this, &CoinControlDialog::lockCoin);
     unlockAction = contextMenu->addAction(tr("&Unlock unspent"), this, &CoinControlDialog::unlockCoin);
@@ -180,7 +180,7 @@ void CoinControlDialog::showMenu(const QPoint &point)
         // disable some items (like Copy Transaction ID, lock, unlock) for tree roots in context menu
         if (item->data(COLUMN_ADDRESS, TxHashRole).toString().length() == 64) // transaction hash is 64 characters (this means it is a child node, so it is not a parent node in tree mode)
         {
-            copyTransactionHashAction->setEnabled(true);
+            m_copy_transaction_outpoint_action->setEnabled(true);
             if (model->wallet().isLockedCoin(COutPoint(uint256S(item->data(COLUMN_ADDRESS, TxHashRole).toString().toStdString()), item->data(COLUMN_ADDRESS, VOutRole).toUInt())))
             {
                 lockAction->setEnabled(false);
@@ -194,7 +194,7 @@ void CoinControlDialog::showMenu(const QPoint &point)
         }
         else // this means click on parent node in tree mode -> disable all
         {
-            copyTransactionHashAction->setEnabled(false);
+            m_copy_transaction_outpoint_action->setEnabled(false);
             lockAction->setEnabled(false);
             unlockAction->setEnabled(false);
         }
@@ -228,10 +228,14 @@ void CoinControlDialog::copyAddress()
         GUIUtil::setClipboard(contextMenuItem->text(COLUMN_ADDRESS));
 }
 
-// context menu action: copy transaction id
-void CoinControlDialog::copyTransactionHash()
+// context menu action: copy transaction id and vout index
+void CoinControlDialog::copyTransactionOutpoint()
 {
-    GUIUtil::setClipboard(contextMenuItem->data(COLUMN_ADDRESS, TxHashRole).toString());
+    const QString address = contextMenuItem->data(COLUMN_ADDRESS, TxHashRole).toString();
+    const QString vout = contextMenuItem->data(COLUMN_ADDRESS, VOutRole).toString();
+    const QString outpoint = QString("%1:%2").arg(address).arg(vout);
+
+    GUIUtil::setClipboard(outpoint);
 }
 
 // context menu action: lock coin
--- a/src/qt/coincontroldialog.h
+++ b/src/qt/coincontroldialog.h
@@ -63,7 +63,7 @@ private:
 
     QMenu *contextMenu;
     QTreeWidgetItem *contextMenuItem;
-    QAction *copyTransactionHashAction;
+    QAction* m_copy_transaction_outpoint_action;
     QAction *lockAction;
     QAction *unlockAction;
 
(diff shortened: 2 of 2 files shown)


laanwj:
ACK c4ea11c919a49322578db549f31c6e68a2560540
FWIW we used to append the vout to the id when copying from the normal transaction list as well, a long time ago, but it always got a lot of flak from confused users. But I think it's fine here, coin control is an advanced feature anyway.

jarolrod (commented):
Concept ACK

Can we do something to document this change in the text of the related context menu action? If the text states `copy transaction ID` then a user will expect ONLY the txid. We should set the correct expectations of what the action will do.

maybe the following?
```diff
@@ -55,7 +55,7 @@ CoinControlDialog::CoinControlDialog(CCoinControl& coin_control, WalletModel* _m
     contextMenu->addAction(tr("&Copy address"), this, &CoinControlDialog::copyAddress);
     contextMenu->addAction(tr("Copy &label"), this, &CoinControlDialog::copyLabel);
     contextMenu->addAction(tr("Copy &amount"), this, &CoinControlDialog::copyAmount);
-    copyTransactionHashAction = contextMenu->addAction(tr("Copy transaction &ID"), this, &CoinControlDialog::copyTransactionHash);
+    copyTransactionHashAction = contextMenu->addAction(tr("Copy transaction &ID/vout"), this, &CoinControlDialog::copyTransactionHash);
     contextMenu->addSeparator();
     lockAction = contextMenu->addAction(tr("L&ock unspent"), this, &CoinControlDialog::lockCoin);
     unlockAction = contextMenu->addAction(tr("&Unlock unspent"), this, &CoinControlDialog::unlockCoin);
```

meshcollider:
Modified the context menu text as per @jarolrod's suggestion

src/qt/coincontroldialog.cpp
```diff
@@ -231,7 +231,8 @@ void CoinControlDialog::copyAddress()
 // context menu action: copy transaction id
 void CoinControlDialog::copyTransactionHash()
```
kristapsk, line 232: Method should be renamed as well as comment above, as it no more copies just transaction hash id alone.

meshcollider: Done :)

src/qt/coincontroldialog.h
```diff
@@ -63,7 +63,7 @@ class CoinControlDialog : public QDialog
 
     QMenu *contextMenu;
     QTreeWidgetItem *contextMenuItem;
-    QAction *copyTransactionHashAction;
+    QAction *copyTransactionOutpointAction;
```
shaavan, line 66: The Clang format checker suggested the following change:

```suggestion
    QAction* copyTransactionOutpointAction;
```

src/qt/coincontroldialog.cpp
```diff
@@ -229,9 +229,10 @@ void CoinControlDialog::copyAddress()
 }
 
 // context menu action: copy transaction id
```
shaavan, line 231: I guess you missed updating the comment.

```suggestion
// context menu action: copy transaction id and output index
```

shaavan (commented):
Concept ACK

This PR adds the functionality of copying output index and transaction ID as a context menu option from the Coin Selection dialog box. I was able to successfully test this PR on Ubuntu 20.04 (Using Qt version 5.12.8).

**The Difference with Master:**

|Master|PR|
|---------|----|
|9f4cc49b1294299ae9bba26e5b48b82a032222f8c46d05e455602b298ff8bcba|9f4cc49b1294299ae9bba26e5b48b82a032222f8c46d05e455602b298ff8bcba:1|

The changes in functions’ names aptly represent the change in functionality. I have mentioned some minor nits that shall be looked into.

Btw, I was just curious. Would it be beneficial to apply this functionality to the context menu action under the transaction tab?

src/qt/coincontroldialog.cpp
```diff
@@ -229,9 +229,10 @@ void CoinControlDialog::copyAddress()
...
-void CoinControlDialog::copyTransactionHash()
+void CoinControlDialog::copyTransactionOutpoint()
 {
-    GUIUtil::setClipboard(contextMenuItem->data(COLUMN_ADDRESS, TxHashRole).toString());
+    const QString outpoint = contextMenuItem->data(COLUMN_ADDRESS, TxHashRole).toString() + ":" + contextMenuItem->data(COLUMN_ADDRESS, VOutRole).toString();
```
promag, line 234: nit, maybe a bit more clear:

```cpp
const auto address = contextMenuItem->data(COLUMN_ADDRESS, TxHashRole).toString();
const auto vout = contextMenuItem->data(COLUMN_ADDRESS, VOutRole).toString();
const auto outpoint = QString("%1:%2").arg(address).arg(vout);
```

src/qt/coincontroldialog.h
promag: nit, also use `m_` prefix and snake case.

promag (commented):
Code review ACK 949e700149a7d84e7b507f9c8d5dd44af1d82ef2.

@laanwj I think we can have both actions copy-txid and copy-outpoint. 

Left some nits.

meshcollider:
Thanks for the reviews, will address the nits shortly.

> Would it be beneficial to apply this functionality to the context menu action under the transaction tab?

IMO no, because unlike the coin selection which deals with UTXOs, transactions can have multiple outputs and most non-technical users don't understand the idea of outputs anyway, so it would just be confusing.

meshcollider:
Addressed nits

kristapsk (approved):
ACK 10c6929d55ba9bc203bbadfb834537445dbd67ce

src/qt/coincontroldialog.cpp
```diff
@@ -55,7 +55,7 @@ CoinControlDialog::CoinControlDialog(CCoinControl& coin_control, WalletModel* _m
     contextMenu->addAction(tr("&Copy address"), this, &CoinControlDialog::copyAddress);
     contextMenu->addAction(tr("Copy &label"), this, &CoinControlDialog::copyLabel);
     contextMenu->addAction(tr("Copy &amount"), this, &CoinControlDialog::copyAmount);
-    copyTransactionHashAction = contextMenu->addAction(tr("Copy transaction &ID"), this, &CoinControlDialog::copyTransactionHash);
+    m_copy_transaction_outpoint_action = contextMenu->addAction(tr("Copy transaction &ID and output index"), this, &CoinControlDialog::copyTransactionOutpoint);
```
promag, line 58: Anything against having both actions?

meshcollider: @promag seems a bit more GUI clutter for no purpose. Its a manual action touching the clipboard so it can't interfere with any automation, so the user can always just manually delete the `:n` on the end if they don't want it. IMO txid alone is useless in this situation because this is a specific outpoint selection dialog.

hebasto:
Concept ACK.

hebasto (approved):
ACK 10c6929d55ba9bc203bbadfb834537445dbd67ce, tested on Linux Mint 20.2 (Qt 5.12.8).

shaavan (approved):
ACK 10c6929

Changes since my last review:
- `copyTransactionOutpointAction` has been renamed to `m_copy_transaction_outpoint_action`.
- Comment has been aptly updated to match the change in functionality.
- The `copyTransactionOutpoint` function has been expanded to make it more clear.
- PR has been rebased on the current master.
- Formatting has now been corrected.

**Difference with Master (Screenshot):**
|Master|PR|
|---------|----|
|Master|PR|

PR #439, opened 2021-09-29
Do not show unused widgets at startup

### LINKED ISSUE #428: Weird status icons stacking - ghost icon
There are at least up to six icons in order:

1. unit: btc, sat...
2. HD: disabled/enabled.
3. wallet lock.
4. proxy (disabled/enabled ?).
5. network status and number of connections.
6. syncing/synced.

I think there's a way to make a ghostly placeholder appear between the unit (btc, sat...) and the proxy "P".

**Expected behavior**

A normal stacking of icons.

**Actual behavior**

Too much spacing between the unit and the proxy "P".

**To reproduce**

Boot Bitcoin Core Qt without a wallet loaded. Just happens, in my case...

**System information**

Self-compiled v22.0 tag.

Contemporary Debian GNU/Linux, KDE 5.54.0, Qt 5.11.3, xcb.

**Other**

I might have some time to bisect this.

hebasto:
On master (8d83f9c1d19d9e701291b9906723dfb548d3dc5f), when starting without wallets the `labelWalletEncryptionIcon` and
`labelWalletHDStatusIcon` widgets are not used but still visible as empty space:

Screenshot from 2021-09-29 21-59-22

If one opens any wallet then closes it, the widget layout becomes densed:

Screenshot from 2021-09-29 22-05-31

This PR makes widget layout densed at startup.

Fixes #428.

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -171,7 +171,9 @@ BitcoinGUI::BitcoinGUI(interfaces::Node& node, const PlatformStyle *_platformSty
         frameBlocksLayout->addWidget(unitDisplayControl);
         frameBlocksLayout->addStretch();
         frameBlocksLayout->addWidget(labelWalletEncryptionIcon);
+        labelWalletEncryptionIcon->hide();
         frameBlocksLayout->addWidget(labelWalletHDStatusIcon);
+        labelWalletHDStatusIcon->hide();
     }
     frameBlocksLayout->addWidget(labelProxyIcon);
     frameBlocksLayout->addStretch();


katesalazar:
Only watched it in action on macOS, but it works. Thank you!

jarolrod (commented):
ACK 489060dcaf45e66c9af4c94cd9376f3f225da18b

Confirmed the related bug (#428) and tested that this fixes it. The issue here is that we aren't hiding the unused icons. The PR fixes the issue by hiding them. When they are needed again (when you create a wallet), they are properly shown. Tested on macOS 12. 

| Bug on master | PR | Create Wallet, Icons will still Appear |
|----------------|----|-------------------------------------|
| Screen Shot 2021-09-29 at 9 15 46 PM | Screen Shot 2021-09-29 at 9 16 55 PM | Screen Shot 2021-09-29 at 9 17 21 PM |

promag (commented):
Code review ACK 489060dcaf45e66c9af4c94cd9376f3f225da18b.

katesalazar:
Port it back to 22 please? And maybe 0.21 as well?

PR #440, opened 2021-09-29
Do not require restart if overridden option is modified

hebasto:
The options set via the GUI can be overridden by settings in the `bitcoin.conf` or by command-line options. If such an option is overridden, its modification in the GUI has no effect. Therefore, no need to require the client restart.

### DIFF
--- a/src/qt/optionsdialog.cpp
+++ b/src/qt/optionsdialog.cpp
@@ -200,21 +200,75 @@ void OptionsDialog::setModel(OptionsModel *_model)
     /* warn when one of the following settings changes by user action (placed here so init via mapper doesn't trigger them) */
 
     /* Main */
-    connect(ui->prune, &QCheckBox::clicked, this, &OptionsDialog::showRestartWarning);
+    connect(ui->prune, &QCheckBox::clicked, [this] {
+        if (!model->getOverriddenByCommandLine().contains("-prune")) {
+            showRestartWarning();
+        }
+    });
     connect(ui->prune, &QCheckBox::clicked, this, &OptionsDialog::togglePruneWarning);
-    connect(ui->pruneSize, qOverload<int>(&QSpinBox::valueChanged), this, &OptionsDialog::showRestartWarning);
-    connect(ui->databaseCache, qOverload<int>(&QSpinBox::valueChanged), this, &OptionsDialog::showRestartWarning);
-    connect(ui->externalSignerPath, &QLineEdit::textChanged, [this]{ showRestartWarning(); });
-    connect(ui->threadsScriptVerif, qOverload<int>(&QSpinBox::valueChanged), this, &OptionsDialog::showRestartWarning);
+    connect(ui->pruneSize, qOverload<int>(&QSpinBox::valueChanged), [this] {
+        if (!model->getOverriddenByCommandLine().contains("-prune")) {
+            showRestartWarning();
+        }
+    });
+
+    connect(ui->databaseCache, qOverload<int>(&QSpinBox::valueChanged), [this] {
+        if (!model->getOverriddenByCommandLine().contains("-dbcache")) {
+            showRestartWarning();
+        }
+    });
+
+    connect(ui->externalSignerPath, &QLineEdit::textChanged, [this] {
+        if (!model->getOverriddenByCommandLine().contains("-signer")) {
+            showRestartWarning();
+        }
+    });
+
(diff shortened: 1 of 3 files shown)


jonatack:
Concept ACK.

jarolrod:
concept ack

shaavan:
Concept ACK

shaavan (approved):
tACK 17cea05b4c23606f3d7b2be9e1c01f9a8188dee9

This PR allows not displaying restart warning for change in options, whose value is already overridden under the bitcoin.conf file.

This is done by adding a conditional statement that displays the warning only when, setting’s name is not present in the bitcoin.conf file (**edit:** or in command-line options). I checked if there is a way to write only the setting’s name in bitcoin.conf file without specifying the value. There is none. So the logic of code is exemplary here.

The formatting of the code is correct, and the ordering of settings correctly matches the order in which they are displayed. I was able to test this PR on Ubuntu 20.04 successfully. I was able to try all the individual settings successfully. Let me added an example screenshot pair.

**Screenshots (for prune option):**

|Before updating the bitcoin.conf file|After Updating the bitcoin.conf file|
|-----------------------------------------------|---------------------------------------------|
|Before|After|

I agree with the changes suggested in this PR. The settings overridden by the bitcoin.conf file will not be affected by the change of setting in GUI, and hence they don’t need a restart warning for them.

hebasto:
@shaavan 
> ... is not present in the bitcoin.conf file...

Or in command-line options.

luke-jr:
Wouldn't it make more sense to instead have a warning message that restart is needed *without* the overriding option, to make it have effect?

jarolrod (commented):
tACK 17cea05

Tested on macOS 12. Conceptually, I agree with this PR. But, the current situation/UX around settings is not ideal. While this change is a step in the right direction, we should still do more.

In regards to this change, we could follow up with some change that makes it clear to the user that a setting is currently overriden by the `bitcoin.conf`. It could use similar logic as this pr, utilizing: `model->getOverriddenByCommandLine().contains("-prune")`.

ryanofsky:
I think this PR seems a little dangerous and is is not really an improvement in it's current form, but it could be a good change with some tweaks. 

The problem I see is the same one luke-jr and jarolrod pointed out, that **this is not providing the user a clear enough warning that the change they just tried to make will be ignored**. Even if the "This change would require a restart" message was not strictly accurate in all cases (because restarting could be necessary but not sufficient to apply the changes), at least the "This change would require a restart" message gave immediate feedback to the user that the changes they just made would not take effect. After this PR, there is no prominent feedback about the changes being ignored, except for the easy-to-miss "Options set in this dialog are overridden" blurb.

I think if the goal of this PR is to remove ambiguity of the "This change would require a restart message", instead of dropping it would be better to replace it with less ambiguous messages depending on the circumstance:

- "Changes will not take effect until bitcoin is restarted" in the normal case.
- "Changes will not take effect until bitcoin is restarted and command line or configuration file overrides are removed" in the special case addressed by this PR.

> In regards to this change, we could follow up with some change that makes it clear to the user that a setting is currently overriden by the `bitcoin.conf`. It could use similar logic as this pr, utilizing: `model->getOverriddenByCommandLine().contains("-prune")`.

This is really a broader question, and probably best to discuss in a new issue. (There is also a wiki page https://github.com/bitcoin-core/bitcoin-devwiki/wiki/Settings-design-questions to explore tradeoffs, but it hasn't really been filled out or updated.)

Two things are worth noting though:

- [bitcoin/bitcoin#15936](https://github.com/bitcoin/bitcoin/pull/15936) changes behavior so that settings changed in the GUI will take priority over setting specified in `bitcoin.conf`, and will be used by `bitcoind`. I think this makes GUI settings dialog more useful and powerful and consistent and reduces complexity here.

- [bitcoin/bitcoin#15936](https://github.com/bitcoin/bitcoin/pull/15936) does not change the behavior of command line settings overriding GUI settings, which is a good thing, but still leaves open the question of how GUI should reflect the overrides. I don't think anybody thinks the "Options set in this dialog are overridden" blurb is a great solution. Another solution could be to show the overridden settings as disabled. Another interesting solution is one Sjors implemented in [bitcoin/bitcoin#12833](https://github.com/bitcoin/bitcoin/pull/12833) which is to drop the "Options set in this dialog are overridden" blurb entirely, show active command line values in the settings dialog, but treat the settings as temporary and do not persist them unless the user actually sets them.

PR #441, opened 2021-09-30
Add Create Unsigned button to SendConfirmationDialog

achow101:
Instead of having different buttons or changing button behavior for making a PSBT, just have SendConfirmationDialog return whether the user wants a PSBT or a broadcasted transaction. Since this dialog is used by both the bumpFeeAction and the SendCoinsDialog, changes to both to support the different behavior is needed. They will check the return value of the SendConfirmationDialog for whether a PSBT needs to be created instead of checking whether private keys are disabled.

Strings used in this dialog are being slightly modified to work with both private keys enabled and disabled wallets.

Moved from https://github.com/bitcoin/bitcoin/pull/18789

### DIFF
--- a/src/qt/forms/optionsdialog.ui
+++ b/src/qt/forms/optionsdialog.ui
@@ -252,6 +252,16 @@
             </property>
            </widget>
           </item>
+          <item>
+           <widget class="QCheckBox" name="m_enable_psbt_controls">
+            <property name="text">
+             <string extracomment="An options window setting to enable PSBT controls.">Enable &amp;PSBT controls</string>
+            </property>
+            <property name="toolTip">
+             <string extracomment="Tooltip text for options window setting that enables PSBT controls.">Whether to show PSBT controls.</string>
+            </property>
+           </widget>
+          </item>
          </layout>
         </widget>
        </item>
--- a/src/qt/optionsmodel.cpp
+++ b/src/qt/optionsmodel.cpp
@@ -83,6 +83,11 @@ void OptionsModel::Init(bool resetSettings)
         settings.setValue("fCoinControlFeatures", false);
     fCoinControlFeatures = settings.value("fCoinControlFeatures", false).toBool();
 
+    if (!settings.contains("enable_psbt_controls")) {
+        settings.setValue("enable_psbt_controls", false);
+    }
+    m_enable_psbt_controls = settings.value("enable_psbt_controls", false).toBool();
+
     // These are shared with the core or have a command-line parameter
     // and we want command-line parameters to overwrite the GUI settings.
     //
@@ -360,6 +365,8 @@ QVariant OptionsModel::data(const QModelIndex & index, int role) const
             return m_use_embedded_monospaced_font;
         case CoinControlFeatures:
             return fCoinControlFeatures;
+        case EnablePSBTControls:
+            return settings.value("enable_psbt_controls");
         case Prune:
             return settings.value("bPrune");
         case PruneSize:
@@ -506,6 +513,10 @@ bool OptionsModel::setData(const QModelIndex & index, const QVariant & value, in
             fCoinControlFeatures = value.toBool();
             settings.setValue("fCoinControlFeatures", fCoinControlFeatures);
             Q_EMIT coinControlFeaturesChanged(fCoinControlFeatures);
+            break;
+        case EnablePSBTControls:
+            m_enable_psbt_controls = value.toBool();
+            settings.setValue("enable_psbt_controls", m_enable_psbt_controls);
             break;
         case Prune:
             if (settings.value("bPrune") != value) {
--- a/src/qt/optionsmodel.h
+++ b/src/qt/optionsmodel.h
@@ -70,6 +70,7 @@ public:
         SpendZeroConfChange,    // bool
         Listen,                 // bool
         Server,                 // bool
+        EnablePSBTControls,     // bool
         OptionIDRowCount,
     };
 
@@ -91,6 +92,7 @@ public:
     bool getUseEmbeddedMonospacedFont() const { return m_use_embedded_monospaced_font; }
     bool getCoinControlFeatures() const { return fCoinControlFeatures; }
     bool getSubFeeFromAmount() const { return m_sub_fee_from_amount; }
+    bool getEnablePSBTControls() const { return m_enable_psbt_controls; }
     const QString& getOverriddenByCommandLine() { return strOverriddenByCommandLine; }
 
     /* Explicit setters */
@@ -116,6 +118,7 @@ private:
     bool m_use_embedded_monospaced_font;
     bool fCoinControlFeatures;
     bool m_sub_fee_from_amount;
+    bool m_enable_psbt_controls;
     /* settings that were overridden by command-line */
     QString strOverriddenByCommandLine;
 
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -324,16 +324,17 @@ bool SendCoinsDialog::PrepareSendText(QString& question_string, QString& informa
         formatted.append(recipientElement);
     }
 
-    if (model->wallet().privateKeysDisabled() && !model->wallet().hasExternalSigner()) {
-        question_string.append(tr("Do you want to draft this transaction?"));
-    } else {
-        question_string.append(tr("Are you sure you want to send?"));
-    }
-
+    /*: Message displayed when attempting to create a transaction. Cautionary text to prompt the user to verify
+        that the displayed transaction details represent the transaction the user intends to create. */
+    question_string.append(tr("Do you want to create this transaction?"));
     question_string.append("<br /><span style='font-size:10pt;'>");
     if (model->wallet().privateKeysDisabled() && !model->wallet().hasExternalSigner()) {
+        /*: Text to inform a user attempting to create a transaction of their current options. At this stage,
+            a user can only create a PSBT. This string is displayed when private keys are disabled and an external
+            signer is not available. */
         question_string.append(tr("Please, review your transaction proposal. This will produce a Partially Signed Bitcoin Transaction (PSBT) which you can save or copy and then sign with e.g. an offline %1 wallet, or a PSBT-compatible hardware wallet.").arg(PACKAGE_NAME));
     } else {
+        /*: Text to prompt a user to review the details of the transaction they are attempting to send. */
         question_string.append(tr("Please, review your transaction."));
     }
     question_string.append("</span>%1");
@@ -397,21 +398,20 @@ void SendCoinsDialog::sendButtonClicked([[maybe_unused]] bool checked)
     if (!PrepareSendText(question_string, informative_text, detailed_text)) return;
     assert(m_current_transaction);
 
-    const QString confirmation = model->wallet().privateKeysDisabled() && !model->wallet().hasExternalSigner() ? tr("Confirm transaction proposal") : tr("Confirm send coins");
-    const QString confirmButtonText = model->wallet().privateKeysDisabled() && !model->wallet().hasExternalSigner() ? tr("Create Unsigned") : tr("Sign and send");
-    auto confirmationDialog = new SendConfirmationDialog(confirmation, question_string, informative_text, detailed_text, SEND_CONFIRM_DELAY, confirmButtonText, this);
+    const QString confirmation = tr("Confirm send coins");
+    auto confirmationDialog = new SendConfirmationDialog(confirmation, question_string, informative_text, detailed_text, SEND_CONFIRM_DELAY, !model->wallet().privateKeysDisabled(), true, this);
     confirmationDialog->setAttribute(Qt::WA_DeleteOnClose);
     // TODO: Replace QDialog::exec() with safer QDialog::show().
     const auto retval = static_cast<QMessageBox::StandardButton>(confirmationDialog->exec());
 
-    if(retval != QMessageBox::Yes)
+    if(retval != QMessageBox::Yes && retval != QMessageBox::Save)
     {
         fNewRecipientAllowed = true;
         return;
     }
 
     bool send_failure = false;
-    if (model->wallet().privateKeysDisabled()) {
+    if (retval == QMessageBox::Save) {
         CMutableTransaction mtx = CMutableTransaction{*(m_current_transaction->getWtx())};
         PartiallySignedTransaction psbtx(mtx);
         bool complete = false;
@@ -512,6 +512,7 @@ void SendCoinsDialog::sendButtonClicked([[maybe_unused]] bool checked)
             assert(false);
         } // msgBox.exec()
     } else {
+        assert(!model->wallet().privateKeysDisabled());
         // now send the prepared transaction
         WalletModel::SendCoinsReturn sendStatus = model->sendCoins(*m_current_transaction);
         // process sendStatus and on error generate message shown to user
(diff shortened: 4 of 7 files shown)


src/qt/sendcoinsdialog.cpp
```diff
@@ -324,16 +324,22 @@ bool SendCoinsDialog::PrepareSendText(QString& question_string, QString& informa
...
-    if (model->wallet().privateKeysDisabled() && !model->wallet().hasExternalSigner()) {
-        question_string.append(tr("Do you want to draft this transaction?"));
-    } else {
-        question_string.append(tr("Are you sure you want to send?"));
-    }
-
     question_string.append("<br /><span style='font-size:10pt;'>");
+    /*: Message displayed when attempting to create a transaction. Cautionary text to prompt the user to verify
+        that the displayed transaction details represent the transaction the user intends to create. */
+    question_string.append(tr("Do you want to create this transaction?"));
```
shaavan, line 330: Though the line is correctly displayed in the GUI, there is a lack of space after it, which seems odd.

**Screenshot (PR):**
Note: the slight white highlighting is done to emphasize and was not present while testing.
Correction

Either a space should be added after the statement so that the spacing is corrected. Or it should be displayed as a header text as is done on the master branch:

**Screenshot (Master):**
Master

src/qt/optionsmodel.h
```diff
@@ -70,6 +70,7 @@ class OptionsModel : public QAbstractListModel
         SpendZeroConfChange,    // bool
         Listen,                 // bool
         Server,                 // bool
+        EnablePSBTControls,     // bool
```
shaavan, line 73: The Clang format check provided the following corrections.

```diff
-        StartAtStartup,         // bool
-        ShowTrayIcon,           // bool
-        MinimizeToTray,         // bool
-        MapPortUPnP,            // bool
-        MapPortNatpmp,          // bool
-        MinimizeOnClose,        // bool
-        ProxyUse,               // bool
-        ProxyIP,                // QString
-        ProxyPort,              // int
-        ProxyUseTor,            // bool
-        ProxyIPTor,             // QString
-        ProxyPortTor,           // int
-        DisplayUnit,            // BitcoinUnits::Unit
-        ThirdPartyTxUrls,       // QString
-        Language,               // QString
+        StartAtStartup,            // bool
+        ShowTrayIcon,              // bool
+        MinimizeToTray,            // bool
+        MapPortUPnP,               // bool
+        MapPortNatpmp,             // bool
+        MinimizeOnClose,           // bool
+        ProxyUse,                  // bool
+        ProxyIP,                   // QString
+        ProxyPort,                 // int
+        ProxyUseTor,               // bool
+        ProxyIPTor,                // QString
+        ProxyPortTor,              // int
+        DisplayUnit,               // BitcoinUnits::Unit
+        ThirdPartyTxUrls,          // QString
+        Language,                  // QString
         UseEmbeddedMonospacedFont, // bool
-        CoinControlFeatures,    // bool
-        SubFeeFromAmount,       // bool
-        ThreadsScriptVerif,     // int
-        Prune,                  // bool
-        PruneSize,              // int
-        DatabaseCache,          // int
-        ExternalSignerPath,     // QString
-        SpendZeroConfChange,    // bool
-        Listen,                 // bool
-        Server,                 // bool
-        EnablePSBTControls,     // bool
+        CoinControlFeatures,       // bool
+        SubFeeFromAmount,          // bool
+        ThreadsScriptVerif,        // int
+        Prune,                     // bool
+        PruneSize,                 // int
+        DatabaseCache,             // int
+        ExternalSignerPath,        // QString
+        SpendZeroConfChange,       // bool
+        Listen,                    // bool
+        Server,                    // bool
+        EnablePSBTControls,        // bool
         OptionIDRowCount,
     };

```

shaavan (changes requested):
Concept ACK

I agree with the idea behind this PR.
Though I have not completed my test, I have found some corrections here and there in the code that will be great for starters. I shall post my thorough review after completing the testings later.

src/qt/optionsmodel.cpp
```diff
@@ -83,6 +83,11 @@ void OptionsModel::Init(bool resetSettings)
         settings.setValue("fCoinControlFeatures", false);
     fCoinControlFeatures = settings.value("fCoinControlFeatures", false).toBool();
 
+    if (!settings.contains("fEnablePSBTControls")) {
+        settings.setValue("fEnablePSBTControls", false);
+    }
```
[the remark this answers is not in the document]
achow101: While that may be technically correct, code written that way can be more confusing. Our style guide states that brackets must be included even for code like this.

src/qt/forms/optionsdialog.ui
```diff
@@ -252,6 +252,16 @@
             </property>
            </widget>
           </item>
+          <item>
+           <widget class="QCheckBox" name="m_enable_psbt_controls">
```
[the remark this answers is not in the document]
achow101: This follows the new naming convention that the style guide specifies. There is generally an effort to migrate to a new style, so new code may not match the style of surrounding older code.

src/qt/optionsmodel.h
achow101: This are unrelated changes so I will not include them.

src/qt/sendcoinsdialog.cpp
achow101: Fixed

src/qt/forms/optionsdialog.ui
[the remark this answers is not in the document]
shaavan: I was not aware of the change in writing style. That's something new I learned today!

src/qt/optionsmodel.cpp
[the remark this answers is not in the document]
shaavan: I might be wrong, and it might be a new change in writing style. But the code above it doesn't use brackets for one-liner **if-else**.

```
    if (!settings.contains("strThirdPartyTxUrls"))
        settings.setValue("strThirdPartyTxUrls", "");
    strThirdPartyTxUrls = settings.value("strThirdPartyTxUrls", "").toString();

    if (!settings.contains("fCoinControlFeatures"))
        settings.setValue("fCoinControlFeatures", false);
    fCoinControlFeatures = settings.value("fCoinControlFeatures", false).toBool();
```

src/qt/optionsmodel.h
shaavan: Makes Sense 👍

src/qt/optionsmodel.cpp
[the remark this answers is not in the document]
achow101: We are trying to change code style gradually. Existing code may not adhere to the style guide, but modifications to it and newly added code must.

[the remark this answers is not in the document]
shaavan: > We are trying to change code style gradually. Existing code may not adhere to the style guide, but modifications to it and newly added code must.

Understood.

> Also, personally, I hate the bracket-less style so I'm not going to do this even if the style guide allows it. There is no requirement that such code needs to be written this way.

I get it. Bracket-less style have a chance of inducing unnecessary errors in case of adding some line inside the conditional statement.

src/qt/optionsmodel.h
```diff
@@ -116,6 +118,7 @@ class OptionsModel : public QAbstractListModel
     bool m_use_embedded_monospaced_font;
     bool fCoinControlFeatures;
     bool m_sub_fee_from_amount;
+    bool fEnablePSBTControls;
```
laanwj, line 121: Any specific reason for using the antiquated `fEnablePSBTControls` convention here, or was this an oversight? (I see you tried to use `m_` elsewhere)

src/qt/forms/optionsdialog.ui
```diff
@@ -252,6 +252,16 @@
...
+            <property name="text">
+             <string extracomment="An options window setting to enable advanced PSBT controls.">Enable advanced &amp;PSBT controls</string>
```
ryanofsky, line 258: In commit "qt: hide Create Unsigned button behind an expert mode option" (790733e54c3daff547acbc90fb179b48b6d2fb3c)

This is just a personal opinion, but I don't like the "advanced" weasel wording that makes the option more mysterious, implies that if you are an "advanced" user you should already know what this does without any explanation, and implies that if you don't have some outside knowledge what this does, then maybe you are some kind of simpleton.

Easy to fix this by replacing: "Enable advanced PBST controls" with "Enable PSBT controls". Or just literally "Enable create unsigned transaction feature when spending".

src/qt/optionsmodel.cpp
```diff
@@ -372,6 +377,8 @@ QVariant OptionsModel::data(const QModelIndex & index, int role) const
             return settings.value("fListen");
         case Server:
             return settings.value("server");
+        case EnablePSBTControls:
```
ryanofsky, line 380: In commit "qt: hide Create Unsigned button behind an expert mode option" (790733e54c3daff547acbc90fb179b48b6d2fb3c)

Suggest moving this case up near `case CoinControlFeatures` above, so this is grouped together with GUI options, not node options and it is easier ensure that all these options are handled consistently in the future. Same suggestion as well for `OptionsModel::setData`

src/qt/sendcoinsdialog.cpp
```diff
@@ -329,8 +329,17 @@ bool SendCoinsDialog::PrepareSendText(QString& question_string, QString& informa
     question_string.append(tr("Do you want to create this transaction?"));
     question_string.append("<br /><span style='font-size:10pt;'>");
     if (model->wallet().privateKeysDisabled() && !model->wallet().hasExternalSigner()) {
+        /*: Text to inform a user attempting to create a transaction of their current options. At this stage,
```
ryanofsky, line 332: In commit "qt: hide Create Unsigned button behind an expert mode option" (790733e54c3daff547acbc90fb179b48b6d2fb3c)

Minor: This changes seem like they should be part of the previous commit. The previous commit is adding the "create unsigned" button unconditionally, and the new commit is adding the option to hide the button. It seems like the instructions for using the button should be added in the same commit that adds the button.

This is nitpicking though, so ignore unless there's some need to change.

src/qt/sendcoinsdialog.cpp
```diff
@@ -397,7 +406,11 @@ void SendCoinsDialog::sendButtonClicked([[maybe_unused]] bool checked)
     const QString confirmation = tr("Confirm send coins");
     const QString confirmButtonText = tr("Send");
     const QString psbt_button_text = tr("Create Unsigned");
-    auto confirmationDialog = new SendConfirmationDialog(confirmation, question_string, informative_text, detailed_text, SEND_CONFIRM_DELAY, confirmButtonText, !model->wallet().privateKeysDisabled(), psbt_button_text, this);
+
+    // Show the Create Unsigned button if set in the options or if private keys are disabled.
+    bool show_unsigned_button = model->getOptionsModel()->getEnablePSBTControls() || model->wallet().privateKeysDisabled();
```
ryanofsky, line 411: In commit "qt: hide Create Unsigned button behind an expert mode option" (790733e54c3daff547acbc90fb179b48b6d2fb3c)

It seems clumsy for this show unsigned logic to have to be repeated two places (in spending and fee bumping) and for all these extra calls to privateKeysDisabled to be needed in the code. It would be clearer to replace the `show_unsigned` constructor argument with an `always_show_unsigned` constructor argument and pass `getEnablePSBTControls` directly instead of `getEnablePSBTControls || privateKeysDisabled`, since SendConfirmationDialog already knows the `privateKeysDisabled` value

Also it would make the PR clearer if instead of changing the `SendConfirmationDialog` gradually in two commits, you just implemented it fully in the first commit passing `true` for `always_show_unsigned`, and then simply replaced `true` with `getEnablePSBTControls` in the second commit.

ryanofsky (approved):
Code review ACK 790733e54c3daff547acbc90fb179b48b6d2fb3c. I guess I'm grumpy and left some complaints about the code and use of the word "advanced" in option description. But overall this change looks very good and makes the UI more consistent and understandable while adding a new feature. Please ignore any feedback below that is not helpful!

Would recommend adding release notes to this PR, and maybe including the nice screenshots from shaavan in the PR description.

src/qt/optionsmodel.h
achow101: Changed to `m_enable_psbt_controls`.

src/qt/forms/optionsdialog.ui
achow101: Removed "advanced"

src/qt/optionsmodel.cpp
achow101: Done

src/qt/optionsmodel.cpp
```diff
@@ -83,6 +83,11 @@ void OptionsModel::Init(bool resetSettings)
         settings.setValue("fCoinControlFeatures", false);
     fCoinControlFeatures = settings.value("fCoinControlFeatures", false).toBool();
 
+    if (!settings.contains("m_enable_psbt_controls")) {
+        settings.setValue("m_enable_psbt_controls", false);
+    }
+    m_enable_psbt_controls = settings.value("m_enable_msbt_controls", false).toBool();
```
ryanofsky, line 89: In commit "qt: hide Create Unsigned button behind an expert mode option" (9d8cf228a339101cc4956b32b39e377f5b5c21be)

Bug: spelling "msbt" should be "psbt".

Separate, minor suggestion: I'd also suggest dropping the "m_" prefix in the setting name (not member name). Other setting names don't have "m_" prefixes. Settings names are persisted externally and survive across releases, while member names are changeable implementation details. Also (this part is just my opinion), having redundant `OptionsModel` member variables like `m_enable_psbt_controls` and `fCoinControlFeatures` even exist is a mistake and a footgun to begin with. `OptionsModel` only needs member variables for specific settings  with complicated external interactions, like the pruning setting, not for plain settings that can be get and set with `data()` and `setData()` calls. I'm not suggesting getting rid of the members in this PR, just saying it would be strange later if the members were simplified away and some settings still had m_ prefixes.

ryanofsky (commented):
There is a small bug second commit (`msbt_controls` spelling) but than that 9d8cf228a339101cc4956b32b39e377f5b5c21be looks very good, and thanks for the updates!

src/qt/sendcoinsdialog.h
```diff
@@ -114,18 +114,21 @@ class SendConfirmationDialog : public QMessageBox
...
     int exec() override;
 
 private Q_SLOTS:
     void countDown();
-    void updateYesButton();
+    void updateButtons();
 
 private:
     QAbstractButton *yesButton;
+    QAbstractButton *m_psbt_button;
```
hebasto, line 126: style nit:
```suggestion
    QAbstractButton* m_psbt_button;
```

src/qt/sendcoinsdialog.h
```diff
@@ -114,18 +114,21 @@ class SendConfirmationDialog : public QMessageBox
...
     QTimer countDownTimer;
     int secDelay;
-    QString confirmButtonText;
+    QString confirmButtonText{tr("Send")};
+    bool m_enable_send;
+    QString m_psbt_button_text{tr("Create Unsigned")};
```
hebasto, line 131: nit:
```suggestion
    const QString confirmButtonText{tr("Send")};
    const bool m_enable_send;
    const QString m_psbt_button_text{tr("Create Unsigned")};
```

src/qt/walletmodel.cpp
```diff
@@ -523,7 +522,7 @@ bool WalletModel::bumpFee(uint256 hash, uint256& new_hash)
     }
 
     // Short-circuit if we are returning a bumped transaction PSBT to clipboard
-    if (create_psbt) {
+    if (retval == QMessageBox::Save) {
```
hebasto, line 525: As `m_psbt_button = button(QMessageBox::Save);` is not near, maybe add a comment which describes that this code actually handles "Create Unsigned" button press?

Is `// Short-circuit if we are returning a bumped transaction PSBT to clipboard` comment above still relevant?

PR #444, opened 2021-10-03
Network Watch tool

luke-jr:
Now that the crash bug (https://github.com/bitcoinknots/bitcoin/issues/4) has been tracked down, this seems ready for re-opening for Core.

(Originally https://github.com/bitcoin/bitcoin/pull/9849)

### DIFF
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -56,6 +56,7 @@ QT_MOC_CPP = \
   qt/moc_macdockiconhandler.cpp \
   qt/moc_macnotificationhandler.cpp \
   qt/moc_modaloverlay.cpp \
+  qt/moc_netwatch.cpp \
   qt/moc_notificator.cpp \
   qt/moc_openuridialog.cpp \
   qt/moc_optionsdialog.cpp \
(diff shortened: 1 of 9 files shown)


jarolrod:
Concept ACK

In general, I'm in favor of not having tons of pop-out windows. May be worth exploring this as a new tab. 

If possible, can you include screenshots of this in action in the OP?

luke-jr:
Not sure how to add inline images on GitHub, but there's a sceenshot on the wiki: https://en.bitcoin.it/wiki/File:Bitcoinknots-netwatch.png

luke-jr:
As for "pop out windows", I think that's the correct UX for features like this. Having a tab in the main window just doesn't seem to fit well.

Maybe a new debug-window-alike for statistics?

jarolrod (commented):
Have not yet reviewed the code, but: 

Commit 47711fdbc4b5d7f53aa778bcb1af34fc699ab81f must be broken off and opened in the core repo if it is needed. Then this can be based on the PR that would be open. The change to to `validationinterface` cannot be done in this repo.

Additionally, the commit history is not clean. The last 4 commits could be squashed onto ae6804e1a1e52611c7f98c6194a8384dda7abe60

rebroad:
@luke-jr  how about a screenshot?

rebroad:
concept NACK - at least, so far, I'm struggling to understand the use case for this. I'd have thought something more command line would make sense, as surely there'd be so much data that it would scroll by too quickly to be useful.

jonatack:
Concept ACK if I'm understanding from the earlier PR. Suggest describing what it does and the use case (what and why) in the pull description.

hebasto:

PR #447, opened 2021-10-05
Never disable HD status icon

stratospher:
[#8517](https://github.com/bitcoin/bitcoin/pull/8517) introduced the HD status icon in the bottom menu bar. At the time, it may have made sense to disable the HD status icon (which would display the icon at 50% opacity) when the wallet is not HD or watch-only. 
 
Since then, we have had many changes to our UI. Namely, the ability to switch between dark and light mode on macOS. Additionally, since version 0.16, the Bitcoin Core client only [generates HD wallets](https://github.com/bitcoin/bitcoin/issues/12547). Non-HD Wallets cannot be created anymore, only imported.
 
This PR proposes never to disable the HD status icon in the bottom menu bar. Instead, we always leave the HD status icon enabled for a better user interface. There’s no good reason for it to remain disabled, and it’s hard to see the watch-only icon in macOS dark mode. If a user has imported a non-HD wallet, we also want the icon to be as clear as possible in the bottom menu bar.

- on macOS

|                        |      Master      |         PR           |
| ------------- | ------------- | ------------- |
|  Dark Mode    | unnamed  | unnamed-2 |
|  Light Mode   |unnamed-3| unnamed-4|
- on ubuntu 20.04

|      Master      |         PR           |
| ------------- | ------------- |
|   Screenshot 2021-10-06 at 3 15 12 AM |  Screenshot 2021-10-06 at 3 18 20 AM|

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -1304,8 +1304,6 @@ void BitcoinGUI::setHDStatus(bool privkeyDisabled, int hdEnabled)
     labelWalletHDStatusIcon->setThemedPixmap(privkeyDisabled ? QStringLiteral(":/icons/eye") : hdEnabled ? QStringLiteral(":/icons/hd_enabled") : QStringLiteral(":/icons/hd_disabled"), STATUSBAR_ICONSIZE, STATUSBAR_ICONSIZE);
     labelWalletHDStatusIcon->setToolTip(privkeyDisabled ? tr("Private key <b>disabled</b>") : hdEnabled ? tr("HD key generation is <b>enabled</b>") : tr("HD key generation is <b>disabled</b>"));
     labelWalletHDStatusIcon->show();
-    // eventually disable the QLabel to set its opacity to 50%
-    labelWalletHDStatusIcon->setEnabled(hdEnabled);
 }
 
 void BitcoinGUI::setEncryptionStatus(int status)


jarolrod (commented):
ACK 35e814c1cfedf17e7751b730ae1e36099801be4c

Tested on macOS 12. I think this is nice & small UI improvement. It's a bit of an eyesore when using a watch-only wallet. One nice thing about this change is that the icon (in the cases of a watch-only and non-hd wallet) will be properly colorized when switching between light and dark mode on macOS. As shown in your screenshots with Ubuntu, it's a small UI improvement regardless of platform.

**Watch-Only Icon**

| master | pr |
|--------|---|
| Screen Shot 2021-10-05 at 10 35 10 PM | Screen Shot 2021-10-05 at 10 30 48 PM |
 


Since I don't have a non-HD wallet available to import, I looked at the visual changes this PR would have on the HD disabled icon with the following patch:

```diff
@@ -1301,7 +1301,7 @@ bool BitcoinGUI::handlePaymentRequest(const SendCoinsRecipient& recipient)
 
 void BitcoinGUI::setHDStatus(bool privkeyDisabled, int hdEnabled)
 {
-    labelWalletHDStatusIcon->setThemedPixmap(privkeyDisabled ? QStringLiteral(":/icons/eye") : hdEnabled ? QStringLiteral(":/icons/hd_enabled") : QStringLiteral(":/icons/hd_disabled"), STATUSBAR_ICONSIZE, STATUSBAR_ICONSIZE);
+    labelWalletHDStatusIcon->setThemedPixmap(QStringLiteral(":/icons/hd_disabled"), STATUSBAR_ICONSIZE, STATUSBAR_ICONSIZE);
     labelWalletHDStatusIcon->setToolTip(privkeyDisabled ? tr("Private key <b>disabled</b>") : hdEnabled ? tr("HD key generation is <b>enabled</b>") : tr("HD key generation is <b>disabled</b>"));
```

**HD disabled icon**
| master + patch | pr + patch |
| ----------------|-----------|
| Screen Shot 2021-10-05 at 10 46 00 PM | Screen Shot 2021-10-05 at 10 41 31 PM |

hebasto:
Screenshots in the OP seem misleading as they do not show the`labelWalletHDStatusIcon`.

> Additionally, since version 0.16, the Bitcoin Core client only [generates HD wallets](https://github.com/bitcoin/bitcoin/issues/12547). Non-HD Wallets cannot be created anymore, only imported.

It does not mean people stopped using non-HD wallets.

Concept ACK. The `labelWalletHDStatusIcon` is not clickable, therefore, disabling it makes no sense.

Is a "non-HD" label is more clear to read and understand than a crossed "HD" one?

hebasto:
> > Screenshots in the OP seem misleading as they do not show the`labelWalletHDStatusIcon`.
> 
> They certainly are showing the `labelWalletHDStatusIcon`, but in the included screenshots it is currently set to the watch-only icon.

Right.

luke-jr:
>As you might notice, the icons displayed in the last two scenarios in the PR have become precisely the same, which will probably introduce unnecessary confusion. And I don’t think inducing any kind of UX confusion, however small, is favorable.

Is there in fact a meaningful distinction? HD deals with how addresses are generated from private keys, but watch-only doesn't *have* private keys... If there's a reason to distinguish, please open an issue before this gets lost. ;)

luke-jr:
Watch-only *wallets* were introduced in 0.17, while the ability to create non-HD wallets was removed in 0.16, so even if it's possible to make a watch-only non-HD wallet (I don't think it is), it should be weird enough to not care about how it gets displayed in the GUI.

PR #448, opened 2021-10-07
Add helper to load font

promag:
Originally submitted as https://github.com/bitcoin-core/gui-qml/pull/49.

### DIFF
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -46,7 +46,6 @@
 
 #include <QApplication>
 #include <QDebug>
-#include <QFontDatabase>
 #include <QLatin1String>
 #include <QLibraryInfo>
 #include <QLocale>
@@ -492,7 +491,7 @@ int GuiMain(int argc, char* argv[])
 #endif
 
     BitcoinApplication app;
-    QFontDatabase::addApplicationFont(":/fonts/monospace");
+    GUIUtil::LoadFont(QStringLiteral(":/fonts/monospace"));
 
     /// 2. Parse command-line options. We do this after qt in order to show an error if there are problems parsing these
     // Command-line options take precedence:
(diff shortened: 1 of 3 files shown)


src/qt/guiutil.cpp
```diff
@@ -272,6 +272,12 @@ bool hasEntryData(const QAbstractItemView *view, int column, int role)
     return !selection.at(0).data(role).toString().isEmpty();
 }
 
+void loadFont(const QString& file_name)
```
hebasto, line 275: Why not return `id` for generality?

From Qt [docs](https://doc.qt.io/qt-5.12/qfontdatabase.html#addApplicationFont):
>  An ID is returned that can be used to remove the font again with `removeApplicationFont()` or to retrieve the list of family names contained in the font.

src/qt/guiutil.cpp
```diff
@@ -272,6 +272,12 @@ bool hasEntryData(const QAbstractItemView *view, int column, int role)
...
+{
+    int id = QFontDatabase::addApplicationFont(file_name);
```
hebasto, line 277:
```suggestion
    const int id = QFontDatabase::addApplicationFont(file_name);
```
?

src/qt/bitcoin.cpp
```diff
@@ -492,7 +491,7 @@ int GuiMain(int argc, char* argv[])
 #endif
 
     BitcoinApplication app;
-    QFontDatabase::addApplicationFont(":/fonts/monospace");
+    GUIUtil::loadFont(":/fonts/monospace");
```
hebasto, line 494:
```suggestion
    GUIUtil::loadFont(QStringLiteral(":/fonts/monospace"));
```

luke-jr:
Why do we want to abort if it fails??

PR #449, opened 2021-10-07
Restore "S" accelerator for "Start on system login" option

luke-jr:
bitcoin-core/gui#416 changed the option assigned to accelerator key "S", but there's no rationale given.

Best to leave it alone, and give the new option a new accelerator key.

Since "R" is already taken for Reset, this shifts the new RPC server option to use "P" instead

### DIFF
--- a/src/qt/forms/optionsdialog.ui
+++ b/src/qt/forms/optionsdialog.ui
@@ -33,7 +33,7 @@
           <string>Automatically start %1 after logging in to the system.</string>
          </property>
          <property name="text">
-          <string>Start %1 on system &amp;login</string>
+          <string>&amp;Start %1 on system login</string>
          </property>
         </widget>
        </item>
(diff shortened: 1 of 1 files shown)


shaavan (commented):
No strong opinion for now
 
The purpose of the mnemonic shortcut is to increase the speed of workflow. For that, they need to be such that they are easily recognizable and easy to remember.
So for the RPC server setting, the most natural shortcut would be the letter R. But since that is already taken, we must use another letter. According to me, the second-best choice is to use the first letter of the second word (i.e., **s**) instead of the second letter of the first word (i.e., **P**).

The reason why I think so is:
1. The corresponding command line setting for this option is **-server**, which makes the shortcut **s** quickly form an association in our mind with this setting, making the shortcut easy to remember.
2. The second reason being how our brain remembers and correlates things. I am not proficient in its science, but as far as I know, our brain tends to associate a word with its first letter rather than the latter. This is very similar to how we can just quickly recall a complete song in our head if someone tells the few starting words and rhythm of the song, whereas we struggle to recall if someone starts the music from the middle. 
From this perspective, the shortcut **P** does not have an as strong association with the word **RPC** in our mind as the shortcut **s** has with the word **server**.

These were just my two cents. I might be wrong in some of my reasoning. In that case, please correct me. But for now, I think the shortcuts are good as they currently are.

hebasto (approved):
ACK 25a581419d10b3c7d99789da18afd51f2984fbc8, tested on Linux Mint 20.2 (Qt 5.12.8).

PR #450, opened 2021-10-09
Prefix makefile variables with QT_

promag:
Improves consistency readability if future QML variables are added.


promag:
Moved to https://github.com/bitcoin/bitcoin/pull/23242 as requested by @hebasto.

PR #457, opened 2021-10-25
Add abitlity to resize peerTable and banTable

shaavan:
Till now, peerTable and banTable under the peerTab were fixed in height and cannot be resized. This creates unnecessary inconvenience, especially when you have many items on one table and not many on another.

This PR proposes to add a splitter between these two tables, allowing them to be resized freely.

|Master|PR|
|---------|----|
|Screenshot from 2021-10-25 21-26-09|Screenshot from 2021-11-01 17-43-01|

### DIFF
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -889,101 +889,110 @@
           </property>
           <layout class="QVBoxLayout" name="verticalLayout_9">
            <item>
-            <widget class="QTableView" name="peerWidget">
-             <property name="sizePolicy">
-              <sizepolicy hsizetype="Expanding" vsizetype="MinimumExpanding">
-               <horstretch>0</horstretch>
-               <verstretch>1</verstretch>
-              </sizepolicy>
+            <widget class="QSplitter" name="tableSplitter">
+             <property name="orientation">
+              <enum>Qt::Vertical</enum>
              </property>
-             <property name="tabKeyNavigation">
+             <property name="handleWidth">
+              <number>4</number>
+             </property>
+             <property name="childrenCollapsible">
               <bool>false</bool>
              </property>
-             <property name="alternatingRowColors">
-              <bool>true</bool>
-             </property>
-             <property name="textElideMode">
-              <enum>Qt::ElideMiddle</enum>
-             </property>
-             <property name="sortingEnabled">
-              <bool>true</bool>
-             </property>
-             <property name="wordWrap">
-              <bool>false</bool>
-             </property>
-             <attribute name="horizontalHeaderHighlightSections">
-              <bool>false</bool>
-             </attribute>
-            </widget>
-           </item>
-           <item>
-            <widget class="QWidget" name="banTable" native="true">
-             <property name="sizePolicy">
-              <sizepolicy hsizetype="Preferred" vsizetype="Minimum">
-               <horstretch>0</horstretch>
-               <verstretch>0</verstretch>
-              </sizepolicy>
-             </property>
-             <layout class="QVBoxLayout" name="verticalLayout_7">
-              <item>
-               <widget class="QLabel" name="banHeading">
-                <property name="sizePolicy">
-                 <sizepolicy hsizetype="Preferred" vsizetype="Minimum">
-                  <horstretch>0</horstretch>
-                  <verstretch>0</verstretch>
-                 </sizepolicy>
-                </property>
-                <property name="minimumSize">
-                 <size>
-                  <width>0</width>
-                  <height>32</height>
-                 </size>
-                </property>
-                <property name="maximumSize">
-                 <size>
-                  <width>16777215</width>
-                  <height>32</height>
-                 </size>
-                </property>
-                <property name="font">
-                 <font>
-                  <pointsize>12</pointsize>
-                 </font>
-                </property>
-                <property name="cursor">
-                 <cursorShape>IBeamCursor</cursorShape>
-                </property>
-                <property name="text">
-                 <string>Banned peers</string>
-                </property>
-                <property name="alignment">
-                 <set>Qt::AlignBottom|Qt::AlignLeading|Qt::AlignLeft</set>
-                </property>
-                <property name="wordWrap">
-                 <bool>true</bool>
-                </property>
-                <property name="textInteractionFlags">
-                 <set>Qt::NoTextInteraction</set>
-                </property>
-               </widget>
-              </item>
-              <item>
-               <widget class="QTableView" name="banlistWidget">
-                <property name="tabKeyNavigation">
-                 <bool>false</bool>
-                </property>
-                <property name="alternatingRowColors">
-                 <bool>true</bool>
-                </property>
-                <property name="sortingEnabled">
-                 <bool>true</bool>
-                </property>
-                <attribute name="horizontalHeaderHighlightSections">
-                 <bool>false</bool>
-                </attribute>
-               </widget>
-              </item>
-             </layout>
+             <widget class="QTableView" name="peerWidget">
+              <property name="sizePolicy">
+               <sizepolicy hsizetype="Expanding" vsizetype="MinimumExpanding">
+                <horstretch>0</horstretch>
+                <verstretch>1</verstretch>
+               </sizepolicy>
+              </property>
+              <property name="tabKeyNavigation">
+               <bool>false</bool>
+              </property>
+              <property name="alternatingRowColors">
+               <bool>true</bool>
+              </property>
+              <property name="textElideMode">
+               <enum>Qt::ElideMiddle</enum>
+              </property>
+              <property name="sortingEnabled">
+               <bool>true</bool>
+              </property>
+              <property name="wordWrap">
+               <bool>false</bool>
+              </property>
+              <attribute name="horizontalHeaderHighlightSections">
+               <bool>false</bool>
+              </attribute>
+             </widget>
+             <widget class="QWidget" name="banTable" native="true">
+              <property name="sizePolicy">
+               <sizepolicy hsizetype="Preferred" vsizetype="Minimum">
+                <horstretch>0</horstretch>
+                <verstretch>0</verstretch>
+               </sizepolicy>
+              </property>
+              <layout class="QVBoxLayout" name="verticalLayout_7">
+               <item>
+                <widget class="QLabel" name="banHeading">
+                 <property name="sizePolicy">
+                  <sizepolicy hsizetype="Preferred" vsizetype="Minimum">
+                   <horstretch>0</horstretch>
+                   <verstretch>0</verstretch>
+                  </sizepolicy>
+                 </property>
+                 <property name="minimumSize">
+                  <size>
+                   <width>0</width>
+                   <height>32</height>
+                  </size>
+                 </property>
+                 <property name="maximumSize">
+                  <size>
+                   <width>16777215</width>
+                   <height>32</height>
+                  </size>
+                 </property>
+                 <property name="font">
+                  <font>
+                   <pointsize>12</pointsize>
+                  </font>
+                 </property>
+                 <property name="cursor">
+                  <cursorShape>IBeamCursor</cursorShape>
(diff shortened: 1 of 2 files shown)


src/qt/forms/debugwindow.ui
```diff
@@ -887,85 +887,113 @@
...
              </property>
-             <attribute name="horizontalHeaderHighlightSections">
-              <bool>false</bool>
-             </attribute>
+             <widget class="QWidget" name="peerTable" native="true">
+              <layout class="QVBoxLayout" name="verticalLayout_9">
+               <item>
+                <widget class="QLabel" name="peerHeading_2">
+                 <property name="text">
+                  <string>Connected peers</string>
```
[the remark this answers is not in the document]
shaavan: > This shouldn't display until there is a Banned Peers table
Additionally, I shouldn't be able to hide any of the tables

These are valid suggestions. Let me take a look into implementing these.

>I don't think this should be a part of this PR considering the title suggests you want to just add functionality to resize tabled

Ok then, I shall remove the name change from this PR and may open it as another pull request in the future (if need be)

shaavan:
Updated 5e54cfe to c09e6ad ([pr457.01](https://github.com/shaavan/gui/tree/pr457.01) -> [pr457.02](https://github.com/shaavan/gui/tree/pr457.02))
Addressed @jarolrod comments

Updates:
1. Considering that naming peerTable was not gathering a lot of support. I have removed that change from this PR.
2. Made the peerTable and banTable non-collapsible.
3. Increased splitter’s height from `3px` to `4px`.

> This shouldn't display until there is a Banned Peers table

@jarolrod. As you rightly mentioned, showing splitter when no BanTable was looking kind of off. But in the new iteration of this PR, the splitter seems like an element of peerTable and looks aesthetically pleasing even when there is no banTable. I have updated the PR description with the new Screenshots. You can look at them to better get the feel of what I am trying to say.

jarolrod (commented):
Getting there, but the splitter should behave like the splitter that currently exists between the `Peers Table` and the `Peer Information Pane`.

When I select a Peer, I get the `Peer Information Pane`, as well as a splitter where I can resize as needed. When the peer is de-selected, the splitter disappears; because there is nothing to resize.

Similarly, until I have a banned peer and a banned peer table, the splitter being introduced here should not be displayed at all because there is nothing to resize. Additionally, if i unban a peer and subsequently no longer have a banned peers table, the splitter should disappear as well.

shaavan:
Updated from c09e6ad39da24a7a969533f6ce5c9cea14e8c9ce to aa6229e5d2bebe20d3e8e3eded8ff3ad7841bfcb ([pr457.02](https://github.com/shaavan/gui/tree/pr457.02) -> [pr457.03](https://github.com/shaavan/gui/tree/pr457.03), [diff](https://github.com/shaavan/gui/compare/pr457.02..pr457.03))
Addressed @jarolrod comments

**Changes:**

- Instead of hiding individual children of banTable. The code is now updated to hide the banTable widget. This will make sure the splitter will not be displayed when there is no banTable.
- Updated sizePolicy to better reflect the peerWidget and the banTable widget.
- Added code to save and restore the state (position) of the splitter. The state is remembered even in the case when GUI is shut down and restarted.

shaavan:
cc @jarolrod. PR ready for another review!

src/qt/forms/debugwindow.ui
```diff
@@ -6,13 +6,16 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>740</width>
-    <height>430</height>
+    <width>744</width>
+    <height>531</height>
```
kristapsk, line 10: Why this change?

shaavan: I am not sure of the reason. But I observed they are set to these values (744, 531) each time debugwindow.ui is opened in QTCreator even if we try to change them. However, this does not cause any functional change.

kristapsk: Ok, then seems reasonable to change to these values.

kristapsk (approved):
ACK aa6229e5d2bebe20d3e8e3eded8ff3ad7841bfcb (tested under Linux / Xfce4)

src/qt/forms/debugwindow.ui
```diff
@@ -887,85 +890,112 @@
...
-              <set>Qt::AlignBottom|Qt::AlignLeading|Qt::AlignLeft</set>
-             </property>
-             <property name="wordWrap">
-              <bool>true</bool>
-             </property>
-             <property name="textInteractionFlags">
-              <set>Qt::NoTextInteraction</set>
+            <widget class="QSplitter" name="tableSplitter">
+             <property name="styleSheet">
+              <string notr="true">QSplitter::handle{background: #a6a6a6;}</string>
```
promag, line 897: Why this style?

promag (commented):
Concept ACK.

src/qt/forms/debugwindow.ui
shaavan: I did this to make the splitter visible and looks aesthetically pleasing.
Do you recommend keeping the splitter invisible?

w0xlt (approved):
tACK aa6229e . Tested on Ubuntu 21.10.

src/qt/forms/debugwindow.ui
shaavan: What do you think about this? @promag

hebasto:
Concept ACK.

src/qt/forms/debugwindow.ui
hebasto: > Do you recommend keeping the splitter invisible?

I don't think the hardcoded color will work for any platform and appearance (dark/light), no?

shaavan: > I don't think the hardcoded color will work for any platform and appearance (dark/light), no?

The spiller looks fine for the default light mode (in Linux) and dark mode in macOS, which can be seen in the screenshots of the following comments: [light](https://github.com/bitcoin-core/gui/pull/457#issue-1035432748); [dark](https://github.com/bitcoin-core/gui/pull/457#pullrequestreview-791298779).

However, this doesn't guarantee that this will work for any custom appearance. Do you recommend removing the hardcoded color and making the splitter invisible again?

src/qt/forms/debugwindow.ui
RandyMcMillan: We manually edit these values to remove this diff junk - this way it stays consistent.

RandyMcMillan: Including - resetting the currentIndex to zero (below)

src/qt/forms/debugwindow.ui
hebasto: Why not keeping the same look as for horizontal `splitter`?

> Do you recommend keeping the splitter invisible?

What does make you think so?

src/qt/rpcconsole.cpp
```diff
@@ -750,6 +754,9 @@ void RPCConsole::setClientModel(ClientModel *model, int bestblock_height, int64_
         ui->startupTime->setText(model->formatClientStartupTime());
         ui->networkName->setText(QString::fromStdString(Params().NetworkIDString()));
 
+        //Setup Peer and Ban Table Splitter
+        ui->tableSplitter->setChildrenCollapsible(false);
+
```
hebasto, line 759: This code could be a part of the `debugwindow.ui` file as it is for the horizontal `splitter`.

Qt Designer tool could very helpful for such tasks.

src/qt/forms/debugwindow.ui
```diff
@@ -36,7 +39,7 @@
    <item>
     <widget class="QTabWidget" name="tabWidget">
      <property name="currentIndex">
-      <number>0</number>
+      <number>3</number>
```
hebasto, line 42: Why this change?

src/qt/forms/debugwindow.ui
shaavan: > Do you recommend keeping the splitter invisible?

By this, I meant to say, keeping it the same as the horizontal splitter.

> Why not keep the same look as for horizontal splitter?

I think this will be a more optimal approach to go with.

hebasto: But the horizontal splitter is not invisible :)

And having different looks for similar GUI elements does not as an optimal design for me.

hebasto: At least, introducing a new GUI element, and changing thuis GUI element style across the GUI are two different tasks.

shaavan:
Updated from aa6229e5d2bebe20d3e8e3eded8ff3ad7841bfcb to 0c364201a2e16355cd6424c6137761c7967db4e0([pr457.04](https://github.com/shaavan/gui/tree/pr457.04) -> [pr457.05](https://github.com/shaavan/gui/tree/pr457.05))
Addressed @hebasto and @randymacmilan comments

- Removed custom color of the added splitter
- Moved `childrenCollapsible=False` to debugwindow.ui file.
- Removed auto-generated changes by QtCreator.

src/qt/forms/debugwindow.ui
shaavan: This was an auto-generated change by QtCreator. Fixed in 0c364201a2e16355cd6424c6137761c7967db4e0

src/qt/rpcconsole.cpp
shaavan: Addressed in 0c364201a2e16355cd6424c6137761c7967db4e0

src/qt/forms/debugwindow.ui
shaavan: Fixed splitter color to default palette in 0c364201a2e16355cd6424c6137761c7967db4e0

src/qt/forms/debugwindow.ui
shaavan: Resetted these changes to the master's value in 0c364201a2e16355cd6424c6137761c7967db4e0

src/qt/forms/debugwindow.ui
```diff
@@ -887,85 +887,112 @@
             <height>0</height>
            </size>
           </property>
-          <layout class="QVBoxLayout" name="verticalLayout_7">
+          <layout class="QVBoxLayout" name="verticalLayout_9">
```
hebasto, line 890: Is this change required?

hebasto (commented):
Approach ACK 0c364201a2e16355cd6424c6137761c7967db4e0.

Could you please separate a refactoring commit which introduces the `banTable` widget?

src/qt/forms/debugwindow.ui
shaavan: Since this PR introduced a new sub-Layout, the introduced sub-Layout was named in the chronological order of its introduction. And since **verticalLayout_7** already exists (on Line 935), giving this the same name leads to a warning message while compiling:

```
qt/forms/debugwindow.ui: Warning: The name 'verticalLayout_7' (QVBoxLayout) is already in use, defaulting to 'verticalLayout_71'.
```

So, I am letting this change be as it is for now, and I welcome any suggestions you might have.

shaavan:
Updated from 0c364201a2e16355cd6424c6137761c7967db4e0 to 5a9bdce0ad753919522b8bc627c1c12d92d7cae8 ([pr457.05](https://github.com/shaavan/gui/commits/pr457.05) -> [pr457.06](https://github.com/shaavan/gui/commits/pr457.06), [diff](https://github.com/shaavan/gui/compare/pr457.05..pr457.06))

Addressed @hebasto comment

Changes:
- Rebased PR over the current master.
- Separated adding of **banTable** and **tableSplitter** in separate commits.
- Removed unnecessary `width`/`height` changes.

jarolrod (commented):
While the resizing works, the UI here does not. The UI element representing the splitter is too small and is pinned to the bottom of `peerTable`.

| linux | macOS |
| ----- | ------- |
Screenshot from 2022-08-06 18-41-26 | Screen Shot 2022-08-06 at 6 49 31 PM |

hebasto:

PR #459, opened 2021-10-29
Address type dropdown, add Taproot (Receive tab)

Sjors:
This PR replaces the Bech32 checkbox with an address type dropdown It "Use Taproot" checkbox to the receive screen for any wallet with a Taproot descriptor. The Bech32m option is hidden for wallets that don't have taproot descriptors.

Bech32 is kept as the default even for Taproot enabled wallets until it's more widely supported.

(an earlier attempt of this PR added a second checkbox)

Schermafbeelding 2021-12-21 om 11 44 05

**Suggested testing**
* notice that the Bech32m entry only appears for wallet with a taproot descriptor
* verify that it generates a bc1p instead of bc1q address
1. Legacy wallet
2. Default descriptor wallet (current does not have taproot descriptor)
3. Wallet with taproot descriptor:
   * just create a new descriptor wallet

Replaces https://github.com/bitcoin/bitcoin/pull/22260

### DIFF
--- a/doc/release-notes-gui-459.md
+++ b/doc/release-notes-gui-459.md
@@ -1,0 +1,4 @@
+GUI changes
+-----------
+
+- The Bech32 checkbox has been replaced with a dropdown for all address types, including the new Bech32m (BIP-350) standard for Taproot enabled wallets.
--- a/src/qt/receivecoinsdialog.cpp
+++ b/src/qt/receivecoinsdialog.cpp
@@ -87,10 +87,18 @@ void ReceiveCoinsDialog::setModel(WalletModel *_model)
             &QItemSelectionModel::selectionChanged, this,
             &ReceiveCoinsDialog::recentRequestsView_selectionChanged);
 
-        if (model->wallet().getDefaultAddressType() == OutputType::BECH32) {
-            ui->useBech32->setCheckState(Qt::Checked);
-        } else {
-            ui->useBech32->setCheckState(Qt::Unchecked);
+        // Populate address type dropdown and select default
+        auto add_address_type = [&](OutputType type, const QString& text, const QString& tooltip) {
+            const auto index = ui->addressType->count();
+            ui->addressType->addItem(text, (int) type);
+            ui->addressType->setItemData(index, tooltip, Qt::ToolTipRole);
+            if (model->wallet().getDefaultAddressType() == type) ui->addressType->setCurrentIndex(index);
+        };
+        add_address_type(OutputType::LEGACY, "Base58 (Legacy)", "Not recommended due to higher fees and less protection against typos.");
+        add_address_type(OutputType::P2SH_SEGWIT, "Base58 (P2SH-SegWit)", "Generates an address compatible with older wallets.");
+        add_address_type(OutputType::BECH32, "Bech32 (SegWit)", "Generates a native segwit address (BIP-173). Some old wallets don't support it.");
+        if (model->wallet().taprootEnabled()) {
+            add_address_type(OutputType::BECH32M, "Bech32m (Taproot)", "Bech32m (BIP-350) is an upgrade to Bech32, wallet support is still limited.");
         }
 
         // Set the button to be enabled or disabled based on whether the wallet can give out new addresses.
@@ -144,15 +152,7 @@ void ReceiveCoinsDialog::on_receiveButton_clicked()
     QString address;
     QString label = ui->reqLabel->text();
     /* Generate new receiving address */
-    OutputType address_type;
-    if (ui->useBech32->isChecked()) {
-        address_type = OutputType::BECH32;
-    } else {
-        address_type = model->wallet().getDefaultAddressType();
-        if (address_type == OutputType::BECH32) {
-            address_type = OutputType::P2SH_SEGWIT;
-        }
-    }
+    const OutputType address_type = (OutputType)ui->addressType->currentData().toInt();
     address = model->getAddressTableModel()->addRow(AddressTableModel::Receive, label, "", address_type);
 
     switch(model->getAddressTableModel()->getEditStatus())
--- a/src/wallet/interfaces.cpp
+++ b/src/wallet/interfaces.cpp
@@ -461,6 +461,11 @@ public:
     bool canGetAddresses() override { return m_wallet->CanGetAddresses(); }
     bool hasExternalSigner() override { return m_wallet->IsWalletFlagSet(WALLET_FLAG_EXTERNAL_SIGNER); }
     bool privateKeysDisabled() override { return m_wallet->IsWalletFlagSet(WALLET_FLAG_DISABLE_PRIVATE_KEYS); }
+    bool taprootEnabled() override {
+        if (m_wallet->IsLegacy()) return false;
+        auto spk_man = m_wallet->GetScriptPubKeyMan(OutputType::BECH32M, /*internal=*/false);
+        return spk_man != nullptr;
+    }
     OutputType getDefaultAddressType() override { return m_wallet->m_default_address_type; }
     CAmount getDefaultMaxTxFee() override { return m_wallet->m_default_max_tx_fee; }
     void remove() override
--- a/src/interfaces/wallet.h
+++ b/src/interfaces/wallet.h
@@ -256,6 +256,9 @@ public:
     // Return whether private keys enabled.
     virtual bool privateKeysDisabled() = 0;
 
+    // Return whether the wallet contains a Taproot scriptPubKeyMan
+    virtual bool taprootEnabled() = 0;
+
     // Return whether wallet uses an external signer.
     virtual bool hasExternalSigner() = 0;
 
--- a/src/qt/forms/receivecoinsdialog.ui
+++ b/src/qt/forms/receivecoinsdialog.ui
@@ -195,7 +195,7 @@
            </widget>
           </item>
           <item>
-           <widget class="QCheckBox" name="useBech32">
+           <widget class="QComboBox" name="addressType">
             <property name="sizePolicy">
              <sizepolicy hsizetype="Fixed" vsizetype="Fixed">
               <horstretch>0</horstretch>
@@ -210,12 +210,6 @@
             </property>
             <property name="focusPolicy">
              <enum>Qt::StrongFocus</enum>
-            </property>
-            <property name="toolTip">
-             <string>Native segwit addresses (aka Bech32 or BIP-173) reduce your transaction fees later on and offer better protection against typos, but old wallets don't support them. When unchecked, an address compatible with older wallets will be created instead.</string>
-            </property>
-            <property name="text">
-             <string>Generate native segwit (Bech32) address</string>
             </property>
            </widget>
           </item>
@@ -366,7 +360,7 @@
  <tabstops>
   <tabstop>reqLabel</tabstop>
   <tabstop>reqAmount</tabstop>
-  <tabstop>useBech32</tabstop>
+  <tabstop>addressType</tabstop>
   <tabstop>reqMessage</tabstop>
   <tabstop>receiveButton</tabstop>
   <tabstop>clearButton</tabstop>


Sjors:
They're not mutually exclusive. I'm intentionally trying not to bother the user with the distinction between bech32 and bech32m (except in the tooltip). The bech32 button covers both. The new button is about using Taproot, as opposed to v0 SegWit. If you uncheck bech32 it will disable the Taproot checkbox, because there is no p2sh version of taproot.

luke-jr:
IMO having 2 checkboxes like this is even more confusing.

Sjors:
We show all address types: legacy, p2sh-segwit, SegWit v0 and Taproot. Perhaps the least confusing way to do that would be a dropdown that defaults to SegWit v0.

prayank23:
> We could show all address types: legacy, p2sh-segwit, SegWit v0 and Taproot. Perhaps the least confusing way to do that would be a dropdown that defaults to SegWit v0.

This makes sense

sipa:
I think it's wrong to have it list "taproot" explicitly (or witness v0, for that matter); in descriptor wallets that's an implementation aspect the user shouldn't care about. What they care is legacy(base58) vs. bech32 vs bech32m, as those correspond to what senders supports. If a witness v2 softfork were to occur, hopefully bech32m can be reused unmodified for its addresses.

src/wallet/interfaces.cpp
```diff
@@ -458,6 +458,11 @@ class WalletImpl : public Wallet
     bool canGetAddresses() override { return m_wallet->CanGetAddresses(); }
     bool hasExternalSigner() override { return m_wallet->IsWalletFlagSet(WALLET_FLAG_EXTERNAL_SIGNER); }
     bool privateKeysDisabled() override { return m_wallet->IsWalletFlagSet(WALLET_FLAG_DISABLE_PRIVATE_KEYS); }
+    bool taprootEnabled() override {
+        if (m_wallet->IsLegacy()) return false;
```
MarcoFalke, line 462: Is this special case needed? Wouldn't it return nullptr below?

Sjors: Maybe not, but since legacy wallets don't have descriptors, it does seem cleaner.

MarcoFalke:
> FWIW, I think we should get rid of the LEGACY/P2SH_SEGWIT OutputTypes, and merge them into a BASE58 one or so.

In that case we should also never ever send change to LEGACY again, unless manually done so by the user?

luke-jr:
Then how do users continue avoiding Segwit? (Perhaps there's no reason to avoid Taproot, but there isn't a Base58 option for that, and it requires a new wallet)

Sjors:
Based on the above discussion, I'm starting to like my own approach more :-)

In the long run, e.g. by the time there's a SegWit v2 softfork, I'm hoping the distinction between bech32 and bech32m is no longer relevant because the ecosystem supports both. But my guess is that we'll have a transitional period between when that soft-fork activates and when we feel comfortable using it by default.

On the way to that future we have this Taproot checkbox, presenting it as a new feature. In a later release we make it the default.

Then, unless there is strong opposition to using Taproot, we drop the checkbox altogether. At that point the interface is back where it was before the PR.  "bech32" would then mean "bech32m" and we could even rename it. Unchecking it would produce a p2sh-segwit address which should only be needed when dealing with dinosaur exchanges.

The story then repeats for a future SegWit v2 soft fork: we first add a check-box to opt-in, make it the default and then later drop the checkbox.

sipa:
@luke-jr I just think the choice of what script construction is used should be a wallet-level configuration option, not something selected at address creation time. In descriptor wallets the answer would be: if you want P2PKH, you use a `pkh` descriptor for the base58 address type. For legacy wallets it could be a config option or so.

ajtowns:
@sipa That sounds to me like there should be some extra wallet-level configuration option that says "for base58 addresses, I know I can sign for both `pkh(X)` and `sh(wpkh(X))` for both receiving(non-internal) and change(internal), but I prefer `pkh(X)`". eg if you upgrade to segwit, receive some funds via `sh(wpkh(X))` then decide you don't like segwit and want to keep using `pkh(X)` by preference, not having that option would mean you'd have to move the `sh(wpkh(X))` descriptor to a different wallet, which would probably be pretty annoying. And I think you'd get similar scenarios if you come up with new descriptors (eg switching between 4/5 and 3/5 to sign, or a musig threshold vs a checkmultisig threshold) then decided you didn't really like them.

@Sjors Having an "address type" drop down of:

 * base58 (legacy)
 * bech32 (segwit)
 * bech32m (taproot)

(matching `-addresstype`) seems like it would be better to me? Implying that taproot uses bech32 is a bit misleading and seems likely to cause confusion even with a tool tip...

With the "I prefer descriptor X for bech32 addresses" option above, you don't need to do anything when "segwit v2" comes along -- that just adds a new descriptor that you won't generate addresses for until you update your config to say you prefer it.

(I don't think the gui really needs a "use this specific descriptor to generate this address"; if you want to be that fine-grained, use the deriveaddresses RPC?)

Maybe also upgrade `-addresstype` to be something settable in the gui as well?

achow101:
I would also prefer that this use a drop down rather than adding more boolean options.

Rspigler:
tACK

1.  v22.0 Legacy:  Address option is Native Segwit/Bech32 checkbox (defaults to enabled). p2sh-segwit address generated if unchecked.

2.  v22.0 Descriptor Wallet.  Same options/addresses, but with external/internal pkh, sh-wpkh, and wpkh descriptors.

3.  PR https://github.com/bitcoin/bitcoin/pull/22364:  Again same options, but generates external/internal pkh, sh-wpkh, wpkh, and now tr descriptors (if descriptor wallet created).

4.  PR #459  + https://github.com/bitcoin/bitcoin/pull/22364:  Checkboxes for Native Segwit/Bech32 and Taproot (as pictured in OP).  When both enabled, generates bech32m address from tr descriptor.  With just Native Segwit/Bech32 enabled, bech32 address generated from wpkh descriptor, as expected.  Unchecking Native Segwit/Bech32 disables 'Use Taproot', and generates p2sh-segwit address from sh-wpkh descriptor.


I also agree that a dropdown would be better. While Taproot is a segwit version, I don't think the end user needs to be burdened with this fact by how it is currently set up. I like @ajtowns suggestion for Address_Format (Soft Fork Name). But maybe it'd be better as:

base58 (Segwit)
bech32 (Native Segwit)
bech32m (Taproot)

Unless you wanted to include a legacy as well as p2sh-segwit address option.  

This labeling format also allows the distinction @Sjors was looking for in a future Segwit v2 softfork, if bech32m format is reused:

...
bech32m (taproot)
bech32m (future_fork)


If @sipa wants this to be a wallet level config option, maybe the user's preference can be set there for the default, but then  can still be changed at address creation time if necessary.

Rspigler:
Should we include a legacy address option if it has already been deprecated in the GUI?

luke-jr:
It's curious that people are treating "base58" as if it's a single address type. The difference between single-key base58 and P2SH is similar to the difference between Bech32 and Bech32m ;)

sipa:
> It's curious that people are treating "base58" as if it's a single address type. The difference between single-key base58 and P2SH is similar to the difference between Bech32 and Bech32m ;)

You're right; they're different formats for the purpose of senders. What unites them is the fact that both have been around for nearly 10 years.

kristapsk (approved):
ACK 4c87ff08632a89437d97f8c0546bc1997cfbf435 (tested on signet with both descriptors and legacy wallets).

Sjors:
(it's now a dropdown)

I think the CI failures are spurious.

Rmf77:
I claim my funds

Rmf77 (approved):

src/qt/receivecoinsdialog.cpp
```diff
@@ -87,11 +87,18 @@ void ReceiveCoinsDialog::setModel(WalletModel *_model)
             &QItemSelectionModel::selectionChanged, this,
             &ReceiveCoinsDialog::recentRequestsView_selectionChanged);
 
-        if (model->wallet().getDefaultAddressType() == OutputType::BECH32) {
-            ui->useBech32->setCheckState(Qt::Checked);
-        } else {
-            ui->useBech32->setCheckState(Qt::Unchecked);
+        // Populate address type dropdown and select default
+        ui->addressType->insertItem((int)OutputType::LEGACY, "Base58 (Legacy)");
+        ui->addressType->setItemData((int)OutputType::LEGACY, "Not recommended due to higher fees and less protection against typos.", Qt::ToolTipRole);
```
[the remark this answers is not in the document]
Sjors: It becomes a problem when the user sets `-addresstype=legacy`, because then nothing would be selected by default. Previously the checkbox would toggle between legacy and bech32 in that case.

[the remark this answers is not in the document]
MarcoFalke: When unchecked, `-addresstype=legacy` would map to `p2sh-witness` in the GUI, not `legacy`. So I think we should preserve that, no?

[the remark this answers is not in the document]
kristapsk: I think Bitcoin Core users are already above average Bitcoin users and mostly know what they are doing, don't think we should limit people from creating legacy P2PKH addresses, if they want to.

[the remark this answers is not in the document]
MarcoFalke: No one has been using this for the past 4 years, and I am not aware of anyone complaining about the lack of it? So it seems odd to add an option that has no use case.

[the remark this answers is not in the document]
kristapsk: I have used this by using RPC. Would have used in GUI if it would have been available.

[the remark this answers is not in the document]
Sjors: Correction, the checkbox is not disabled on master when using  `-addresstype=legacy`. The original PR description was wrong (or the behavior was changed later).

On master you can test that, when launched with  `-addresstype=legacy`, it would produce a P2PKH address, not wrapped segwit. This happened in the code below:

Schermafbeelding 2021-12-10 om 19 45 45

MarcoFalke (approved):
ACK

kristapsk (approved):
re-ACK f5ba3d9394e47fa737429ecbb43a91bb767e8c59

doc/release-notes-gui-459.md
```diff
@@ -0,0 +1,4 @@
+GUI changes
+-----------
+
+- The Bech32 checkbox has been replaced with a dropdown for all address types, including the new Bech32m (BIP-350) standard for Taproot enabled wallets. The default remains Bech32 (SegWit).
```
MarcoFalke, line 4:
```suggestion
- The Bech32 checkbox has been replaced with a dropdown for all address types, including the new Bech32m (BIP-350) standard for Taproot enabled wallets.
```

nit: I don't think we need to mention stuff that doesn't change. Also, this is wrong? The default is whatever `-addresstype` is set to?

Sjors: Fixed

promag (commented):
Tested ACK 79580c09ff9d3c9c845e02fbcf98f8187e7e9fea.

My only comment is that combo box entries are mapped to `OutputType` values - meaning we can't reorder. The following commit uses `UserRole` to store the type instead of combo box index https://github.com/promag/gui/commit/b164039aa2b6462b6b1eac6bc79b40cd211b41b4 (https://github.com/promag/gui/tree/pr/459).

Sjors:
@promag thanks! Let's keep that commit in mind in case we actually want to reorder, which I don't think is very likely.

Sjors:
OTOH it's actually looks a bit cleaner too, so I squashed it in

Rspigler:
Will re-test

Rspigler:
I like the way this looks, but there is a typo in 1 tooltip, and I think other tooltips can be improved.

Base58 (Legacy):	"Not recommended due to higher fees and less protection against typos." -> "Not recommended."
It is also not recommended b/c of security issues with multisig, and scaling issues.  No reason to list them all. If an advanced user wants to use an old address, they will know why they need it.

Base58 (P2SH-SegWit):	"Generates an address compatible with older wallets."  :heavy_check_mark: 

Bech32 (SegWit):	"Generates a native segwit addresses (BIP-173), which reduces your transaction fees later on and offers better protection against typos, but some old wallets don't support it."  There is a typo: "Generates a native segwit addresses"->"Generates a native segwit address".  This is also a very long tooltip and looks bad IMO.  Perhaps:
"Generates a native segwit address (BIP-173) which offers various improvements."

Bech32m (Taproot):  "Although Bech32m (BIP-350) looks similar to Bech32, there is a slight difference and only some newer wallets support it."  I don't see a reason to take up space discussing how it looks, but I can't think of any improvement to offer right now that sounds good, other than basically what is said for SegWit :grimacing:

Sjors:
I fixed the typos and shortened the tooltips a bit.

>  security issues with multisig

That's not an issue for PKH (1...) addresses.

Scaling is a problem for the ecosystem, but the two issues I put in the tooltip (fees and typo protection) are relevant to individual users. If they encounter a wallet or service insisting on this, they should be appropriately annoyed :-) 

> Bech32 (SegWit)
> This is also a very long tooltip and looks bad IMO

I agree it looks ugly, but that's mainly because I wasn't able to add line breaks. That might be fixable though.

It's a bit shorter now though.

> I don't see a reason to take up space discussing how it looks

I reworded it a bit. The problem is that a user will put the bech32m address into a different wallet / exchange and get an error message like "Invalid (bech32) address".

Rspigler:
tACK  c9a77e227eecf357c7dd550efb8c1827bb99a5de.  I like the changes made now, thanks!

kristapsk (approved):
re-ACK c9a77e227eecf357c7dd550efb8c1827bb99a5de

src/qt/receivecoinsdialog.cpp
```diff
@@ -87,10 +87,18 @@ void ReceiveCoinsDialog::setModel(WalletModel *_model)
...
-        } else {
-            ui->useBech32->setCheckState(Qt::Unchecked);
+        // Populate address type dropdown and select default
+        auto add_address_type = [&](OutputType type, const QString& text, const QString& tooltip) {
+            const auto index = ui->addressType->count();
+            ui->addressType->addItem(text, (int) type);
+            ui->addressType->setItemData(index, tooltip, Qt::ToolTipRole);
+            if (model->wallet().getDefaultAddressType() == type) ui->addressType->setCurrentIndex(index);
+        };
+        add_address_type(OutputType::LEGACY, "Base58 (Legacy)", "Not recommended due to higher fees and less protection against typos.");
```
MarcoFalke, line 97: This is missing a check (like the one for taproot)?

Locally, I am getting:

Screenshot from 2021-12-22 12-26-46

src/qt/receivecoinsdialog.cpp
```diff
@@ -87,10 +87,18 @@ void ReceiveCoinsDialog::setModel(WalletModel *_model)
...
+        add_address_type(OutputType::P2SH_SEGWIT, "Base58 (P2SH-SegWit)", "Generates an address compatible with older wallets.");
+        add_address_type(OutputType::BECH32, "Bech32 (SegWit)", "Generates a native segwit address (BIP-173). Some old wallets don't support it.");
```
MarcoFalke, line 99: nit: maybe replace "wallets" with "services" or "third-party services" or "third-party wallets".

Otherwise the user  might get confused into thinking that an old wallet.dat can't generate this, which is not true?

src/qt/receivecoinsdialog.cpp
Sjors: I assume you manually created a wallet without a `pkh()` descriptor? We could add checks for the existence of each output type to the interface and then remove them from the dropdown. However, if someone manually creates a wallet, they would also understand the error message, and it might even better than a missing item?

src/qt/receivecoinsdialog.cpp
Sjors: I'm not sure if they're more likely to think of an upgrade scenario than about receiving coins from another wallet they have? "third-party" is a bit of an tech industry term, and probably doesn't translate well. Paging @jonatack.

PR #469, opened 2021-11-12
Load Base64 PSBT string from file

achow101:
Some .psbt files may have the PSBT as a base64 string instead of in binary. We should be able to load those files.

### DIFF
--- a/src/qt/walletframe.cpp
+++ b/src/qt/walletframe.cpp
@@ -215,6 +215,14 @@ void WalletFrame::gotoLoadPSBT(bool from_clipboard)
         }
         std::ifstream in{filename.toLocal8Bit().data(), std::ios::binary};
         data.assign(std::istream_iterator<unsigned char>{in}, {});
+
+        // Some psbt files may be base64 strings in the file rather than binary data
+        std::string b64_str{data.begin(), data.end()};
+        b64_str.erase(b64_str.find_last_not_of(" \t\n\r\f\v") + 1); // Trim trailing whitespace
+        auto b64_dec = DecodeBase64(b64_str);
+        if (b64_dec.has_value()) {
+            data = b64_dec.value();
+        }
     }
 
     std::string error;


hebasto:
> Some .psbt files may have the PSBT as a base64 string instead of in binary.

How to reliably create a correct such file?

achow101:
> How to reliably create a correct such file?

Get Base64 PSBT string. Make a new file. Paste string into file.

src/qt/walletframe.cpp
```diff
@@ -212,6 +212,15 @@ void WalletFrame::gotoLoadPSBT(bool from_clipboard)
         }
         std::ifstream in(filename.toLocal8Bit().data(), std::ios::binary);
         data = std::string(std::istreambuf_iterator<char>{in}, {});
+
+        // Some psbt files may be base64 strings in the file rather than binary data
+        bool invalid = false;
+        std::string b64_dec = data;
+        b64_dec.erase(b64_dec.find_last_not_of(" \t\n\r\f\v") + 1); // Trim trailing whitespace
+        b64_dec = DecodeBase64(b64_dec, &invalid);
```
luke-jr, line 220: Base64 files standardly have newlines every N charaters, but our `DecodeBase64` doesn't tolerate them currently...

jarolrod (approved):
tACK d54f498f8abbc49ff179e965e421cb0d8d5f5baa

Tested by saving a base64 psbt using sparrow wallet. Loading the file fails on master with `Unable to decode PSBT
Invalid PSBT magic bytes: iostream error`. On this PR branch, i can load the PSBT successfully.

src/qt/walletframe.cpp
```diff
@@ -212,6 +212,15 @@ void WalletFrame::gotoLoadPSBT(bool from_clipboard)
         }
         std::ifstream in(filename.toLocal8Bit().data(), std::ios::binary);
         data = std::string(std::istreambuf_iterator<char>{in}, {});
+
+        // Some psbt files may be base64 strings in the file rather than binary data
+        bool invalid = false;
+        std::string b64_dec = data;
+        b64_dec.erase(b64_dec.find_last_not_of(" \t\n\r\f\v") + 1); // Trim trailing whitespace
```
promag, line 219: We have `Trim` in src/util/string.h:

 ```cpp
bool invalid;
const auto decoded = DecodeBase64(Trim(data), &invalid);
if (!invalid) data = decoded
```

**Review by promag — COMMENTED** [review]
Concept ACK.

**Comment by jb55** [comment]
Concept ACK

File: src/qt/walletframe.cpp [hunk 3]
```diff
@@ -215,6 +215,16 @@ void WalletFrame::gotoLoadPSBT(bool from_clipboard)
...
+        b64_dec = DecodeBase64(b64_dec, &invalid);
+        if (!invalid) {
+            data = b64_dec;
+        }
+>>>>>>> d54f498f8 (gui: Load Base64 PSBT string from file)
```
luke-jr, line 227: o.O

achow101: Oops. fixed.

jarolrod (commented):
tACK 2c3ee4c347838ecadb17a011932dffc077e46630

Tested with same method as stated here: https://github.com/bitcoin-core/gui/pull/469#pullrequestreview-839000783

Could this use a release note?

src/qt/walletframe.cpp
```diff
@@ -215,6 +215,14 @@ void WalletFrame::gotoLoadPSBT(bool from_clipboard)
         }
         std::ifstream in{filename.toLocal8Bit().data(), std::ios::binary};
         data.assign(std::istream_iterator<unsigned char>{in}, {});
+
+        // Some psbt files may be base64 strings in the file rather than binary data
```
laanwj, line 219: I dislike this kind of ambiguity in file formats, it's a thing that can bite you in unexpected ways at some point, but I understand the reason. A valid binary PSBT can never be valid b64, I hope?

shaavan (approved):
ACK 2c3ee4c347838ecadb17a011932dffc077e46630

- I was able to load the Base64 PSBT transaction from the file through this PR, which the master failed to do so.

src/qt/walletframe.cpp
achow101: > A valid binary PSBT can never be valid b64, I hope?

Yes. The magic in binary are the characters `psbt`, but are `cHn...` in base64. Interpreting the binary as base64 or vice versa would result in the wrong magic.

PR #473, opened 2021-11-15
Enable a non-linear network traffic option (click to toggle between linear and non-linear)

rebroad:
Helps to make visible the low bandwidth traffic and high bandwidth (or spikes) concurrently. (Previously the low bandwidth traffic becomes invisible in the presence of spikes).

Old format: 
image

New format:
image

Old format:
image

New format:
image

(clicking on the graph will toggle between the new format and the old format).

### DIFF
--- a/src/qt/trafficgraphwidget.cpp
+++ b/src/qt/trafficgraphwidget.cpp
@@ -47,16 +47,22 @@ int TrafficGraphWidget::getGraphRangeMins() const
     return nMins;
 }
 
+int TrafficGraphWidget::y_value(float value)
+{
+    int h = height() - YMARGIN * 2;
+    return YMARGIN + h - (h * 1.0 * value / fMax);
+}
+
 void TrafficGraphWidget::paintPath(QPainterPath &path, QQueue<float> &samples)
 {
     int sampleCount = samples.size();
-    if(sampleCount > 0) {
+    if(sampleCount > 0 && fMax > 0) {
         int h = height() - YMARGIN * 2, w = width() - XMARGIN * 2;
         int x = XMARGIN + w;
         path.moveTo(x, YMARGIN + h);
         for(int i = 0; i < sampleCount; ++i) {
             x = XMARGIN + w - w * i / DESIRED_SAMPLES;
-            int y = YMARGIN + h - (int)(h * samples.at(i) / fMax);
+            int y = y_value(samples.at(i));
             path.lineTo(x, y);
         }
         path.lineTo(x, YMARGIN + h);
@@ -84,7 +90,7 @@ void TrafficGraphWidget::paintEvent(QPaintEvent *)
 
     // draw lines
     painter.setPen(axisCol);
-    painter.drawText(XMARGIN, YMARGIN + h - h * val / fMax-yMarginText, QString("%1 %2").arg(val).arg(units));
+    painter.drawText(XMARGIN, y_value(val)-yMarginText, QString("%1 %2").arg(val).arg(units));
     for(float y = val; y < fMax; y += val) {
         int yy = YMARGIN + h - h * y / fMax;
         painter.drawLine(XMARGIN, yy, width() - XMARGIN, yy);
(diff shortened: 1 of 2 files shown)


katesalazar:
Love the idea, the problem with these things is
there's gonna be someone who is annoyed by the behavior change.

When I do things like this, I always maintain the old behavior somehow,
so that kind of reaction against gets suppressed more easily.

laanwj:
Concept ACK, this seems like a good idea to me, given that, what you really want to see in graphs like this is the gross magnitude and not so much very small variations in the throughput.

rebroad:
> Love the idea, the problem with these things is there's gonna be someone who is annoyed by the behavior change.
> 
> When I do things like this, I always maintain the old behavior somehow, so that kind of reaction against gets suppressed more easily.

I agree, which is why I've now made it toggleable - just click on the graph to switch between the two display options.

katesalazar:
Concept ACK.

> > Love the idea, the problem with these things is there's gonna be someone who is annoyed by the behavior change.
> > When I do things like this, I always maintain the old behavior somehow, so that kind of reaction against gets suppressed more easily.
> 
> I agree, which is why I've now (latest commit) made it toggleable - just click on the graph to switch between the two display options.

Obscure, but OK.

src/qt/trafficgraphwidget.cpp
```diff
@@ -50,19 +50,28 @@ int TrafficGraphWidget::getGraphRangeMins() const
...
             x = XMARGIN + w - w * i / DESIRED_SAMPLES;
-            int y = YMARGIN + h - (int)(h * samples.at(i) / fMax);
+            int y = YMARGIN + h - (int)(h * 1.0 * (fToggle ? (pow(samples.at(i), 0.3) / pow(fMax, 0.3)) : (samples.at(i) / fMax)));
             path.lineTo(x, y);
         }
         path.lineTo(x, YMARGIN + h);
     }
 }
 
+void TrafficGraphWidget::mousePressEvent(QMouseEvent *event)
```
jarolrod, line 66: This isn't a great UI decision for end users. This functionality is hidden from users. If they discover it, it would be by accident; after having clicked on the Network Traffic widget. 

Let's make this feature more accessible and documented by adding two buttons next to the `reset` button:  one that sets the existing format and another button that enables the new format introduced in this PR.

The `fToggle` variable here should then be made into a persistent setting.

Additionally, when a longer sampling period has been set with the slider; it can take a minute or longer for the format to switch. There are optimizations that need to be done. I have not investigated this.

jarolrod (commented):
concept ack

hebasto:
Concept ACK.

src/qt/trafficgraphwidget.cpp
rebroad: @jarolrod What options are available so far for making GUI changes persistent? Regarding the delay when the slider is moved, I fixed that in a later update.

What more obvious way might there be to reveal the Y axis is toggleable? A tooltip on the graph perhaps? I am reluctant to add extra buttons given the slider is already providing many values in so little space, so I do prefer the clicking on the graph to change it. It may be that someone finds it by accident, but then this could be a pleasant surprise that - an "easter egg" feature!

katesalazar: @rebroad 

> What more obvious way might there be to reveal the Y axis is toggle-able?

Yes, button, of course! Checkbox, radio, all those are good.

> A tooltip on the graph perhaps?

Improves from no tooltip, but no,
that's not enough improvement.

> I am reluctant to add extra buttons given the slider is already providing many values in so little space, so I do prefer the clicking on the graph to change it.

What? No, that's an huge slider. An unnecessarily large one.
Sliders can be as small as necessary when they are
combined with a text-box presenting the slided value.
There's a huge empty space below the legend too.
There's plenty of avail space to discard invisible toggles.

> It may be that someone finds it by accident, but then this could be a pleasant surprise that - an "easter egg" feature!

No, an easter egg is meant to be fun or charming in
some way. Hiding actual feature in the same way like
one could do with a true easter egg is transgredes
the principle of least surprise.

Cheers!

RandyMcMillan:
Concept ACK

More useful information "at a glance".

RandyMcMillan:
Adding a label to the highest horizontal line would be useful.

rebroad:
> Adding a label to the highest horizontal line would be useful.

I did consider this, but as it's not the current functionality, and it would make the highest point on the graph no longer reach the top (as the label would need extra space) I decided against it.  Could be done as a separate pull request though.

rebroad:
@jarolrod I found a way to make it update the graph immediately now - using QWidget::update(). Updated to do this.

src/qt/trafficgraphwidget.cpp
rebroad: @jarolrod The latest commit makes the update immediate now (using QWidget::update).

rebroad:
> Adding a label to the highest horizontal line would be useful.

@RandyMcMillan #492 also reduces the need for this as now you can point to a point on the graph to get details.

RandyMcMillan:
@rebroad - added this in PR #539 
https://github.com/bitcoin-core/gui/pull/539/commits/10f8e797bcd6d8f8910b7560a6e132d197b5f97b

hebasto:

PR #476, opened 2021-11-20
src/net.cpp: Fail to listen alert text break fix

RandyMcMillan:


RandyMcMillan:
When another instance of bitcoin is running:

Screen Shot 2021-11-19 at 10 09 11 PM

Before:
Screen Shot 2021-11-19 at 10 09 18 PM

After:
Screen Shot 2021-11-19 at 10 10 37 PM

shaavan (approved):
ACK a5097f37ffead4790c11da66a8156b828f812ec5

Simple but necessary change. Before line break, the usage of `-listen=0` was not evident, and `-` could have confused a simple text hyphen. This PR removes the scope for that doubt.
Nice catch, @RandyMcMillan!

katesalazar:
Concept ACK.

hebasto:
Why don't use https://github.com/bitcoin-core/gui/blob/19815b096c6a2b2e0bcb802e9e5842e67952993e/src/qt/rpcconsole.cpp#L497 
which we already use in our code?

hebasto:
@RandyMcMillan 

Could you move this PR into the main repo because it changes the code outside the `src/qt/` directory?

Issue #273, opened 2021-04-06
Ugly RPC Console Output on macOS


jarolrod:
**Edit**: This is a macOS specific issue. A monospace font is used on Linux and windows, but not on macOS. Add to the list of reasons to embed a regular monospace font.

The help output for any RPC command is formatted differently in the macOS GUI `console` versus a Linux GUI `console`.

With the Linux GUI `console`, the `type/descriptions` for an `argument/result` are shown at a uniform distance from the `argument/result` itself, thanks to the use of a monospace font.

With the macOS GUI `console`, the `type/descriptions` do not line up because a monospace font is not used. This leads to 'ugly' output.

Example: `help setwalletflag`

**macOS GUI RPC Console**
![Screen Shot 2021-04-06 at 7 07 06 PM](https://user-images.githubusercontent.com/23396902/113788966-36386780-970c-11eb-81a5-e9e210cf1a23.png)

**Linux GUI RPC Console**
![Screen Shot 2021-04-08 at 11 49 15 AM](https://user-images.githubusercontent.com/23396902/114057420-8a546080-9860-11eb-9051-9dc8dabcfa2f.png)

hebasto:
> A monospace font is used on Linux and windows, but not on macOS.

Actually, the font family in the default stylesheet https://github.com/bitcoin-core/gui/blob/f0fa32450ec35056b3e1025262aeaef4a24c35ee/src/qt/rpcconsole.cpp#L805-L817
is set correctly to `.AppleSystemUIFontMonospaced`.

But this approach (using styleesheets) does [not work](https://github.com/bitcoin-core/gui/pull/79/commits/22e0114d055b0ac6cdc47af2968b5dc742ba055a) on macOS. Probably, the reason of such behavior is the fact that `.AppleSystemUIFontMonospaced` is a [private](https://doc.qt.io/qt-5/qfontdatabase.html#isPrivateFamily) font family.

hebasto:
Maybe, the following patch
```diff
diff --git a/src/qt/rpcconsole.cpp b/src/qt/rpcconsole.cpp
index 34d055e5a..1acfaa3b2 100644
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -803,7 +803,7 @@ void RPCConsole::clear(bool clearHistory)
     }
 
     // Set default style sheet
-    QFontInfo fixedFontInfo(GUIUtil::fixedPitchFont());
[...]

PR #477, opened 2021-11-20
Monospaced output in Console on macOS

RandyMcMillan:
This PR addresses issue https://github.com/bitcoin-core/gui/issues/273 
A monospace font is used on Linux and Windows for the console output - but not on MacOS.
This change forces the MacOS GUI to use the embedded RobotoMono-Bold.ttf font,
which is defined as the GUIUtil::fixedPitchFont()

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -866,7 +866,11 @@ void RPCConsole::clear(bool keep_prompt)
     }
 
     // Set default style sheet
+#ifdef Q_OS_MAC
+    QFontInfo fixedFontInfo(GUIUtil::fixedPitchFont(/*use_embedded_font=*/true));
+#else
     QFontInfo fixedFontInfo(GUIUtil::fixedPitchFont());
+#endif
     ui->messagesWidget->document()->setDefaultStyleSheet(
         QString(
                 "table { }"


RandyMcMillan:
REF: Issue: #273

Per @hebasto 's suggestion

Before:
Screen Shot 2021-11-19 at 11 15 18 PM

Screen Shot 2021-11-19 at 11 16 32 PM

RandyMcMillan:
A few more examples:

Screen Shot 2021-11-20 at 12 29 39 AM

RandyMcMillan:
Screen Shot 2021-11-20 at 12 45 06 AM

shaavan (commented):
Concept ACK

This is a definite improvement over the currently displayed font.
However, I think the newly displayed font is too bold and looks very bulky when there is a lot of text on the screen. **Like in this screenshot:**

image

What do you think about experimenting with a lighter version of the monospace font?

For example, in #273, the screenshot of **Linux GUI RPC Console** displays a lighter version of the monospace font. That text looks very light while each word is also easily readable.
**Screenshot:**
image

RandyMcMillan:
This issue has been posted since April (7 months) - if you want to focus on a bunch of minutia - feel free to on your own time. I am not burning a bunch of time on a 30 second edit that anybody else could have done.

shaavan:
Well, I was not expecting such an aggressive response. I was under the impression that the goal of this PR was to fix the problem entirely and not just to fix a tiny issue that anybody could have fixed.

RandyMcMillan:
@shaavan - may apologize for the previous response...

The whole embedding fonts approach opens up a whole can of worms - which I prefer to avoid.
This PR is a simple "fix" avoiding all that controversy.

There is clearly a more elegant fix which you may want to implement and I would be open to reviewing and contributing to.

With that said - this PR is a minimal fix which shouldn't evoke the font issues.
Feel free to review previous discussions on the topic.

RandyMcMillan:
Update: embed Regular Weight Font

Screen Shot 2021-11-20 at 4 48 12 PM

Screen Shot 2021-11-20 at 4 48 33 PM

Screen Shot 2021-11-20 at 4 48 58 PM

shaavan (approved):
> may apologizes for the tone of my previous response...

It's okay. Sometimes all of us feel a little out of place and may lose our cool. The good thing is to accept our mistakes humbly. So I appreciate that.

> With that said - this PR is a minimal fix which shouldn't evoke the font issues.

I understand the context of this PR now. And IMO this PR accomplishes the goal it set out to achieve.

ACK 4516e39c0a9376385ad72689361992040defa2ff

katesalazar:
ACK 4516e39c0a9376385ad72689361992040defa2ff.

katesalazar:
I would be happier enclosing the new code under __APPLE__ ifdefs though.

src/qt/rpcconsole.cpp
```diff
@@ -866,7 +866,7 @@ void RPCConsole::clear(bool keep_prompt)
     }
 
     // Set default style sheet
-    QFontInfo fixedFontInfo(GUIUtil::fixedPitchFont());
+    QFontInfo fixedFontInfo(GUIUtil::fixedPitchFont(true));
```
hebasto, line 869: To make code more readable and easier to reason about:
```suggestion
    QFontInfo fixedFontInfo(GUIUtil::fixedPitchFont(/*use_embedded_font=*/true));
```

See https://github.com/bitcoin/bitcoin/pull/23545

hebasto (commented):
Concept ACK.

> I would be happier enclosing the new code under __APPLE__ ifdefs though.

Agree that there are no reasons to change look in systems that do not require any change.

Would you mind to provide the pr description? And refer to #273 in it?

src/qt/rpcconsole.cpp
RandyMcMillan: @katesalazar @hebasto - thanks for the feedback!

katesalazar:
ACK 9e6775d424ad931d3d94755dddf0563fe27b8ded.

I'd be happier with a commit message bringing
a thorough description copying bits from here and the parent issue.
Maybe in the future Apple fixes this and the two lines do the same,
I wouldn't want to check the issue tracker to know why this code,
only VCS history.
Lossless VCS history migration is far easier than lossless issue tracker migration.
In other words, GitHub is temporary, VCS can be forever.

src/qt/rpcconsole.cpp
```diff
@@ -866,7 +866,11 @@ void RPCConsole::clear(bool keep_prompt)
     }
 
     // Set default style sheet
+#ifdef __APPLE__
```
hebasto, line 869:
```suggestion
#ifdef Q_OS_MAC
```

as we use it in our Qt code (sorry for not mentioning this earlier).

hebasto (commented):
Approach ACK 9e6775d424ad931d3d94755dddf0563fe27b8ded.

hebasto:
ACK b9f0aff6b4a64c75d8f937fb9a3546c96d6f5a01

shaavan (approved):
reACK b9f0aff6b4a64c75d8f937fb9a3546c96d6f5a01

Changes since my last review:

- Applied `ifdef`, `endif` conditional statement to prevent this change from affecting other Operating Systems. This change prevents the risk of having unintentional behavior changes on some OS.
- Added `/*use_embedded_font=*/` in the argument of fixedPitchFont function. This is done to make the code more readable.

jarolrod (commented):
tACK b9f0aff6b4a64c75d8f937fb9a3546c96d6f5a01

katesalazar:
Well a couple of late points... I'm always swamped y'know.

* Won't ACK b9f0aff6b4a64c7 only because
I'm not familiar enough with `Q_OS_MAC`.

* I tested this, and (only after testing) I have to agree to
  [the boldness point written by shaavan][0], I can't avoid
  writing it, but

  * in my system the boldness can be made away with
    this patch immediately on top of b9f0aff6b4a64c7:

        diff --git a/src/qt/guiutil.cpp b/src/qt/guiutil.cpp
        index 4262866f32..847e73c7ae 100644
        --- a/src/qt/guiutil.cpp
        +++ b/src/qt/guiutil.cpp
        @@ -88,7 +88,11 @@ QString dateTimeStr(qint64 nTime)
         QFont fixedPitchFont(bool use_embedded_font)
         {
             if (use_embedded_font) {
        +#ifdef __APPLE__
        +        return {"PT Mono"};
        +#else
                 return {"Roboto Mono"};
        +#endif
             }
             return QFontDatabase::systemFont(QFontDatabase::FixedFont);
         }

    I think this is strongly more "UI-idiomatic" than the
    b9f0aff6b4a64c7 state of things, I'd recommend macOS
    users to try it.

  I'm not familiar enough with macOS development as to know
  what's the more appropriate monospaced font on macOS
  (or if such thing depends on macOS versions).

Cheers!

[0]: https://github.com/bitcoin-core/gui/pull/477#pullrequestreview-811806941

RandyMcMillan:
I will take another look - to see how the boldness can be addressed. :)

luke-jr:
>I will take another look - to see how the boldness can be addressed. :)

After #497, it should be simpler to make the console font user-configurable.

katesalazar:
current Roboto Mono on Dark Appearance:

Screen Shot 2022-01-04 at 21 41 26

available PT Mono on Dark Appearance:

Screen Shot 2022-01-04 at 21 42 53

PR #481, opened 2021-11-22
rpcconsole:accessibility - new shortcuts enable single hand gesture

RandyMcMillan:


RandyMcMillan:
Accessibility improvement:

CMD/CTRL + 1
CMD/CTRL + 2
CMD/CTRL + 3
CMD/CTRL + 4

are easily accessed by a single left hand motion and greatly simplifies a 2 handed gesture.

katesalazar:
If you look at this menu from GNU/Linux:

> Screen Shot 2021-11-22 at 9 27 34 AM

And hit the "alt" key, then W (for Window), M (for Minimize), S (for Sending addresses), R (for Receiving addresses), I (for Information), C (for Console), N (for Network Traffic), and P (for Peers) get conveniently highlighted.
Have you considered that?

I would have just checked my mac build too, but is broken for now.

RandyMcMillan:
I agree - more short cuts can be added.

katesalazar:
IMO "Receiving addresses" is much more important than "Network Traffic", for one.

If you are going to stand for this,
I'd say propose the change only for macs and 
add numeric shortcuts to the other controls.

RandyMcMillan:
From an accessibility perspective - it is ideal to keep shortcuts to a 2 key combination CNTL + <key> 

3 key combos: 

SHIFT + CNTL + <key> are not friendly to impaired users.

RandyMcMillan:
The "tab through" functionality can also be optimized...

These types of changes usually generate a lot of "diff junk" in the QT .ui files and is frowned upon - but i believe this is an important topic that warrants a continued discussion, research and scrutiny. :)

katesalazar:
I ran this on KDE, maybe it felt a bit weird, but not enough to NAK.

RandyMcMillan:
If you think about it from a single left handed perspective - it ought to make sense...

Then approach it from a two handed perspective...

RandyMcMillan:
More generally - try to use the gui - with only one hand - you will quickly see how badly the GUI falls short...

jarolrod (commented):
In terms of real world usability, considering all types of different keyboards that can be used as input, CTRL+ALPHA is better. 

For example, I do not have a number row on my keyboard. I can access numeric keys, but they are hidden on a different layer after pressing a modifier key. In terms of usability, this would be a downgrade for me.

If there is a good argument that this is better for some users who are disabled, then this feature could be delegated behind a setting that can toggle between mnemonic shortcuts and accessibility oriented shortcuts.

RandyMcMillan:
I greatly appreciate your time and feedback...

While thinking more about this - I noticed that...

ALT + 1
ALT + 2
ALT + 3
ALT + 4 

Doesn't raise the wallet interface and set focus - as it clearly should...

And there is no indication of this functionality in the MacOS interface?

katesalazar:
Seen yesterday those combinations circle through the 4 wallet tabs in at
least one desktop environment interface.
If it doesnt do for every, I guess it could be bug, if so, separate from
this PR.

katesalazar:
I was reserving about concept ACKing too early because I was seeing
something like this probably coming...

... related to that, so I was thinking about configurable key shortcuts,
and was already imaging this usual needlessly stupid GUI dev hell;
like the case for detecting caps lock before an alpha key is pressed,
when entering a password; to be expected here. :pray:🙏

So good luck if you have a look at any of that! :laughing:😂

luke-jr (changes requested):
Concept NACK. Standard shortcuts like this are Ctrl+Letter (like we have right now).

Furthermore, there's no particular reason for this order of debug window tabs, and they could very well be reordered (eg, by inserting one in the middle) in the future.

shaavan (commented):
Concept NACK

I was concept 0 on this PR, to begin with. Because:
1. Many bitcoin-core GUI users would be pretty comfortable with current letter shortcuts and have a kind of muscle memory for them. Changing those would create unnecessary strain in their workflow.
2. To press `ctrl + number` with one hand, you have to bend your hand in kind of an awkward and out-of-place way, especially on an external full-size keyboard. `Alt + number` would improve over the current `ctrl + number` implementation. But it seems like `Alt + number` doesn’t seem to work (as discussed [here](https://github.com/bitcoin-core/gui/pull/481#issuecomment-977824764)).

But after reading @luke-jr’s thoughts:

> Furthermore, there's no particular reason for this order of debug window tabs, and they could very well be reordered (e.g., by inserting one in the middle) in the future.

I agree with him. I think the above statement is a strong argument against this PR. In my view, the improved accessibility doesn’t balance the unnecessary maintenance burden this PR introduces. That’s why I am concept NACK on this PR.

PR #483, opened 2021-11-25
Make network graph slider easier to use

rebroad:
Currently the network graph slider has 288 possible positions, and most of these are likely never used, due to obscure durations.

This pull request simplifies the slider to have only 12 positions, making it much easier to use, and setting the durations to more likely-wanted settings: 1m, 2m, 5m, 10m, 20m, 30m, 1h, 2h, 3h, 6h, 12h, 24h.

This change also allows easier rearranging of the window, for example, adding an extra button next to the reset button, shrinking the slider, and therefore making it still easy to use when smaller in length.

### DIFF
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -633,10 +633,10 @@
               <number>1</number>
              </property>
              <property name="maximum">
-              <number>288</number>
+              <number>12</number>
              </property>
              <property name="pageStep">
-              <number>12</number>
+              <number>2</number>
              </property>
              <property name="value">
               <number>6</number>
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -1129,8 +1129,8 @@ void RPCConsole::scrollToEnd()
 
 void RPCConsole::on_sldGraphRange_valueChanged(int value)
 {
-    const int multiplier = 5; // each position on the slider represents 5 min
-    int mins = value * multiplier;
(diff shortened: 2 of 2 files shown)


shaavan (approved):
I like the idea of reducing the number of options and removing the redundant ones. I also think these 12 options are sufficient for any general user’s need. 

I was able to test this PR successfully on Ubuntu 20.04. Here’s the screenshot of the Network Traffic window taken of the PR branch (slider at 20 minutes).

Screenshot from 2021-11-26 15-50-06

katesalazar:
Concept NAK, 12 sliders are too few positions for such a large slider.

I booted it and tested the 12 sliders. It feels like few positions.
Please iterate.

jarolrod (commented):
I think this is fine to do, but I'm concept 0 on this.
> Currently the network graph slider has 288 possible positions, and most of these are likely never used, due to obscure durations.

I think this is true.

> This pull request simplifies the slider to have only 12 positions, making it much easier to use

How exactly does this make the slider "easier to use"? The workflow of using the slider as a slider remains untouched

rebroad:
> Concept NAK, 12 sliders are too few positions for such a large slider.
> 
> I booted it and tested the 12 sliders. It feels like few positions. Please iterate.

This is intentional as there is talk about adding an extra button, and so the slider will be reduced in size in the future, but I do agree with you as it is.

rebroad:
> How exactly does this make the slider "easier to use"? The workflow of using the slider as a slider remains untouched

@jarolrod because it is faster to use, i.e. less moving left and right to get the exact duration desired (usually a round number).

katesalazar:
nice, looking forward to the button

hebasto:
> There hasn't been much activity lately. What is the status here?

Closing due to lack of interest.

PR #484, opened 2021-11-25
Don't clear traffic graph when changing interval

rebroad:
This change stops the graph from being reset when the duration is changed. If the user wants it to be cleared, then they can click the reset button that is already present.

### DIFF
--- a/src/qt/trafficgraphwidget.cpp
+++ b/src/qt/trafficgraphwidget.cpp
@@ -159,8 +159,7 @@ void TrafficGraphWidget::setGraphRangeMins(int mins)
     int msecsPerSample = nMins * 60 * 1000 / DESIRED_SAMPLES;
     timer->stop();
     timer->setInterval(msecsPerSample);
-
-    clear();
+    timer->start();
 }
 
 void TrafficGraphWidget::clear()


shaavan:
Strong Concept ACK

It felt kind of annoying when you had network traffic data for the last 30 mins, and you, by mistake or intentionally moved the slider, and all data was lost.
I didn't think it was such a simple fix. Thanks for catching it @rebroad!
Let me do the testing and post the tested review soon!

shaavan (commented):
Approach NACK

The (x-axis) scale of the graph doesn’t update accordingly with the change in timer. 
For example,

When I set the timer to 45 min the graph looks like following:

Screenshot from 2021-11-26 16-06-54

But when I changed the scale to far-right, i.e., 1d. The graph remained the same:

Screenshot from 2021-11-26 16-07-00

The change done in this PR allows displaying the graph in the future according to the latest timer setting. However, since the time-scaling of the graph that is already displayed is not changing accordingly, it could lead to wrong conclusions on the user’s side.

The optimal behavior would be to expand or contract the displayed graph according to the latest time-scale setting.

jarolrod (commented):
Approach NACK, instead you should do more sampling and caching. Then display the appropriate data as the slider moves instead of clearing and drawing.

katesalazar:
Postponed until [approach NAK][0] is tackled
or fighted.

[0]: https://github.com/bitcoin-core/gui/pull/484#pullrequestreview-817113598

rebroad:
@shaavan PR #492 hopefully addresses your concerns raised here.

rebroad:
> The (x-axis) scale of the graph doesn’t update accordingly with the change in timer. For example,

I would argue that the user would have to be extraordinarily unintelligent to have an expectation of the X-axis being linear, if it is the same user that changed the duration and could see (at the time) that the pre-existing data was not resampled.

The only scenario I can see confusion arising is it someone viewed the graph without being aware that the scale has been changed mid-way through the data. I think there ought to be a limit to the extent we (developers) take responsibility for all the ways people can misinterpret data that has been manipulated by others.

The assumption here is there is only one user - the same user viewing the data as the user configuring the sample rate of that data.

And even if a 3rd party was involved - there's really very little "damage" that could occur from any false assumptions.



> 
> When I set the timer to 45 min the graph looks like following:
> 
> Screenshot from 2021-11-26 16-06-54
> 
> But when I changed the scale to far-right, i.e., 1d. The graph remained the same:
> 
> Screenshot from 2021-11-26 16-07-00
> 
> The change done in this PR allows displaying the graph in the future according to the latest timer setting. However, since the time-scaling of the graph that is already displayed is not changing accordingly, it could lead to wrong conclusions on the user’s side.
> 
> The optimal behavior would be to expand or contract the displayed graph according to the latest time-scale setting.

rebroad:
> Approach NACK, instead you should do more sampling and caching. Then display the appropriate data as the slider moves instead of clearing and drawing.

Personally, I would not want this functionality (I did consider it). I much prefer maximizing the use of the data already sampled (which resampling would not achieve).

shaavan:
> PR #492 hopefully addresses your concerns raised here.

#492 would be an improvement over the current implementation and can be helpful in many cases other than this PR.
However, it still does not fix the issue I described in its entirety.

Usually, we don't take the cursor to all the positions on the graph and read the specific time of that instance with focus. We typically glance over the chart as a whole to get a gist of the network traffic. So there is a risk of possible misinterpretation, even for an individual user.

Though @reboard has addressed this in [this](https://github.com/bitcoin-core/gui/pull/484#issuecomment-983795490) comment, I still think this is a bad UX design. And personally, for me, the possible declination in the user's experience outweighs the functionality this PR is introducing.

promag:
How about replacing the slider with something like image

Then it just needs some vertical lines with timestamps on the x-axis.

This suggestion is also an alternative to #483.

RandyMcMillan:
REF: https://github.com/bitcoin-core/gui/issues/532

RandyMcMillan:
Concept ACK

RandyMcMillan:
@rebroad - Please rebase - I prefer to test PR on a more recent commit. (Even though it is a trivial change) -thanks
NOTE: current /gui master passes CI 9ec3991ad3b2ae91997c696a4c2f187fe538eff0

RandyMcMillan:
Dynamic scaling based on QSlider value -> https://github.com/RandyMcMillan/gui/tree/1643137693-retain-network-graph-on-interval-change

RandyMcMillan:
@rebroad - I recommend renaming this PR to 
```
gui: Network Traffic Graph - dynamic scaling view
```
or something similar.

https://user-images.githubusercontent.com/152159/151081341-78ef4334-21e4-4919-9265-4502d1e338ee.mov

RandyMcMillan:
@rebroad - I submitted a PR to your repo or you can fetch my commit and cherry-pick it on top of yours.
Thanks - https://github.com/RandyMcMillan/gui/commit/00c016f1a0a74ff5bedf28f10dc9ea3071f6b648

RandyMcMillan:
The scaling is working...

https://user-images.githubusercontent.com/152159/151122794-dd3d44c3-0687-4739-9016-6b80b67615e8.mp4

RandyMcMillan:
@rebroad - I based this PR on the commit in this PR...

https://github.com/bitcoin-core/gui/pull/539

hebasto:

PR #492, opened 2021-12-01
Show ToolTip on Network Traffic graph

rebroad:
Enables a tooltip on the graph to make it easier to see what point in time a part of the graph corresponds to.

image

This PR includes the change in #484, but also deals with the potential issues that the simplicity of 484 introduced.

TODO:
- [x] make it easier to select a point on the graph by factoring in the vertical position of the mouse pointer also.
- [ ] maybe - add ridges on the X-axis to show where an interval change or time jump occurs.
- [x] make the Units translateable.

### DIFF
--- a/src/qt/trafficgraphwidget.cpp
+++ b/src/qt/trafficgraphwidget.cpp
@@ -47,16 +47,22 @@ int TrafficGraphWidget::getGraphRangeMins() const
     return nMins;
 }
 
+int TrafficGraphWidget::y_value(float value)
+{
+    int h = height() - YMARGIN * 2;
+    return YMARGIN + h - (h * 1.0 * value / fMax);
+}
+
 void TrafficGraphWidget::paintPath(QPainterPath &path, QQueue<float> &samples)
 {
     int sampleCount = samples.size();
-    if(sampleCount > 0) {
+    if(sampleCount > 0 && fMax > 0) {
         int h = height() - YMARGIN * 2, w = width() - XMARGIN * 2;
         int x = XMARGIN + w;
         path.moveTo(x, YMARGIN + h);
         for(int i = 0; i < sampleCount; ++i) {
             x = XMARGIN + w - w * i / DESIRED_SAMPLES;
-            int y = YMARGIN + h - (int)(h * samples.at(i) / fMax);
+            int y = y_value(samples.at(i));
             path.lineTo(x, y);
         }
         path.lineTo(x, YMARGIN + h);
@@ -84,7 +90,7 @@ void TrafficGraphWidget::paintEvent(QPaintEvent *)
 
     // draw lines
     painter.setPen(axisCol);
-    painter.drawText(XMARGIN, YMARGIN + h - h * val / fMax-yMarginText, QString("%1 %2").arg(val).arg(units));
+    painter.drawText(XMARGIN, y_value(val)-yMarginText, QString("%1 %2").arg(val).arg(units));
     for(float y = val; y < fMax; y += val) {
         int yy = YMARGIN + h - h * y / fMax;
         painter.drawLine(XMARGIN, yy, width() - XMARGIN, yy);
@@ -94,13 +100,13 @@ void TrafficGraphWidget::paintEvent(QPaintEvent *)
         axisCol = axisCol.darker();
         val = pow(10.0f, base - 1);
         painter.setPen(axisCol);
-        painter.drawText(XMARGIN, YMARGIN + h - h * val / fMax-yMarginText, QString("%1 %2").arg(val).arg(units));
+        painter.drawText(XMARGIN, y_value(val)-yMarginText, QString("%1 %2").arg(val).arg(units));
         int count = 1;
         for(float y = val; y < fMax; y += val, count++) {
             // don't overwrite lines drawn above
             if(count % 10 == 0)
                 continue;
-            int yy = YMARGIN + h - h * y / fMax;
+            int yy = y_value(y);
             painter.drawLine(XMARGIN, yy, width() - XMARGIN, yy);
         }
     }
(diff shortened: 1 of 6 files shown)


src/qt/trafficgraphwidget.cpp
```diff
@@ -63,6 +66,54 @@ void TrafficGraphWidget::paintPath(QPainterPath &path, QQueue<float> &samples)
...
+    }
+    if (x == -1) return;
+
+    int w = width() - XMARGIN * 2;
+    int sampleCount = vTimeStamp.size();
+    int real_i = (w + XMARGIN - x) * DESIRED_SAMPLES / w;
+    static int i = real_i;
+    int new_x = x;
+    if (x == last_x && y == last_y) { // Follow ToolTip value if mouse has not moved
+        if (fShiftLeft && i>=0 && i < sampleCount) i++;
```
rebroad, line 91: FYI - the reason this code is here is to add the possibility to draw on the graph the point that the tooltip is referring to. But this could be added in a later PR.

src/qt/trafficgraphwidget.cpp
```diff
@@ -126,12 +177,17 @@ void TrafficGraphWidget::updateRates()
 {
     if(!clientModel) return;
 
+    static int64_t nTime = GetTimeMillis();
+    static int64_t nLastTime = nTime - timer->interval();
+    int nRealInterval = nTime - nLastTime;
```
rebroad, line 182: This code also helps to fix the "spike" issue that can happen when the real interval is different to the timer->interval() value.

shaavan:
Concept ACK

I think this PR is an improvement from a UX perspective. I shall be testing this PR when the TODO is completed.

rebroad:
> Concept ACK
> 
> I think this PR is an improvement from a UX perspective. I shall be testing this PR when the TODO is completed.

TODO of point selection now complete.

Still TODO - add ridges on the X-axis to show where an interval change or time jump change occurs.

rebroad:
Having to keep changing things as the test is failing (it makes false assumptions).

src/util/strencodings.cpp
```diff
@@ -369,6 +369,36 @@ std::string FormatParagraph(const std::string& in, size_t width, size_t indent)
...
+    }
+    if (value < 1) return strprintf(strprintf("%%.%df %s%%s", dp, letter), value, strUnit);
+    if (value < 10) return strprintf(strprintf("%%.%df %s%%s", std::max(dp - 1, 0), letter), value, strUnit);
+    if (value < 100) return strprintf(strprintf("%%.%df %s%%s", std::max(dp - 2, 0), letter), value, strUnit);
+    if (value < 1'000) return strprintf(strprintf("%%.%df %s%%s", std::max(dp - 3, 0), letter), value, strUnit);
+
+    return strprintf(strprintf("%%.%df %s%%s", std::max(dp - 4, 0),  letter), value, strUnit);
+}
+
+std::string strBps(float bits) {
```
rebroad, line 394: This isn't used by this PR, but is planned to be used in #305

katesalazar:
I'd fancy reviewing this,
but my time is very limited.
As you keep finishing PRs, I could use some meta issue
(or something) helping me realize the dependency tree
between PRs without me having to spend time figuring it out.
Cheers,

rebroad:
> I'd fancy reviewing this, but my time is very limited. As you keep finishing PRs, I could use some meta issue (or something) helping me realize the dependency tree between PRs without me having to spend time figuring it out. Cheers,

Hi. The latest push to this PR ought to be the last one - I had been trying to get the ToolTips to behave better, which they now do. They no longer time-out (disappear) and they also can now moved along with the graph regardless of the duration/speed selected. Also, the yellow circle will disappear more reliably now if the ToolTip also disappears (due to the mouse pointing elsewhere). It took a while to refine this functionality, which in the latest commit required an additional timer (tt_timer) to check the ToolTip status at an interval independent of the graph being updated, or the mouse moving.

katesalazar:
Built and tested this, seems cool.

src/qt/trafficgraphwidget.cpp
```diff
@@ -120,18 +174,63 @@ void TrafficGraphWidget::paintEvent(QPaintEvent *)
         painter.setPen(Qt::red);
         painter.drawPath(p);
     }
+    int sampleCount = vTimeStamp.size();
+    if (ttpoint >= 0 && ttpoint < sampleCount) {
+        painter.setPen(Qt::yellow);
+        int w = width() - XMARGIN * 2;
+        int x = XMARGIN + w - w * ttpoint / DESIRED_SAMPLES;
+        int y = y_value(floatmax(vSamplesIn.at(ttpoint), vSamplesOut.at(ttpoint)));
+        painter.drawEllipse(QPointF(x, y), 3, 3);
```
katesalazar, line 183: What about fencing these in a function? Maybe a macro?
Maybe add a couple of comments?

src/qt/trafficgraphwidget.cpp
```diff
@@ -262,8 +262,7 @@ void TrafficGraphWidget::setGraphRangeMins(int mins)
     int msecsPerSample = nMins * 60 * 1000 / DESIRED_SAMPLES;
     timer->stop();
     timer->setInterval(msecsPerSample);
-
-    clear();
+    timer->start();
```
katesalazar, line 265: Why did you separate this change in a rev of its own?

CryptoGodfatherVA4 (approved):

CryptoGodfatherVA4 (approved):

CryptoGodfatherVA4 (approved):

CryptoGodfatherVA4 (approved):

hebasto:

katesalazar:
#473 is similarly inactive.

PR #493, opened 2021-12-01
qt: Handle Android back key in the Node window

hebasto:
On master (4633199cc8a466b8a2cfa14ba9d7793dd4c469f4) there are no means to return from the Node window to the main one on Android.

This PR assigns this functionality to the Android back key:

Screenshot_1638395318

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -14,6 +14,7 @@
 #include <netbase.h>
 #include <qt/bantablemodel.h>
 #include <qt/clientmodel.h>
+#include <qt/guiutil.h>
 #include <qt/peertablesortproxy.h>
 #include <qt/platformstyle.h>
 #include <qt/walletmodel.h>
@@ -910,8 +911,7 @@ void RPCConsole::clear(bool keep_prompt)
 
 void RPCConsole::keyPressEvent(QKeyEvent *event)
 {
-    if(windowType() != Qt::Widget && event->key() == Qt::Key_Escape)
-    {
+    if (windowType() != Qt::Widget && GUIUtil::IsEscapeOrBack(event->key())) {
         close();
     }
 }
--- a/src/qt/guiutil.h
+++ b/src/qt/guiutil.h
@@ -428,6 +428,15 @@ namespace GUIUtil
      */
     void ShowModalDialogAndDeleteOnClose(QDialog* dialog);
 
+    inline bool IsEscapeOrBack(int key)
+    {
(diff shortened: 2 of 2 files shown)


hebasto:
cc @icota

src/qt/rpcconsole.cpp
```diff
@@ -910,8 +910,14 @@ void RPCConsole::clear(bool keep_prompt)
 
 void RPCConsole::keyPressEvent(QKeyEvent *event)
 {
-    if(windowType() != Qt::Widget && event->key() == Qt::Key_Escape)
-    {
+    const auto close_key =
+#ifdef Q_OS_ANDROID
```
icota, line 914: I don't like the `#ifdef`. Any way to remap this globally?

icota (changes requested):

src/qt/rpcconsole.cpp
hebasto: @icota Could you elaborate? I'm not sure about `Qt::Key_Back` on non-Android platforms.

icota: One point is that this should be done globally rather than just in `rpcconsole`. In our case `Qt::Key_Back` and `Qt::Key_Escape` will always serve the same purpose.

It also should not be an either-or because Android phones can have keyboards hooked up to them and ChromeOS can run APKs as well.

I guess my suggestion would be something like a global macro/function that checks if its escape or back button. If this is truly the only place where we need to check for the back button then `if (event->key == Qt::Key_Escape || event->key == Qt::Key_Back)` is also cool.

hebasto:
Updated fe035a5fd95944f8c654ee46ba3a4aae245560b3 -> a56a1049380b0acb532681484fbb675c3b2ff365 ([pr493.01](https://github.com/hebasto/gui/commits/pr493.01) -> [pr493.02](https://github.com/hebasto/gui/commits/pr493.02)).

@icota 
> I guess my suggestion would be something like a global macro/function that checks if its escape or back button.

Done.

I didn't find other cases when Android back button does not work as expected, but maybe I just missed them.

src/qt/rpcconsole.cpp
hebasto: Thanks! [Implemented](https://github.com/bitcoin-core/gui/pull/493#issuecomment-985075240).

icota:
utACK https://github.com/bitcoin-core/gui/commit/a56a1049380b0acb532681484fbb675c3b2ff365

PR #496, opened 2021-12-02
Added some safety features from permissions.

KambizAsadzadeh:
Users can now view the permissions requests themselves.

In the Android section, it was observed that no application for a permission has been applied! The codes I sent are enough to solve this problem.

### DIFF
--- a/src/qt/android/src/org/bitcoincore/qt/BitcoinQtActivity.java
+++ b/src/qt/android/src/org/bitcoincore/qt/BitcoinQtActivity.java
@@ -10,6 +10,8 @@ import java.io.File;
 
 public class BitcoinQtActivity extends QtActivity
 {
+
+    final private int REQUEST_CODE_ASK_MULTIPLE_PERMISSIONS = 124;
     @Override
     public void onCreate(Bundle savedInstanceState)
     {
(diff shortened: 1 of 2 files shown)


src/qt/android/src/org/bitcoincore/qt/BitcoinQtActivity.java
```diff
@@ -25,5 +27,65 @@ public void onCreate(Bundle savedInstanceState)
...
+             fuckMarshMallow();
+         }
+    }
+
+@Override
+public void onRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) {
+    switch (requestCode) {
+        case REQUEST_CODE_ASK_MULTIPLE_PERMISSIONS: {
+            Map<String, Integer> perms = new HashMap<String, Integer>();
+            perms.put(Manifest.permission.ACCESS_FINE_LOCATION, PackageManager.PERMISSION_GRANTED);
```
icota, line 42: Why do we need `ACCESS_FINE_LOCATION`?

src/qt/android/src/org/bitcoincore/qt/BitcoinQtActivity.java
```diff
@@ -25,5 +27,65 @@ public void onCreate(Bundle savedInstanceState)
...
+            }
+        }
+        break;
+        default:
+            super.onRequestPermissionsResult(requestCode, permissions, grantResults);
     }
 }
+
+@TargetApi(Build.VERSION_CODES.M)
+private void fuckMarshMallow() {
```
icota, line 61: I like swearing too but it belongs in the comments. Name the function after what it does.

src/qt/android/src/org/bitcoincore/qt/BitcoinQtActivity.java
```diff
@@ -25,5 +27,65 @@ public void onCreate(Bundle savedInstanceState)
...
+    List<String> permissionsNeeded = new ArrayList<String>();
+    final List<String> permissionsList = new ArrayList<String>();
+    if (!addPermission(permissionsList, Manifest.permission.ACCESS_FINE_LOCATION))
+        permissionsNeeded.add("Show Location");
```
icota, line 65: Again I don't understand why we need location.

src/qt/bitcoin.cpp
```diff
@@ -486,6 +486,14 @@ int GuiMain(int argc, char* argv[])
     QApplication::setAttribute(Qt::AA_DontUseNativeMenuBar);
     QApplication::setAttribute(Qt::AA_DontCreateNativeWidgetSiblings);
     QApplication::setAttribute(Qt::AA_DontUseNativeDialogs);
+
+    // Add new safety native permission dialog for R/W actions.
+    static const QString write_permission = "android.permission.WRITE_EXTERNAL_STORAGE";
+    if (QtAndroid::checkPermission(write_permission) == QtAndroid::PermissionResult::Denied)
```
icota, line 492: Is the rationale for this to be able to write to SD card?

icota (changes requested):

src/qt/android/src/org/bitcoincore/qt/BitcoinQtActivity.java
KambizAsadzadeh: Ope! This is my mistake, We don't need to this permission! I'll remove this page n the next commit.

src/qt/bitcoin.cpp
KambizAsadzadeh: Yes, we need to this for internal and removable volumes, such as an SD card, appear in the file system as part of external storage.

KambizAsadzadeh: @icota Thank you for review!
I checked the code again, there is no need for the Java method, and the last changes I made are enough on the C++ side.

src/qt/android/src/org/bitcoincore/qt/BitcoinQtActivity.java
KambizAsadzadeh: This was my mistake, I didn't pay attention to it, you are right.

icota:
Thank you for cleaning up the Java part @KambizAsadzadeh. 

Can you please fix the CI and give us a brief rationale for the PR? What are you able to do with these changes that you weren't able to do before?

KambizAsadzadeh:
> Thank you for cleaning up the Java part @KambizAsadzadeh.
You're welcome!

> Can you please fix the CI and give us a brief rationale for the PR? 
> What are you able to do with these changes that you weren't able to do before?

Check this please:
https://github.com/bitcoin-core/gui/blob/57982f419e36d0023c83af2dd0d683ca3160dc2a/src/qt/android/AndroidManifest.xml#L6

In this section there is a request for permission to write on storage right? Well, In this case a native dialog must be displayed to accept or reject this request and this is important in terms of privacy. This is not possible by default and you must do this manually. :)

We have two ways to do this, one is to use Java code or JNI inside C++ and the other is to use method checkPermission on QtAndroid class. this method checks if the permission was granted or not. This function should be called every time when the application starts for needed permissions, as the users might disable them from Android Settings.

icota:
Thank you for clarifying @KambizAsadzadeh. This is something that is needed but if we follow good practices probably doesn't belong at startup as not all users will opt to use external storage.

According to [the docs](https://developer.android.com/training/permissions/requesting#workflow_for_requesting_permissions):
> Wait for the user to invoke the task or action in your app that requires access to specific private user data. At that time, your app can request the runtime permission that's required for accessing that data.

IMO there should be a GUI prompt/explainer and only then should we trigger the OS modal to ask for these permission. Something to discuss with the design team?

KambizAsadzadeh:
> Thank you for clarifying @KambizAsadzadeh. This is something that is needed but if we follow good practices probably doesn't belong at startup as not all users will opt to use external storage.
> 
> According to [the docs](https://developer.android.com/training/permissions/requesting#workflow_for_requesting_permissions):
> 
> > Wait for the user to invoke the task or action in your app that requires access to specific private user data. At that time, your app can request the runtime permission that's required for accessing that data.
> 

You're welcome, Thank you for your good comments.
I agree with you regarding UX terms, In my opinion, it's better to call the `QtAndroid ::checkPermission` method where this permission is needed instead of startup. Can you show me the part you used for writing external storage?

**Rule 4:**
>Check whether the user has already granted the runtime permission that your app requires. If so, your app can access the private user data. If not, continue to the next step.
You must check whether you have that permission every time you perform an operation that requires that permission.

icota:
> Can you show me the part you used for writing external storage?

I never used external storage, my hardware doesn't even have such an option. 

As it stands right now we use the `.bitcoin` folder in app sandbox. I _think_ I enabled `WRITE_EXTERNAL_STORAGE` because I believed this would be granted without user intervention. In that case one could change the conf and use the SD.

I realise that external storage is a big use-case (people repurposing old phones) so moving forward I suggest we either:
- Document how to enable this manually (long press on the app icon -> go to permissions)
- Design some sort of UI around this (prompt user if wants to use external storage -> trigger permissions modal)

PR #497, opened 2021-12-03
Enable users to configure their monospace font specifically

luke-jr:
This replaces the overly-verbose radio-button font setting (which only allows embedded or autodetected from system) with a simple combobox providing both existing options as well as a custom option to allow the user to select any font of their choice/style.

### DIFF
--- a/src/qt/optionsdialog.cpp
+++ b/src/qt/optionsdialog.cpp
@@ -198,6 +198,15 @@ void OptionsDialog::setModel(OptionsModel *_model)
         setMapper();
         mapper->toFirst();
 
+        const auto& font_for_money = _model->data(_model->index(OptionsModel::FontForMoney, 0), Qt::EditRole).value<OptionsModel::FontChoice>();
+        if (std::holds_alternative<OptionsModel::FontChoiceAbstract>(font_for_money)) {
+            ui->embeddedFont_radioButton->setChecked(font_for_money != OptionsModel::UseBestSystemFont);
+            ui->systemFont_radioButton->setChecked(font_for_money == OptionsModel::UseBestSystemFont);
+        } else {
+            ui->embeddedFont_radioButton->setChecked(false);
+            ui->systemFont_radioButton->setChecked(false);
+        }
+
         updateDefaultProxyNets();
     }
 
@@ -275,7 +284,6 @@ void OptionsDialog::setMapper()
     mapper->addMapping(ui->lang, OptionsModel::Language);
     mapper->addMapping(ui->unit, OptionsModel::DisplayUnit);
     mapper->addMapping(ui->thirdPartyTxUrls, OptionsModel::ThirdPartyTxUrls);
-    mapper->addMapping(ui->embeddedFont_radioButton, OptionsModel::UseEmbeddedMonospacedFont);
 }
 
 void OptionsDialog::setOkButtonState(bool fState)
@@ -337,6 +345,12 @@ void OptionsDialog::on_openBitcoinConfButton_clicked()
 
 void OptionsDialog::on_okButton_clicked()
 {
+    if (ui->embeddedFont_radioButton->isChecked()) {
+        model->setData(model->index(OptionsModel::FontForMoney, 0), QVariant::fromValue(OptionsModel::FontChoice{OptionsModel::FontChoiceAbstract::EmbeddedFont}));
+    } else if (ui->systemFont_radioButton->isChecked()) {
+        model->setData(model->index(OptionsModel::FontForMoney, 0), QVariant::fromValue(OptionsModel::FontChoice{OptionsModel::FontChoiceAbstract::BestSystemFont}));
+    }
+
     mapper->submit();
     accept();
     updateDefaultProxyNets();
--- a/src/qt/optionsdialog.cpp
+++ b/src/qt/optionsdialog.cpp
@@ -23,13 +23,69 @@
 
 #include <chrono>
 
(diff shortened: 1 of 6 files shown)


jarolrod (commented):
Concept ACK, a follow-up can be to use the selected font wherever monospace fonts are used such as the console.

luke-jr:
> a follow-up can be to use the selected font wherever monospace fonts are used such as the console.

A separate configuration for that may make sense , but yes, this makes it much simpler to do that. Figured it would be a bit too much for this PR though :)

RandyMcMillan:
Concept ACK

This should allow users to select a dyslexic font as well (if they have it installed) - which can be helpful when dealing with numbers.

src/qt/optionsdialog.cpp
```diff
@@ -19,15 +19,71 @@
...
 #include <QDir>
+#include <QFontDialog>
 #include <QIntValidator>
 #include <QLocale>
 #include <QMessageBox>
 #include <QSettings>
 #include <QSystemTrayIcon>
 #include <QTimer>
 
+int setFontChoice(QComboBox* cb, const OptionsModel::FontChoice& fc)
```
promag, line 33: move under anonymous namespace?

src/qt/optionsdialog.cpp
```diff
@@ -19,15 +19,71 @@
...
+{
+    int i;
```
promag, line 35:
```
int i = cb->count();
while (i--) {
```

src/qt/optionsdialog.cpp
```diff
@@ -19,15 +19,71 @@
...
+
+    const auto& on_font_choice_changed = [cb, preview](int index) {
+        static int previous_index = -1;
+        QVariant item_data = cb->itemData(index);
+        QFont f;
+        if (item_data.canConvert<OptionsModel::FontChoice>()) {
+            f = OptionsModel::getFontForChoice(item_data.value<OptionsModel::FontChoice>());
+        } else {
+            bool ok;
+            f = QFontDialog::getFont(&ok, GUIUtil::fixedPitchFont(false), cb->parentWidget());
```
promag, line 71: This creates a nested event loop which we are trying to avoid. At least there should be a follow-up refactor to avoid this helper.

promag (commented):
Concept ACK.

src/qt/optionsdialog.cpp
luke-jr: The current code is better/clearer.

hebasto:
@luke-jr still working on this?

luke-jr:
@hebasto Yes, but I don't understand the review comment.

src/qt/optionsdialog.cpp
hebasto: @luke-jr 
> What do you mean?

I think, @promag means an [event loop created by a dialog](https://doc.qt.io/qt-5/qdialog.html#exec).

RandyMcMillan:
Concept ACK

https://github.com/RandyMcMillan/gui/runs/6019248786

Screen Shot 2022-04-14 at 2 58 41 AM
Screen Shot 2022-04-14 at 2 59 07 AM
Screen Shot 2022-04-14 at 2 59 44 AM

src/qt/optionsdialog.cpp
luke-jr: I don't see a reason to avoid it. The caution in the link is not applicable.

hebasto:
@luke-jr  Rebase?

luke-jr:
Rebased

hebasto:
Approach ACK. Tested on Windows 11. Observing kind of jittering of the vertical position of the "Font in the Overview tab:" label when choosing different fonts. Could it be avoided?

luke-jr:
I'm guessing it's the preview font changing? How would you suggest avoiding it?

hebasto:

luke-jr:
Rebased

pablomartin4btc (commented):
tested ACK  a17fd33edd1374145fd6986fbe352295355fde4f

Current UI without this PR:

image

With this PR:

image

The label is moving vertically with some font selection as @hebasto [pointed it out](https://github.com/bitcoin-core/gui/pull/497#issuecomment-1341222673) and even the window gets resized sometimes.

Perhaps to avoid it, we could just leave the view as we have it today but instead of a radio button we fit the combo box in there so there's enough space for the text preview.

In the dialog that's shown once we click on "custom...", shouldn't we filter by "monospace" font types only? @RandyMcMillan, I'm not sure if this could be an issue, however I checked that some dyslexic fonts that are non-monospaced (eg OpenDyslexic) have monospaced versions as well.  

And perhaps limit the size?

image

When reviewing the commit messages, except for the 3rd one (3a6757eed9a24e91e7d800d8026cc3a5c4840141), please take into account the title's length compared to the recommendation in our [guidelines](https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#committing-patches) (eg `GUI: Move "embedded font or not" decision into new OptionsModel::getFontForMoney method`) perhaps when the title doesn't fit you can shorten them with the intention and add both technical and code-specific details in the commit body?

hebasto (approved):
ACK a17fd33edd1374145fd6986fbe352295355fde4f, I have reviewed the code and tested it on Ubuntu 22.04. This is a UX improvement. https://github.com/bitcoin-core/gui/pull/497#issuecomment-1341222673 might be addressed in a follow-up.

PR #506, opened 2021-12-13
QRImageWidget: Sizing and font fixups

luke-jr:
- Margin of QRCode should be constant, not change based on complexity of QRCode
- Text region should be sized based on height of actual text, not a constant guess
- Width of image should not be adjusted based on vertical text size/margins
- Adjusted constant QRCode and margin sizes to visually match what we had previously
- If text won't fit, grow up to 25% wider, or split across multiple lines

(Note: Out of scope for this PR, after #497, I think we should allow customising the font used here and default to the embedded monospace for better privacy)

### DIFF
--- a/src/qt/qrimagewidget.cpp
+++ b/src/qt/qrimagewidget.cpp
@@ -83,7 +83,14 @@ bool QRImageWidget::setQR(const QString& data, const QString& text)
         // Plan how many lines are needed
         QFontMetrics fm(font);
         const int text_width = GUIUtil::TextWidth(fm, text);
-        text_lines = (text_width + max_text_width - 1) / max_text_width;
+        if (text_width > max_text_width && text_width < max_text_width * 5 / 4) {
+            // Allow the image to grow up to 25% wider
+            qr_image_width = text_width + (2 * QR_IMAGE_TEXT_MARGIN);
+            qr_image_x_margin = (qr_image_width - QR_IMAGE_SIZE) / 2;
+            text_lines = 1;
+        } else {
+            text_lines = (text_width + max_text_width - 1) / max_text_width;
+        }
         qr_image_height += (fm.height() * text_lines) + QR_IMAGE_TEXT_MARGIN;
     }
     QImage qrAddrImage(qr_image_width, qr_image_height, QImage::Format_RGB32);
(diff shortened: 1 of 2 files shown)


promag (commented):
Concept ACK, however, I'll try a different approach for the multi-line case.

promag:
@luke-jr feel free to try the commit 01730333996c614dfff12d13037e6673be4c2da1 ([branch](https://github.com/promag/gui/tree/pr/506)) which avoids manual text wrap by using `Qt::TextWrapAnywhere` flag.

Another thing I noticed is that the default minimum font size in `calculateIdealFontSize` is too small.

luke-jr:
```c++
        auto rect = fm.boundingRect(max_rect, Qt::AlignCenter, text);
        if (rect.width() > max_rect.width() * 5 / 4) {
```

Isn't this impossible? I find the unnecessary QRect usage hard to follow.

>Another thing I noticed is that the default minimum font size in calculateIdealFontSize is too small.

IMO out of scope for this PR

promag:
> Isn't this impossible?

No because it doesn't wrap.

> I find the unnecessary QRect usage hard to follow.

And I think it's unnecessary to calculate lines and manually wrap the text.

hebasto:
@luke-jr Screenshots "before" and "after" in the PR description could be useful for further discussion.

jarolrod (commented):
Concept ACK, stylistic changes look fine to me.

| master | PR (rebased) |
| ------- | ------------- |
Screen Shot 2022-06-28 at 2 11 01 AM | Screen Shot 2022-06-28 at 2 12 19 AM |

hebasto (commented):
Tested 4c5143d56c98d4f9f961059692f155bbe2248c44.

This PR effectively makes font size smaller that makes it harder to read.

- master branch:
image

- this PR:
Screenshot from 2022-07-19 15-20-36

pablomartin4btc (commented):
Concept ACK, I agree with the improvements on achieving a more consistent adjustment of the QR code image size. However, it appears that there is still an issue with the font size, which makes the address in the picture difficult to read. Is there a way to improve the legibility of the address text?

pablomartin4btc (commented):
tested ACK 4c5143d56c98d4f9f961059692f155bbe2248c44

- master:

Screenshot from 2023-09-17 18-52-45

- this PR (rebased):

Screenshot from 2023-09-18 16-41-03

PR #509, opened 2021-12-17
Respect dialog modality and fix a regression in wallet unlock

hebasto:
As pointed in bitcoin/bitcoin#23790 a regression in wallet unlock was introduced in bitcoin-core/gui#336 when a synchronous `AskPassphraseDialog` has been replaced with an asynchronous one.

This PR reverts a call back to a synchronous mode.

To make synchronous dialogs behave nice during shutdown some additional changes were made.

Please note that disabling the tray icon menu when a modal dialog is active is useful itself as on master (4ad59042b359f473d5888ecee0c9288dcf98f1c9) it is possible to switch to the "Receive" tab while the GUI is waiting for a password for the "Send" tab:

Screenshot from 2021-12-17 18-59-51

This is confusing and must be avoided.

Fixes bitcoin/bitcoin#23790.

### DIFF
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -268,7 +268,11 @@ void BitcoinApplication::createWindow(const NetworkStyle *networkStyle)
     connect(window, &BitcoinGUI::quitRequested, this, &BitcoinApplication::requestShutdown);
 
     pollShutdownTimer = new QTimer(window);
-    connect(pollShutdownTimer, &QTimer::timeout, window, &BitcoinGUI::detectShutdown);
+    connect(pollShutdownTimer, &QTimer::timeout, [this]{
+        if (!QApplication::activeModalWidget()) {
+            window->detectShutdown();
+        }
+    });
 }
 
 void BitcoinApplication::createSplashScreen(const NetworkStyle *networkStyle)
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -818,16 +818,26 @@ void BitcoinGUI::createTrayIconMenu()
         // Using QSystemTrayIcon::Context is not reliable.
         // See https://bugreports.qt.io/browse/QTBUG-91697
         trayIconMenu.get(), &QMenu::aboutToShow,
-        [this, show_hide_action, send_action, receive_action, sign_action, verify_action] {
+        [this, show_hide_action, send_action, receive_action, sign_action, verify_action, options_action, node_window_action, quit_action] {
             if (show_hide_action) show_hide_action->setText(
                 (!isHidden() && !isMinimized() && !GUIUtil::isObscured(this)) ?
                     tr("&Hide") :
                     tr("S&how"));
-            if (enableWallet) {
-                send_action->setEnabled(sendCoinsAction->isEnabled());
-                receive_action->setEnabled(receiveCoinsAction->isEnabled());
-                sign_action->setEnabled(signMessageAction->isEnabled());
-                verify_action->setEnabled(verifyMessageAction->isEnabled());
+            if (QApplication::activeModalWidget()) {
+                for (QAction* a : trayIconMenu.get()->actions()) {
+                    a->setEnabled(false);
+                }
+            } else {
+                if (show_hide_action) show_hide_action->setEnabled(true);
+                if (enableWallet) {
+                    send_action->setEnabled(sendCoinsAction->isEnabled());
+                    receive_action->setEnabled(receiveCoinsAction->isEnabled());
+                    sign_action->setEnabled(signMessageAction->isEnabled());
+                    verify_action->setEnabled(verifyMessageAction->isEnabled());
+                }
+                options_action->setEnabled(optionsAction->isEnabled());
+                node_window_action->setEnabled(openRPCConsoleAction->isEnabled());
+                if (quit_action) quit_action->setEnabled(true);
             }
         });
 }
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -784,33 +784,16 @@ void BitcoinGUI::createTrayIcon()
 void BitcoinGUI::createTrayIconMenu()
 {
 #ifndef Q_OS_MAC
-    // return if trayIcon is unset (only on non-macOSes)
-    if (!trayIcon)
-        return;
+    if (!trayIcon) return;
+#endif // Q_OS_MAC
 
-    trayIcon->setContextMenu(trayIconMenu.get());
-    connect(trayIcon, &QSystemTrayIcon::activated, [this](QSystemTrayIcon::ActivationReason reason) {
-        if (reason == QSystemTrayIcon::Trigger) {
-            // Click on system tray icon triggers show/hide of the main window
-            toggleHidden();
-        }
-    });
-#else
-    // Note: On macOS, the Dock icon is used to provide the tray's functionality.
-    MacDockIconHandler *dockIconHandler = MacDockIconHandler::instance();
-    connect(dockIconHandler, &MacDockIconHandler::dockIconClicked, [this] {
-        show();
-        activateWindow();
-    });
-    trayIconMenu->setAsDockMenu();
-#endif
-
-    // Configuration of the tray icon (or Dock icon) menu
+    // Configuration of the tray icon (or Dock icon) menu.
 #ifndef Q_OS_MAC
     // Note: On macOS, the Dock icon's menu already has Show / Hide action.
     trayIconMenu->addAction(toggleHideAction);
     trayIconMenu->addSeparator();
-#endif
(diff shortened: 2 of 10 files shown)


prayank23:
I was unable to reproduce the issue mentioned in https://github.com/bitcoin/bitcoin/issues/23790 using PR branch, so issue might be fixed. There was a password prompt each time I tried to send bitcoin from an encrypted wallet.

image

Would like to test few more things and go through each commit since there are 11 commits in this PR before ACK

willyko:
testedACK for the issue mentioned in https://github.com/bitcoin/bitcoin/issues/23790

hebasto:
Updated 702a243d34f8ef1ca1789ee0c9ee70aef7ec821e -> 80e7e303262e149df78cc8f0eb3ce357b03184ca ([pr509.01](https://github.com/hebasto/gui/commits/pr509.01) -> [pr509.02](https://github.com/hebasto/gui/commits/pr509.02), [diff](https://github.com/hebasto/gui/compare/pr509.01..pr509.02)).

Dropped [`QAction::setStatusTip()`](https://doc.qt.io/qt-5/qaction.html#statusTip-prop) calls:
- status tip works only for top-level parent widget, tray icon's `QMenu` instance is not a top-level one
- even on master this feature does not work consistently across all supported platforms
- the tray icon context menu is not local to the main window status bar, so status tips are likely to be unnoticed by users

prayank23:
Is this error related to this PR in any way? 
image

I get this when I try to load a PSBT for signing. This does not happen in v22.0

prayank23 (approved):
tACK https://github.com/bitcoin-core/gui/pull/509/commits/80e7e303262e149df78cc8f0eb3ce357b03184ca

This PR fixes the bug reported in https://github.com/bitcoin/bitcoin/issues/23790 and I did not find any issues with the code or testing other things.

prayank23:
There is some error in CI though which I missed earlier: https://github.com/bitcoin-core/gui/pull/509/checks?check_run_id=4570631121

hebasto:
Updated 80e7e303262e149df78cc8f0eb3ce357b03184ca -> 356e21f68dc953b2efa95fe11954266f655e3eed ([pr509.02](https://github.com/hebasto/gui/commits/pr509.02) -> [pr509.03](https://github.com/hebasto/gui/commits/pr509.03), [diff](https://github.com/hebasto/gui/compare/pr509.02..pr509.03)).

Silenced `-Wunused-variable` warnings for macOS.

prayank23 (approved):
reACK https://github.com/bitcoin-core/gui/pull/509/commits/356e21f68dc953b2efa95fe11954266f655e3eed

promag:
@hebasto have you tried to fix the regression without reverting the other change?

hebasto:
@promag 
> @hebasto have you tried to fix the regression without reverting the other change?

Yes. Although, did not finish it. Changes got more and more evolved.

promag:
@hebasto I can see why, there are multiple cases. Not sure how I didn't  catch `unlockWallet` blocking case on https://github.com/bitcoin-core/gui/pull/336#pullrequestreview-766699378 review.

I think the best course of action is to revert and follow up with necessary refactors.

Concept ACK.

w0xlt (approved):
tACK https://github.com/bitcoin-core/gui/pull/509/commits/356e21f68dc953b2efa95fe11954266f655e3eed

Verified that the error mentioned in https://github.com/bitcoin/bitcoin/issues/23790 happens in master branch and this PR fixes it.

src/qt/bitcoingui.cpp
```diff
@@ -804,16 +804,26 @@ void BitcoinGUI::createTrayIconMenu()
...
                 (!isHidden() && !isMinimized() && !GUIUtil::isObscured(this)) ?
                     tr("&Hide") :
                     tr("S&how"));
-            if (enableWallet) {
-                send_action->setEnabled(sendCoinsAction->isEnabled());
-                receive_action->setEnabled(receiveCoinsAction->isEnabled());
-                sign_action->setEnabled(signMessageAction->isEnabled());
-                verify_action->setEnabled(verifyMessageAction->isEnabled());
+            if (QApplication::activeModalWidget()) {
+                for (QAction* a : trayIconMenu.get()->actions()) {
```
Sjors, line 813: 97692afeb2849b3bd673b2908fd91473b37dbbbb: I don't see any difference in behavior on macOS; before and after this commit, if I open coin selection, it disables everything except the macOS standard items (Options, Show all windows, Hide, Stop)

src/qt/bitcoin.cpp
```diff
@@ -265,7 +265,11 @@ void BitcoinApplication::createWindow(const NetworkStyle *networkStyle)
     connect(window, &BitcoinGUI::quitRequested, this, &BitcoinApplication::requestShutdown);
 
     pollShutdownTimer = new QTimer(window);
-    connect(pollShutdownTimer, &QTimer::timeout, window, &BitcoinGUI::detectShutdown);
+    connect(pollShutdownTimer, &QTimer::timeout, [this]{
```
Sjors, line 268: cda881a9ced639d5f22031cf334140a151d9b505 : on macOS when I'm in coin control, before and after this commit, Command + Q is ignored (not delayed). Same with Stop from the tray menu.

But when I hit ctrl + c in the terminal, before this commit it would shut QT down, with this commit it indeed delays the shutdown until the dialog is closed.

Sjors (commented):
a632303a67e6642aa793af1e74522801b5a12093 indeed fixes the _no lock after send_ regression for me. But it introduces some (much less problematic) regressions, at least on macOS.

Sjors (approved):
tACK 74e638afb6e310ef55b07b74fcf388a82030dacc

src/qt/bitcoingui.cpp
```diff
@@ -790,7 +785,7 @@ void BitcoinGUI::createTrayIconMenu()
     }
 
     // Note: On macOS, the Dock icon's menu already has Show / Hide action.
-    trayIconMenu->addAction(toggleHideAction);
+    QAction* show_hide_action = trayIconMenu->addAction(tr("Show / &Hide"), this, &BitcoinGUI::toggleHidden);
```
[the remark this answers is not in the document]
hebasto: Commit message updated.

hebasto:
@Sjors 

> you may want to delay this keyboard shortcut, because we're past the translation cut-off?

Change is small, so it's okay to get it merged before string freeze (now we are in "soft translation string freeze").

@prayank23 @w0xlt @promag 

Mind having another look into this PR?

hebasto:
On macOS another behavior change noticed. On master, due to https://github.com/bitcoin-core/gui/blob/b7942c94824e4e06489417ee587d1f3ac631fce0/src/qt/bitcoingui.cpp#L316 `optionsAction` has forcibly been hidden in the Dock menu:

Screenshot from 2022-02-09 06-49-29

With this PR `options_action` has no `QAction::PreferencesRole` set, and it is active:

Screenshot from 2022-02-09 06-56-51

@Sjors What do think about this behavior change?

Sjors:
It seems confusing to have "Options..." and "Options ->" in a single dropdown. I don't care deeply, but maybe just keep the original behavior?

prayank23:
> @prayank23 @w0xlt @promag
> 
> Mind having another look into this PR?

I don't care about UI issues with Mac but this pull request addresses an issue which affects security on every platform, works as expected so I have no issues with it. Would have been better if reverting changes were simple and less line of code in this PR. Hopefully we don't ship v23.0 with a known issue that affects security.

ACK https://github.com/bitcoin-core/gui/pull/509/commits/74e638afb6e310ef55b07b74fcf388a82030dacc

hebasto:
Updated 74e638afb6e310ef55b07b74fcf388a82030dacc -> f730bd7d580502ae3c3b5953ada3724b59f5cd9b ([pr509.04](https://github.com/hebasto/gui/commits/pr509.04) -> [pr509.05](https://github.com/hebasto/gui/commits/pr509.05), [diff](https://github.com/hebasto/gui/compare/pr509.04..pr509.05)).

> It seems confusing to have "Options..." and "Options ->" in a single dropdown. I don't care deeply, but maybe just keep the original behavior?

Fixed.

hebasto:
@prayank23 @Sjors 

Would you mind re-acking?

prayank23 (approved):
tACK https://github.com/bitcoin-core/gui/pull/509/commits/f730bd7d580502ae3c3b5953ada3724b59f5cd9b

Tried on Fedora and found no issues.


image

Sjors:
Post merge tACK, [only](https://github.com/hebasto/gui/compare/pr509.04..pr509.05) difference is the `PreferencesRole`.

jarolrod:
Post merge tACK, I was able to reproduce the original issue and I've done testing to ensure that this PR fixes this.

Sjors:
I wonder if this was also a problem when _creating_ a wallet, see #571.

PR #516, opened 2021-12-31
Change 'Show' icon

w0xlt:
This PR changes the 'Show' icon in `receivecoinsdialog.{ui,cpp}`.

The current icon for the 'Show' button is the edit icon, which makes it look like records can be edited on this screen (which is not the case).

The icon already available that seems to be the most suitable for this case is the "eye", so this PR makes this change.

| PR  | Master |
| ------------- | ------------- |
| PR |  master  |

### DIFF
--- a/src/qt/forms/receivecoinsdialog.ui
+++ b/src/qt/forms/receivecoinsdialog.ui
@@ -308,7 +308,7 @@
           </property>
           <property name="icon">
            <iconset resource="../bitcoin.qrc">
-            <normaloff>:/icons/edit</normaloff>:/icons/edit</iconset>
+            <normaloff>:/icons/eye</normaloff>:/icons/eye</iconset>
           </property>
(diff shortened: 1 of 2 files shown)


kristapsk (approved):
utACK 1831d43e8fb73358858907e1d95ce54e1f42b2e7

shaavan (approved):
ACK 1831d43e8fb73358858907e1d95ce54e1f42b2e7

- The `eye` icon makes more sense than a `pen` icon for the "show" option.
- Tested successfully on Ubuntu 20.04 (Using Qt version 5.12.8).

**Screenshot:**

|Master|PR|
|---------|----|
|Master|Screenshot from 2022-01-04 11-59-47|

hebasto (approved):
ACK 1831d43e8fb73358858907e1d95ce54e1f42b2e7

PR #517, opened 2022-01-06
refactor, qt: Use std::chrono for parameters of QTimer methods

hebasto:
Since Qt 5.8 `QTimer` methods have overloads that accept `std::chrono::milliseconds` arguments:
- [`QTimer::singleShot`](https://doc.qt.io/archives/qt-5.9/qtimer.html#singleShot-8)
- [`QTimer::start`](https://doc.qt.io/archives/qt-5.9/qtimer.html#start-2)

### DIFF
--- a/src/qt/clientmodel.cpp
+++ b/src/qt/clientmodel.cpp
@@ -17,6 +17,7 @@
 #include <netbase.h>
 #include <util/system.h>
 #include <util/threadnames.h>
+#include <util/time.h>
 #include <validation.h>
 
 #include <stdint.h>
@@ -288,7 +289,7 @@ static void BlockTipChanged(ClientModel* clientmodel, SynchronizationState sync_
     const bool throttle = (sync_state != SynchronizationState::POST_INIT && !fHeader) || sync_state == SynchronizationState::INIT_REINDEX;
     const int64_t now = throttle ? GetTimeMillis() : 0;
     int64_t& nLastUpdateNotification = fHeader ? nLastHeaderTipUpdateNotification : nLastBlockTipUpdateNotification;
-    if (throttle && now < nLastUpdateNotification + MODEL_UPDATE_DELAY) {
+    if (throttle && now < nLastUpdateNotification + count_milliseconds(MODEL_UPDATE_DELAY)) {
         return;
     }
 
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -41,6 +41,7 @@
 #endif // ENABLE_WALLET
 
 #include <boost/signals2/connection.hpp>
+#include <chrono>
 #include <memory>
 
 #include <QApplication>
@@ -410,10 +411,10 @@ void BitcoinApplication::initializeResult(bool success, interfaces::BlockAndHead
             connect(paymentServer, &PaymentServer::message, [this](const QString& title, const QString& message, unsigned int style) {
                 window->message(title, message, style);
             });
-            QTimer::singleShot(100, paymentServer, &PaymentServer::uiReady);
+            QTimer::singleShot(100ms, paymentServer, &PaymentServer::uiReady);
         }
 #endif
-        pollShutdownTimer->start(200);
+        pollShutdownTimer->start(200ms);
     } else {
         Q_EMIT splashFinished(); // Make sure splash screen doesn't stick around during shutdown
         requestShutdown();
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -414,7 +414,7 @@ void BitcoinApplication::initializeResult(bool success, interfaces::BlockAndHead
             QTimer::singleShot(100ms, paymentServer, &PaymentServer::uiReady);
         }
 #endif
-        pollShutdownTimer->start(200ms);
+        pollShutdownTimer->start(SHUTDOWN_POLLING_DELAY);
     } else {
         Q_EMIT splashFinished(); // Make sure splash screen doesn't stick around during shutdown
         requestShutdown();
--- a/src/qt/guiconstants.h
+++ b/src/qt/guiconstants.h
@@ -5,10 +5,13 @@
 #ifndef BITCOIN_QT_GUICONSTANTS_H
 #define BITCOIN_QT_GUICONSTANTS_H
 
+#include <chrono>
 #include <cstdint>
 
-/* Milliseconds between model updates */
-static const int MODEL_UPDATE_DELAY = 250;
+using namespace std::chrono_literals;
+
+/* A delay between model updates */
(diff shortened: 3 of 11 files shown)


hebasto:
cc @shaavan @promag @jarolrod

promag (commented):
Code review ACK 1009baf2fc5f67d1761ca8474e33198086f82981-

src/qt/clientmodel.cpp
```diff
@@ -288,7 +289,7 @@ static void BlockTipChanged(ClientModel* clientmodel, SynchronizationState sync_
     const bool throttle = (sync_state != SynchronizationState::POST_INIT && !fHeader) || sync_state == SynchronizationState::INIT_REINDEX;
     const int64_t now = throttle ? GetTimeMillis() : 0;
     int64_t& nLastUpdateNotification = fHeader ? nLastHeaderTipUpdateNotification : nLastBlockTipUpdateNotification;
-    if (throttle && now < nLastUpdateNotification + MODEL_UPDATE_DELAY) {
+    if (throttle && now < nLastUpdateNotification + MODEL_UPDATE_DELAY.count()) {
```
shaavan, line 292: nit:
```suggestion
    if (throttle && now < nLastUpdateNotification + count_milliseconds(MODEL_UPDATE_DELAY)) {
```

We could clarify the unit.

src/qt/clientmodel.cpp
```diff
@@ -288,7 +289,7 @@ static void BlockTipChanged(ClientModel* clientmodel, SynchronizationState sync_
     const bool throttle = (sync_state != SynchronizationState::POST_INIT && !fHeader) || sync_state == SynchronizationState::INIT_REINDEX;
     const int64_t now = throttle ? GetTimeMillis() : 0;
```
shaavan, line 290: Could we do the following change?
```suggestion
    const auto now = throttle ? GetTime<std::chrono::milliseconds>() : 0ms;
```

shaavan (approved):
Code Review ACK 1009baf2fc5f67d1761ca8474e33198086f82981

- I prefer using **std::chrono** over **int64_t** for time variables wherever possible. So I agree with the PR's concept.
- Also, the refactor looks clean, and makes code more readable.

I have added some suggestions that might help make this PR even better.

p.s. While reviewing this PR, I came across two functions in the codebase **GetTime** and **GetSystemTime**. So I was curious whether these have different functionalities or are similar in their use case.

src/qt/clientmodel.cpp
jarolrod: We can't do this because we want sys time not mockable time

jarolrod (commented):
ACK 1009baf2fc5f67d1761ca8474e33198086f82981

Code Review and tested ACK. I've verified that all instances of ConfirmMessage have been updated. 

in 94f914bbb9a46d5d595bf5fc6575719e48deac10, since the commit message states `Use std::chrono in parameters of QTimer methods` you could update all instances to use the std::chrono overload, even the 0 ones. Otherwise, this commit is reallly just `Use std::chrono in parameters of QTimer methods time is > 0`

src/qt/clientmodel.cpp
jarolrod: I think this is a fine clarification if others think it makes this code more readable, additionally there is this line of guidance: 

https://github.com/bitcoin-core/gui/blob/6182e5086fc82b49a5f0d2abdab4bc7e2a4686ef/src/util/time.h#L22-L24

src/qt/clientmodel.cpp
jarolrod: We don't want to do this because we want sys time, not mockable time.

shaavan: > we want sys time, not mockable time

That makes sense. Thanks for the clarification.

hebasto:
Updated 1009baf2fc5f67d1761ca8474e33198086f82981 -> f3bdc143b67e8a5e763071a0774f6d994ca35c57 ([pr517.01](https://github.com/hebasto/gui/commits/pr517.01) -> [pr517.02](https://github.com/hebasto/gui/commits/pr517.02)):

- rebased due to the conflict with #441
- addressed @shaavan's [comment](https://github.com/bitcoin-core/gui/pull/517#discussion_r780249318)
- addressed @jarolrod's [comment](https://github.com/bitcoin-core/gui/pull/517#pullrequestreview-847191362)

@jarolrod 
> you could update all instances to use the std::chrono overload, even the 0 ones.

Don't want to choose units for arguments :)

src/qt/clientmodel.cpp
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui/pull/517#issuecomment-1008325109).

jarolrod:
re-ACK f3bdc14

promag (commented):
Code review ACK f3bdc143b67e8a5e763071a0774f6d994ca35c57.

But there are other cases like
https://github.com/bitcoin-core/gui/blob/1f7acfdcca3525edf3eba23535240d23ff0c4fc9/src/qt/trafficgraphwidget.cpp#L161
https://github.com/bitcoin-core/gui/blob/1f7acfdcca3525edf3eba23535240d23ff0c4fc9/src/qt/transactionview.cpp#L118
https://github.com/bitcoin-core/gui/blob/1f7acfdcca3525edf3eba23535240d23ff0c4fc9/src/qt/transactionview.cpp#L126

hebasto:
Thanks @promag 

> But there are other cases like
> https://github.com/bitcoin/bitcoin/blob/1f7acfdcca3525edf3eba23535240d23ff0c4fc9/src/qt/trafficgraphwidget.cpp#L161
> https://github.com/bitcoin/bitcoin/blob/1f7acfdcca3525edf3eba23535240d23ff0c4fc9/src/qt/transactionview.cpp#L118
> https://github.com/bitcoin/bitcoin/blob/1f7acfdcca3525edf3eba23535240d23ff0c4fc9/src/qt/transactionview.cpp#L126

Add a commit which addresses the case in `TransactionView::TransactionView`.

Refactoring of `TrafficGraphWidget` I leaved for an follow up, as it looks non-trivial.

promag (commented):
Code review ACK 51250b0906e56b39488304208ad119c951b4ae7d.

shaavan (approved):
reACK 51250b0906e56b39488304208ad119c951b4ae7d

Changes since my last review:

- `MODEL_UPDATE_DELAY.count()` -> `count_milliseconds(MODEL_UPDATE_DELAY)`
- Consequently `#include <chrono>` -> `#include <util/time.h>` in src/qt/clientmodel.h
- Variable `input_filter_delay` has been converted from **int** to **std::chrono**.

The update takes care of the input_filter_delay. And as for the **TrafficGraphWidget** variable, let me try giving it a shot.

PR #524, opened 2022-01-12
Replace int with std::chrono in for the timer->setInterval() argument

shaavan:
The PR is a follow-up to #517

- It addresses the change suggested in [this](https://github.com/bitcoin-core/gui/pull/517#pullrequestreview-850260826) comment.
- This PR changes the type of `msecsPerSample` from **int** to **std::chrono::minutes** and makes other relevant subsequent changes that were limited to the **trafficgraphwidget** file.

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -54,6 +54,8 @@
 #include <QTimer>
 #include <QVariant>
 
+#include <chrono>
+
 const int CONSOLE_HISTORY = 50;
 const int INITIAL_TRAFFIC_GRAPH_MINS = 30;
 const QSize FONT_RANGE(4, 40);
@@ -1140,7 +1142,7 @@ void RPCConsole::on_sldGraphRange_valueChanged(int value)
 
 void RPCConsole::setTrafficGraphRange(int mins)
 {
-    ui->trafficGraph->setGraphRangeMins(mins);
+    ui->trafficGraph->setGraphRange(std::chrono::minutes{mins});
     ui->lblGraphRange->setText(GUIUtil::formatDurationStr(std::chrono::minutes{mins}));
 }
 
--- a/src/qt/trafficgraphwidget.cpp
+++ b/src/qt/trafficgraphwidget.cpp
@@ -11,6 +11,7 @@
 #include <QColor>
 #include <QTimer>
 
+#include <chrono>
 #include <cmath>
 
 #define DESIRED_SAMPLES         800
@@ -22,7 +23,6 @@ TrafficGraphWidget::TrafficGraphWidget(QWidget *parent) :
     QWidget(parent),
     timer(nullptr),
     fMax(0.0f),
-    nMins(0),
     vSamplesIn(),
     vSamplesOut(),
     nLastBytesIn(0),
@@ -42,10 +42,7 @@ void TrafficGraphWidget::setClientModel(ClientModel *model)
     }
 }
 
-int TrafficGraphWidget::getGraphRangeMins() const
-{
-    return nMins;
-}
+std::chrono::minutes TrafficGraphWidget::getGraphRange() const { return m_range; }
 
 void TrafficGraphWidget::paintPath(QPainterPath &path, QQueue<float> &samples)
 {
@@ -153,12 +150,12 @@ void TrafficGraphWidget::updateRates()
     update();
 }
 
-void TrafficGraphWidget::setGraphRangeMins(int mins)
+void TrafficGraphWidget::setGraphRange(std::chrono::minutes new_range)
 {
-    nMins = mins;
-    int msecsPerSample = nMins * 60 * 1000 / DESIRED_SAMPLES;
+    m_range = new_range;
+    const auto msecs_per_sample{std::chrono::duration_cast<std::chrono::milliseconds>(m_range) / DESIRED_SAMPLES};
     timer->stop();
-    timer->setInterval(msecsPerSample);
+    timer->setInterval(msecs_per_sample);
 
     clear();
 }
--- a/src/qt/trafficgraphwidget.h
+++ b/src/qt/trafficgraphwidget.h
@@ -8,6 +8,8 @@
 #include <QWidget>
 #include <QQueue>
 
+#include <chrono>
+
 class ClientModel;
 
 QT_BEGIN_NAMESPACE
@@ -22,14 +24,14 @@ class TrafficGraphWidget : public QWidget
 public:
     explicit TrafficGraphWidget(QWidget *parent = nullptr);
     void setClientModel(ClientModel *model);
-    int getGraphRangeMins() const;
+    std::chrono::minutes getGraphRange() const;
 
 protected:
     void paintEvent(QPaintEvent *) override;
 
 public Q_SLOTS:
     void updateRates();
-    void setGraphRangeMins(int mins);
+    void setGraphRange(std::chrono::minutes new_range);
     void clear();
 
 private:
@@ -37,7 +39,7 @@ private:
 
     QTimer *timer;
     float fMax;
-    int nMins;
+    std::chrono::minutes m_range{0};
     QQueue<float> vSamplesIn;
     QQueue<float> vSamplesOut;
     quint64 nLastBytesIn;


src/qt/trafficgraphwidget.h
```diff
@@ -22,7 +24,7 @@ class TrafficGraphWidget : public QWidget
 public:
     explicit TrafficGraphWidget(QWidget *parent = nullptr);
     void setClientModel(ClientModel *model);
-    int getGraphRangeMins() const;
+    std::chrono::minutes getGraphRangeMins() const;
```
hebasto, line 27:
```suggestion
    std::chrono::minutes getGraphRange() const;
```

Also s/`setGraphRangeMins`/`setGraphRange`/

src/qt/trafficgraphwidget.h
```diff
@@ -37,7 +39,7 @@ public Q_SLOTS:
 
     QTimer *timer;
     float fMax;
-    int nMins;
+    std::chrono::minutes nMins;
```
hebasto, line 42:
```suggestion
    std::chrono::minutes m_range;
```

src/qt/trafficgraphwidget.cpp
```diff
@@ -11,18 +11,21 @@
...
 #define YMARGIN                 10
 
+using namespace std::chrono_literals;
+
 TrafficGraphWidget::TrafficGraphWidget(QWidget *parent) :
     QWidget(parent),
     timer(nullptr),
     fMax(0.0f),
-    nMins(0),
+    nMins(0min),
```
hebasto, line 28: Move into a member initializer?

src/qt/trafficgraphwidget.h
shaavan: Addressed in 9bb48bf657c7f2d36309807fbc7b55d0ab581a76

src/qt/trafficgraphwidget.h
shaavan: Addressed in 9bb48bf657c7f2d36309807fbc7b55d0ab581a76

src/qt/trafficgraphwidget.cpp
shaavan: Addressed in 9bb48bf657c7f2d36309807fbc7b55d0ab581a76

jonatack (commented):
Concept ACK 9bb48bf657c7f2d36309807fbc7b55d0ab581a76

RandyMcMillan:
nACK 9bb48bf657c7f2d36309807fbc7b55d0ab581a76

This is breaking the Network Traffic tab on MacOS X86_64 &  arm64

---
Darwin ₿ 19.6.0 Darwin Kernel Version 19.6.0: Tue Oct 12 18:34:05 PDT 2021; root:xnu-6153.141.43~1/RELEASE_X86_64 x86_64

Screen Shot 2022-01-14 at 12 17 27 AM
---
Darwin ₿ 20.5.0 Darwin Kernel Version 20.5.0: Sat May  8 05:10:31 PDT 2021; root:xnu-7195.121.3~9/RELEASE_ARM64_T8101 arm64

Screen Shot 2022-01-14 at 12 17 34 AM
---

hebasto (commented):
Tested 9bb48bf657c7f2d36309807fbc7b55d0ab581a76 on Linux Mint 20.3 (Qt 5.12.8) -- the "Network Traffic" tab shows a black window.

src/qt/trafficgraphwidget.cpp
```diff
@@ -153,10 +155,10 @@ void TrafficGraphWidget::updateRates()
     update();
 }
 
-void TrafficGraphWidget::setGraphRangeMins(int mins)
+void TrafficGraphWidget::setGraphRange(std::chrono::minutes mins)
 {
-    nMins = mins;
-    int msecsPerSample = nMins * 60 * 1000 / DESIRED_SAMPLES;
+    m_range = std::chrono::minutes(mins);
+    auto msecsPerSample = m_range * 60 * 1000 / DESIRED_SAMPLES;
```
hebasto, line 161: https://github.com/bitcoin-core/gui/pull/524#pullrequestreview-853590647
> Tested [9bb48bf](https://github.com/bitcoin-core/gui/commit/9bb48bf657c7f2d36309807fbc7b55d0ab581a76) on Linux Mint 20.3 (Qt 5.12.8) -- the "Network Traffic" tab shows a black window.

Could be fixed with
```suggestion
    const auto msecs_per_sample{std::chrono::duration_cast<std::chrono::milliseconds>(m_range) / DESIRED_SAMPLES};
```

shaavan:
Updated from ([pr524.02](https://github.com/shaavan/gui/tree/pr524.02) -> [pr524.03](https://github.com/shaavan/gui/tree/pr524.03))
Addressed @jonatack and @hebasto’s suggestions.

- Done proper casting of **m_range** variable from **mins** to **milliseconds**.
- Used `duration_casting` instead of `static_casting`, as suggested by @hebasto. The reason for doing so is explained in [this](https://stackoverflow.com/a/44277832) article.
- Incorporated other **nit** suggestions.

Tested that the network traffic graph is working correctly now.

src/qt/trafficgraphwidget.cpp
shaavan: Thanks, @hebasto for catching it and giving suggestions on how to correct this misbehavior

Addressed in 5e6ad435b0ed634fca5745668bf949799b306bff

src/qt/rpcconsole.cpp
```diff
@@ -1140,7 +1140,7 @@ void RPCConsole::on_sldGraphRange_valueChanged(int value)
 
 void RPCConsole::setTrafficGraphRange(int mins)
 {
-    ui->trafficGraph->setGraphRangeMins(mins);
+    ui->trafficGraph->setGraphRange(std::chrono::minutes(mins));
```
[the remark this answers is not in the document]
shaavan: Thanks, @jonatack, for nit suggestions, as well as catching the missing header. I intended to add the header, but for some reason, I missed adding it before updating the PR for some reason.

Addressed in 5e6ad43

src/qt/trafficgraphwidget.cpp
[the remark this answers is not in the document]
shaavan: Addressed in 5e6ad43

src/qt/trafficgraphwidget.cpp
```diff
@@ -42,9 +44,9 @@ void TrafficGraphWidget::setClientModel(ClientModel *model)
     }
 }
 
-int TrafficGraphWidget::getGraphRangeMins() const
+std::chrono::minutes TrafficGraphWidget::getGraphRange() const
 {
-    return nMins;
+    return m_range;
```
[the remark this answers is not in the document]
shaavan: I think it's better to incorporate the correct way of writing the code from the get-go.

Addressed in 5e6ad43

src/qt/trafficgraphwidget.cpp
```diff
@@ -11,18 +11,20 @@
...
+#include <chrono>
 #include <cmath>
 
 #define DESIRED_SAMPLES         800
 
 #define XMARGIN                 10
 #define YMARGIN                 10
 
+using namespace std::chrono_literals;
+
```
hebasto, line 23: Why this change?

src/qt/trafficgraphwidget.cpp
```diff
@@ -153,12 +152,12 @@ void TrafficGraphWidget::updateRates()
     update();
 }
 
-void TrafficGraphWidget::setGraphRangeMins(int mins)
+void TrafficGraphWidget::setGraphRange(std::chrono::minutes mins)
 {
-    nMins = mins;
-    int msecsPerSample = nMins * 60 * 1000 / DESIRED_SAMPLES;
+    m_range = std::chrono::minutes{mins};
```
hebasto, line 157: Maybe just
```suggestion
    m_range = mins;
```
?

hebasto (commented):
Approach ACK 5e6ad435b0ed634fca5745668bf949799b306bff, tested on Linux Mint 20.3 (Qt 5.12.8).

About commit message:
- please use "qt: " prefix (see https://github.com/bitcoin-core/gui/blob/master/CONTRIBUTING.md#creating-the-pull-request)
- what about "qt, refactor: Use std::chrono in TrafficGraphWidget class"?
- references like `PR#517` do not work well with separated repositories that use the monotree; you could use `bitcoin-core/gui#517`, or just drop it

src/qt/trafficgraphwidget.h
```diff
@@ -22,22 +24,22 @@ class TrafficGraphWidget : public QWidget
...
-    int getGraphRangeMins() const;
+    std::chrono::minutes getGraphRange() const;
 
 protected:
     void paintEvent(QPaintEvent *) override;
 
 public Q_SLOTS:
     void updateRates();
-    void setGraphRangeMins(int mins);
+    void setGraphRange(std::chrono::minutes mins);
```
hebasto, line 34: Maybe
```suggestion
    void setGraphRange(std::chrono::minutes new_range);
```
?

shaavan:
Updated from  5e6ad43 to f7a19ef ([pr524.03](https://github.com/shaavan/gui/tree/pr524.03) -> [pr524.04](https://github.com/shaavan/gui/tree/pr524.04))
Addressed @hebasto suggestions

Changes:
- Removed the redundant `using namespace std::chrono_literals;` statement.
- Removed unnecessary type conversion of `chrono::minutes` variable.
- Renamed `mins` -> `new_range`.
- Made appropriate changes to commit message as suggested [here](https://github.com/bitcoin-core/gui/pull/524#pullrequestreview-853649608).

src/qt/trafficgraphwidget.cpp
shaavan: I checked through the file and made sure that this statement is not needed.

Removed in f7a19ef774ef92ce348215593e3590a750c345e1

src/qt/trafficgraphwidget.cpp
shaavan: The type-conversion here seems redundant.

Addressed in f7a19ef774ef92ce348215593e3590a750c345e1

src/qt/trafficgraphwidget.h
shaavan: I think `new_range` is a more suitable name for the argument, and now its unit is cleared by its type (i.e., minutes)

Addressed in f7a19ef774ef92ce348215593e3590a750c345e1

hebasto (approved):
ACK f7a19ef774ef92ce348215593e3590a750c345e1

RandyMcMillan:
tACK f7a19ef774ef92ce348215593e3590a750c345e1

on MacOS x86_64 and Arm64

Gemk83 (approved):

src/qt/trafficgraphwidget.h
```diff
@@ -22,22 +24,22 @@ class TrafficGraphWidget : public QWidget
 public:
     explicit TrafficGraphWidget(QWidget *parent = nullptr);
     void setClientModel(ClientModel *model);
-    int getGraphRangeMins() const;
+    std::chrono::minutes getGraphRange() const;
```
promag, line 27: Just remove? It's not used.

promag (commented):
Code review ACK f7a19ef774ef92ce348215593e3590a750c345e1.

src/qt/trafficgraphwidget.h
RandyMcMillan: @promag @shaavan 

I concur - It appears to be part of the original implementation but never used for anything...

it can and should be removed...

https://github.com/bitcoin-core/gui/blame/16781e1bc9f8ffc721ebea73434e0066957bc959/src/qt/trafficgraphwidget.cpp#L45


Screen Shot 2022-01-20 at 1 16 34 PM

RandyMcMillan: @promag - great catch!

shaavan:
Updated from ([pr524.04](https://github.com/shaavan/gui/tree/pr524.04) -> [pr524.05](https://github.com/shaavan/gui/tree/pr524.05))
Addressed @promag suggestions

Changes:
- Removed the unused function `getGraphRange()` from the **TrafficGraphWidget** class.

src/qt/trafficgraphwidget.h
shaavan: Thanks @promag for catching this. Addressed in 5efd391eb1ea1a3551b2335827e4b5577be2aca3

w0xlt (approved):
tACK 5efd391 on Ubuntu 21.10 (Qt 5.15.2).

hebasto (approved):
re-ACK 5efd391eb1ea1a3551b2335827e4b5577be2aca3

src/qt/trafficgraphwidget.cpp
```diff
@@ -153,12 +148,12 @@ void TrafficGraphWidget::updateRates()
     update();
 }
 
-void TrafficGraphWidget::setGraphRangeMins(int mins)
+void TrafficGraphWidget::setGraphRange(std::chrono::minutes new_range)
```
promag, line 151: Drop `m_range` and rename this arg to `range`.

promag: Unless you early return like `if (m_range == new_range) return`, but I think this won't happen in practice.

hebasto: > Drop `m_range` and rename this arg to `range`.

In a separate (first?) commit, no?

promag: Why not the same commit? Just drop `nMin` instead of rename.

hebasto: (a) getting rid of `TrafficGraphWidget::getGraphRangeMins()` and `TrafficGraphWidget::m_range`, and (b) changing type of `TrafficGraphWidget::setGraphRange()` parameter looks really as two focused commits, no?

promag: Either way loos fine to me 👍

shaavan:
Updated from 5efd391 to bdf7cb9 ([pr524.05](https://github.com/shaavan/gui/tree/pr524.05) -> [pr524.06](https://github.com/shaavan/gui/tree/pr524.06))
Addressed @promag suggestion

Changes:
- Removed `m_range` as a member variable of the class TrafficGraphWidget and replaced it with a local `range` variable.

src/qt/trafficgraphwidget.cpp
shaavan: I think this is a valid suggestion. Better to get rid of a redundant member variable.

Addressed in bdf7cb9f0c17bdeeb8389d4b9140aca65e08ac71

@hebasto
> (a) getting rid of TrafficGraphWidget::getGraphRangeMins() and TrafficGraphWidget::nMin, and (b) refactoring of TrafficGraphWidget::setGraphRangeMins() (renaming and parameter type changing) looks really as two focused commits, no?

I agree with your reasoning. However, (a) and (b) are not mutually exclusive, and hence I was having a hard time splitting these two changes while keeping the commits meaningful. Therefore, I reckoned it would be best to make these changes in one commit.

RandyMcMillan:
Approach ACK
Removing the TrafficGraphWidget::getGraphRangeMins() function nACK

I use int TrafficGraphWidget::getGraphRangeMins() const in PR #539 (which fixes issue #532)

I have added a variation to of this PR to #539 https://github.com/bitcoin-core/gui/pull/539/commits/c25d3f42084926c92ae2051278cd785cb1907d56 to ensure compatibility with the chrono lib.

https://github.com/bitcoin-core/gui/blob/c25d3f42084926c92ae2051278cd785cb1907d56/src/qt/trafficgraphwidget.cpp#L46

### Note

https://github.com/bitcoin-core/gui/pull/524/files#diff-da38549cd85e0c422891b8ef1eb3cc33e76f0b80d15ca63aa7ad6f065c8ba694L44

src/qt/trafficgraphwidget.h
RandyMcMillan: > Approach ACK Removing the TrafficGraphWidget::getGraphRangeMins() function nACK
> 
> I use int TrafficGraphWidget::getGraphRangeMins() const in PR #539 (which fixes issue #532)
> 
> I have added a variation to of this PR to #539 [c25d3f4](https://github.com/bitcoin-core/gui/commit/c25d3f42084926c92ae2051278cd785cb1907d56) to ensure compatibility with the chrono lib.
> 
> https://github.com/bitcoin-core/gui/blob/c25d3f42084926c92ae2051278cd785cb1907d56/src/qt/trafficgraphwidget.cpp#L46
> 
> ### Note
> https://github.com/bitcoin-core/gui/pull/524/files#diff-da38549cd85e0c422891b8ef1eb3cc33e76f0b80d15ca63aa7ad6f065c8ba694L44

shaavan:
> I use int TrafficGraphWidget::getGraphRangeMins() const in PR #539 (which fixes issue #532)

I looked through #532, but I couldn't find any use case of `getGraphRangeMins()` there. So I don't think there should be an issue removing this function in this PR.

> I have added a variation of this PR to #539 c25d3f4 to ensure compatibility with the chrono lib.

I looked through it, and it looked good. However, I reckon the primary purpose of #539 is to fix an issue by adding a scaling feature. And the purpose of this PR is refactoring. So I think it would be best to keep these PRs separate.

RandyMcMillan:
If you insist - but to scale the graph - we will have to know the current value of the slider - otherwise the graph can't calculate how many samples to display. It will have to be re-added - it is easier to leave it in. 

https://github.com/bitcoin-core/gui/pull/539

shaavan:
> It will have to be re-added - it is easier to leave it in.

I think you are right. Let me keep the `GetGraphRange()` function and make it consistent with the previous changes.

shaavan:
Updated from bdf7cb9 to cad0e0a ([pr524.06](https://github.com/shaavan/gui/tree/pr524.06) -> [pr524.07](https://github.com/shaavan/gui/tree/pr524.07))
Addressed @RandyMcMillan comment

Changes:
- Restored the `getGraphRange()` function, as it is used in #539. So removing and re-adding the function doesn’t seem optimal.

The updated PR correctly compiles on Ubuntu 20.04, with the Network Traffic widget showing the correct graph.

shaavan:
Reverted back to f7a19ef774ef92ce348215593e3590a750c345e1 ([pr524.07](https://github.com/shaavan/gui/tree/pr524.07) -> [pr524.04](https://github.com/shaavan/gui/tree/pr524.04), [diff](https://github.com/shaavan/gui/compare/pr524.07..pr524.04))

Reason:
- The is almost no difference between pr524.07 and pr524.04 (see the [diff](https://github.com/shaavan/gui/compare/pr524.07..pr524.04)).
- f7a19ef774ef92ce348215593e3590a750c345e1 had quite a few ACKs behind it, by:
    - @hebasto: https://github.com/bitcoin-core/gui/pull/524#pullrequestreview-855595100
    - @RandyMcMillan: https://github.com/bitcoin-core/gui/pull/524#issuecomment-1015866815
    - @promag: https://github.com/bitcoin-core/gui/pull/524#pullrequestreview-858455886

P.S.: To avoid any unintended changes between the PR and f7a19ef774ef92ce348215593e3590a750c345e1, I used the following command to update the PR:
```
git reset --hard HEAD^
git cherry-pick f7a19ef774ef92ce348215593e3590a750c345e1
```
The commit id has changed (which I was not able to restore) but the changes are identical f7a19ef774ef92ce348215593e3590a750c345e1

shaavan:
Restored PR to previous commit f7a19ef774ef92ce348215593e3590a750c345e1. The reason for doing so is explained here: https://github.com/bitcoin-core/gui/pull/524#issuecomment-1028906108

P.s.: Thanks, @hebasto, for your help in restoring this commit.

RandyMcMillan:
@shaavan - thanks for leaving that function in - @rebroad and I have work that utilizes this.

PR #533, opened 2022-01-21
gui: add more detailed address error message

w0xlt:
When reviewing https://github.com/bitcoin/bitcoin/pull/24106, I noticed that GUI does not provide a clear reason for invalid addresses. Basically it just changes the `QLineEdit` to red color.

This PR adds a more detailed error message. The motivations are:

. Provide more information about the error and improve the user experience.
. When a more subtle error occurs, as the one detailed in https://github.com/bitcoin/bitcoin/pull/24106, the user can understand what is happening.

To test this PR, enter invalid addresses in the `Pay To` field on the `Send` tab and also in the address manager (icon next to `Pay To`).

| PR  | 
| ------------- |
| addr_invalid_base58| 
| addr_invalid_bech32 |
| send_invalid_base58 |
| send_invalid_bech32 |
| send_invalid_checksum |


prayank23:
Concept ACK

src/qt/forms/editaddressdialog.ui
```diff
@@ -53,6 +53,16 @@
        </property>
       </widget>
      </item>
+     <item row="2" column="1">
+      <widget class="QLabel" name="errorMessage">
+       <property name="styleSheet">
+        <string notr="true">color:red;font-weight:bold;</string>
```
prayank23, line 59: nit:
```suggestion
        <string notr="true">color:red;</string>
```

src/qt/forms/sendcoinsentry.ui
```diff
@@ -149,71 +217,13 @@
...
-      </property>
-      <property name="alignment">
-       <set>Qt::AlignRight|Qt::AlignTrailing|Qt::AlignVCenter</set>
-      </property>
-      <property name="buddy">
-       <cstring>payAmount</cstring>
+    <item row="0" column="1">
+     <widget class="QLabel" name="errorMessage">
+      <property name="styleSheet">
+       <string notr="true">color:red;font-weight:bold;</string>
```
prayank23, line 223: nit:
```suggestion
       <string notr="true">color:red;</string>
```

src/qt/forms/sendcoinsentry.ui
```diff
@@ -33,20 +33,78 @@
...
+       <widget class="BitcoinAmountField" name="payAmount">
+        <property name="toolTip">
+         <string>The amount to send in the selected unit</string>
+        </property>
+       </widget>
+      </item>
+      <item>
+       <widget class="QCheckBox" name="checkboxSubtractFeeFromAmount">
+        <property name="toolTip">
+         <string>The fee will be deducted from the amount being sent. The recipient will receive less bitcoins than you enter in the amount field. If multiple recipients are selected, the fee is split equally.</string>
```
prayank23, line 78:
```suggestion
         <string>The fee will be deducted from the amount being sent. The recipient will receive less bitcoin than you enter in the amount field. If multiple recipients are selected, the fee is split equally.</string>
```

prayank23:
Tested with and without BOLD error message:

1.

image

2.

image

I would prefer 2 but no strong opinion. There is already so much red in that UI once user moves the cursor with wrong address that bold looks weird.

prayank23:
Some errors in CI: https://github.com/bitcoin-core/gui/pull/533/checks?check_run_id=4895179955

w0xlt:
Forced-pushed with following changes:
. Addressed @prayank23 suggestions in https://github.com/bitcoin-core/gui/pull/533#discussion_r789588871, https://github.com/bitcoin-core/gui/pull/533#discussion_r789589090 and https://github.com/bitcoin-core/gui/pull/533#discussion_r789589875 .

. Manually changed `qt/forms/sendcoinsentry.ui`. Qt Creator is messing up the file with unnecessary code and components like `QPalette`. This caused errors in CI:

```
In file included from qt/sendcoinsentry.cpp:10:
./qt/forms/ui_sendcoinsentry.h: In member function ‘void Ui_SendCoinsEntry::setupUi(QStackedWidget*)’:
./qt/forms/ui_sendcoinsentry.h:231:54: error: ‘PlaceholderText’ is not a member of ‘QPalette’
         palette.setBrush(QPalette::Active, QPalette::PlaceholderText, brush7);
```

src/qt/forms/editaddressdialog.ui
w0xlt: Agreed. Done. Thanks.

src/qt/forms/sendcoinsentry.ui
w0xlt: Done. Thanks.

src/qt/forms/sendcoinsentry.ui
w0xlt: Done. Thanks.

kristapsk:
Concept ACK

jarolrod (commented):
I don't think, at the GUI level, we need to present the user with such detail. A translatable message that conveys the entered address is not correct is enough. If presenting the internal error message is useful for developing purposes, perhaps detailed error messages can be enabled through GUI settings added under a developer mode tab.

We can't display a translated error message with this approach and per our [translation policy](https://github.com/bitcoin-core/gui/blob/master/doc/translation_strings_policy.md)

https://github.com/bitcoin-core/gui/blob/e3ce019667fba2ec50a59814a26566fb67fa9125/doc/translation_strings_policy.md#L26-L27

prayank23:
> Do not translate technical or extremely rare errors. 

These are technical errors so policy is still followed.

katesalazar:
I built and ran this, I think it's an interesting experiment,
but it seems there's reserved space for the label even
when it wouldn't be used. Sorry to say
I would love if the space for the label isn't there
until necessary. 🤷‍♀️

src/qt/forms/editaddressdialog.ui
```diff
@@ -53,6 +53,16 @@
        </property>
       </widget>
      </item>
+     <item row="2" column="1">
+      <widget class="QLabel" name="errorMessage">
+       <property name="styleSheet">
+        <string notr="true">color:red;</string>
```
hebasto, line 59: Is hardcoded red color visually compatible with dark theme? Maybe add some relevant screenshots?

src/qt/forms/sendcoinsentry.ui
jonatack: Per `git grep "less bitcoin"` the codebase currently uses "bitcoins" in this context.

src/qt/forms/editaddressdialog.ui
w0xlt: There is a screenshot posted by @jonatack here:
https://github.com/bitcoin-core/gui/pull/534#pullrequestreview-860753411

But this style is already used in `sendcoinsdialog.ui` and `sendcoinsdialog.cpp` for invalid custom change addresses and insufficient funds , so it seems to be a well tested configuration.

achow101:
I agree with others that it looks a bit weird when there is no error. Perhaps move it to the tooltip?

w0xlt:
@achow101 I proposed an alternative version in #560, which keeps the layout consistent.

Rspigler:
I've changed my opinion on this and I agree with @jarolrod, I don't think this is necessary at the GUI level, and could just confuse users further.  All that is necessary is the address is invalid. Concept NACK from me.

I think #560 is a much better approach

hebasto:
Closing due to lack of interest.

PR #536, opened 2022-01-24
Add address error location to GUI

### LINKED ISSUE #527: Bech32 typo detection
It would be nice to have bech32 typo detection in the GUI.

sipa wrote a demo for this years ago:

Schermafbeelding 2019-09-01 om 11 44 44

http://bitcoin.sipa.be/bech32/demo/demo.html

Moved from https://github.com/bitcoin/bitcoin/issues/16779

meshcollider:
Closes #527. May supersede #533, depending on whether people feel the error reason is better displayed immediately or in this validation dialog.

This is an alternative to #537: I believe we should only show the user potential errors if they ask for it, as the errors are not guaranteed to be correct and enabling/encouraging the user to bruteforce the errors may lead to loss of funds.

Adds a context menu item "Attempt error location" to the address inputs:
image

Which, when clicked, opens a dialog with the error message and highlighted detected errors in red:
image

### DIFF
--- a/src/qt/qvalidatedlineedit.cpp
+++ b/src/qt/qvalidatedlineedit.cpp
@@ -7,10 +7,15 @@
 #include <qt/bitcoinaddressvalidator.h>
 #include <qt/guiconstants.h>
 
+#include <QContextMenuEvent>
+#include <QMenu>
+#include <QMessageBox>
+
 QValidatedLineEdit::QValidatedLineEdit(QWidget *parent) :
     QLineEdit(parent),
     valid(true),
-    checkValidator(nullptr)
+    checkValidator(nullptr),
+    errorLocator(nullptr)
 {
     connect(this, &QValidatedLineEdit::textChanged, this, &QValidatedLineEdit::markValid);
 }
@@ -108,6 +113,11 @@ void QValidatedLineEdit::setCheckValidator(const QValidator *v)
     checkValidator = v;
 }
 
+void QValidatedLineEdit::setErrorLocator(const ErrorLocator *e)
+{
+    errorLocator = e;
+}
+
 bool QValidatedLineEdit::isValid()
 {
     // use checkValidator in case the QValidatedLineEdit is disabled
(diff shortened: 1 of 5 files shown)


src/qt/qvalidatedlineedit.cpp
```diff
@@ -121,3 +131,45 @@ bool QValidatedLineEdit::isValid()
...
+    if (error_message.empty()) {
+        QMessageBox::information(this, "Error locator", "No errors detected.", QMessageBox::Ok);
+    } else {
+        QString error_text = "Error: " + QString::fromStdString(error_message) + "<br />";
+        error_text += "<pre>";
+        if (!error_locations.empty()) {
+            int prev_loc = 0;
+            for (int loc : error_locations) {
+                error_text += input.mid(prev_loc, loc - prev_loc);
+                error_text +=  "<FONT COLOR='#ff0000'>" + input[loc] + "</FONT>";
```
katesalazar, line 154: [is something like this possible here?](https://github.com/bitcoin/bitcoin/blob/e3de7cb9039770e0fd5b8bb8a5cba35c87ae8f00/src/qt/rpcconsole.cpp#L877-L883)

meshcollider: Sure, done

luke-jr:
Approach NACK. User shouldn't need to actively interact like this.

See https://github.com/bitcoin/bitcoin/compare/22.x...luke-jr:gui_bech32_errpos-22+knots for a better approach (I'll rebase it soon)

meshcollider:
I don't think it's as safe to point out potential errors (which may be incorrect) unless the user asks for it.

src/qt/qvalidatedlineedit.cpp
katesalazar: I mean calling red what is red

Sjors:
This seems overly cautious, and I tend to prefer #537.

First of all I suspect most users in most cases will copy-paste addresses, so error correction is only needed in a minority of cases. Bech32 allows [up to 4 typos](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#checksum-design) without ambiguity.

Perhaps we could hide the positions behind a menu like this if there are more than 5 typos.

meshcollider:
> Bech32 allows up to 4 typos without ambiguity.

Only in detection, not in correction. We can only accurately correct two errors, and if more than that are made, the results could be incorrect. This is also only substitution errors - deletion and insertion errors are not handled well.

> Perhaps we could hide the positions behind a menu like this if there are more than 5 typos.

There is no way for the code to know how many errors are made. Anything could happen if more than that are made, including that you chance upon another valid address by accident.

luke-jr:
>There is no way for the code to know how many errors are made. Anything could happen if more than that are made, including that you chance upon another valid address by accident.

Which is why expecting error detection to be perfect is unreasonable, and efforts to hide it behind a menu just a nuisance. ;)

Sjors:
> Only in detection, not in correction. 

Right, but we're not doing correction. That feature might indeed be better for a dropdown menu with a warning when there's more than 2 characters.

> This is also only substitution errors - deletion and insertion errors are not handled well.

For now those impact the length, in which case we mark the whole thing invalid. (though that leaves swapping characters)

> There is no way for the code to know how many errors are made. Anything could happen if more than that are made, including that you chance upon another valid address by accident.

So this is a scenario where the user makes so many typos that they are, in error correcting space, closer to a different address? In that case the user would see a number of characters marked as incorrect, but those would actually not be incorrect. Without autocorrect, a user would then take another look at their piece of paper to see if they misread those characters. That would be confusing.

It could also be dangerous if, rather than looking at the paper, they start randomly permutating the wrong character (assuming it's only 1, maybe 2 if they're really patient).

What are the odds the above scenario. @sipa?

meshcollider:
@Sjors sorry, this I worded things a bit confusingly.

> Right, but we're not doing correction.

We actually are. The error location code is actually computing the specific errors themselves, and would be able to correct them. We deliberately don't show those corrections to the user, just which characters (we think) are wrong. This can only be done with up to 2 characters.

Detection is just "is this whole string valid or not." That's what already exists in master right now (whether the checksum verifies or not). It cannot say anything about the specific location of those errors.

Sjors:
In that case, maybe we should do two things when error positions are found:
1. Highlight them, but not in red
2. Put a warning below the input: "Please check these characters. If they are correct, re-type the whole address. Do not attempt to guess."

achow101:
Concept ACK, although having to get the info by opening a menu item is rather clunky. What about putting the error information in the tooltip?

Also, I think there needs to be a warning that the found errors are not necessarily correct.

Rspigler:
> It could also be dangerous if, rather than looking at the paper, they start randomly permutating the wrong character (assuming it's only 1, maybe 2

I don't see what reasonable person would do this, rather than checking the address they are supposed to be copying. But I agree that something like @Sjors suggestion (https://github.com/bitcoin-core/gui/pull/536#issuecomment-1022137943) would be a good fix

jb55:
>  Although having to get the info by opening a menu item is rather clunky

Agreed that the context menu is clunky and very unintuitive. Perhaps there could be a warning icon that appears next to the input box that you can click to open the dialog. I do like the dialog because it would allow us to put in @Sjors full warning message next to the marked error locations:

> "Please check these characters. If they are correct, re-type the whole address. Do not attempt to guess. The marked characters are our best guess, but it's possible the mistake is somewhere else. Randomly trying different characters may lead to a loss of funds. Please carefully check the original address, e.g. the paper note."

PR #543, opened 2022-02-03
peers-tab: add connection duration column to tableview

RandyMcMillan:
This change adds an "Age" column to the peers table view,
which displays the duration of each peer's connection.

### DIFF
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -79,6 +79,8 @@
 void ForceActivation();
 #endif
 
+using namespace std::chrono_literals;
+
 namespace GUIUtil {
 
 QString dateTimeStr(const QDateTime &date)
@@ -725,6 +727,16 @@ QString formatDurationStr(std::chrono::seconds dur)
     const auto s2{s.count()};
     if (s2 || str_list.empty()) str_list.append(QObject::tr("%1 s").arg(s2));
     return str_list.join(" ");
+}
+
+QString FormatPeerAge(std::chrono::seconds time_connected)
+{
+    const auto time_now{GetTime<std::chrono::seconds>()};
+    const auto age{time_now - time_connected};
+    if (age >= 24h) return QObject::tr("%1 d").arg(age / 24h);
+    if (age >= 1h) return QObject::tr("%1 h").arg(age / 1h);
+    if (age >= 1min) return QObject::tr("%1 m").arg(age / 1min);
+    return QObject::tr("%1 s").arg(age / 1s);
 }
 
 QString formatServicesStr(quint64 mask)
--- a/src/qt/peertablemodel.cpp
+++ b/src/qt/peertablemodel.cpp
@@ -71,6 +71,8 @@ QVariant PeerTableModel::data(const QModelIndex& index, int role) const
         switch (column) {
         case NetNodeId:
             return (qint64)rec->nodeStats.nodeid;
+        case Age:
+            return GUIUtil::FormatPeerAge(rec->nodeStats.m_connected);
         case Address:
             return QString::fromStdString(rec->nodeStats.m_addr_name);
         case Direction:
@@ -96,6 +98,7 @@ QVariant PeerTableModel::data(const QModelIndex& index, int role) const
     } else if (role == Qt::TextAlignmentRole) {
         switch (column) {
         case NetNodeId:
+        case Age:
             return QVariant(Qt::AlignRight | Qt::AlignVCenter);
         case Address:
             return {};
--- a/src/qt/peertablemodel.h
+++ b/src/qt/peertablemodel.h
@@ -47,6 +47,7 @@ public:
 
     enum ColumnIndex {
         NetNodeId = 0,
+        Age,
         Address,
         Direction,
         ConnectionType,
@@ -82,6 +83,9 @@ private:
         /*: Title of Peers Table column which contains a
             unique number used to identify a connection. */
         tr("Peer"),
+        /*: Title of Peers Table column which indicates the duration (length of time)
+            since the peer connection started. */
+        tr("Age"),
         /*: Title of Peers Table column which contains the
             IP/Onion/I2P address of the connected peer. */
         tr("Address"),
--- a/src/qt/peertablesortproxy.cpp
+++ b/src/qt/peertablesortproxy.cpp
@@ -24,6 +24,8 @@ bool PeerTableSortProxy::lessThan(const QModelIndex& left_index, const QModelInd
     switch (static_cast<PeerTableModel::ColumnIndex>(left_index.column())) {
     case PeerTableModel::NetNodeId:
         return left_stats.nodeid < right_stats.nodeid;
+    case PeerTableModel::Age:
+        return left_stats.m_connected > right_stats.m_connected;
     case PeerTableModel::Address:
         return left_stats.m_addr_name.compare(right_stats.m_addr_name) < 0;
     case PeerTableModel::Direction:
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -693,6 +693,7 @@ void RPCConsole::setClientModel(ClientModel *model, int bestblock_height, int64_
             ui->peerWidget->setColumnWidth(PeerTableModel::Subversion, SUBVERSION_COLUMN_WIDTH);
             ui->peerWidget->setColumnWidth(PeerTableModel::Ping, PING_COLUMN_WIDTH);
         }
+        ui->peerWidget->horizontalHeader()->setSectionResizeMode(PeerTableModel::Age, QHeaderView::ResizeToContents);
         ui->peerWidget->horizontalHeader()->setStretchLastSection(true);
         ui->peerWidget->setItemDelegateForColumn(PeerTableModel::NetNodeId, new PeerIdViewDelegate(this));
 
@@ -725,6 +726,7 @@ void RPCConsole::setClientModel(ClientModel *model, int bestblock_height, int64_
             ui->banlistWidget->setColumnWidth(BanTableModel::Address, BANSUBNET_COLUMN_WIDTH);
             ui->banlistWidget->setColumnWidth(BanTableModel::Bantime, BANTIME_COLUMN_WIDTH);
         }
+        ui->banlistWidget->horizontalHeader()->setSectionResizeMode(BanTableModel::Address, QHeaderView::ResizeToContents);
         ui->banlistWidget->horizontalHeader()->setStretchLastSection(true);
 
         // create ban table context menu
--- a/src/qt/guiutil.h
+++ b/src/qt/guiutil.h
@@ -223,6 +223,9 @@ namespace GUIUtil
     /** Convert seconds into a QString with days, hours, mins, secs */
     QString formatDurationStr(std::chrono::seconds dur);
 
+    /** Convert peer connection time to a QString denominated in the most relevant unit. */
+    QString FormatPeerAge(std::chrono::seconds time_connected);
+
     /** Format CNodeStats.nServices bitmask into a user-readable string */
     QString formatServicesStr(quint64 mask);
 


RandyMcMillan:
Screen Shot 2022-02-02 at 9 00 10 PM

RandyMcMillan:
Screenshot of [363f181](https://github.com/bitcoin-core/gui/pull/543/commits/363f1818523dbeb9a0be27b9b0bb0de8bf7084df)

Screen Shot 2022-02-03 at 6 52 47 AM

RandyMcMillan:
commit 281b65b69e894bff136533770ee8eb57d53646a9

we make adjustments so that the peers tab collapses down to a useful "desktop widget".

Screen Shot 2022-02-03 at 8 32 21 AM

RandyMcMillan:
This PR and https://github.com/bitcoin-core/gui/pull/540 will work well together. Both the network graph and peers tab will be useful in a minimized window size and present information in a well formatted way. 
Screen Shot 2022-02-03 at 8 40 42 AM

src/qt/peertablemodel.h
```diff
@@ -90,6 +91,9 @@ public Q_SLOTS:
         tr("Direction"),
         /*: Title of Peers Table column which describes the type of
             peer connection. The "type" describes why the connection exists. */
+        tr("Duration"),
+        /*: Title of Peers Table column which describes the duration of
```
jarolrod, line 95: You added this translator comment in reverse, the comment on lines 95 and 96 should be above line 94 in this diff.

jarolrod (commented):
What is the reason to add this column to the peers table? Why is it a useful metric for a user of the GUI?

RandyMcMillan:
> What is the reason to add this column to the peers table? Why is it a useful metric for a user of the GUI?

I guess one could ask - why `peer id` `address` `direction` `type` `network` `ping` `sent` `received` or `user agent` are useful.

And the fact that there is some much unused visual real estate not being used for anything at all - I am not even sure what the concern or hesitation to adding more columns to the table view would be. On wider screens more than half of the available real-estate  is LITERALLY EMPTY SPACE. 

Screen Shot 2022-02-04 at 6 09 10 AM

RandyMcMillan:
If, by your assertion, that connection duration is NOT USEFUL - why track it? Why not rip it out of the Client Model - a system could use that memory and resources for something else.

src/qt/peertablemodel.h
RandyMcMillan: Thanks for your attention on this.

RandyMcMillan:
The command line equivalent `bitcoin-cli -netinfo 5` we report `age` to the user. So adding "duration" to the gui tableview is a parallel to the command line report. Changing the column name to `age` seems more appropriate than debating whether or not it should be added. I was just following the naming convention in the code. 

Screen Shot 2022-02-04 at 7 22 57 AM

RandyMcMillan:
Note: I like the idea of verbosity in the command line version.
`bitcoin-cli -netinfo`
`bitcoin-cli -netinfo 1`
`bitcoin-cli -netinfo 2`
etc...

If defending the nuance and minutia of gui design changes wasn't so ridiculous (in this repo) - we could consider adding similar functionality to the gui  (levels of verbosity to the table) - but I literally don't get paid enough (approaching nothing) to deal with that type of design process unless every one else was already on board.

prayank23:
Concept ACK

> The connection duration is a useful metric to see when viewing the peer table - sorting the tableview with the connection duration is made possible.

Duration is also useful combined with other things because users can see how much data was sent and received in that duration.

jonatack:
Concept ACK. I do look quite a lot of at age in -netinfo. The direction and type columns should remain together, as combined they form the connection type, like `<->` and `type` in -netinfo, see `enum class ConnectionType` in `src/net.h`. I see age/duration as being next to the id column, like in -netinfo, or the data sent/recv ones as @prayank23 mentioned. It may be good to make the output less verbose somehow (seconds seems unneeded). If you use signet, you won't need to black out sensitive parts of your screenshots.

src/qt/peertablemodel.h
```diff
@@ -88,6 +89,9 @@ public Q_SLOTS:
         /*: Title of Peers Table column which indicates the direction
             the peer connection was initiated from. */
         tr("Direction"),
+        /*: Title of Peers Table column which indicates the duration (length of time)
+            since the peer connection started. */
+        tr("Duration"),
```
jonatack, line 94: I like Age, but if you don't, maybe consider Uptime.

src/qt/rpcconsole.cpp
```diff
@@ -723,6 +724,7 @@ void RPCConsole::setClientModel(ClientModel *model, int bestblock_height, int64_
             ui->banlistWidget->setColumnWidth(BanTableModel::Address, BANSUBNET_COLUMN_WIDTH);
             ui->banlistWidget->setColumnWidth(BanTableModel::Bantime, BANTIME_COLUMN_WIDTH);
         }
+        ui->banlistWidget->horizontalHeader()->setSectionResizeMode( 0, QHeaderView::ResizeToContents);
```
jonatack, line 727: Here and line 694 as well
```suggestion
        ui->banlistWidget->horizontalHeader()->setSectionResizeMode(0, QHeaderView::ResizeToContents);
```

RandyMcMillan:
@jonatack - thanks for the constructive feedback - I appreciate your reasoning. I will re-implement accordingly.

src/qt/rpcconsole.cpp
RandyMcMillan: removed white space

src/qt/peertablemodel.h
RandyMcMillan: used Age as column name

RandyMcMillan:
commit 22b1a40 implements suggested changes.
Squashed commits per feedback

RandyMcMillan:
Screenshots of 22b1a40

Screen Shot 2022-02-04 at 10 09 00 PM

Screen Shot 2022-02-04 at 10 09 11 PM

If anybody wants to send a patch truncating the time format - I will gladly add it with attribution.

prayank23 (approved):
utACK https://github.com/bitcoin-core/gui/pull/543/commits/22b1a40537ca7a62e271cb3962c04cdf2066edec

luke-jr:
Concept ACK

I do note the sort will be the same as sorting by ID...

RandyMcMillan:
@luke-jr - good observation - thanks again for your great feed back - (my apologize for the other thing - I will make the mempool-tab repo more "normal" 😄 )

jonatack (commented):
Debug build is clean and seems to be working. Can you update, squash and fix the commit names, descriptions, and code organization? i.e. make this ready for more final review.

> sorting the tableview with the connection duration is made possible.

Suggest removing the above from the pull description, as it is the same as the existing sort by peer id. Albeit with the current version, the sort orders of Peer and Age are inversed.

RandyMcMillan:
Commit: d01c0b36d11ebd761054bec2c51599ee141238fc

Cleaned up commit structure and messages.
The `Age` column now displays the connection duration in seconds for the first minute - then switches to hour/minute format when connection age is greater than 60 seconds. 

Screen Shot 2022-02-10 at 11 34 03 PM

src/qt/peertablemodel.cpp
```diff
@@ -65,12 +65,19 @@ QVariant PeerTableModel::data(const QModelIndex& index, int role) const
         return QVariant();
 
     CNodeCombinedStats *rec = static_cast<CNodeCombinedStats*>(index.internalPointer());
-
+    const auto time_now{GetTime<std::chrono::seconds>()};
```
jonatack, line 68: 0af493d include chrono
```diff
--- a/src/qt/peertablemodel.cpp
+++ b/src/qt/peertablemodel.cpp
@@ -9,11 +9,14 @@
 
 #include <interfaces/node.h>
 
+#include <chrono>
 #include <utility>
```

src/qt/peertablemodel.cpp
```diff
@@ -65,12 +65,19 @@ QVariant PeerTableModel::data(const QModelIndex& index, int role) const
...
         switch (column) {
         case NetNodeId:
             return (qint64)rec->nodeStats.nodeid;
+        case Age:
+            if(std::chrono::duration_cast<std::chrono::minutes>(time_now-rec->nodeStats.m_connected) < std::chrono::seconds{60}){
+                /* display connection "Age" in seconds for first minute */
+                return GUIUtil::formatDurationStr(std::chrono::duration_cast<std::chrono::seconds>(time_now-rec->nodeStats.m_connected));
+            }else{
+                return GUIUtil::formatDurationStr(std::chrono::duration_cast<std::chrono::minutes>(time_now-rec->nodeStats.m_connected));
+            }
```
jonatack, line 80: d01c0b36 suggestions

- clang-format
- use chrono literals, e.g. `60s` instead of `std::chrono::seconds{60}`
- drop unneeded casts and comment that echoes the code (and shouldn't be in doxygen format here)

<details><summary>suggested diff</summary><p>

```diff 
+using namespace std::chrono_literals;
+
 PeerTableModel::PeerTableModel(interfaces::Node& node, QObject* parent) :
     QAbstractTableModel(parent),
     m_node(node),
@@ -72,11 +75,11 @@ QVariant PeerTableModel::data(const QModelIndex& index, int role) const
         case NetNodeId:
             return (qint64)rec->nodeStats.nodeid;
         case Age:
-            if(std::chrono::duration_cast<std::chrono::minutes>(time_now-rec->nodeStats.m_connected) < std::chrono::seconds{60}){
-                /* display connection "Age" in seconds for first minute */
-                return GUIUtil::formatDurationStr(std::chrono::duration_cast<std::chrono::seconds>(time_now-rec->nodeStats.m_connected));
-            }else{
-                return GUIUtil::formatDurationStr(std::chrono::duration_cast<std::chrono::minutes>(time_now-rec->nodeStats.m_connected));
+            if (time_now - rec->nodeStats.m_connected < 60s) {
+                return GUIUtil::formatDurationStr(time_now - rec->nodeStats.m_connected);
+            } else {
+                return GUIUtil::formatDurationStr(
+                    std::chrono::duration_cast<std::chrono::minutes>(time_now - rec->nodeStats.m_connected));
             }
```
</p></details>

If this becomes larger, say, to display in DD:HH format after 24 hours it may make sense to extract it to a helper function.

src/qt/peertablesortproxy.cpp
```diff
@@ -24,6 +24,8 @@ bool PeerTableSortProxy::lessThan(const QModelIndex& left_index, const QModelInd
     switch (static_cast<PeerTableModel::ColumnIndex>(left_index.column())) {
     case PeerTableModel::NetNodeId:
         return left_stats.nodeid < right_stats.nodeid;
+    case PeerTableModel::Age:
+        return left_stats.m_connected > right_stats.m_connected;
```
jonatack, line 28: The first commit doesn't build cleanly without the second commit adding this line; they should be squashed into the same commit.

```
qt/peertablesortproxy.cpp:24:13: warning: enumeration value 'Age' not handled in switch [-Wswitch]
    switch (static_cast<PeerTableModel::ColumnIndex>(left_index.column())) {
            ^
1 warning generated.
```

src/qt/rpcconsole.cpp
```diff
@@ -723,6 +724,7 @@ void RPCConsole::setClientModel(ClientModel *model, int bestblock_height, int64_
             ui->banlistWidget->setColumnWidth(BanTableModel::Address, BANSUBNET_COLUMN_WIDTH);
             ui->banlistWidget->setColumnWidth(BanTableModel::Bantime, BANTIME_COLUMN_WIDTH);
         }
+        ui->banlistWidget->horizontalHeader()->setSectionResizeMode(0, QHeaderView::ResizeToContents);
```
jonatack, line 727: da2dc046 for each of these two lines, it might be good to comment which column (Age, IP/Netmask) it corresponds to and to update the value if the column order changes (if I understand correctly)

jonatack (commented):
Tested approach ACK for a bit over one hour, this seems much improved.

src/qt/rpcconsole.cpp
RandyMcMillan: I added a generic reference to qt/peertablemodel.h - so future column changes doesn't add to endless revisions to this comment.

src/qt/peertablemodel.cpp
RandyMcMillan: applied in commit 4cacb8c0463307dfc02f077af5df2f79974ff9d2

src/qt/peertablesortproxy.cpp
RandyMcMillan: I re-reviewed [CONTRIBUTING.md](https://github.com/bitcoin-core/gui/blob/master/CONTRIBUTING.md) - it has been awhile - and will adjust my workflow and PR structure accordingly - thanks for the correction.

src/qt/peertablemodel.cpp
RandyMcMillan: The ResizeToContents settings should accommodate the duration string as the connection persists longer than 24 hours.

I applied your [suggested diff](https://github.com/bitcoin-core/gui/commit/cca835e812b875b2ed50ecda9c879f0f79b89f00.diff) with attribution.

src/qt/peertablesortproxy.cpp
jonatack: Great. Yes, each commit should be hygienic (clean debug build and all tests green).

jonatack (commented):
ACK 795a03da76b64e898fdeb5f36be2eb9a4600546e

RandyMcMillan:
Commit [a2e44c1](https://github.com/bitcoin-core/gui/commit/a2e44c1e3838195d0c0187a5f112015fbb1a8fd6)
[add reusable FormatPeerAge() utility helper](https://github.com/bitcoin-core/gui/pull/543/commits/349c597475d68e90596d165ee1cd5f08bbe59e2b) - simplifying over all implementation.

src/qt/peertablemodel.cpp
```diff
@@ -96,10 +98,12 @@ QVariant PeerTableModel::data(const QModelIndex& index, int role) const
     } else if (role == Qt::TextAlignmentRole) {
         switch (column) {
         case NetNodeId:
+        case Age:
             return QVariant(Qt::AlignRight | Qt::AlignVCenter);
         case Address:
             return {};
         case Direction:
+            return QVariant(Qt::AlignCenter);
```
jonatack, line 106: f25a6bae sorry for not noticing before; this line can be dropped

jonatack (commented):
ACK a2e44c1e3838195d0c0187a5f112015fbb1a8fd6

modulo one comment

prayank23 (approved):
tACK https://github.com/bitcoin-core/gui/pull/543/commits/a2e44c1e3838195d0c0187a5f112015fbb1a8fd6

image

src/qt/guiutil.cpp
```diff
@@ -732,6 +734,18 @@ QString formatDurationStr(std::chrono::seconds dur)
...
+QString FormatPeerAge(std::chrono::seconds time_connected)
+{
+    const auto time_now{GetTime<std::chrono::seconds>()};
+    const auto age{time_now - time_connected};
+    if (age < 60s) return QObject::tr("%1 s").arg(age.count());
+    const auto hours{std::chrono::duration_cast<std::chrono::hours>(age).count()};
+    if (hours >= 24) return QObject::tr("%1 d").arg(hours / 24);
+    if (hours) return QObject::tr("%1 h").arg(hours);
+    const auto minutes{std::chrono::duration_cast<std::chrono::minutes>(age).count()};
+    return QObject::tr("%1 m").arg(minutes);
```
jonatack, line 746: @RandyMcMillan, looking at this further, I think the function in 349c597475d68e90596d165ee1cd5f08bbe59e2b can be simplified and improved as follows:

```diff
@@ -736,12 +736,10 @@ QString FormatPeerAge(std::chrono::seconds time_connected)
 {
     const auto time_now{GetTime<std::chrono::seconds>()};
     const auto age{time_now - time_connected};
-    if (age < 60s) return QObject::tr("%1 s").arg(age.count());
-    const auto hours{std::chrono::duration_cast<std::chrono::hours>(age).count()};
-    if (hours >= 24) return QObject::tr("%1 d").arg(hours / 24);
-    if (hours) return QObject::tr("%1 h").arg(hours);
-    const auto minutes{std::chrono::duration_cast<std::chrono::minutes>(age).count()};
-    return QObject::tr("%1 m").arg(minutes);
+    if (age >= 24h) return QObject::tr("%1 d").arg(age / 24h);
+    if (age >= 1h) return QObject::tr("%1 h").arg(age / 1h);
+    if (age >= 1min) return QObject::tr("%1 m").arg(age / 1min);
+    return QObject::tr("%1 s").arg(age / 1s);
 } 
```

jonatack: I repushed an update to https://github.com/jonatack/gui/commits/gui-peers-tab-age-column with the changes in https://github.com/bitcoin-core/gui/pull/543#discussion_r806000745 and https://github.com/bitcoin-core/gui/pull/543#discussion_r805658460 if you want to cherry-pick them.

RandyMcMillan:
Rebased to 25a91a571a4f3453f7e0e9299ee7a40a61d04f19
commit abf93ddcdf5802f9b76e747a7669bc65bbda92cc includes a more compact implementation of the
`QString FormatPeerAge(std::chrono::seconds time_connected)`
helper function.

jonatack:
ACK 35d685a7db407161d267abb699b3181f05e998bb

prayank23 (approved):
reACK https://github.com/bitcoin-core/gui/pull/543/commits/35d685a7db407161d267abb699b3181f05e998bb

RandyMcMillan:
This has been a fun collaborative effort. 😄 

I enjoyed watching a rough idea get polished into an "ACK-able" state.

src/qt/guiutil.cpp
luke-jr: Should probably add translator notes too

shaavan (approved):
ACK 35d685a7db407161d267abb699b3181f05e998bb

I agree with the concept of adding an **Age** column in the Node window table because I agree with all the reasons for its approval discussed above. To briefly reiterate them:
1. We have unused screen real estate which could be used more productively.
2. The **Age** column is displayed in the `-netinfo` command for `bitcoin-cli`. It makes logical sense to keep our GUI consistent with the cli.
3. **Age** column is frequently used by users through `-netinfo` command. So it would be a good idea to add the GUI.

Regarding the code:
- I like the structuring of the `FormatPeerPage` function. The code is clean and easy to understand.
- I like the idea of displaying only the highest degree of time, instead of precise time. This gives users a general idea of how a node has been contacted to them, while also not taking too much screen real estate, and user’s attention.
- I think naming the column “Age” seems appropriate to me. This keeps it consistent with bitcoin-cli's naming.

Here are some screenshots I have added showing the correct working of this PR, on Ubuntu 20.04

|Master|PR|
|---------|----|
|Screenshot from 2022-02-15 17-05-17|PR|

src/qt/guiutil.cpp
jonatack: > Should probably add translator notes too

Agree, am doing this for several of these (FormatPeerAge, formatDurationStr, TimeDurationField, etc.) as a follow-up.

w0xlt (approved):
tACK 35d685a on Ubuntu 21.10 Qt 5.15.2

src/qt/guiutil.cpp
```diff
@@ -732,6 +734,16 @@ QString formatDurationStr(std::chrono::seconds dur)
     return strList.join(" ");
 }
 
+QString FormatPeerAge(std::chrono::seconds time_connected)
```
promag, line 737: abf93ddcdf5802f9b76e747a7669bc65bbda92cc

I think this could be idempotent `QString FormatAge(std::chrono::seconds age)`, to avoid `GetTime`.

Note that `PeerTableModel` has a timer to refresh the data, just call `GetTime` once there.

src/qt/peertablemodel.cpp
```diff
@@ -71,6 +71,8 @@ QVariant PeerTableModel::data(const QModelIndex& index, int role) const
         switch (column) {
         case NetNodeId:
             return (qint64)rec->nodeStats.nodeid;
+        case Age:
+            return GUIUtil::FormatPeerAge(rec->nodeStats.m_connected);
```
promag, line 75: 852b036a7cc0d38f00558125e04528d4cd0fa2dd

Following my previous comment, this could be
```cpp
return GUIUtil::FormatAge(m_now - rec->nodeStats.m_connected);
```

promag (commented):
Concept ACK.

src/qt/guiutil.cpp
jonatack: @promag I agree this would be an improvement and am looking into it, but am not sure how to integrate GetTime into the QTimer code. Pointers welcome.

jonatack:
This pull has 4 ACKs and could perhaps be merged now that v23 has been branched off, which would enable moving forward on planned follow-ups like https://github.com/bitcoin-core/gui/pull/543#discussion_r806779134, while still allowing a possible improvement per https://github.com/bitcoin-core/gui/pull/543#discussion_r820256862.

src/qt/rpcconsole.cpp
```diff
@@ -693,6 +693,8 @@ void RPCConsole::setClientModel(ClientModel *model, int bestblock_height, int64_
             ui->peerWidget->setColumnWidth(PeerTableModel::Subversion, SUBVERSION_COLUMN_WIDTH);
             ui->peerWidget->setColumnWidth(PeerTableModel::Ping, PING_COLUMN_WIDTH);
         }
+        // reference qt/peertablemodel.h for column indexes
+        ui->peerWidget->horizontalHeader()->setSectionResizeMode(1, QHeaderView::ResizeToContents);
```
hebasto, line 697: Drop "magic" numbers?
```suggestion
        ui->peerWidget->horizontalHeader()->setSectionResizeMode(PeerTableModel::Age, QHeaderView::ResizeToContents);
```

src/qt/rpcconsole.cpp
```diff
@@ -725,6 +727,8 @@ void RPCConsole::setClientModel(ClientModel *model, int bestblock_height, int64_
             ui->banlistWidget->setColumnWidth(BanTableModel::Address, BANSUBNET_COLUMN_WIDTH);
             ui->banlistWidget->setColumnWidth(BanTableModel::Bantime, BANTIME_COLUMN_WIDTH);
         }
+        // reference qt/peertablemodel.h for column indexes
+        ui->banlistWidget->horizontalHeader()->setSectionResizeMode(0, QHeaderView::ResizeToContents);
```
hebasto, line 731: Drop "magic" numbers?
```suggestion
        ui->banlistWidget->horizontalHeader()->setSectionResizeMode(BanTableModel::Address, QHeaderView::ResizeToContents);
```

hebasto (changes requested):

RandyMcMillan:
7ddff98 merged requested changes.
Will squash after re-review.

jonatack:
ACK but squashing your changes into the relevant commit (in this case, the last commit 35d685a7db407161d267abb699b3181f05e998bb) would save time for reviewers and speed up the merge.

Jamewood:
(PR #543)

Jamewood:
51708c4

RandyMcMillan:
@Jamewood - thank you for the ACK if this was your intention - please ACK in the appropriate format - and remove the other comments. 😀

shaavan (approved):
reACK 51708c4516cb9d52e84dc8850d93f556dda1a75b

Changes since my last review:

- Dropped “Magic numbers”. i.e.,
`1` → `PeerTableModel::Age`
`0` → `BanTableModel::Address`
- Rebased over master.

I agree with the idea of dropping “magic” numbers because:

1. It makes code easier to understand and reason with.
2. It makes code robust to change in enum ordering under the class’s definition.

hebasto (approved):
ACK 51708c4516cb9d52e84dc8850d93f556dda1a75b, I have reviewed the code and it looks OK, I agree it can be merged.

Issue #545, opened 2022-02-09
SIGABRT on macOS after Dock menu --> Quit


hebasto:
macOS Big Sur 11.6.3 (20G415) + master branch (b7942c94824e4e06489417ee587d1f3ac631fce0):

```
% src/qt/bitcoin-qt -signet        
zsh: abort      src/qt/bitcoin-qt -signet
```
```
% lldb src/qt/bitcoin-qt -- -signet
...
Process 5957 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x00007fff2060f92e libsystem_kernel.dylib`__pthread_kill + 10
libsystem_kernel.dylib`__pthread_kill:
->  0x7fff2060f92e <+10>: jae    0x7fff2060f938            ; <+20>
    0x7fff2060f930 <+12>: movq   %rax, %rdi
    0x7fff2060f933 <+15>: jmp    0x7fff20609ad9            ; cerror_nocancel
    0x7fff2060f938 <+20>: retq   
Target 0: (bitcoin-qt) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0x00007fff2060f92e libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff2063e5bd libsystem_pthread.dylib`pthread_kill + 263
    frame #2: 0x00007fff20593406 libsystem_c.dylib`abort + 125
    frame #3: 0x00000001016f9ca9 QtCore`___lldb_unnamed_symbol171$$QtCore + 9
    frame #4: 0x0000000101a28608 QtCore`QMessageLogger::fatal(char const*, ...) const + 196
    frame #5: 0x0000000101a28abe QtCore`___lldb_unnamed_symbol4042$$QtCore + 46
    frame #6: 0x00000001017015a8 QtCore`QThread::~QThread() + 152
    frame #7: 0x000000010006fcfe bitcoin-qt`RPCConsole::~RPCConsole(this=0x000000010a8b9600) at rpcconsole.cpp:605:1 [opt]
    frame #8: 0x0000000100070030 bitcoin-qt`RPCConsole::~RPCConsole() [inlined] RPCConsole::~RPCConsole(this=0x000000010a8b9600) at rpcconsole.cpp:585:1 [opt]
[...]

PR #547, opened 2022-02-12
Override BitcoinApplication::event() to handle QEvent::Quit

hebasto:
bitcoin-core/gui#336 introduced a regression when termination requests from a platform are not handled properly.

This PR fixes this regression. On macOS shutdown after clicking "Quit" in Dock icon menu, and during logout works again.

Fixes bitcoin-core/gui#545.

### DIFF
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -450,6 +450,16 @@ WId BitcoinApplication::getMainWinId() const
     return window->winId();
 }
 
+bool BitcoinApplication::event(QEvent* e)
+{
+    if (e->type() == QEvent::Quit) {
+        requestShutdown();
+        return true;
+    }
+
+    return QApplication::event(e);
+}
+
 static void SetupUIArgs(ArgsManager& argsman)
 {
     argsman.AddArg("-choosedatadir", strprintf("Choose data directory on startup (default: %u)", DEFAULT_CHOOSE_DATADIR), ArgsManager::ALLOW_ANY, OptionsCategory::GUI);
--- a/src/qt/bitcoin.h
+++ b/src/qt/bitcoin.h
@@ -92,6 +92,9 @@ Q_SIGNALS:
     void splashFinished();
     void windowShown(BitcoinGUI* window);
 
+protected:
+    bool event(QEvent* e) override;
+
 private:
     std::optional<InitExecutor> m_executor;
     OptionsModel *optionsModel;


RandyMcMillan:
tACK e7fc50681e99e3c726db2bc4d3d425ed8a0fc6b3

Tested on MacOS Darwin ₿ 19.6.0 Darwin Kernel Version 19.6.0: Sun Nov 14 19:58:51 PST 2021; root:xnu-6153.141.50~1/RELEASE_X86_64 x86_64


Tested on MacOS Darwin m1.deepspace.host 21.3.0 Darwin Kernel Version 21.3.0: Wed Jan  5 21:37:58 PST 2022; root:xnu-8019.80.24~20/RELEASE_ARM64_T8101 arm64

Sjors:
tACK e7fc50681e99e3c726db2bc4d3d425ed8a0fc6b3 (rebased on master) indeed fixes the crash described in #545

Sjors (approved):
tACK e7fc50681e99e3c726db2bc4d3d425ed8a0fc6b3 (rebased on master) indeed fixes the crash described in #545

src/qt/bitcoin.cpp
```diff
@@ -450,6 +450,16 @@ WId BitcoinApplication::getMainWinId() const
     return window->winId();
 }
 
+bool BitcoinApplication::event(QEvent* e)
+{
+    if (e->type() == QEvent::Quit) {
```
luke-jr, line 455: Is there a reason `QEvent::Quit` is undocumented (and apparently unused in anything else Google finds)?

hebasto: I don't know.

saadbitcoin:
Hi
I am can not mining bitcoin cor

src/qt/bitcoin.cpp
luke-jr: Would be nice if there's a way we can fix this without undocumented (undefined?) behaviour.

Worst case, partially reverting #336 instead? :/

hebasto: > Is there a reason `QEvent::Quit` is undocumented (and apparently unused in anything else Google finds)?

> Would be nice if there's a way we can fix this without undocumented (undefined?) behaviour.

Qt codease, as every other codebase, is not perfect. And if documentation diverges from code, it is an bug in documentation (https://bugreports.qt.io/browse/QTBUG-92122) which has already been fixed but not backported.

> Worst case, partially reverting #336 instead? :/

Yes, this solution is also possible. Only there are two concerns:
- do we really want to return to undocumented way to interrupt the main event loop during application run?
- is it feasible for 23.0 considering [release schedule](https://github.com/bitcoin/bitcoin/issues/22969)?

luke-jr: >it is an bug in documentation (https://bugreports.qt.io/browse/QTBUG-92122) which has already been fixed but not backported.

Ok, seems fine if that's all it is

promag (approved):
Tested ACK e7fc50681e99e3c726db2bc4d3d425ed8a0fc6b3 on macOS 10.15 with Qt 5.15.2.

PR #549, opened 2022-02-15
refactor: use std::chrono for formatDurationStr() helper

jonatack:
Updates `formatDurationStr()` to use the `chrono` standard lib. No change in behavior.

### DIFF
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -713,23 +713,18 @@ QString ConnectionTypeToQString(ConnectionType conn_type, bool prepend_direction
 
 QString formatDurationStr(std::chrono::seconds dur)
 {
-    const auto secs = count_seconds(dur);
-    QStringList strList;
-    int days = secs / 86400;
-    int hours = (secs % 86400) / 3600;
-    int mins = (secs % 3600) / 60;
-    int seconds = secs % 60;
-
-    if (days)
-        strList.append(QObject::tr("%1 d").arg(days));
-    if (hours)
-        strList.append(QObject::tr("%1 h").arg(hours));
-    if (mins)
-        strList.append(QObject::tr("%1 m").arg(mins));
-    if (seconds || (!days && !hours && !mins))
-        strList.append(QObject::tr("%1 s").arg(seconds));
-
-    return strList.join(" ");
+    using days = std::chrono::duration<int, std::ratio<86400>>; // can remove this line after C++20
+    const auto d{std::chrono::duration_cast<days>(dur)};
+    const auto h{std::chrono::duration_cast<std::chrono::hours>(dur - d)};
+    const auto m{std::chrono::duration_cast<std::chrono::minutes>(dur - d - h)};
+    const auto s{std::chrono::duration_cast<std::chrono::seconds>(dur - d - h - m)};
+    QStringList str_list;
+    if (auto d2{d.count()}) str_list.append(QObject::tr("%1 d").arg(d2));
+    if (auto h2{h.count()}) str_list.append(QObject::tr("%1 h").arg(h2));
+    if (auto m2{m.count()}) str_list.append(QObject::tr("%1 m").arg(m2));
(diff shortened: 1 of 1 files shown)


shaavan (approved):
ACK 6f2593dc23565abaa3d176595cba6e07883f512e

The updated code looks clean and is straightforward to understand. I was able to compile the PR branch successfully and test that updated `formatDurationStr` is providing correct results for the `ui->peerConnTime`.

Here is a screenshot of the correct working of the PR.

PR

As a small nit suggestion, I think it would be a good idea to space out the code a little bit into logical sections. But this is a non-critical suggestion, and you may take it up in case you are going to update the PR for some other reasons.

w0xlt (approved):
tACK 6f2593d on Ubuntu 21.10 Qt 5.15.2

src/qt/guiutil.cpp
```diff
@@ -713,23 +713,18 @@ QString ConnectionTypeToQString(ConnectionType conn_type, bool prepend_direction
...
-        strList.append(QObject::tr("%1 s").arg(seconds));
-
-    return strList.join(" ");
+    using days = std::chrono::duration<int, std::ratio<86400>>; // can remove this line after C++20
+    const auto d{std::chrono::duration_cast<days>(dur)};
+    const auto h{std::chrono::duration_cast<std::chrono::hours>(dur - d)};
+    const auto m{std::chrono::duration_cast<std::chrono::minutes>(dur - d - h)};
+    const auto s{std::chrono::duration_cast<std::chrono::seconds>(dur - d - h - m)};
+    QStringList str_list;
+    if (auto d2{d.count()}) str_list.append(QObject::tr("%1 d").arg(d2));
```
hebasto, line 722: Maybe add translator comments, here and after?

jonatack: Yes, doing for several of these (FormatPeerAge, formatDurationStr, TimeDurationField, etc.) as a follow-up to this and to #543 (see https://github.com/bitcoin-core/gui/pull/543#discussion_r806779134).

src/qt/guiutil.cpp
```diff
@@ -713,23 +713,18 @@ QString ConnectionTypeToQString(ConnectionType conn_type, bool prepend_direction
...
+    if (auto h2{h.count()}) str_list.append(QObject::tr("%1 h").arg(h2));
+    if (auto m2{m.count()}) str_list.append(QObject::tr("%1 m").arg(m2));
+    const auto s2{s.count()};
+    if (s2 || str_list.empty()) str_list.append(QObject::tr("%1 s").arg(s2));
```
promag, line 726: 6f2593dc23565abaa3d176595cba6e07883f512e

nit, suggestion:
```cpp
if (auto s2{s.count()}) str_list.append(QObject::tr("%1 s").arg(s2));
return str_list.empty() ? "0 s" : str_list.join(" ");
```

promag (approved):
Code review ACK 6f2593dc23565abaa3d176595cba6e07883f512e.

Maybe leave a comment that `<format>` can be used once we jump to C++20.

src/qt/guiutil.cpp
jonatack: Sure, will do if have to retouch or will sneak it into the translator comments follow-up.

RandyMcMillan: Note: When following up - "Connection Age" will bring the naming convention together - in the gui as well as CL

[tr("Age")](https://github.com/bitcoin-core/gui/pull/543/files#diff-3fcbd1d49f399ac9b4656011ea03e7ef80a855482d6de33855e6326ec065321fR88
)

RandyMcMillan:
tACK 6f2593dc23565abaa3d176595cba6e07883f512e

tested on macOS x86_64, Arm64

Screen Shot 2022-03-01 at 3 17 51 PM

Screen Shot 2022-03-01 at 3 19 26 PM

PR #553, opened 2022-02-18
Change address / amount error background

w0xlt:
This PR proposes a small change in QLineEdit when there is an error in the input.

master | 
--- |
image 

PR |
--- | 
image |

This also shows good results when combined with other open PRs.

#537 | 
--- |
image

#533 |
--- |
image |

### DIFF
--- a/src/qt/guiconstants.h
+++ b/src/qt/guiconstants.h
@@ -25,7 +25,7 @@ static const int STATUSBAR_ICONSIZE = 16;
 static const bool DEFAULT_SPLASHSCREEN = true;
 
 /* Invalid field background style */
-#define STYLE_INVALID "background:#FF8080"
+#define STYLE_INVALID "border: 3px solid #FF8080"
 
 /* Transaction list -- unconfirmed transaction */
 #define COLOR_UNCONFIRMED QColor(128, 128, 128)


promag:
IMO this improves text contrast.

However, other improvements could be made as well, some suggestions:
 - consistent error message near the invalid fields, for instance, the error message for invalid change address is displayed on the right while the upcoming bech32 error shows on the top
 - the amount field only turns red when clicking send and it's zero
 - the send button remains enabled while a field is invalid, maybe it should be disabled

jarolrod (commented):
ACK fe7c81e34e2e16c4a5ec967645ebb49e161d3a25

I'm ACK on this change specifically, there could be an accessibility improvement to this change (ping @RandyMcMillan)

I disagree that, in the case of #537, this "shows good results when combined with other open PRs" :)

RandyMcMillan:
@jarolrod - thanks for pinging me on this - my apologizes for not seeing it sooner...

@w0xlt,


Just to define the issue with more clarity,

We can see using this color contrast calculator that in the light theme (on Mac) that the black text on the [rgb(255,128,128)](https://colordesigner.io/#FF8080) is ok.

Screen Shot 2022-03-16 at 5 59 59 AM

But when we switch to the dark theme (on Mac) the contrast becomes unsatisfactory from a contrast perspective and is apparent just by looking at it. 

Screen Shot 2022-03-16 at 6 00 29 AM

RandyMcMillan:
When fe7c81e34e2e16c4a5ec967645ebb49e161d3a25 is tested against light/dark theme changes (on Mac) we get...

Screen Shot 2022-03-16 at 6 41 19 AM

Screen Shot 2022-03-16 at 6 42 05 AM

which doesn't work because the test color is changed bases on system theme...

RandyMcMillan:
Screen Shot 2022-03-16 at 7 26 55 AM

Screen Shot 2022-03-16 at 7 27 09 AM

Screen Shot 2022-03-16 at 7 27 35 AM

RandyMcMillan:
Here are a couple patches to experiment with [palleteRole](https://doc.qt.io/qt-5/stylesheet-reference.html#paletterole) in Qt:
You will have to call the text field to update during the system theme change - other wise it will only update when it receives focus.

```shell
diff --git a/src/qt/guiconstants.h b/src/qt/guiconstants.h
index 60e8ee2420..8b37c5dfb4 100644
--- a/src/qt/guiconstants.h
+++ b/src/qt/guiconstants.h
@@ -25,7 +25,7 @@ static const int STATUSBAR_ICONSIZE = 16;
 static const bool DEFAULT_SPLASHSCREEN = true;
 
 /* Invalid field background style */
-#define STYLE_INVALID "border: 3px solid #FF8080"
+#define STYLE_INVALID "color: palette(highlighted-text); background-color: pallet(shadow); border: 3px solid #FF8080"
 
 /* Transaction list -- unconfirmed transaction */
 #define COLOR_UNCONFIRMED QColor(128, 128, 128)

```

```shell
diff --git a/src/qt/guiconstants.h b/src/qt/guiconstants.h
index 60e8ee2420..dca896ca7e 100644
--- a/src/qt/guiconstants.h
+++ b/src/qt/guiconstants.h
@@ -25,7 +25,7 @@ static const int STATUSBAR_ICONSIZE = 16;
 static const bool DEFAULT_SPLASHSCREEN = true;
 
 /* Invalid field background style */
-#define STYLE_INVALID "border: 3px solid #FF8080"
+#define STYLE_INVALID "color: palette(highlighted-text); background-color: pallet(window); border: 3px solid #FF8080"
 
 /* Transaction list -- unconfirmed transaction */
 #define COLOR_UNCONFIRMED QColor(128, 128, 128)

```

w0xlt:
@RandyMcMillan  Thanks for the detailed review and suggestions.
I will apply and test them.

jarolrod (approved):
Revisited this, and it is in fact an improvement. Specifically, there's is an improvement for text contrast when in dark mode. Additionally it makes the UI language around a sending error consistent.

improvement with text contrast in dark mode:
|  master | PR |
| ------- | --- |
| Screen Shot 2022-06-27 at 9 45 50 PM | Screen Shot 2022-06-27 at 9 46 19 PM |

shaavan (approved):
ACK fe7c81e34e2e16c4a5ec967645ebb49e161d3a25

- This change has improved over the current way of handling erroneous values.
- Objectively, this PR shows definite accessibility improvements, as shown by @RandyMcMillan's thorough review.
- On a subjective note, I find the PR's way of handling errors cleaner than the master's way. 

Screenshot:

|Master|PR|
|---------|----|
|Screenshot from 2022-07-01 14-43-58|Screenshot from 2022-07-01 17-48-54|

w0xlt:
@RandyMcMillan  I tried your suggestions (`palette(window)`, `palette(shadow)`) and also `palette(base)`, but none of them worked.
However I didn't test on MacOS, I did it on Ubuntu by [setting the theme using Qt5 Settings](https://www.reddit.com/r/Ubuntu/comments/qbam6e/why_isnt_dark_theme_working_on_all_application/).

Rspigler:
I actually prefer the red highlighting. IMO it's more contrasting and more of an alert

hebasto:
@GBKS

Mind sharing your designer's opinion regarding this PR approach.

GBKS:
ACK fe7c81e

Just got the DrahtBot notification and realized I was asked for my input. Sorry, for the delay on this. I agree with this change. The text remains more legible. Also visually less jarring with the outline.

If one wanted to keep a red background, it could also be a red with transparency (15-25% might be enough). That would avoid the thick border, still draw enough attention, and not impact contrast so much.

PR #526, opened 2022-01-20
Add address relay/processed/rate-limited fields to peer details

jonatack:
This pull adds the following address fields in rpc getpeerinfo and cli -netinfo to the gui peers details:

- Address Relay (Yes/No)
- Addresses Processed (integer)
- Addresses Rate-Limited (integer)

and uses the additional horizontal space to display "Last Transaction" (instead of "Last Tx").

Screenshot from 2022-01-21 00-05-49

PR #554, opened 2022-02-20
Add and improve translator comments and tooltips for peers tab address fields

jonatack:
Per translator feedback in this thread: https://github.com/bitcoin-core/gui/pull/526#discussion_r809237830

*"The lack of string context in Transifex is a real problem for this project, as proper context (dev notes and/or screenshots) are essential to achieve quality translations."*

This pull adds developer notes for transifex translators via `extracomment` tags, and it improves the existing ones and their tooltips with more context, clarity and completeness for the following peer tab fields as a follow-up to bitcoin-core/gui#526:

- address relay
- addresses processed
- addressed rate-limited

It looks like only six lines of diff, but they are loooong lines.

If this is the right direction, the same can be done for other fields in follow-ups.

### DIFF
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -1620,10 +1620,10 @@
                <item row="24" column="0">
                 <widget class="QLabel" name="peerAddrProcessedLabel">
                  <property name="toolTip">
-                  <string extracomment="Tooltip text for the Addresses Processed field in the peer details area.">Total number of addresses processed, excluding those dropped due to rate-limiting.</string>
+                  <string extracomment="Tooltip text for the Addresses Processed field in the peer details area, which displays the total number of addresses received from this peer that were processed (excludes addresses that were dropped due to rate-limiting).">The total number of addresses received from this peer that were processed (excludes addresses that were dropped due to rate-limiting).</string>
                  </property>
                  <property name="text">
-                  <string>Addresses Processed</string>
+                  <string extracomment="Text title for the Addresses Processed field in the peer details area, which displays the total number of addresses received from this peer that were processed (excludes addresses that were dropped due to rate-limiting).">Addresses Processed</string>
                  </property>
                 </widget>
                </item>
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -1646,10 +1646,10 @@
                <item row="25" column="0">
                 <widget class="QLabel" name="peerAddrRateLimitedLabel">
                  <property name="toolTip">
-                  <string extracomment="Tooltip text for the Addresses Rate-Limited field in the peer details area.">Total number of addresses dropped due to rate-limiting.</string>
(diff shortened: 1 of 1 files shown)


src/qt/forms/debugwindow.ui
```diff
@@ -1594,10 +1594,10 @@
                <item row="23" column="0">
                 <widget class="QLabel" name="peerAddrRelayEnabledLabel">
                  <property name="toolTip">
-                  <string extracomment="Tooltip text for the Address Relay field in the peer details area.">Whether we relay addresses to this peer.</string>
+                  <string extracomment="Tooltip text for the Address Relay field in the peer details area, which displays whether we relay addresses to this peer (Yes/No).">Whether we relay addresses to this peer (Yes/No).</string>
```
promag, line 1597: Why the 2nd Yes/No?

promag (commented):
Concept ACK, seems to be in line with what was suggested/requested.

src/qt/forms/debugwindow.ui
jonatack: Which one would you remove?

promag: The second, leave the string unchanged.

jonatack: Thanks, done.

jonatack:
Updated per @promag feedback (thanks!)

jarolrod (approved):
ACK bcd72ef18f14d7a457db6c8b835a571dec1f2beb

I've been collecting thoughts on how to formulate translator comments [here](https://github.com/bitcoin-core/gui-qml/blob/main/src/qml/doc/translator-comments.md) and the translator comments introduced here tick the boxes. 

in reference to the translator comment documentation I've prepared: in 1df9daaa457a57554ba2b334f989bf5d599177bf you allude to what the value of the field is, I wonder if I should include that point as required context in the doc.

If you retouch, and definitely not necessary, you could rename the PR to `Add and improve peers tab address translator comments and tooltips` and similarly update the commits to state they are updating translator comments instead of extracomment.

jonatack:
> ACK [bcd72ef](https://github.com/bitcoin-core/gui/commit/bcd72ef18f14d7a457db6c8b835a571dec1f2beb)
> 
> in reference to the translator comment documentation I've prepared: in [1df9daa](https://github.com/bitcoin-core/gui/commit/1df9daaa457a57554ba2b334f989bf5d599177bf) you allude to what the value of the field is, I wonder if I should include that point as required context in the doc.

Describing the value of the field where relevant and helpful seems like a good idea.

> If you retouch, and definitely not necessary, you could rename the PR to `Add and improve peers tab address translator comments and tooltips` and similarly update the commits to state they are updating translator comments instead of extracomment.

Done!

jarolrod (commented):
re-ACK [4d2b503](https://github.com/bitcoin-core/gui/commit/4d2b503d6cbb593cf75cf053fd480de64b3f4fc0)

Thanks for picking up the suggestions!

hebasto (approved):
ACK 4d2b503d6cbb593cf75cf053fd480de64b3f4fc0, I have reviewed the code and it looks OK, I agree it can be merged.

PR #555, opened 2022-02-21
Restore Send button when using external signer

### LINKED ISSUE #551: External signer Send button disabled
When using an external signer, the Create Unsigned button actually signs on the hardware device and the Send button is incorrectly disabled. Probably a regression from #441.

Schermafbeelding 2022-02-18 om 17 27 00

cc @achow101

Sjors:
Fixes #551

For the simplest use case of a wallet with one external signer and "PSBT Controls" disabled in settings (the default), the send dialog will behave the same as when using a wallet with private keys. I.e. there's no "Create Unsigned" button.

When PSBT controls are turned on, you can now actually make a PSBT with signing it; before this PR that button would trigger a sign event and would broadcast the transaction.

In case a multisig, the Send button will sign on the device, and then fall back to presenting a PSBT (same behavior as before #441).

This PR starts with two refactoring commits to move some stuff into a helper function for improved readability in general, and to make the main commit easier to review.

### DIFF
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -447,6 +447,34 @@ void SendCoinsDialog::presentPSBT(PartiallySignedTransaction& psbtx)
     } // msgBox.exec()
 }
 
+bool SendCoinsDialog::signWithExternalSigner(PartiallySignedTransaction& psbtx, CMutableTransaction& mtx, bool& complete) {
+    TransactionError err;
+    try {
+        err = model->wallet().fillPSBT(SIGHASH_ALL, /*sign=*/true, /*bip32derivs=*/true, /*n_signed=*/nullptr, psbtx, complete);
+    } catch (const std::runtime_error& e) {
+        QMessageBox::critical(nullptr, tr("Sign failed"), e.what());
+        return false;
+    }
+    if (err == TransactionError::EXTERNAL_SIGNER_NOT_FOUND) {
+        //: "External signer" means using devices such as hardware wallets.
+        QMessageBox::critical(nullptr, tr("External signer not found"), "External signer not found");
+        return false;
+    }
+    if (err == TransactionError::EXTERNAL_SIGNER_FAILED) {
+        //: "External signer" means using devices such as hardware wallets.
+        QMessageBox::critical(nullptr, tr("External signer failure"), "External signer failure");
+        return false;
+    }
+    if (err != TransactionError::OK) {
+        tfm::format(std::cerr, "Failed to sign PSBT");
+        processSendCoinsReturn(WalletModel::TransactionCreationFailed);
+        return false;
+    }
+    // fillPSBT does not always properly finalize
+    complete = FinalizeAndExtractPSBT(psbtx, mtx);
+    return true;
+}
+
 void SendCoinsDialog::sendButtonClicked([[maybe_unused]] bool checked)
 {
     if(!model || !model->getOptionsModel())
@@ -479,33 +507,7 @@ void SendCoinsDialog::sendButtonClicked([[maybe_unused]] bool checked)
         assert(!complete);
         assert(err == TransactionError::OK);
         if (model->wallet().hasExternalSigner()) {
-            try {
-                err = model->wallet().fillPSBT(SIGHASH_ALL, true /* sign */, true /* bip32derivs */, nullptr, psbtx, complete);
-            } catch (const std::runtime_error& e) {
-                QMessageBox::critical(nullptr, tr("Sign failed"), e.what());
-                send_failure = true;
-                return;
-            }
-            if (err == TransactionError::EXTERNAL_SIGNER_NOT_FOUND) {
-                //: "External signer" means using devices such as hardware wallets.
-                QMessageBox::critical(nullptr, tr("External signer not found"), "External signer not found");
-                send_failure = true;
-                return;
-            }
-            if (err == TransactionError::EXTERNAL_SIGNER_FAILED) {
-                //: "External signer" means using devices such as hardware wallets.
-                QMessageBox::critical(nullptr, tr("External signer failure"), "External signer failure");
-                send_failure = true;
-                return;
-            }
-            if (err != TransactionError::OK) {
-                tfm::format(std::cerr, "Failed to sign PSBT");
-                processSendCoinsReturn(WalletModel::TransactionCreationFailed);
-                send_failure = true;
-                return;
-            }
-            // fillPSBT does not always properly finalize
-            complete = FinalizeAndExtractPSBT(psbtx, mtx);
+            send_failure = !signWithExternalSigner(psbtx, mtx, complete);
         }
 
         // Broadcast transaction if complete (even with an external signer this
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -401,6 +401,52 @@ bool SendCoinsDialog::PrepareSendText(QString& question_string, QString& informa
     return true;
 }
 
+void SendCoinsDialog::presentPSBT(PartiallySignedTransaction& psbtx)
+{
+    // Serialize the PSBT
+    CDataStream ssTx(SER_NETWORK, PROTOCOL_VERSION);
+    ssTx << psbtx;
+    GUIUtil::setClipboard(EncodeBase64(ssTx.str()).c_str());
+    QMessageBox msgBox;
+    msgBox.setText("Unsigned Transaction");
+    msgBox.setInformativeText("The PSBT has been copied to the clipboard. You can also save it.");
+    msgBox.setStandardButtons(QMessageBox::Save | QMessageBox::Discard);
+    msgBox.setDefaultButton(QMessageBox::Discard);
+    switch (msgBox.exec()) {
+    case QMessageBox::Save: {
+        QString selectedFilter;
+        QString fileNameSuggestion = "";
+        bool first = true;
+        for (const SendCoinsRecipient &rcp : m_current_transaction->getRecipients()) {
+            if (!first) {
+                fileNameSuggestion.append(" - ");
+            }
+            QString labelOrAddress = rcp.label.isEmpty() ? rcp.address : rcp.label;
+            QString amount = BitcoinUnits::formatWithUnit(model->getOptionsModel()->getDisplayUnit(), rcp.amount);
+            fileNameSuggestion.append(labelOrAddress + "-" + amount);
+            first = false;
+        }
+        fileNameSuggestion.append(".psbt");
+        QString filename = GUIUtil::getSaveFileName(this,
+            tr("Save Transaction Data"), fileNameSuggestion,
+            //: Expanded name of the binary PSBT file format. See: BIP 174.
+            tr("Partially Signed Transaction (Binary)") + QLatin1String(" (*.psbt)"), &selectedFilter);
+        if (filename.isEmpty()) {
+            return;
+        }
+        std::ofstream out{filename.toLocal8Bit().data(), std::ofstream::out | std::ofstream::binary};
+        out << ssTx.str();
+        out.close();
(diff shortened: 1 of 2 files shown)


src/qt/sendcoinsdialog.cpp
```diff
@@ -502,19 +502,32 @@ void SendCoinsDialog::sendButtonClicked([[maybe_unused]] bool checked)
...
     bool send_failure = false;
-    if (retval == QMessageBox::Save) {
+    if (retval == QMessageBox::Save) { // Create Unsigned clicked
         CMutableTransaction mtx = CMutableTransaction{*(m_current_transaction->getWtx())};
         PartiallySignedTransaction psbtx(mtx);
+        // Copy PSBT to clipboard and offer to save
+        presentPSBT(psbtx);
+    } else { // Send clicked
+        assert(!model->wallet().privateKeysDisabled() || model->wallet().hasExternalSigner());
         bool complete = false;
```
promag, line 512: a49bb0c8328d1e0ecf0ee80916594ff260f79613

I think it makes more sense to refactor this to:
```cpp
bool broadcast = true;
if (model->wallet().hasExternalSigner()) {
    // ...
    bool complete;
    send_failure = !signWithExternalSigner(psbtx, mtx, complete);
    if (complete) {
        // ...
    } else {
        // ...
        broadcast = false
    }
}
if (broadcast) {
    // ...
}

**Review by promag — COMMENTED** [review]
Concept ACK

**Comment by Sjors** [comment]
Addressed @promag's feedback.

**Review by promag — COMMENTED** [review]
@Sjors it will broadcast even if you reject signing on the device, so it's missing `broadcast = false` in that case.

File: src/qt/sendcoinsdialog.cpp [hunk 2]
```diff
@@ -502,19 +502,34 @@ void SendCoinsDialog::sendButtonClicked([[maybe_unused]] bool checked)
     }
 
     bool send_failure = false;
-    if (retval == QMessageBox::Save) {
+    bool broadcast = true;
```
promag, line 505: 33d165d802846943c0ba70be4e643e3b50c6634e

```diff
diff --git a/src/qt/sendcoinsdialog.cpp b/src/qt/sendcoinsdialog.cpp
index c420aa3581..87a0e4a098 100644
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -502,7 +502,6 @@ void SendCoinsDialog::sendButtonClicked([[maybe_unused]] bool checked)
     }

     bool send_failure = false;
-    bool broadcast = true;
     if (retval == QMessageBox::Save) { // Create Unsigned clicked
         CMutableTransaction mtx = CMutableTransaction{*(m_current_transaction->getWtx())};
         PartiallySignedTransaction psbtx(mtx);
@@ -510,11 +509,13 @@ void SendCoinsDialog::sendButtonClicked([[maybe_unused]] bool checked)
         presentPSBT(psbtx);
     } else { // Send clicked
         assert(!model->wallet().privateKeysDisabled() || model->wallet().hasExternalSigner());
+        bool broadcast = true;
         if (model->wallet().hasExternalSigner()) {
             CMutableTransaction mtx = CMutableTransaction{*(m_current_transaction->getWtx())};
             PartiallySignedTransaction psbtx(mtx);
             bool complete = false;
             send_failure = !signWithExternalSigner(psbtx, mtx, complete);
+            broadcast = complete;
             // A transaction signed with an external signer is not always complete,
             // e.g. in a multisig wallet.
             if (complete) {
@@ -523,7 +524,6 @@ void SendCoinsDialog::sendButtonClicked([[maybe_unused]] bool checked)
                 m_current_transaction->setWtx(tx);
             } else if (!send_failure) {
                 presentPSBT(psbtx);
-                broadcast = false;
             }
         }
```

Sjors: Oops...

Sjors: If the user rejects the transaction then `complete = false` and it will present the PSBT. We can't tell the difference between rejection and failure, so that's difficult to avoid. I don't see how this patch would change that.

jonatack:
Concept ACK

src/qt/sendcoinsdialog.cpp
promag: If user rejects then `complete = false` and `send_failure = true`, so it doesn't `presentPSBT()` and because 
`broadcast = true` it calls `model->sendCoins()`

Sjors: I see. I restructured this code and fix a bit...

kristapsk:
Concept ACK

src/qt/sendcoinsdialog.cpp
promag: Thanks, LGTM, will re-test.

promag (approved):
Tested ACK 86e3e2cda8f2ed91f13ea0cf7c0bf291083c1970.

src/qt/sendcoinsdialog.cpp
```diff
@@ -411,7 +490,7 @@ void SendCoinsDialog::sendButtonClicked([[maybe_unused]] bool checked)
     assert(m_current_transaction);
 
     const QString confirmation = tr("Confirm send coins");
-    auto confirmationDialog = new SendConfirmationDialog(confirmation, question_string, informative_text, detailed_text, SEND_CONFIRM_DELAY, !model->wallet().privateKeysDisabled(), model->getOptionsModel()->getEnablePSBTControls(), this);
+    auto confirmationDialog = new SendConfirmationDialog(confirmation, question_string, informative_text, detailed_text, SEND_CONFIRM_DELAY, /*enable_send=*/!model->wallet().privateKeysDisabled() || model->wallet().hasExternalSigner(), /*always_show_unsigned=*/model->getOptionsModel()->getEnablePSBTControls(), this);
```
jonatack, line 493: 33d165d readability suggestion, replace named args with localvars of same name here

```diff
+    const bool enable_send{!model->wallet().privateKeysDisabled() || model->wallet().hasExternalSigner()};
+    const bool always_show_unsigned{model->getOptionsModel()->getEnablePSBTControls()};
```

```suggestion
    auto confirmationDialog = new SendConfirmationDialog(confirmation, question_string, informative_text, detailed_text, SEND_CONFIRM_DELAY, enable_send, always_show_unsigned, this);
```

src/qt/sendcoinsdialog.cpp
```diff
@@ -423,50 +502,38 @@ void SendCoinsDialog::sendButtonClicked([[maybe_unused]] bool checked)
...
         PartiallySignedTransaction psbtx(mtx);
-        bool complete = false;
-        // Always fill without signing first. This prevents an external signer
-        // from being called prematurely and is not expensive.
-        TransactionError err = model->wallet().fillPSBT(SIGHASH_ALL, false /* sign */, true /* bip32derivs */, nullptr, psbtx, complete);
-        assert(!complete);
-        assert(err == TransactionError::OK);
+        // Copy PSBT to clipboard and offer to save
+        presentPSBT(psbtx);
+    } else { // Send clicked
```
jonatack, line 510: 33d165d readability suggestion

```diff
     bool send_failure = false;
-    if (retval == QMessageBox::Save) { // Create Unsigned clicked
+    if (retval == QMessageBox::Save) {
+        // "Create Unsigned" clicked
         CMutableTransaction mtx = CMutableTransaction{*(m_current_transaction->getWtx())};
         PartiallySignedTransaction psbtx(mtx);
         // Copy PSBT to clipboard and offer to save
         presentPSBT(psbtx);
-    } else { // Send clicked
+    } else {
+        // "Send" clicked
         assert(!model->wallet().privateKeysDisabled() || model->wallet().hasExternalSigner());
```

src/qt/sendcoinsdialog.h
```diff
@@ -117,6 +128,8 @@ class SendConfirmationDialog : public QMessageBox
 
 public:
     SendConfirmationDialog(const QString& title, const QString& text, const QString& informative_text = "", const QString& detailed_text = "", int secDelay = SEND_CONFIRM_DELAY, bool enable_send = true, bool always_show_unsigned = true, QWidget* parent = nullptr);
+    /* Returns QMessageBox::Cancel, QMessageBox::Yes when send is
+       clicked and QMessageBox::Save when Create Unsigned is clicked. */
```
jonatack, line 132: 33d165d maybe s/send/Send/, and wrap Send and Create Unsigned in quotes

jonatack (commented):
Untested ACK 86e3e2cda8f2ed91f13ea0cf7c0bf291083c1970 code review and debug build, the approach and logic looks correct

A few style and readability suggestions, happy to re-review if you update.

luke-jr (commented):
utACK 86e3e2cda8f2ed91f13ea0cf7c0bf291083c1970 except for 1 part I'm unsure on

src/qt/sendcoinsdialog.cpp
```diff
@@ -401,6 +401,80 @@ bool SendCoinsDialog::PrepareSendText(QString& question_string, QString& informa
...
+        break;
+    default:
+        assert(false);
+    } // msgBox.exec()
+}
+
+bool SendCoinsDialog::signWithExternalSigner(PartiallySignedTransaction& psbtx, CMutableTransaction& mtx, bool& complete) {
+    TransactionError err;
+    try {
+        err = model->wallet().fillPSBT(SIGHASH_ALL, /*sign=*/false, /*bip32derivs=*/true, /*n_signed=*/nullptr, psbtx, complete);
```
luke-jr, line 453: `sign` was `true` here - why was it changed to `false`?

luke-jr (changes requested):

src/qt/sendcoinsdialog.cpp
Sjors: Yikes, I broke that during the style fix.

luke-jr (approved):
utACK 2efdfb88aab6496dcf2b98e0de30635bc6bade85

src/qt/sendcoinsdialog.cpp
```diff
@@ -423,50 +502,38 @@ void SendCoinsDialog::sendButtonClicked([[maybe_unused]] bool checked)
...
 
     bool send_failure = false;
-    if (retval == QMessageBox::Save) {
+    if (retval == QMessageBox::Save) { // Create Unsigned clicked
         CMutableTransaction mtx = CMutableTransaction{*(m_current_transaction->getWtx())};
         PartiallySignedTransaction psbtx(mtx);
-        bool complete = false;
-        // Always fill without signing first. This prevents an external signer
-        // from being called prematurely and is not expensive.
-        TransactionError err = model->wallet().fillPSBT(SIGHASH_ALL, false /* sign */, true /* bip32derivs */, nullptr, psbtx, complete);
```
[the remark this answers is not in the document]
jonatack: Yes, as a potential follow-up it looks like this code block could be potentially de-duped by extracting to a helper method.
```cpp
        CMutableTransaction mtx = CMutableTransaction{*(m_current_transaction->getWtx())};
        PartiallySignedTransaction psbtx(mtx);
        bool complete = false;
        TransactionError err = model->wallet().fillPSBT(SIGHASH_ALL, /*sign=*/false, /*bip32derivs=*/true, /*n_signed=*/nullptr, psbtx, complete);
        assert(!complete);
        assert(err == TransactionError::OK);
```

jonatack:
utACK 2efdfb88aab6496dcf2b98e0de30635bc6bade85 diff review since my last review, code re-review, rebased to current master, verified clean debug build of each commit

src/qt/sendcoinsdialog.cpp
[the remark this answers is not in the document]
Sjors: I'm going to avoid touching this now, but I agree it would be good to move into a helper.

hebasto:
Backported in https://github.com/bitcoin/bitcoin/pull/24596.

jonatack:
It might be good to have another tested ACK of the latest push.

Sjors:
@achow101?

PR #557, opened 2022-02-22
Revert "qt: Do not use QObject::tr plural syntax for numbers with a unit symbol"

luke-jr:
Apparently this got forgotten. Maybe too late for 23.x (it's a bugfix, but changes translation strings).

This reverts commit 3adde72bc99215062c8dabd38f8c34ad093452b5 (#296)

per [GChuf](https://github.com/bitcoin-core/gui/pull/296#issuecomment-962516055)

>I can confirm for slovenian and other slavic languages that we do have 3 or 4 different ways of saying "%n GB needed%, depending on the actual number of gigabytes. Similar to english "is/are". There's no way to cover all cases ... this is exactly why transifex allows you to have more than 2 options.

### DIFF
--- a/src/qt/intro.cpp
+++ b/src/qt/intro.cpp
@@ -298,12 +298,12 @@ void Intro::setStatus(int status, const QString &message, quint64 bytesAvailable
 
 void Intro::UpdateFreeSpaceLabel()
 {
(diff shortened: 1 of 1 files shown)


hebasto (approved):
ACK 0c64401324b03f5576149bf27138cdb103dae934, I have reviewed the code and it looks OK, I agree it can be merged.

Languages indeed are so different.

PR #539, opened 2022-01-27
Scale network graph based on time interval

RandyMcMillan:
The Network Graph is updated to scale based on the time interval selected by the QSlider.
The graph now displays a sample history that is retained when the QSlider value is changed.
Fixes issue #532

PR #559, opened 2022-02-24
Network graph improvements

### LINKED ISSUE #532: Debug Window: Network Traffic cleared
Dragging the slider should not erase the current data.

Moved from: https://github.com/bitcoin/bitcoin/issues/7481.

rebroad:
This combines #492, #483, #484, and #473 and also fixes #532 by adding the ability to change between graph durations without losing the data (it's sampled at each interval). Also added is some graphical animations when the X or Y scale changes. (it might be necessary to make this optional on lower performing hardware, such as the raspberry pi).

One more feature is that the graph will start on the smallest duration, and automatically increase the graph duration once the graph is filled.

Some examples can be seen in my comment at https://github.com/bitcoin-core/gui/pull/539#issuecomment-1031852315 

I'm expecting there may be issues with my coding style, so I'll create this PR as a draft at this stage.

### DIFF
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -630,19 +630,25 @@
            <item>
             <widget class="QSlider" name="sldGraphRange">
              <property name="minimum">
-              <number>1</number>
+              <number>0</number>
              </property>
              <property name="maximum">
-              <number>288</number>
+              <number>2400</number>
              </property>
              <property name="pageStep">
-              <number>12</number>
+              <number>200</number>
              </property>
              <property name="value">
-              <number>6</number>
+              <number>0</number>
              </property>
              <property name="orientation">
               <enum>Qt::Horizontal</enum>
+             </property>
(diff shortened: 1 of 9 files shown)


katesalazar:
What's the rationale for combining those PRs in this one?

rebroad:
> What's the rationale for combining those PRs in this one?

Because they are difficult to separate - given they each kinda tread on the toes of the others, I thought better to PR the whole foot rather than the individual toes. e.g. more than one feature uses the y_value() function, the tool-tip change and the linear/non-linear change both need an extra commit to be able work together.

katesalazar:
This was a bad idea.
If there wasn't enough developer time to review your scattered PRs,
a larger combined PR shall only have larger reviewing problems.
I'd suggest choosing one of your small PRs and to pay for reviews.
Get it definitely NACKed or ACKed and merged, then repeat with a different
PR.
Choose PR order wisely.
Good luck!

PR #534, opened 2022-01-23
Add translatable address error message 

w0xlt:
This PR is an alternative to https://github.com/bitcoin-core/gui/pull/533 that addresses https://github.com/bitcoin-core/gui/pull/533#pullrequestreview-860261164 .

While this PR no longer provides a detailed error message, it still brings some UI improvements:

. Improve user experience by showing an error message instead of just changing the color of `QLineEdit`
. Currently, when an invalid address is entered in `custom change address`, an error message is displayed. But the same does not happen in the `Pay To` field or when adding a new address in the address manager. This PR standardizes error handling for all address fields.
. The message added in this PR is the same as used in the `custom change address`, which is already translated into several languages.

| PR  | 
| -------------| 
| invalid_new_address | 
| invalid_pay_to |

PR #560, opened 2022-02-26
Make error message layout consistent

w0xlt:
Currently, the error message for invalid change address is displayed on the right while in the "Pay to" and "Address" fields do not show the error message, as commented in https://github.com/bitcoin-core/gui/pull/553#issuecomment-1047306976.

This PR makes the address error feedback consistent across all fields.

Supersede https://github.com/bitcoin-core/gui/pull/534

master |
--- |
master_pay_to |
master_addr_edit |

PR |
--- |
pr_pay_to |
pr_addr_edit |

### DIFF
--- a/src/qt/editaddressdialog.cpp
+++ b/src/qt/editaddressdialog.cpp
@@ -3,9 +3,10 @@
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
 #include <qt/editaddressdialog.h>
-#include <qt/forms/ui_editaddressdialog.h>
 
+#include <key_io.h>
 #include <qt/addresstablemodel.h>
+#include <qt/forms/ui_editaddressdialog.h>
 #include <qt/guiutil.h>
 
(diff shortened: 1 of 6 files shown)


katesalazar:
I assume in this figure there is a static placeholder for the red label (normally empty):



I assume that's regular P2WPKH invoice address you are showing.

Would mean the destination address box be shorter. So what happens if destination address is a longer P2WSH address? Would it then clip while with the current code it would not?

RandyMcMillan:
Concept ACK

Rspigler:
tACK commit 9d75184fe7d915e05ac01d50cec90881594ef19a 

Tested all address types in the 'Pay To', 'Custom Change Address', and 'New Send Address' fields.

Just noting that it says "Warning: Invalid Bitcoin address" immediately, but only highlights red when you click out of the text box. I don't know if I have a preference there.  But I like the change.

hebasto:

PR #561, opened 2022-03-01
Remove unnecessary validation

w0xlt:
`if (valid)` validation in line 266 seems unnecessary.
`valid` is initialized as `true` in line 255 and is still unchanged when line 266 is reached.

### DIFF
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -263,7 +263,7 @@ bool SendCoinsDialog::PrepareSendText(QString& question_string, QString& informa
             {
                 recipients.append(entry->getValue());
             }
-            else if (valid)
+            else
             {
                 ui->scrollArea->ensureWidgetVisible(entry);
                 valid = false;


src/qt/sendcoinsdialog.cpp
```diff
@@ -263,7 +263,7 @@ bool SendCoinsDialog::PrepareSendText(QString& question_string, QString& informa
             {
                 recipients.append(entry->getValue());
             }
-            else if (valid)
+            else
```
laanwj, line 266: The `if (valid)` makes sure that `ensureWidgetVisible` is only called once, even if there are invalid entries. I'm not sure if this is important, either for performance or correctness, but it's something to consider.

laanwj:
Slight NACK on this change, the change does subtly change behavior, and seems unnecessary.

w0xlt:
Closing this PR. As mentioned by @laanwj , the idea is to call `ensureWidgetVisible `on the first invalid entry only and there is no reason to change this behavior.

Issue #573, opened 2022-04-04
Remove explicit DBUS notificator code?


laanwj:
Back in the day I implemented DBUS support in `notificator.cpp` because the default Qt notifications used a deprecated notification through X11, which didn't show up in Ubuntu's GUI anymore. However I just found in the [Qt 5.5 release notes](https://qthub.com/static/doc/qt5/qtdoc/whatsnew55.html) that 

> On supported desktops, the xcb plugin now uses the D-Bus based org.kde.StatusNotifier protocol for system tray icons, and org.freedesktop.Notifications for notifications.

(for comparison, we require Qt *5.11.3*)

From this I understand that Qt will already automatically do the right thing with regard to notifications. Unless we do something custom not supported by this, it may be possible to get rid of our custom code for this.

It would remove the (direct) dependency on QtDbus. This will need to be tested carefully, of course.

(something similar may be true for MacOS notifications; I did not look into this)

fanquake:
Concept ACK

hebasto:
Commenting here as #575 has been closed.

@laanwj https://github.com/bitcoin-core/gui/pull/575#issuecomment-1092932810:
> I expect this to give too much complaints from people that are currently using notifications without a system tray. There's no way to do what within Qt, so my dbus code is still the only viable option.

Moreover, our code requires (needlessly, I guess) existence of an icon. The following patch:
```diff
--- a/src/qt/notificator.cpp
+++ b/src/qt/notificator.cpp
@@ -37,7 +37,7 @@ Notificator::Notificator(const QString &_programName, QSystemTrayIcon *_trayIcon
     ,interface(nullptr)
 #endif
 {
-    if(_trayIcon && _trayIcon->supportsMessages())
+    if(QSystemTrayIcon::supportsMessages())
     {
         mode = QSystemTray;
     }
```
should works for such cases.

https://github.com/bitcoin-core/gui/pull/575#issuecomment-1090534172:
> I  am not sure this is working as expected.

The same behavior could be observed on the master branch after configuring `--without-qtdbus`. The reason of such behavior is that the [`Notificator` constructor](https://github.com/bitcoin/bitcoin/blob/87c7dcc60d62a5bc6b7092b63dba41b55baccf20/src/qt/notificator.cpp#L30-L58) prioritize `DBus` over `QSystemTrayIcon`. I don't know why this design has been chosen.

laanwj:
[...]

PR #574, opened 2022-04-04
qt: Equalize notifySystray and notifyDBus interface

laanwj:
Make `notifySystray` and `notifyDBus` feature-equivalent by making `notifySystray` support custom icons. This uses the `QSystemTrayIcon::showMessage` overload introduced in Qt 5.9.

This is an initial step towards bitcoin-core/gui#573.

### DIFF
--- a/src/qt/notificator.cpp
+++ b/src/qt/notificator.cpp
@@ -147,6 +147,27 @@ QVariant FreedesktopImage::toVariant(const QImage &img)
     return QVariant(FreedesktopImage::metaType(), &fimg);
 }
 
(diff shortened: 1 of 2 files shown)


laanwj:
I guess the best way of testing this on Linux is to comment out the DBUS notificator code. And generate notifications somehow.

Also, it should be tested on Windows. DBUS is irrelevant on Windows, but the QTrayIcon code is used there. Removed the Linux label.

laanwj:
Closing, it looks like we're not using the custom icon functionality at all. Will remove.

PR #575, opened 2022-04-06
qt: Remove dbus notification code

### LINKED ISSUE #573: Remove explicit DBUS notificator code?

laanwj:
As of Qt 5.5, Qt will automatically use Freedesktop notifications when available and relevant to the platform. There is no more need to roll our own, so remove the dbus notification code.

This change also removes any mention of the QtDbus library from the build system. We do not need to link against it explicitly anymore. I have however not removed Qt dbus flags from the depends system, because it needs to be available internally for this to work.

The feature to set a custom QIcon for notifications is also removed, as it was only available for DBUS and besides, isn't used at all!

This change only affects Linux. Closes #573.

## Known limitation

So there's one expected case where this fails to show a notification, where the old code does. If there is no recognized tray (`QSystemTrayIcon::isSystemTrayAvailable`), there is no tray icon, and no notification will be shown (as Qt's notifications go through the tray icon).

I'm not sure this is a blocking problem. It may actually be an advantage as it avoids the problem underlying #575. It won't try to connect to the notification service at all if there is likely no support.

### DIFF
--- a/build-aux/m4/bitcoin_qt.m4
+++ b/build-aux/m4/bitcoin_qt.m4
@@ -78,21 +78,6 @@ AC_DEFUN([BITCOIN_QT_INIT],[
   AC_ARG_WITH([qt-translationdir],[AS_HELP_STRING([--with-qt-translationdir=PLUGIN_DIR],[specify qt translation path (overridden by pkgconfig)])], [qt_translation_path=$withval], [])
   AC_ARG_WITH([qt-bindir],[AS_HELP_STRING([--with-qt-bindir=BIN_DIR],[specify qt bin path])], [qt_bin_path=$withval], [])
 
-  AC_ARG_WITH([qtdbus],
-    [AS_HELP_STRING([--with-qtdbus],
-    [enable DBus support (default is yes if qt is enabled and QtDBus is found, except on Android)])],
-    [use_dbus=$withval],
-    [use_dbus=auto])
-
-  dnl Android doesn't support D-Bus and certainly doesn't use it for notifications
-  case $host in
-    *android*)
-      if test "$use_dbus" != "yes"; then
-        use_dbus=no
-      fi
-    ;;
-  esac
-
   AC_SUBST(QT_TRANSLATION_DIR,$qt_translation_path)
 ])
 
@@ -102,7 +87,7 @@ dnl   BITCOIN_QT_CONFIGURE([MINIMUM-VERSION])
 dnl
 dnl Outputs: See _BITCOIN_QT_FIND_LIBS
 dnl Outputs: Sets variables for all qt-related tools.
-dnl Outputs: bitcoin_enable_qt, bitcoin_enable_qt_dbus, bitcoin_enable_qt_test
+dnl Outputs: bitcoin_enable_qt, bitcoin_enable_qt_test
 AC_DEFUN([BITCOIN_QT_CONFIGURE],[
   qt_version=">= $1"
   qt_lib_prefix="Qt5"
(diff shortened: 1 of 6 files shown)


fanquake:
Concept ACK

laanwj:
The following patch may be useful for testing. It adds menu items to the "help" menu that launch notifications of every type:
```
commit 9ccd2147f0c581f56f9bbfa714bae31f993d5d96 (HEAD -> master)
Author: laanwj <126646+laanwj@users.noreply.github.com>
Date:   Wed Apr 6 18:44:02 2022 +0200

    notification testing

diff --git a/src/qt/bitcoingui.cpp b/src/qt/bitcoingui.cpp
index 85e3c23085243be8300197fd5c5b5b518df2cf9d..53ebc50fad475bc5ca1f677a2b8bfa04a02a2e4b 100644
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -525,6 +525,25 @@ void BitcoinGUI::createMenuBar()
     help->addSeparator();
     help->addAction(aboutAction);
     help->addAction(aboutQtAction);
+
+    QAction *notifyAction1 = new QAction("Example notification (Information)", this);
+    help->addAction(notifyAction1);
+    connect(notifyAction1, &QAction::triggered, [this] {
+            notificator->notify(Notificator::Information, "Example title", "Example message");
+        }
+    );
+    QAction *notifyAction2 = new QAction("Example notification (Warning)", this);
+    help->addAction(notifyAction2);
+    connect(notifyAction2, &QAction::triggered, [this] {
+            notificator->notify(Notificator::Warning, "Example title", "Example message");
+        }
+    );
+    QAction *notifyAction3 = new QAction("Example notification (Critical)", this);
+    help->addAction(notifyAction3);
+    connect(notifyAction3, &QAction::triggered, [this] {
+            notificator->notify(Notificator::Critical, "Example title", "Example message");
+        }
+    );
 }
```

build-aux/m4/bitcoin_qt.m4
```diff
@@ -369,7 +347,7 @@ dnl _BITCOIN_QT_FIND_LIBS
 dnl ---------------------
 dnl
 dnl Outputs: All necessary QT_* variables are set.
-dnl Outputs: have_qt_test and have_qt_dbus are set (if applicable) to yes|no.
+dnl Outputs: have_qt_test are set (if applicable) to yes|no.
```
jonatack, line 350:
```suggestion
dnl Outputs: have_qt_test is set (if applicable) to yes|no.
```

jonatack (commented):
Concept ACK. Debug build is clean. Tested on same config as https://github.com/bitcoin-core/gui/pull/575#issuecomment-1090534172 with the suggested patch, tried various actions and read documentation but didn't figure out how to see notifications (other than a pop-up dialog box with the critical example, suggestions welcome).

laanwj:
>  tried various actions and read documentation but didn't figure out how to see notifications (other than a pop-up dialog box with the critical example, pointers welcome).

That happens if it doesn't detect a system tray. Normal notifications are ignored but it will popup for critical ones.

I'm going to close this for now, might want to revisit for Qt6.

laanwj:
But thanks for testing anyway and making me realize that :smile:

PR #576, opened 2022-04-06
Add qt unit test runner summary

jonatack:
Append a one-line summary to the output of running `./src/qt/test/test_bitcoin-qt` indicating that all tests passed or showing the number of failing tests. It's currently a bit inconvenient to see this result by eyeballing all of the output.

### DIFF
--- a/src/qt/test/test_main.cpp
+++ b/src/qt/test/test_main.cpp
@@ -21,8 +21,10 @@
 #endif // ENABLE_WALLET
 
 #include <QApplication>
+#include <QDebug>
 #include <QObject>
 #include <QTest>
+
 #include <functional>
 
 #if defined(QT_STATICPLUGIN)
@@ -113,5 +115,10 @@ int main(int argc, char* argv[])
     }
 #endif
 
+    if (fInvalid) {
+        qWarning("\nThere were errors in some of the tests above.\n");
+    } else {
+        qDebug("\nAll tests passed.\n");
+    }
     return fInvalid;
 }
--- a/src/qt/test/test_main.cpp
+++ b/src/qt/test/test_main.cpp
@@ -86,32 +86,32 @@ int main(int argc, char* argv[])
     app.setApplicationName("Bitcoin-Qt-test");
     app.createNode(*init);
 
-    int fInvalid{0};
+    int num_test_failures{0};
 
     AppTests app_tests(app);
-    fInvalid += QTest::qExec(&app_tests);
+    num_test_failures += QTest::qExec(&app_tests);
 
     OptionTests options_tests(app.node());
-    fInvalid += QTest::qExec(&options_tests);
+    num_test_failures += QTest::qExec(&options_tests);
 
     URITests test1;
-    fInvalid += QTest::qExec(&test1);
+    num_test_failures += QTest::qExec(&test1);
 
     RPCNestedTests test3(app.node());
-    fInvalid += QTest::qExec(&test3);
+    num_test_failures += QTest::qExec(&test3);
 
 #ifdef ENABLE_WALLET
     WalletTests test5(app.node());
-    fInvalid += QTest::qExec(&test5);
+    num_test_failures += QTest::qExec(&test5);
 
     AddressBookTests test6(app.node());
-    fInvalid += QTest::qExec(&test6);
+    num_test_failures += QTest::qExec(&test6);
 #endif
 
-    if (fInvalid) {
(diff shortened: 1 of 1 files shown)


jonatack:
A sample patch to generate 3 errors to test the runner output in this case. 

<details><summary>patch</summary><p>

```diff
diff --git a/src/qt/test/optiontests.cpp b/src/qt/test/optiontests.cpp
index 4a943a634..9e6c8a243 100644
--- a/src/qt/test/optiontests.cpp
+++ b/src/qt/test/optiontests.cpp
@@ -28,6 +28,7 @@ void OptionTests::optionTests()
     gArgs.LockSettings([&](util::Settings& settings) {
         settings.rw_settings.erase("prune");
     });
+    QVERIFY(gArgs.IsArgSet("-prune"));
     gArgs.WriteSettingsFile();
 }
 
@@ -51,7 +52,7 @@ void OptionTests::parametersInteraction()
 
     OptionsModel{};
 
-    const bool expected{false};
+    const bool expected{true};
 
     QVERIFY(gArgs.IsArgSet("-listen"));
     QCOMPARE(gArgs.GetBoolArg("-listen", !expected), expected);
diff --git a/src/qt/test/rpcnestedtests.cpp b/src/qt/test/rpcnestedtests.cpp
index 669a05fe0..4395c20cf 100644
--- a/src/qt/test/rpcnestedtests.cpp
+++ b/src/qt/test/rpcnestedtests.cpp
@@ -73,7 +73,7 @@ void RPCNestedTests::rpcNestedTests()
     QVERIFY(result.substr(0,1) == "{");
 
     RPCConsole::RPCExecuteCommandLine(m_node, result, "getblockchaininfo()[\"chain\"]"); //Quote path identifier are allowed, but look after a child containing the quotes in the key
-    QVERIFY(result == "null");
+    QVERIFY(result != "null");
```
</p></details>

src/qt/test/test_main.cpp
```diff
@@ -86,32 +86,32 @@ int main(int argc, char* argv[])
...
     AddressBookTests test6(app.node());
-    if (QTest::qExec(&test6) != 0) {
-        fInvalid = true;
-    }
+    num_test_failures += QTest::qExec(&test6);
 #endif
 
-    return fInvalid;
+    if (num_test_failures) {
+        qWarning("\nFailed tests: %d\n", num_test_failures);
```
shaavan, line 112: nit:
This is less of a suggestion and more of a preference. I don't think this line needs separate line breaks to make it stand out from other lines. I think we can forgo the "\n"s.

shaavan (approved):
ACK d025d7f0251e26b7ab5cf48c236b6b5e46fafe26

It makes sense to give the user an overview of how many tests were failed. Though it won’t wholly communicate the information of which tests failed to the user, it still would make him alert to the situation and allow him to save his efforts of eyeballing in case the “Test failed” message is missing.

I was able to compile the PR and run the `test_bitcoin-qt` file successfully. I have attached the screenshot of the difference in output between the Master and PR branches.

|Master|PR|
|---------|----|
|Master|Screenshot from 2022-04-07 23-32-28|

src/qt/test/test_main.cpp
jonatack: Thanks @shaavan!  I like the line breaks to set off the summary, but if reviewers agree with you I'm ok to drop them.

cheetooooo: thanks @shaavan

jarolrod (approved):
tACK https://github.com/bitcoin-core/gui/commit/d025d7f0251e26b7ab5cf48c236b6b5e46fafe26

I have tested and reviewed the code, I agree this can be merged.

src/qt/test/test_main.cpp
```diff
@@ -86,32 +86,32 @@ int main(int argc, char* argv[])
     app.setApplicationName("Bitcoin-Qt-test");
     app.createNode(*init);
 
+    int num_test_failures{0};
+
     AppTests app_tests(app);
-    if (QTest::qExec(&app_tests) != 0) {
-        fInvalid = true;
-    }
+    num_test_failures += QTest::qExec(&app_tests);
```
hebasto, line 92: [`QTest::qExec`](https://doc.qt.io/qt-5/qtest.html#qExec) does not guarantee return value `1` in case of a test failure, but:
> a value other than 0 if one or more tests failed or in case of unhandled exceptions.

jonatack: Yes, in my testing if there are two failures for instance, it returns 2.

PR #577, opened 2022-04-08
Getting ready to Qt 6 (1/n)

hebasto:
For Qt 5 all changes in this PR are refactoring. But for [Qt 6](https://github.com/bitcoin/bitcoin/pull/24798) they are real bugfixes :)

As I do not provide anyway way to build `bitcoin-qt` against Qt 6.2.4 fir now, suggesting to reviewers to verify changes for Qt 5 only.

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -356,7 +356,7 @@ void BitcoinGUI::createActions()
     showHelpMessageAction->setStatusTip(tr("Show the %1 help message to get a list with possible Bitcoin command-line options").arg(PACKAGE_NAME));
 
     m_mask_values_action = new QAction(tr("&Mask values"), this);
-    m_mask_values_action->setShortcut(QKeySequence(Qt::CTRL + Qt::SHIFT + Qt::Key_M));
+    m_mask_values_action->setShortcut(QKeySequence(Qt::CTRL | Qt::SHIFT | Qt::Key_M));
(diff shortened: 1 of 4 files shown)


shaavan (approved):
ACK 63125752a902de725f8aad1ee5bdaf5cb0c70f91

Going commit wise:

1. The first commit removes that unnecessary importing of the `<QActionGroup>` library in the `walletview.cpp` file where it was unused and add an appropriate import statement in `bitcoingui.cpp` where it is being used in the file. `guiutil.cpp` is using the QStandardPaths library. Hence, it is also appropriately imported into this PR.
2. I successfully verified that the “+” is deprecated for the **QKeySequence** function in Qt 6. I have attached a reference screenshot from the official [Qt 6 documentation](https://doc-snapshots.qt.io/qt6-dev/qkeysequence.html).

Screenshot from 2022-04-10 21-57-50

3. I was able to verify that this `Qt::ItemIsTristate` is deprecated, and `Qt::ItemIsAutoTristate` replaces its usage. I am attaching a reference screenshot from [Qt 5 documentation](https://doc.qt.io/qt-5/qt.html).

Screenshot from 2022-04-10 21-52-00

I successfully compiled and ran the bitcoin core GUI from this PR without any errors or warnings.

jarolrod (approved):
tACK https://github.com/bitcoin-core/gui/commit/63125752a902de725f8aad1ee5bdaf5cb0c70f91

Tested changes on linux, macOS, and windows. Confirmed no change in behavior and confirmed deprecation statuses with Qt docs

PR #579, opened 2022-04-08
Getting ready to Qt 6 (2/n). Remove `QApplication::globalStrut()`

hebasto:
This function has been deprecated in Qt 5.15.0, and has been [removed](https://github.com/qt/qtbase/commit/033d01bd6e2aef740ad1408a04d3ca0ae3b9ba9b) in Qt 6.

### DIFF
--- a/src/qt/bitcoinamountfield.cpp
+++ b/src/qt/bitcoinamountfield.cpp
@@ -141,8 +141,7 @@ public:
 
             opt.rect = rect();
 
-            cachedMinimumSizeHint = style()->sizeFromContents(QStyle::CT_SpinBox, &opt, hint, this)
-                                    .expandedTo(QApplication::globalStrut());
+            cachedMinimumSizeHint = style()->sizeFromContents(QStyle::CT_SpinBox, &opt, hint, this);
         }
         return cachedMinimumSizeHint;
     }


shaavan (commented):
Concept ACK

I successfully verified that the property `globalStrut()` is obeselete in Qt 5. [Refernce](https://doc.qt.io/qt-5/qapplication-obsolete.html#globalStrut-prop). So conceptually (and by approach), I agree with this PR. However, I would like to confirm one thing before ACKing this PR.

You mentioned that this change could have user-faced behavior changes. However, by my understanding:

- `QSize::expandedTo` returns a size max of caller and argument. See for [reference.](https://doc.qt.io/qt-6/qsize.html#expandedTo)
- By default, `GlobalStrut` has 0 widths and height value until it is not changed, which is not being done in the codebase.
- Hence `A.expandedTo(0)` should be equivalent to `A`.
- Hence, removing `expandedTo()` should be a simple refactoring change in this case.

So would you please explain how this PR could have user-faced behavior change?

jarolrod (commented):
ACK 3eaf5dbfe0a0c814116e92f602f3c062259b6ea3

Visually tested on Linux, macOS, and cross compile to windows. I did not notice any visual changes between this PR and master.

I tested dynamic builds on Linux and macOS using Qt 5.15. It would be nice if someone could test this PR using our minimum supported version of 5.11.3. Currently unable to do this testing myself.

luke-jr (approved):
utACK 3eaf5dbfe0a0c814116e92f602f3c062259b6ea3

From my investigation of globalStrut, Qt5 itself never actually supported it correctly in the first place.

hebasto:
@shaavan 
> Concept ACK
> 
> I successfully verified that the property `globalStrut()` is obeselete in Qt 5. [Refernce](https://doc.qt.io/qt-5/qapplication-obsolete.html#globalStrut-prop). So conceptually (and by approach), I agree with this PR. However, I would like to confirm one thing before ACKing this PR.
> 
> You mentioned that this change could have user-faced behavior changes. However, by my understanding:
> 
>    * `QSize::expandedTo` returns a size max of caller and argument. See for [reference.](https://doc.qt.io/qt-6/qsize.html#expandedTo)
> 
>    * By default, `GlobalStrut` has 0 widths and height value until it is not changed, which is not being done in the codebase.
> 
>    * Hence `A.expandedTo(0)` should be equivalent to `A`.
> 
>    * Hence, removing `expandedTo()` should be a simple refactoring change in this case.
> 
> 
> So would you please explain how this PR could have user-faced behavior change?

Thank you for your review. Your analysis is correct. I've removed mention of user-faced behavior change from the PR description.

PR #580, opened 2022-04-09
Getting ready to Qt 6 (3/n). Do not use `QKeyEvent` copy constructor 

hebasto:
This PR is preparation for [Qt 6](https://github.com/bitcoin/bitcoin/pull/24798), and it fixes an experimental build with Qt 6.2.4 as copying of `QEvent` has been [disabled](https://github.com/qt/qtbase/commit/19f9b0d5f54379151eb71e98555b203ad6756276) in Qt 6.0.0.

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -618,17 +618,16 @@ bool RPCConsole::eventFilter(QObject* obj, QEvent *event)
         case Qt::Key_Down: if(obj == ui->lineEdit) { browseHistory(1); return true; } break;
         case Qt::Key_PageUp: /* pass paging keys to messages widget */
         case Qt::Key_PageDown:
-            if(obj == ui->lineEdit)
-            {
-                QApplication::postEvent(ui->messagesWidget, new QKeyEvent(*keyevt));
+            if (obj == ui->lineEdit) {
+                QApplication::sendEvent(ui->messagesWidget, keyevt);
                 return true;
             }
             break;
         case Qt::Key_Return:
         case Qt::Key_Enter:
             // forward these events to lineEdit
-            if(obj == autoCompleter->popup()) {
-                QApplication::postEvent(ui->lineEdit, new QKeyEvent(*keyevt));
+            if (obj == autoCompleter->popup()) {
+                QApplication::sendEvent(ui->lineEdit, keyevt);
                 autoCompleter->popup()->hide();
                 return true;
             }
@@ -642,7 +641,7 @@ bool RPCConsole::eventFilter(QObject* obj, QEvent *event)
                   ((mod & Qt::ShiftModifier) && key == Qt::Key_Insert)))
             {
                 ui->lineEdit->setFocus();
-                QApplication::postEvent(ui->lineEdit, new QKeyEvent(*keyevt));
+                QApplication::sendEvent(ui->lineEdit, keyevt);
                 return true;
             }
         }


shaavan (approved):
ACK d543515e635786214215e5b30a2874c4e8a676dc

- I was able to verify that with Qt 6, the copying of QEvent has been disabled. See for [reference](https://doc.qt.io/qt-6/qtcore-changes-qt6.html#the-qregexp-class) {second and third paragraph}.
- Therefore, instead of copying the key event, this PR uses its data to create a new event.

nit:

As a minor nit, I would suggest creating a variable `type` instead of calling `keyevt→type()` multiple times.

```
QEvent::Type type = keyevt->type()
```

hebasto:
Updated d543515e635786214215e5b30a2874c4e8a676dc -> 8691a444add29d62dade34bf2965cf76a5d6c863 ([pr580.01](https://github.com/hebasto/gui/commits/pr580.01) -> [pr580.02](https://github.com/hebasto/gui/commits/pr580.02), [diff](https://github.com/hebasto/gui/compare/pr580.01..pr580.02)):

- addressed @shaavan's [comment](https://github.com/bitcoin-core/gui/pull/580#pullrequestreview-937312002)

src/qt/rpcconsole.cpp
```diff
@@ -610,6 +610,7 @@ bool RPCConsole::eventFilter(QObject* obj, QEvent *event)
     if(event->type() == QEvent::KeyPress) // Special key handling
     {
         QKeyEvent *keyevt = static_cast<QKeyEvent*>(event);
+        const auto type = keyevt->type();
         int key = keyevt->key();
         Qt::KeyboardModifiers mod = keyevt->modifiers();
```
shaavan, line 615: Since we are refactoring this part of the code, how about we initialize the `key` and `mod` variables as **const** since they are unchanged in their lifetime.

```suggestion
        const int key = keyevt->key();
        const auto mod = keyevt->modifiers();
```

shaavan (approved):
reACK 8691a444add29d62dade34bf2965cf76a5d6c863

Changes since my last review:

- Declared a `type` variable beforehand and used it instead of `keyevt→type()`.

src/qt/rpcconsole.cpp
hebasto: Ok, I'll implement this suggestion if/when other changes will be required.

luke-jr:
Why are we creating a new event instead of forwarding the original one on? Seems like this PR would simply be bypassing the behaviour Qt is trying to prevent...

hebasto:
> Why are we creating a new event instead of forwarding the original one on? Seems like this PR would simply be bypassing the behaviour Qt is trying to prevent...

[`QCoreApplication::postEvent`](https://doc.qt.io/qt-5/qcoreapplication.html#postEvent):
> The event must be allocated on the heap since the post event queue will take ownership of the event and delete it once it has been posted. It is not safe to access the event after it has been posted.

"forwarding the original one on" ends with double deletion of an `QKeyEvent` instance.

luke-jr:
Hmm, how about using https://doc.qt.io/qt-5/qcoreapplication.html#sendEvent ?

hebasto:
Updated 8691a444add29d62dade34bf2965cf76a5d6c863 -> 3ec6504a2e5b4afb7a2719a82191e0b96fe23214 ([pr580.02](https://github.com/hebasto/gui/commits/pr580.02) -> [pr580.03](https://github.com/hebasto/gui/commits/pr580.03), [diff](https://github.com/hebasto/gui/compare/pr580.02..pr580.03)):

- addressed @luke-jr's [comment](https://github.com/bitcoin-core/gui/pull/580#issuecomment-1098274140)

w0xlt (approved):
tACK https://github.com/bitcoin-core/gui/pull/580/commits/3ec6504a2e5b4afb7a2719a82191e0b96fe23214 on Ubuntu 21.10, Qt 5.15.2

shaavan (approved):
reACK 3ec6504a2e5b4afb7a2719a82191e0b96fe23214

Changes since my last review:

- Instead of creating a new event for each call of postEvent, send Event is used with keyevt event.

Since sendEvent does not delete the event argument after the completion of its execution, the same keyevt could be used multiple times instead of creating a separate event for each instance of using postEvent. Therefore I agree with this change.

References:
- postEvent: https://doc.qt.io/qt-6/qcoreapplication.html#postEvent
- sendEvent: https://doc.qt.io/qt-6/qcoreapplication.html#sendEvent
- Comparison between the two: https://doc.qt.io/archives/qq/qq11-events.html (under section **Synthetic Events**)

PR #581, opened 2022-04-10
refactor: Revamp `ClientModel` code to handle core signals

hebasto:
This PR:
- is a pure refactoring with no behavior change
- gets rid of `QMetaObject::invokeMethod()` "dynamic" calls, i.e., without compile-time checks of a called function name and its parameters
- replaces `std::bind`s with lambdas, making parameter permutation (including parameter omitting) explicit
- makes code simpler, more concise, and easier to reason about

Additionally, debug messages have been unified.

### DIFF
--- a/src/qt/clientmodel.cpp
+++ b/src/qt/clientmodel.cpp
@@ -148,11 +148,6 @@ uint256 ClientModel::getBestBlockHash()
     return m_cached_tip_blocks;
 }
 
-void ClientModel::updateNetworkActive(bool networkActive)
-{
-    Q_EMIT networkActiveChanged(networkActive);
-}
-
 void ClientModel::updateAlert()
 {
     Q_EMIT alertsChanged(getStatusBarWarnings());
@@ -231,13 +226,6 @@ void ClientModel::updateBanlist()
 }
 
 // Handlers for core signals
-static void NotifyNetworkActiveChanged(ClientModel *clientmodel, bool networkActive)
-{
-    bool invoked = QMetaObject::invokeMethod(clientmodel, "updateNetworkActive", Qt::QueuedConnection,
-                              Q_ARG(bool, networkActive));
-    assert(invoked);
-}
-
 static void NotifyAlertChanged(ClientModel *clientmodel)
 {
     qDebug() << "NotifyAlertChanged";
@@ -292,7 +280,10 @@ void ClientModel::subscribeToCoreSignals()
         [this](int new_num_connections) {
             Q_EMIT numConnectionsChanged(new_num_connections);
         });
-    m_handler_notify_network_active_changed = m_node.handleNotifyNetworkActiveChanged(std::bind(NotifyNetworkActiveChanged, this, std::placeholders::_1));
+    m_handler_notify_network_active_changed = m_node.handleNotifyNetworkActiveChanged(
+        [this](bool network_active) {
+            Q_EMIT networkActiveChanged(network_active);
+        });
     m_handler_notify_alert_changed = m_node.handleNotifyAlertChanged(std::bind(NotifyAlertChanged, this));
     m_handler_banned_list_changed = m_node.handleBannedListChanged(std::bind(BannedListChanged, this));
     m_handler_notify_block_tip = m_node.handleNotifyBlockTip(std::bind(BlockTipChanged, this, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3, false));
--- a/src/qt/clientmodel.cpp
+++ b/src/qt/clientmodel.cpp
@@ -236,15 +236,6 @@ void ClientModel::updateBanlist()
 }
 
 // Handlers for core signals
-static void ShowProgress(ClientModel *clientmodel, const std::string &title, int nProgress)
-{
-    // emits signal "showProgress"
-    bool invoked = QMetaObject::invokeMethod(clientmodel, "showProgress", Qt::QueuedConnection,
-                              Q_ARG(QString, QString::fromStdString(title)),
-                              Q_ARG(int, nProgress));
-    assert(invoked);
-}
-
 static void NotifyNumConnectionsChanged(ClientModel *clientmodel, int newNumConnections)
 {
     // Too noisy: qDebug() << "NotifyNumConnectionsChanged: " + QString::number(newNumConnections);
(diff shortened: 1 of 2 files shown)


hebasto:
Friendly ping @promag @ryanofsky

src/qt/clientmodel.cpp
```diff
@@ -306,7 +297,10 @@ static void BlockTipChanged(ClientModel* clientmodel, SynchronizationState sync_
 void ClientModel::subscribeToCoreSignals()
 {
     // Connect signals to client
-    m_handler_show_progress = m_node.handleShowProgress(std::bind(ShowProgress, this, std::placeholders::_1, std::placeholders::_2));
+    m_handler_show_progress = m_node.handleShowProgress(
+        [this](const std::string& title, int progress, [[maybe_unused]] bool resume_possible) {
+            Q_EMIT showProgress(QString::fromStdString(title), progress);
```
promag, line 244: 508e2dca5e91c1ff921f01d260fc62f629f1dc9e

This is fine because auto connection is used:
https://github.com/bitcoin-core/gui/blob/747cdf1d652d8587e9f2e3d4436c3ecdbf56d0a5/src/qt/bitcoingui.cpp#L593

src/qt/clientmodel.cpp
```diff
@@ -301,7 +288,10 @@ void ClientModel::subscribeToCoreSignals()
         [this](const std::string& title, int progress, [[maybe_unused]] bool resume_possible) {
             Q_EMIT showProgress(QString::fromStdString(title), progress);
         });
-    m_handler_notify_num_connections_changed = m_node.handleNotifyNumConnectionsChanged(std::bind(NotifyNumConnectionsChanged, this, std::placeholders::_1));
+    m_handler_notify_num_connections_changed = m_node.handleNotifyNumConnectionsChanged(
+        [this](int new_num_connections) {
+            Q_EMIT numConnectionsChanged(new_num_connections);
```
promag, line 248: 639563d7fea6b4d65840625dc466eede32d893cf

Also fine:
https://github.com/bitcoin-core/gui/blob/747cdf1d652d8587e9f2e3d4436c3ecdbf56d0a5/src/qt/bitcoingui.cpp#L580
https://github.com/bitcoin-core/gui/blob/747cdf1d652d8587e9f2e3d4436c3ecdbf56d0a5/src/qt/rpcconsole.cpp#L671

src/qt/clientmodel.cpp
```diff
@@ -284,7 +272,11 @@ void ClientModel::subscribeToCoreSignals()
         [this](bool network_active) {
             Q_EMIT networkActiveChanged(network_active);
         });
-    m_handler_notify_alert_changed = m_node.handleNotifyAlertChanged(std::bind(NotifyAlertChanged, this));
+    m_handler_notify_alert_changed = m_node.handleNotifyAlertChanged(
+        [this]() {
+            qDebug() << "NotifyAlertChanged";
+            Q_EMIT alertsChanged(getStatusBarWarnings());
```
promag, line 257: 9196d8a92837a077f65b3660d5cff523728e7db6

Wrong commit message, should be AlertChanged.

Also using auto connection:
https://github.com/bitcoin-core/gui/blob/747cdf1d652d8587e9f2e3d4436c3ecdbf56d0a5/src/qt/overviewpage.cpp#L254
https://github.com/bitcoin-core/gui/blob/747cdf1d652d8587e9f2e3d4436c3ecdbf56d0a5/src/qt/rpcconsole.cpp#L663

src/qt/clientmodel.cpp
```diff
@@ -277,7 +265,11 @@ void ClientModel::subscribeToCoreSignals()
             qDebug() << "NotifyAlertChanged";
             Q_EMIT alertsChanged(getStatusBarWarnings());
         });
-    m_handler_banned_list_changed = m_node.handleBannedListChanged(std::bind(BannedListChanged, this));
+    m_handler_banned_list_changed = m_node.handleBannedListChanged(
+        [this]() {
+            qDebug() << QString("%1: Requesting update for peer banlist").arg(__func__);
+            banTableModel->refresh();
```
promag, line 262: dd924e3efa021dc2fe50da9fedba4a87ce3026aa

I think this is wrong because `BanTableModel::refresh()` needs to be called on the Qt main thread.

Maybe:
```cpp
    GUIUtil::ObjectInvoke(banTableModel, [=] {
        banTableModel->refresh();
    });
```

promag (commented):
Concept ACK.

shaavan (commented):
Concept ACK

To summarize, this PR is replacing the use of `QMetaObject::invokeMethod()`, `std::bind`, and `std::placeholder` with a more straightforward _lambda_ function and direct `Q_EMIT` signals.

As per my understanding of the master branch’s implementation:

- A function is used in which `QMetaObject::invokeMethod(obj,` member, conn_type, args) is used, which invoked the member signal or slot on the object, with the given args. [Reference](https://doc.qt.io/qt-5/qmetaobject.html#invokeMethod).
- This function is then bounded (using std::bind) to “this” and placeholders if needed, which would be replaced by the required argument for the function. [Reference](https://www.geeksforgeeks.org/bind-function-placeholders-c/).

In this PR’s implementation:

- Instead of creating a function that would `invokeMethod()`, it directly creates it as a _lambda_ function, where the argument values were earlier bounded (using std::bind).
- It uses Q_EMIT to emit the respective signal immediately
- All the Q_EMIT signals are bounded under a `connect` with Qt::ConnectionType as `Qt::AutoConnection`, which will ensure the signal is invoked correctly based on the receiver (if it is in the same thread or not). [Reference](https://doc.qt.io/qt-5/qt.html#ConnectionType-enum).

However, as @promag mentioned, it is not the case with `BanTableModel::refresh()`.
```
void BanTableModel::refresh()
{
    Q_EMIT layoutAboutToBeChanged();
    priv->refreshBanlist(m_node);
    Q_EMIT layoutChanged();
}
```

This function should be invoked under the main thread, but the emit signal it calls (`layoutAboutToBeChanged()`) is not bound by and `connect`. Hence, it would be directly emitted irrespective of the current thread. So this might probably cause some issues. 

I hope my notes and understanding of this PR will help my fellow reviewers. In case any of my analyses are erroneous, please do correct me.


**Testing result:**

I successfully compiled this PR in **Ubuntu 20.04** with **Qt version 5.9.7**.
I experimented with banning a few peers and sorting them (as `BanTableModel::sort()` uses `refresh()`) based on different categories, and I observed no discrepancy in the behavior.
Hence I observed no change in the behavior of GUI.

hebasto:
@shaavan 

Thank you for your review.

> I successfully compiled this PR in **Ubuntu 20.04** with **Qt version 5.9.7**.

Want to update? :)

(bitcoin/bitcoin#24132)

shaavan:
> Want to update? :)

Oops :sweat_smile:! Let me do so now. And retest this PR ASAP!

hebasto:
Updated 78c4e908d985a588aa4a2ee2bb8e0f1ba728d740 -> bcbf982553aba8107fdb0db413d4b9fe8adc8f17 ([pr581.01](https://github.com/hebasto/gui/commits/pr581.01) -> [pr581.02](https://github.com/hebasto/gui/commits/pr581.02), [diff](https://github.com/hebasto/gui/compare/pr581.01..pr581.02)):

- addressed @promag's comments
- unified debug messages

src/qt/clientmodel.cpp
hebasto: Thanks! [Fixed](https://github.com/bitcoin-core/gui/pull/581#issuecomment-1100709598).

src/qt/clientmodel.cpp
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui/pull/581#issuecomment-1100709598).

promag (approved):
Code review ACK bcbf982553aba8107fdb0db413d4b9fe8adc8f17

w0xlt (approved):
tACK https://github.com/bitcoin-core/gui/pull/581/commits/bcbf982553aba8107fdb0db413d4b9fe8adc8f17 on Ubuntu 21.10, Qt 5.15.2.

PR #584, opened 2022-04-13
Getting ready to Qt 6 (5/n). Do not assume `qDBusRegisterMetaType` return type

hebasto:
`qDBusRegisterMetaType` returns:
- [`int`](https://doc.qt.io/qt-5/qdbusargument.html#qDBusRegisterMetaType) in Qt 5
- [`QMetaType`](https://doc.qt.io/qt-6/qdbusargument.html#qDBusRegisterMetaType) in Qt 6

### DIFF
--- a/src/qt/notificator.cpp
+++ b/src/qt/notificator.cpp
@@ -14,8 +14,9 @@
 #include <QTemporaryFile>
 #include <QVariant>
 #ifdef USE_DBUS
-#include <stdint.h>
+#include <QDBusMetaType>
 #include <QtDBus>
+#include <stdint.h>
 #endif
 #ifdef Q_OS_MAC
 #include <qt/macnotificationhandler.h>
@@ -72,8 +73,6 @@ class FreedesktopImage
 public:
     FreedesktopImage() {}
     explicit FreedesktopImage(const QImage &img);
-
-    static int metaType();
 
     // Image to variant that can be marshalled over DBus
     static QVariant toVariant(const QImage &img);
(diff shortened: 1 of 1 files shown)


hebasto:
Friendly ping @laanwj :)

laanwj:
As we don't use custom icons for notifications at all, I think we could get rid of the entire icon data, as well as FreedesktopIcon? so
```
hints["icon_data"] =…
```
and everything that it depend on, including any `const QIcon &icon` parameters.

This is less invasive than removing dbus support completely as in bitcoin-core/gui#575, but will still allow deleting this code.

hebasto:
@laanwj 
> I have no idea whether this is correct. But as we don't use custom icons for notifications at all, I think we could get rid of the entire icon data, as well as FreedesktopIcon? so
> 
> ```
> hints["icon_data"] =…
> ```
> 
> and everything that it depend on, including any `const QIcon &icon` parameters.
> 
> This is less invasive than removing dbus support completely as in #575, but will still allow deleting this code.

On my Ubuntu 22.04 with a [custom testing patch](https://github.com/bitcoin-core/gui/pull/575#issuecomment-1090487621):

Screenshot from 2022-04-16 11-43-42

With the following diff
```diff
--- a/src/qt/notificator.cpp
+++ b/src/qt/notificator.cpp
@@ -193,7 +193,6 @@ void Notificator::notifyDBus(Class cls, const QString &title, const QString &tex
     {
         tmpicon = icon;
     }
-    hints["icon_data"] = FreedesktopImage::toVariant(tmpicon.pixmap(FREEDESKTOP_NOTIFICATION_ICON_SIZE).toImage());
     args.append(hints);
 
     // Timeout (in msec)
```

Screenshot from 2022-04-16 11-42-22

As an Ubuntu user, I'd like to keep the current functionality.

laanwj:
Doesn't the latter icon convey the same information but integrate much better into your desktop theme?

laanwj:
At some point we actually used to have different custom icons for notifications, even different ones for incoming/sent/etc transactions I don't know what happened to that.

hebasto:
> Doesn't the latter icon convey the same information but integrate much better into your desktop theme?

The same icon for message / warning / critical notification.

> But I guess ideally you'd want notifications to have the icon of the program that emits them?

Probably. macOS does like that.

> So yeah if that's what you want then keeping the custom icon functionality (and actually using it) would make sense.

Yes. Let's keep it for now.

hebasto:
@luke-jr @jarolrod @prusnak 

Mind looking into this PR?

prusnak:
Concept ACK

w0xlt (approved):
tACK https://github.com/bitcoin-core/gui/pull/584/commits/6cf4dc7f64b42cbbff6a2ce7616ee625a87a29f5 on Ubuntu 21.10, Qt 5.15.2.

Issue #578, opened 2022-04-08
Replace `QRegExp` with `QRegularExpression`


hebasto:
[`QRegularExpression`](https://doc.qt.io/qt-5/qregularexpression.html)
> implements Perl-compatible regular expressions. It fully supports Unicode.

[`QRegExp`](https://doc.qt.io/qt-5/qregexp.html)
> is modeled on Perl's regexp language. It fully supports Unicode.

> ... the new `QRegularExpression` class ... is recommended in place of `QRegExp`.

The most valued benefit will come when [migrating](https://github.com/bitcoin/bitcoin/pull/24798) to Qt 6, where [`QRegExp`](https://doc.qt.io/qt-6/qregexp.html) requires usage of an additional `Core5Compat` module.

Instead, all use cases of `QRegExp` could be [ported](https://doc.qt.io/qt-6/qregexp.html#porting-to-qregularexpression) to `QRegularExpression`.

#### Useful skills:

basic C++, Qt, regular expressions

#### Want to work on this issue?

For guidance on contributing, please read [CONTRIBUTING.md](https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md) before opening your pull request.

prusnak:
PR in https://github.com/bitcoin-core/gui/pull/585

PR #585, opened 2022-04-15
Replace QRegExp with QRegularExpression

prusnak:
Fixes https://github.com/bitcoin-core/gui/issues/578

Only made sure the code builds, not sure how to test.

### DIFF
--- a/src/qt/addressbookpage.cpp
+++ b/src/qt/addressbookpage.cpp
@@ -18,6 +18,7 @@
 #include <QIcon>
 #include <QMenu>
 #include <QMessageBox>
+#include <QRegularExpression>
 #include <QSortFilterProxyModel>
 
 class AddressBookSortFilterProxyModel final : public QSortFilterProxyModel
@@ -46,12 +47,19 @@ protected:
 
         auto address = model->index(row, AddressTableModel::Address, parent);
 
+    #if (QT_VERSION >= QT_VERSION_CHECK(5, 12, 0))
+        auto match_address = filterRegularExpression().match(model->data(address).toString());
+        auto match_label = filterRegularExpression().match(model->data(label).toString());
+
+        return match_address.hasMatch() || match_label.hasMatch();
+    #else
         if (filterRegExp().indexIn(model->data(address).toString()) < 0 &&
             filterRegExp().indexIn(model->data(label).toString()) < 0) {
             return false;
         }
 
         return true;
+    #endif
     }
 };
 
--- a/src/qt/utilitydialog.cpp
+++ b/src/qt/utilitydialog.cpp
@@ -22,7 +22,7 @@
 #include <QCloseEvent>
 #include <QLabel>
 #include <QMainWindow>
-#include <QRegExp>
+#include <QRegularExpression>
 #include <QTextCursor>
 #include <QTextTable>
 #include <QVBoxLayout>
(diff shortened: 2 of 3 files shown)


hebasto:
Concept ACK.

As a preparation to Qt 6, I think this PR will benefit by incorporating of https://github.com/bitcoin/bitcoin/commit/f6adcadf643c3e3375d234ef0d0987f20bc9869b and https://github.com/bitcoin/bitcoin/commit/b649103d0554055470ea00cb46489c43c90796c9 in any reasonable way.

prusnak:
> As a preparation to Qt 6, I think this PR will benefit by incorporating of [bitcoin/bitcoin@f6adcad](https://github.com/bitcoin/bitcoin/commit/f6adcadf643c3e3375d234ef0d0987f20bc9869b) and [bitcoin/bitcoin@b649103](https://github.com/bitcoin/bitcoin/commit/b649103d0554055470ea00cb46489c43c90796c9) in any reasonable way.


[bitcoin/bitcoin@f6adcad](https://github.com/bitcoin/bitcoin/commit/f6adcadf643c3e3375d234ef0d0987f20bc9869b) is already part of c6b6106

the latter commit I just cherry-picked to this PR

prusnak:
It seems the CI is failing because it does not recognize `filterRegularExpression` and indeed it was introduced in [Qt 5.12](https://doc.qt.io/qt-5/qsortfilterproxymodel.html#filterRegularExpression-prop) while we still do support Qt 5.11.

I updated the second commit of the PR to use `QT_VERSION` check - see 54980a7994a13fa77b80503ba8574490e80c9649

hebasto:
Does it make sense to extract the "Extract first suffix from filter pattern..." functionality into a testable utility function?

hebasto:
> not sure how to test.

Here is a branch with tests for the `AddressBookPage` class -- https://github.com/hebasto/gui/commits/220423-test-ab

And test fails for the 54980a7994a13fa77b80503ba8574490e80c9649 commit.

src/qt/utilitydialog.cpp
```diff
@@ -44,8 +44,7 @@ HelpMessageDialog::HelpMessageDialog(QWidget *parent, bool about) :
         /// HTML-format the license message from the core
         QString licenseInfoHTML = QString::fromStdString(LicenseInfo());
         // Make URLs clickable
-        QRegExp uri("<(.*)>", Qt::CaseSensitive, QRegExp::RegExp2);
-        uri.setMinimal(true); // use non-greedy matching
+        QRegularExpression uri("<(.*)>", QRegularExpression::InvertedGreedinessOption); // use non-greedy matching
```
hebasto, line 47: A comment seems unneeded here because `QRegularExpression::InvertedGreedinessOption` is pretty self-explained.

src/qt/addressbookpage.cpp
```diff
@@ -46,12 +47,19 @@ class AddressBookSortFilterProxyModel final : public QSortFilterProxyModel
 
         auto address = model->index(row, AddressTableModel::Address, parent);
 
+    #if (QT_VERSION >= QT_VERSION_CHECK(5, 12, 0))
```
jarolrod, line 50: slight preference for flipping this around where we're checking if less than qt6, i think this makes the code cleaner and inline with how we've done some other qt6 preparation changes, but not a blocker

prusnak:
I won't be able to work on this PR in the following days/weeks. Feel free to push into my branch (maintainers are allowed) or take over the PR completely if needed.

src/qt/addressbookpage.cpp
hebasto: Current implementation allows to test without Qt 6.

jarolrod:
Picked up in #606

PR #586, opened 2022-04-16
Getting ready to Qt 6 (6/n). Replace `QCoreApplication::quit()` with `QCoreApplication::exit(0)`

hebasto:
### Qt 5:
 - no behavior change.

See https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/kernel/qcoreapplication.cpp?h=5.15#n2012:
```cpp
void QCoreApplication::quit()
{
    exit(0);
}
```

### Qt 6:
 - this change avoids sending a duplicated `QEvent::Quit`

We use `QEvent::Quit` to [handle](https://github.com/bitcoin-core/gui/pull/547) macOS dock menu events. Qt 6 uses `QEvent::Quit` more [widely](https://github.com/qt/qtbase/commit/89f7a2759c6b51343d0a1ca5a82d575abba04e0c). We do not want a duplicated `QEvent::Quit` which fires `Assert(node.args);` in the [`Shutdown()`](https://github.com/bitcoin-core/gui/blob/d1b3dfb275fd98e37cfe8a0f7cea7d03595af2e8/src/init.cpp#L200) function.

### DIFF
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -308,7 +308,9 @@ void BitcoinApplication::startThread()
 
     /*  communication to and from thread */
     connect(&m_executor.value(), &InitExecutor::initializeResult, this, &BitcoinApplication::initializeResult);
-    connect(&m_executor.value(), &InitExecutor::shutdownResult, this, &QCoreApplication::quit);
+    connect(&m_executor.value(), &InitExecutor::shutdownResult, this, [] {
+        QCoreApplication::exit(0);
+    });
     connect(&m_executor.value(), &InitExecutor::runawayException, this, &BitcoinApplication::handleRunawayException);
     connect(this, &BitcoinApplication::requestedInitialize, &m_executor.value(), &InitExecutor::initialize);
     connect(this, &BitcoinApplication::requestedShutdown, &m_executor.value(), &InitExecutor::shutdown);
--- a/src/qt/test/apptests.cpp
+++ b/src/qt/test/apptests.cpp
@@ -119,6 +119,6 @@ AppTests::HandleCallback::~HandleCallback()
     assert(it != callbacks.end());
     callbacks.erase(it);
     if (callbacks.empty()) {
-        m_app_tests.m_app.quit();
+        m_app_tests.m_app.exit(0);
     }
 }


hebasto:
Just for a quick reference, here is a (full?) list of ways to quit the GUI:
1) Close the splash screen window
2) Press `q` on the splash screen
3) Close the main window (if the _"Minimize on close"_ option is disabled)
4) Menu _File --> Exit_. On **macOS** use an application menu
5) Shortcut `Ctrl+Q`
6) System tray icon menu _Exit_. On **macOS** use a Dock icon menu
7) Run `stop` command in the Console tab
8) Receive the `stop` RPC command (if the _"Enable RPC server"_ option is checked)
9) Press "Reset Options" button in the Options tab
10) `Ctrl+C` in a shell which runs `bitcoin-qt`

w0xlt (approved):
tACK https://github.com/bitcoin-core/gui/pull/586/commits/4717ebe71d6c8dad04be51ef00021a8586e2424f on Ubuntu 21.10, Qt 5.15.2.

Tested with all options mentioned in https://github.com/bitcoin-core/gui/pull/586#issuecomment-1100676029

hebasto:
Updated 4717ebe71d6c8dad04be51ef00021a8586e2424f -> e0930ad770623ffad9a83d99846d6cfc20a52e15 ([pr586.01](https://github.com/hebasto/gui/commits/pr586.01) -> [pr586.02](https://github.com/hebasto/gui/commits/pr586.02), [diff](https://github.com/hebasto/gui/compare/pr586.01..pr586.02)):

- added the same changes into the tests

src/qt/bitcoin.cpp
```diff
@@ -308,7 +308,7 @@ void BitcoinApplication::startThread()
 
     /*  communication to and from thread */
     connect(&m_executor.value(), &InitExecutor::initializeResult, this, &BitcoinApplication::initializeResult);
-    connect(&m_executor.value(), &InitExecutor::shutdownResult, this, &QCoreApplication::quit);
+    connect(&m_executor.value(), &InitExecutor::shutdownResult, [] { QCoreApplication::exit(0); });
```
promag, line 311: The signal `shutdownResult` is emitted from executor's thread and this results in calling `exit()` in that thread because the connection lacks a context. But `exit()` must be called in the main thread: https://doc-snapshots.qt.io/qt6-dev/qcoreapplication.html#exit. Maybe:
```cpp
connect(&m_executor.value(), &InitExecutor::shutdownResult, this, [] {
    QCoreApplication::exit(0);
});

hebasto: Why? `QCoreApplication::exit()` is a static member function.

promag: Yes, but according to the doc

> Note also that this function is not thread-safe. It should be called only from the main thread

and by using https://doc.qt.io/qt-5/qobject.html#connect-5 we ensure that lambda is called on the right thread.

hebasto:
Updated e0930ad770623ffad9a83d99846d6cfc20a52e15 -> 252f363f2feff243cae47731d59dfa1b74dd4386 ([pr586.02](https://github.com/hebasto/gui/commits/pr586.02) -> [pr586.03](https://github.com/hebasto/gui/commits/pr586.03), [diff](https://github.com/hebasto/gui/compare/pr586.02..pr586.03)):

- addressed @promag's [comment](https://github.com/bitcoin-core/gui/pull/586#discussion_r878701952)

src/qt/bitcoin.cpp
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui/pull/586#issuecomment-1133648314).

promag (approved):
Code review ACK 252f363f2feff243cae47731d59dfa1b74dd4386.

PR #589, opened 2022-04-20
Getting ready to Qt 6 (7/n). Pass `WalletModel*` pointer to signal only when `ENABLE_WALLET` defined

hebasto:
On master (094d9fda5ccee7d78a2e3d8b1eec17b8b6a33466), the following queued connection https://github.com/bitcoin-core/gui/blob/094d9fda5ccee7d78a2e3d8b1eec17b8b6a33466/src/qt/rpcconsole.cpp#L1107 uses a `const WalletModel*` parameter regardless whether the `ENABLE_WALLET` macro is defined.

Although this code works in Qt 5, it is flawed. On Qt 6, the code gets broken because the fully defined `WalletModel` type is required which is not the case if `ENABLE_WALLET` is undefined.

This PR fixes the issue described above.

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -97,7 +97,11 @@ public:
     explicit RPCExecutor(interfaces::Node& node) : m_node(node) {}
 
 public Q_SLOTS:
-    void request(const QString &command, const WalletModel* wallet_model);
+#ifdef ENABLE_WALLET
+    void request(const QString& command, const WalletModel* wallet_model);
+#else
+    void request(const QString& command);
+#endif // ENABLE_WALLET
 
 Q_SIGNALS:
     void reply(int category, const QString &command);
@@ -416,7 +420,11 @@ bool RPCConsole::RPCParseCommandLine(interfaces::Node* node, std::string &strRes
     }
 }
 
-void RPCExecutor::request(const QString &command, const WalletModel* wallet_model)
+#ifdef ENABLE_WALLET
+void RPCExecutor::request(const QString& command, const WalletModel* wallet_model)
+#else
+void RPCExecutor::request(const QString& command)
+#endif // ENABLE_WALLET
 {
     try
     {
@@ -446,7 +454,12 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
                 "   example:    getblock(getblockhash(0),1)[tx][0]\n\n")));
             return;
         }
+
+#ifdef ENABLE_WALLET
         if (!RPCConsole::RPCExecuteCommandLine(m_node, result, executableCommand, nullptr, wallet_model)) {
+#else
+        if (!RPCConsole::RPCExecuteCommandLine(m_node, result, executableCommand)) {
+#endif // ENABLE_WALLET
             Q_EMIT reply(RPCConsole::CMD_ERROR, QString("Parse error: unbalanced ' or \""));
             return;
         }
@@ -1049,7 +1062,12 @@ void RPCConsole::on_lineEdit_returnPressed()
     //: A console message indicating an entered command is currently being executed.
     message(CMD_REPLY, tr("Executing…"));
     m_is_executing = true;
+
+#ifdef ENABLE_WALLET
     Q_EMIT cmdRequest(cmd, m_last_wallet_model);
+#else
+    Q_EMIT cmdRequest(cmd);
+#endif // ENABLE_WALLET
 
     cmd = QString::fromStdString(strFilteredCmd);
 
--- a/src/qt/rpcconsole.h
+++ b/src/qt/rpcconsole.h
@@ -49,8 +49,11 @@ public:
     }
 
     void setClientModel(ClientModel *model = nullptr, int bestblock_height = 0, int64_t bestblock_date = 0, double verification_progress = 0.0);
-    void addWallet(WalletModel * const walletModel);
+
+#ifdef ENABLE_WALLET
+    void addWallet(WalletModel* const walletModel);
     void removeWallet(WalletModel* const walletModel);
+#endif // ENABLE_WALLET
 
     enum MessageClass {
         MC_ERROR,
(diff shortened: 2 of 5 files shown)


hebasto:
Friendly ping @promag

src/qt/rpcconsole.cpp
```diff
@@ -88,24 +88,6 @@ const QStringList historyFilter = QStringList()
 
 }
 
-/* Object for executing console RPC commands in a separate thread.
```
promag, line 91: 5aa5ed9897cfe3d5832daa9b25835b73da37f5a9

A forward declaration would be fine for the `m_executor` member (so this commit could be dropped)?

src/qt/rpcconsole.cpp
```diff
@@ -1073,11 +1073,11 @@ void RPCConsole::browseHistory(int offset)
 
 void RPCConsole::startExecutor()
 {
-    RPCExecutor *executor = new RPCExecutor(m_node);
-    executor->moveToThread(&thread);
+    m_executor= new RPCExecutor(m_node);
```
promag, line 1079: d36fa9707d9dc00317199633daac03ff0bae9050

nit, space before `=`.

src/qt/rpcconsole.cpp
```diff
@@ -1031,7 +1031,10 @@ void RPCConsole::on_lineEdit_returnPressed()
     //: A console message indicating an entered command is currently being executed.
     message(CMD_REPLY, tr("Executing…"));
     m_is_executing = true;
-    Q_EMIT cmdRequest(cmd, m_last_wallet_model);
+
+    QMetaObject::invokeMethod(m_executor, [this, cmd]() {
+        m_executor->request(cmd, m_last_wallet_model);
```
promag, line 1036: b7257271c533a97ea5e8b675ece6eee61aef23de

I think we should avoid `m_last_wallet_model` since this is executed asynchronously, and in theory `m_last_wallet_model` could change meanwhile. I suggest above:
```cpp
    WalletModel* wallet_model{nullptr};
#ifdef ENABLE_WALLET
    wallet_model = ui->WalletSelector->currentData().value<WalletModel*>();
```
and here:
```cpp
    QMetaObject::invokeMethod(m_executor, [=] {
        m_executor->request(cmd, wallet_model);
    });
```

promag (changes requested):
Concept ACK.

An alternative is to ditch the 2nd argument, like (untested):
```cpp
Q_SIGNALS:
    // For RPC command executor
    void cmdRequest(const QString &command
#ifdef ENABLE_WALLET
        , const WalletModel* wallet_model
#endif
    );
```

```cpp
    Q_EMIT cmdRequest(cmd
#ifdef ENABLE_WALLET
        , m_last_wallet_model
#endif
    );
```

hebasto:
@promag

> Concept ACK.
> 
> An alternative is to ditch the 2nd argument, like (untested):
> 
> ```c++
> Q_SIGNALS:
>     // For RPC command executor
>     void cmdRequest(const QString &command
> #ifdef ENABLE_WALLET
>         , const WalletModel* wallet_model
> #endif
>     );
> ```
> 
> ```c++
>     Q_EMIT cmdRequest(cmd
> #ifdef ENABLE_WALLET
>         , m_last_wallet_model
> #endif
>     );
> ```

That was my initial implementation -- https://github.com/hebasto/gui/commits/pr589.01 :)

But it somehow breaks interaction with `moc` compiler when using MSVC.

hebasto:
@promag 

Thank you for your review! Agree with all of points you mentioned above.

Updated.

src/qt/rpcconsole.cpp
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui/pull/589#issuecomment-1105104759).

src/qt/rpcconsole.cpp
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui/pull/589#issuecomment-1105104759).

src/qt/rpcconsole.cpp
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui/pull/589#issuecomment-1105104759).

src/qt/rpcconsole.h
```diff
@@ -49,8 +53,11 @@ class RPCConsole: public QWidget
     }
 
     void setClientModel(ClientModel *model = nullptr, int bestblock_height = 0, int64_t bestblock_date = 0, double verification_progress = 0.0);
-    void addWallet(WalletModel * const walletModel);
+
+#ifdef ENABLE_WALLET
```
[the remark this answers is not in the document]
promag: 61457c179aec23227dcf3952c575052204103b50

Now I wonder what's the point of this guard 🤔

promag (approved):
Code review ACK 16d136ac51f56d0d5fa799d000d264c685052c4e.

src/qt/rpcconsole.h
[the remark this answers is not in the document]
hebasto: To make `RPCConsole` interface explicit?

src/qt/rpcconsole.cpp
```diff
@@ -1049,7 +1050,10 @@ void RPCConsole::on_lineEdit_returnPressed()
     //: A console message indicating an entered command is currently being executed.
     message(CMD_REPLY, tr("Executing…"));
     m_is_executing = true;
-    Q_EMIT cmdRequest(cmd, m_last_wallet_model);
+
+    QMetaObject::invokeMethod(m_executor, [this, cmd, wallet_model]() {
```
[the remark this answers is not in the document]
hebasto: Updated.

promag (commented):
ACK ab73d5985de5d9c4d1e3fd0f4d9d88a0908ea319

jarolrod (commented):
code review ACK https://github.com/bitcoin-core/gui/commit/ab73d5985de5d9c4d1e3fd0f4d9d88a0908ea319

PR #590, opened 2022-04-23
refactor: Declare `WalletModel` member functions with `const`

hebasto:
After bitcoin/bitcoin#12830 the `WalletModel` class has two member functions: https://github.com/bitcoin-core/gui/blob/be7a5f2fc400e7a3ef72dedbdcf49dd6c96d4f9e/src/qt/walletmodel.h#L81 and https://github.com/bitcoin-core/gui/blob/be7a5f2fc400e7a3ef72dedbdcf49dd6c96d4f9e/src/qt/walletmodel.h#L154

This PR drops the former one as redundant, and declares `WalletModel` member functions with the `const` qualifier where appropriate.

### DIFF
--- a/src/qt/walletmodel.cpp
+++ b/src/qt/walletmodel.cpp
@@ -145,7 +145,7 @@ void WalletModel::updateWatchOnlyFlag(bool fHaveWatchonly)
     Q_EMIT notifyWatchonlyChanged(fHaveWatchonly);
 }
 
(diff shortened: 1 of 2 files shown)


promag (commented):
Code review ACK f70ee34c71aeeb814fe65a69952343dccdb7b906.

kristapsk (approved):
cr ACK f70ee34c71aeeb814fe65a69952343dccdb7b906

PR #591, opened 2022-04-23
test: Add tests for `tableView` in `AddressBookPage` dialog

hebasto:
This PR is a prerequisite for more thorough testing of filtering in the `AddressBookPage` class in context of bitcoin-core/gui#578 and bitcoin-core/gui#585.

Required for bitcoin-core/gui#592.

### DIFF
--- a/src/qt/test/addressbooktests.cpp
+++ b/src/qt/test/addressbooktests.cpp
@@ -8,6 +8,7 @@
 
 #include <interfaces/chain.h>
 #include <interfaces/node.h>
+#include <qt/addressbookpage.h>
 #include <qt/clientmodel.h>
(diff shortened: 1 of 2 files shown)


hebasto:
Friendly ping @promag @prusnak

jarolrod (commented):
Concept ACK

in edae3ab6999ee9e6efabd8403d31e9bd7c84f8a3, we can do this because when not specifying the connection type; there is an `invokeMethod` overload that uses `Qt:AutoConnection`. In this context `Qt::AutoConnection` will really be `Qt::QueuedConnection1` because the receiver and emitter are in different threads.  

https://doc.qt.io/qt-5/qt.html#ConnectionType-enum

PR #592, opened 2022-04-23
Drop wildcard support for address/label search in `AddressBookPage`

hebasto:
Wildcard support for address/label search in `AddressBookPage` was introduced in bitcoin/bitcoin#12080, and it is available in release binaries since v0.17.

But this feature has never being [documented](https://github.com/bitcoin/bitcoin/blob/master/doc/release-notes/release-notes-0.17.0.md) or mentioned in the GUI.

Thus, there are no reasons to believe that users rely on this feature.

Required for [migration](https://github.com/bitcoin/bitcoin/pull/24798) to Qt 6, [because](https://doc.qt.io/qt-6/qregexp.html#porting-to-qregularexpression):
> There is no direct way to do wildcard matching in `QRegularExpression`.

Tests included.

~Based on bitcoin-core/gui#591, so only last 2 commits belong to this PR.~

### DIFF
--- a/src/qt/addressbookpage.cpp
+++ b/src/qt/addressbookpage.cpp
@@ -46,12 +46,9 @@ protected:
 
         auto address = model->index(row, AddressTableModel::Address, parent);
 
-        if (filterRegExp().indexIn(model->data(address).toString()) < 0 &&
-            filterRegExp().indexIn(model->data(label).toString()) < 0) {
-            return false;
-        }
-
-        return true;
+        const auto pattern = filterRegExp();
+        return (model->data(address).toString().contains(pattern) ||
+                model->data(label).toString().contains(pattern));
     }
 };
 
@@ -143,7 +140,9 @@ void AddressBookPage::setModel(AddressTableModel *_model)
     proxyModel = new AddressBookSortFilterProxyModel(type, this);
     proxyModel->setSourceModel(_model);
 
-    connect(ui->searchLineEdit, &QLineEdit::textChanged, proxyModel, &QSortFilterProxyModel::setFilterWildcard);
+    connect(ui->searchLineEdit, &QLineEdit::textChanged, [this](const QString& pattern) {
+        proxyModel->setFilterFixedString(pattern);
+    });
 
     ui->tableView->setModel(proxyModel);
     ui->tableView->sortByColumn(0, Qt::AscendingOrder);
--- a/src/qt/test/addressbooktests.cpp
+++ b/src/qt/test/addressbooktests.cpp
@@ -24,6 +24,7 @@
 #include <chrono>
 
 #include <QApplication>
+#include <QLineEdit>
 #include <QMessageBox>
 #include <QTableView>
 #include <QTimer>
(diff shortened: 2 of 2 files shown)


luke-jr:
cc @promag

luke-jr:
"However, the wildcardToRegularExpression method is provided to translate glob patterns into a Perl-compatible regular expression that can be used for that purpose."

Seems like it should be trivial to retain this feature, if we want to?

laanwj:
~0 on this. As I've said on IRC I don't particularly care for this feature to go (I don't think the entire Address Book is particularly useful, searching in the Transactions tab should cover it?), though "being compatible with Qt 6" is a strange motivation. As @luke-jr says it would still be possible by rewriting the expression.

hebasto:
Rebased on top of the merged #591.

hebasto:
@luke-jr 
> "However, the wildcardToRegularExpression method is provided to translate glob patterns into a Perl-compatible regular expression that can be used for that purpose."
> 
> Seems like it should be trivial to retain this feature, if we want to?

Would you mind provide a working example please?

src/qt/addressbookpage.cpp
```diff
@@ -143,7 +140,9 @@ void AddressBookPage::setModel(AddressTableModel *_model)
     proxyModel = new AddressBookSortFilterProxyModel(type, this);
     proxyModel->setSourceModel(_model);
 
-    connect(ui->searchLineEdit, &QLineEdit::textChanged, proxyModel, &QSortFilterProxyModel::setFilterWildcard);
+    connect(ui->searchLineEdit, &QLineEdit::textChanged, [this](const QString& pattern) {
+        proxyModel->setFilterFixedString(pattern);
```
promag, line 144: In fact `setFilterWildcard` in Qt6 already handles the change noted in the porting documentation:

> There is no direct way to do wildcard matching in [QRegularExpression](https://doc.qt.io/qt-6/qregularexpression.html). However, the wildcardToRegularExpression method is provided to translate glob patterns into a Perl-compatible regular expression that can be used for that purpose.

Qt5
```cpp
void QSortFilterProxyModel::setFilterFixedString(const QString &pattern)
{
    Q_D(QSortFilterProxyModel);
    d->filter_about_to_be_changed();
    QRegExp rx(pattern, d->filter_data.caseSensitivity(), QRegExp::FixedString);
    d->filter_data.setRegExp(rx);
    d->filter_changed();
}
```
Qt6
```cpp
void QSortFilterProxyModel::setFilterWildcard(const QString &pattern)
{
    Q_D(QSortFilterProxyModel);
    d->filter_regularexpression.removeBindingUnlessInWrapper();
    d->filter_about_to_be_changed();
    d->set_filter_pattern(QRegularExpression::wildcardToRegularExpression(
            pattern, QRegularExpression::UnanchoredWildcardConversion));
    d->filter_changed(QSortFilterProxyModelPrivate::Direction::Rows);
    d->filter_regularexpression.notify();
}
```

promag (commented):
Tested ACK 555b673fbc71412908bb73c9c71d49895270335f.

TBH I didn't think about using a wildcard in the search field, so `setFilterFixedString` would be sufficient.

The only issue I have is breaking someone's expectation, as this is a breaking change - using wildcard results in an empty list where before it could list some addresses. I understand it's not documented but it's available for some time.

Finally, if we stick to supporting wildcard then no changes are required as explained in the comment below (but didn't test with Qt6).

src/qt/addressbookpage.cpp
hebasto: https://github.com/bitcoin-core/gui/pull/593#discussion_r878696948

hebasto:
Combined into #593.

PR #594, opened 2022-04-26
scripted-diff: replace deprecated Q_OS_MAC with Q_OS_MACOS

promag:
`Q_OS_MAC` is deprecated but it is also defined when Qt is configured with `-xplatform macx-ios-clang`, and currently it guards some features not available on iOS, like `QProcess`.

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -26,7 +26,7 @@
 #include <qt/walletview.h>
 #endif // ENABLE_WALLET
 
-#ifdef Q_OS_MAC
+#ifdef Q_OS_MACOS
 #include <qt/macdockiconhandler.h>
 #endif
 
@@ -69,7 +69,7 @@
 
 
 const std::string BitcoinGUI::DEFAULT_UIPLATFORM =
-#if defined(Q_OS_MAC)
+#if defined(Q_OS_MACOS)
         "macosx"
 #elif defined(Q_OS_WIN)
         "windows"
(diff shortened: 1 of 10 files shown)


jarolrod (commented):
Concept ACK

This change is ok to do. I've checked qt documentation up to Qt 5.7 that mentions `Q_OS_MAC` is deprecated and to not use. Qt 5.7 docs even mention to use `Q_OS_MACOS` or `Q_OS_DARWIN` if targeting all Darwin platforms (macOS, IOS).

promag:
@fanquake should also define rename this?
https://github.com/bitcoin-core/gui/blob/833add0f48b0fad84d7b8cf9373a349e7aef20b4/build-aux/m4/bitcoin_qt.m4#L234

jarolrod:
> @fanquake should also rename this?
> 
> https://github.com/bitcoin-core/gui/blob/833add0f48b0fad84d7b8cf9373a349e7aef20b4/build-aux/m4/bitcoin_qt.m4#L234

yes, it would be appropriate to update what we want defined here. Should be opened up on the main repo though

jarolrod:
tACK e3daecae0333e5474b54f64619ae6f512b683787

laanwj:
Should some of these be `Q_OS_DARWIN`?
Ping @Sjors he was building for iOS at some point (bitcoin/bitcoin#1255).

katesalazar:
thought the same thing,
didn't tell because I think they can sort those out later 🤔

laanwj:
> didn't tell because I think they can sort those out later thinking

I'm not sure there's such a hurry, might as well do it properly while we're changing this.

Sjors:
I didn't get very far with iOs builds... Specifically I never even tried to compile QT.

However, the more generic `Q_OS_DARWIN` sounds good to me.

src/qt/bitcoingui.cpp
```diff
@@ -26,7 +26,7 @@
 #include <qt/walletview.h>
 #endif // ENABLE_WALLET
 
-#ifdef Q_OS_MAC
+#ifdef Q_OS_MACOS
 #include <qt/macdockiconhandler.h>
```
Sjors, line 30: This is macOS specific, so `Q_OS_MACOS` makes sense.

hebasto (approved):
ACK e3daecae0333e5474b54f64619ae6f512b683787.

Although `Q_OS_DARWIN` is a synonym for the deprecated `Q_OS_MAC`, I agree to narrow guards to `Q_OS_MACOS` as we do not currently support iOS, watchOS, or tvOS.

PR #595, opened 2022-04-27
gui: avoid unclean exit due to permissions issues when setting start on system startup

mruddy:
moving https://github.com/bitcoin/bitcoin/pull/25000 to this repo

Closes https://github.com/bitcoin/bitcoin/issues/24953

This remedies two cases where the GUI can cause the process to exit in an unclean manner. These two cases are cases where the directory path causing the exception is likely/usually not under the node's data and/or blocks directory and thus could be more likely to lack write permissions.

Instead of abruptly exiting the process, this path instead makes log entries such as:

```
2022-04-27T11:43:03Z failed to create autostart directory: filesystem error: cannot create directories: Permission denied [/tmp/crash/autostart]
```

```
2022-04-27T11:44:05Z failed to remove autostart file: filesystem error: cannot remove: Permission denied [/tmp/crash/autostart/bitcoin.desktop]
```

the command line used to test and create the above log entries (with this patch applied) was:
```
mkdir -p /tmp/btc && mkdir -p /tmp/crash && chmod 0500 /tmp/crash && XDG_CONFIG_HOME=/tmp/crash ./src/qt/bitcoin-qt -debug -datadir=/tmp/btc -connect=0
```

The short/easy process to generate the exception is noted in the original issue.

This PR does not address the other cases where `fs::create_directories` may create an unclean exit due to the reasoning that the other cases are generated by paths under the blocks and/or data directories and those directories need to be writable by the bitcoin process, or lots of things fail. This is a targeted PR to avoid a couple of unique recoverable situations.


luke-jr (approved):
utACK

shaavan (approved):
ACK 933d2eddadd2b9ade2f3361f4bdc8fa7741d10f4

- This PR introduces a _try and catch_ statement to catch the `filesystem_error`. And adds a code block to handle this case correctly.
- The code block first clearly describes the cause of the error, and the second returns a **false** bool value.

I was able to compile the PR and run GUI through it successfully.

RandyMcMillan:
utACK 933d2ed

mruddy:
thanks to all of you for the code review. please let me know if there's anything that i can do to move this pr forward.

jonatack:
Concept ACK and code looks reasonable, will review/test.

src/qt/guiutil.cpp
```diff
@@ -613,8 +613,14 @@ bool GetStartOnSystemStartup()
 
 bool SetStartOnSystemStartup(bool fAutoStart)
 {
-    if (!fAutoStart)
-        fs::remove(GetAutostartFilePath());
+    if (!fAutoStart) {
+        try {
+            fs::remove(GetAutostartFilePath());
+        } catch(const fs::filesystem_error& e) {
+            LogPrintf("failed to remove autostart file: %s\n", e.what());
```
hebasto, line 620: Could you please Capitalize the first word in the log message to follow other `LogPrintf()` calls in the code base?

src/qt/guiutil.cpp
```diff
@@ -623,7 +629,12 @@ bool SetStartOnSystemStartup(bool fAutoStart)
             return false;
         pszExePath[r] = '\0';
 
-        fs::create_directories(GetAutostartDir());
+        try {
+            fs::create_directories(GetAutostartDir());
+        } catch(const fs::filesystem_error& e) {
+            LogPrintf("failed to create autostart directory: %s\n", e.what());
```
hebasto, line 635: The same.

hebasto (changes requested):

mruddy:
@hebasto Thanks for looking, I capitalized the first word in the log messages and force pushed it just now as d932157eb79a69555ba52ddc4acd37f16f9eaefc

hebasto (commented):
Tested d932157eb79a69555ba52ddc4acd37f16f9eaefc on Ubuntu 22.04.

Instead of crashes I have nice log messages like that:
```
Failed to remove autostart file: filesystem error: cannot remove: Permission denied [/tmp/crash/autostart/bitcoin-signet.desktop]
```
or
```
Failed to create autostart directory: filesystem error: cannot create directories: Permission denied [/tmp/crash/autostart]
```

But users who run GUI could easily miss such log messages without a dedicated error message dialog.

mruddy:
> But users who run GUI could easily miss such log messages without a dedicated error message dialog.

Maybe, but I'm not too worried about that. I'm happy with the node not crashing and the log mention. If you have a message box that you want me to slap in there, I guess I could add it. If so, then do we need the log messages? It's all the same to me as long as the node doesn't crash and lose data.

hebasto:
> > But users who run GUI could easily miss such log messages without a dedicated error message dialog.
> 
> Maybe, but I'm not too worried about that. I'm happy with the node not crashing and the log mention.

When users check the "Start Bitcoin Core on system login" checkbox, they naturally expect a certain behavior which could silently (with a log message only) fail due to permission issues. We cannot expect that all of the Bitcoin Core GUI users are tech-savvy enough to read a log file to find the reason of the failure.

I'd prefer a crash to behavior that will surely confuse users.

mruddy:
fair enough. i'll leave this open for a couple of weeks and if nobody wants it, i'll close it and the associated issue.

PR #598, opened 2022-05-09
Avoid recalculating the wallet balance - use model cache

furszy:
As per the title says, we are recalculating the entire wallet balance on different situations calling to `wallet().getBalances()`, when should instead make use of the wallet model cached balance.

This has the benefits of (1) not spending resources calculating a balance that we already have cached, and (2) avoid blocking the main thread for a long time, in case of big wallets, walking through the entire wallet's tx map more than what it's really needed.

Changes:

1) Fix: `SendCoinsDialog` was calling `wallet().getBalances()` twice during `setModel`.
2) Use the cached balance if the user did not select any UTXO manually inside the wallet model `getAvailableBalance` call.

-----------------------
As an extra note, this work born in [#25005](https://github.com/bitcoin/bitcoin/pull/25005) but grew out of scope of it.

### DIFF
--- a/src/qt/overviewpage.cpp
+++ b/src/qt/overviewpage.cpp
@@ -148,8 +148,6 @@ OverviewPage::OverviewPage(const PlatformStyle *platformStyle, QWidget *parent) 
 {
     ui->setupUi(this);
 
-    m_balances.balance = -1;
-
     // use a SingleColorIcon for the "out of sync warning" icon
     QIcon icon = m_platform_style->SingleColorIcon(QStringLiteral(":/icons/warning"));
     ui->labelTransactionsStatus->setIcon(icon);
@@ -178,8 +176,9 @@ void OverviewPage::handleTransactionClicked(const QModelIndex &index)
 void OverviewPage::setPrivacy(bool privacy)
 {
     m_privacy = privacy;
-    if (m_balances.balance != -1) {
-        setBalance(m_balances);
+    const auto& balances = walletModel->getCachedBalance();
+    if (balances.balance != -1) {
+        setBalance(balances);
     }
 
     ui->listTransactions->setVisible(!m_privacy);
@@ -198,7 +197,6 @@ OverviewPage::~OverviewPage()
 void OverviewPage::setBalance(const interfaces::WalletBalances& balances)
 {
     BitcoinUnit unit = walletModel->getOptionsModel()->getDisplayUnit();
-    m_balances = balances;
     if (walletModel->wallet().isLegacy()) {
         if (walletModel->wallet().privateKeysDisabled()) {
             ui->labelBalance->setText(BitcoinUnits::formatWithPrivacy(unit, balances.watch_only_balance, BitcoinUnits::SeparatorStyle::ALWAYS, m_privacy));
@@ -306,10 +304,10 @@ void OverviewPage::changeEvent(QEvent* e)
 
 void OverviewPage::updateDisplayUnit()
 {
-    if(walletModel && walletModel->getOptionsModel())
-    {
-        if (m_balances.balance != -1) {
-            setBalance(m_balances);
+    if (walletModel && walletModel->getOptionsModel()) {
+        const auto& balances = walletModel->getCachedBalance();
+        if (balances.balance != -1) {
+            setBalance(balances);
         }
 
         // Update txdelegate->unit with the current unit
--- a/src/qt/overviewpage.cpp
+++ b/src/qt/overviewpage.cpp
@@ -277,14 +277,13 @@ void OverviewPage::setWalletModel(WalletModel *model)
         ui->listTransactions->setModelColumn(TransactionTableModel::ToAddress);
 
         // Keep up to date with wallet
-        interfaces::Wallet& wallet = model->wallet();
-        interfaces::WalletBalances balances = wallet.getBalances();
-        setBalance(balances);
+        setBalance(model->getCachedBalance());
         connect(model, &WalletModel::balanceChanged, this, &OverviewPage::setBalance);
 
         connect(model->getOptionsModel(), &OptionsModel::displayUnitChanged, this, &OverviewPage::updateDisplayUnit);
 
-        updateWatchOnlyLabels(wallet.haveWatchOnly() && !model->wallet().privateKeysDisabled());
+        interfaces::Wallet& wallet = model->wallet();
+        updateWatchOnlyLabels(wallet.haveWatchOnly() && !wallet.privateKeysDisabled());
         connect(model, &WalletModel::notifyWatchonlyChanged, [this](bool showWatchOnly) {
             updateWatchOnlyLabels(showWatchOnly && !walletModel->wallet().privateKeysDisabled());
         });
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -718,7 +718,7 @@ void SendCoinsDialog::setBalance(const interfaces::WalletBalances& balances)
 
 void SendCoinsDialog::refreshBalance()
 {
-    setBalance(model->wallet().getBalances());
+    setBalance(model->getCachedBalance());
     ui->customFee->setDisplayUnit(model->getOptionsModel()->getDisplayUnit());
     updateSmartFeeLabel();
 }
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -164,11 +164,9 @@ void SendCoinsDialog::setModel(WalletModel *_model)
             }
         }
 
-        interfaces::WalletBalances balances = _model->wallet().getBalances();
-        setBalance(balances);
         connect(_model, &WalletModel::balanceChanged, this, &SendCoinsDialog::setBalance);
-        connect(_model->getOptionsModel(), &OptionsModel::displayUnitChanged, this, &SendCoinsDialog::updateDisplayUnit);
-        updateDisplayUnit();
+        connect(_model->getOptionsModel(), &OptionsModel::displayUnitChanged, this, &SendCoinsDialog::refreshBalance);
+        refreshBalance();
 
         // Coin Control
         connect(_model->getOptionsModel(), &OptionsModel::displayUnitChanged, this, &SendCoinsDialog::coinControlUpdateLabels);
@@ -718,7 +716,7 @@ void SendCoinsDialog::setBalance(const interfaces::WalletBalances& balances)
     }
 }
 
-void SendCoinsDialog::updateDisplayUnit()
+void SendCoinsDialog::refreshBalance()
 {
     setBalance(model->wallet().getBalances());
     ui->customFee->setDisplayUnit(model->getOptionsModel()->getDisplayUnit());
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -795,7 +795,7 @@ void SendCoinsDialog::useAvailableBalance(SendCoinsEntry* entry)
     m_coin_control->fAllowWatchOnly = model->wallet().privateKeysDisabled() && !model->wallet().hasExternalSigner();
 
     // Calculate available amount to send.
-    CAmount amount = model->wallet().getAvailableBalance(*m_coin_control);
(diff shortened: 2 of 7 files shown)


src/qt/walletmodel.cpp
```diff
@@ -120,12 +120,23 @@ void WalletModel::pollBalanceChanged()
...
+            QMutexLocker locker(&m_cache_mutex);
+            m_cached_balances = new_balances;
+        }
         Q_EMIT balanceChanged(new_balances);
     }
 }
 
+interfaces::WalletBalances WalletModel::getCachedBalance() const
+{
+    QMutexLocker locker(&m_cache_mutex);
```
promag, line 140: 0ac7d0cd538d79511105e0e4f69dc6e2a8f10a44

Are there concurrent accesses?

furszy: Hey, not yet. It's a glance into the future actually hehe.

At the moment it's only accessed from the main thread. At the following locations:

1) From `pollBalanceChanged` (`QTimer`).
2) From `setWalletModel` -> in `OverviewPage` and `SendCoinsDialog`.
3) From `useAvailableBalance` in `SendCoinsDialog`.
4) From `prepareTransaction` in `SendCoinsDialog`.

Now the rationale:
 I'm locally working on a branch that moves part of the processes to a worker thread thus why the guard. And added that commit here mainly because it would let me work on different flows in different scoped PRs without having to add it to each of them. 
- Not a big deal for me to remove it from this one -

promag: Personally, a good rule to follow is to only push stuff that is used or necessary.

furszy: np 👍🏼, dropped.

src/qt/walletmodel.cpp
```diff
@@ -120,12 +119,18 @@ void WalletModel::pollBalanceChanged()
 
 void WalletModel::checkBalanceChanged(const interfaces::WalletBalances& new_balances)
 {
-    if(new_balances.balanceChanged(m_cached_balances)) {
+    const auto& cached_balace = getCachedBalance();
```
achow101, line 126: In 0da631685626c84934735cf9ddef5cca834180a2 "GUI: add getter for WalletModel::m_cached_balances field."

nit: typo

```suggestion
    const auto& cached_balance = getCachedBalance();
```

src/qt/test/wallettests.cpp
```diff
@@ -193,15 +200,10 @@ void TestGUI(interfaces::Node& node)
...
-        // Check balance in send dialog
-        QLabel* balanceLabel = sendCoinsDialog.findChild<QLabel*>("labelBalance");
-        QString balanceText = balanceLabel->text();
-        BitcoinUnit unit = walletModel.getOptionsModel()->getDisplayUnit();
-        CAmount balance = walletModel.wallet().getBalance();
-        QString balanceComparison = BitcoinUnits::formatWithUnit(unit, balance, false, BitcoinUnits::SeparatorStyle::ALWAYS);
-        QCOMPARE(balanceText, balanceComparison);
-    }
+    // Update walletModel cached balance which will trigger an update for the 'labelBalance' QLabel.
+    walletModel.pollBalanceChanged();
```
achow101, line 204: In 7345a04a2f27d407a66595708b96537f8374be14 "test: GUI WalletTests, trigger walletModel balance changed manually."

It seems potentially bad that the cached balance might be out of sync with the real balance and that `pollBalanceChanged` needs to be called here in the tests?

furszy: It's out of sync merely because we are not starting the polling timer at all in the test. Same as we aren't initializing several parts of the GUI objects. All the objects (wallet, models, views, etc) that are used on this test are manually created instead of using the `WalletController` class flow.

Which I think that is actually what we are looking for. It's an unit test that is focused on verifying that few GUI widgets/views are behaving as expected: updating the presented information, etc. when they receive different signals and/or function calls from outside. It's not about whether the wallet balance polling timer is functioning as expected or not (which we definitely can create a new test case for it in a follow-up work).

furszy:
thanks achow101 for the review!

typo fixed 👍🏼.

hebasto:
The following part of the PR description
> 2\. Guarded `WalletModel::m_cached_balances` with its own mutex, so it can be used from the widgets as well. (the `OverviewPage` and `SendCoinsDialog` now are making use of it instead of calling `wallet().getBalances()` by their own).

looks [outdated](https://github.com/bitcoin-core/gui/pull/598#discussion_r868296044) now.

src/qt/walletmodel.cpp
```diff
@@ -120,12 +123,18 @@ void WalletModel::pollBalanceChanged()
...
 {
-    if(new_balances.balanceChanged(m_cached_balances)) {
+    const auto& cached_balance = getCachedBalance();
+    if (new_balances.balanceChanged(cached_balance)) {
         m_cached_balances = new_balances;
         Q_EMIT balanceChanged(new_balances);
     }
 }
 
+interfaces::WalletBalances WalletModel::getCachedBalance() const
```
hebasto, line 133: 08e74071e3ae30393a80d94a8286536ad7bfab3d

Without having a setter, maybe shorter `cachedBalance()` is better?

src/qt/overviewpage.cpp
```diff
@@ -178,6 +176,7 @@ void OverviewPage::handleTransactionClicked(const QModelIndex &index)
 void OverviewPage::setPrivacy(bool privacy)
 {
     m_privacy = privacy;
+    const auto& m_balances = walletModel->getCachedBalance();
```
hebasto, line 179: 061c6ee98af7185fa340989da8178e0149207e3f

`m_` prefix is used for member variables, not for local ones.

See https://github.com/bitcoin/bitcoin/blob/master/doc/developer-notes.md#coding-style-c

hebasto (commented):
It seems wrong to me when the commit 696683d7c330e2fa1591501aac8bb13324af68f9 _"GUI: use cached balance in overviewpage and sendcoinsdialog."_ breaks GUI `WalletTests`, then another commit fixes the broken test.

If a change is a refactoring then all tests must pass. Otherwise, if a commit introduces a new feature/functionality, it should change both main code and its tests.

src/qt/walletmodel.cpp
```diff
@@ -120,12 +123,18 @@ void WalletModel::pollBalanceChanged()
 
 void WalletModel::checkBalanceChanged(const interfaces::WalletBalances& new_balances)
 {
-    if(new_balances.balanceChanged(m_cached_balances)) {
+    const auto& cached_balance = getCachedBalance();
+    if (new_balances.balanceChanged(cached_balance)) {
```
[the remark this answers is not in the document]
furszy: This was because of https://github.com/bitcoin-core/gui/pull/598#discussion_r868204278.

Before I remove that commit, `getCachedBalance` was guarded by a mutex. Now is not needed anymore 👍🏼.

src/qt/walletmodel.cpp
furszy: At least for me, it's cleared to always use the same naming convention even if we don't have a setter right now.

I mean, generally speaking, if the method is a getter, then why not just name it `getSomething`?
So any dev using the model interface (or any interface) can find it right away without even need to think twice about it.

At least for me, those three characters make searches across large interfaces/projects simpler.

jarolrod (commented):
could break off 20d1b909b4c0836068e18d8982e67edeb593245c into it's own PR in order to make this focused on refactoring to use the model cache

furszy:
> could break off https://github.com/bitcoin-core/gui/commit/20d1b909b4c0836068e18d8982e67edeb593245c into it's own PR in order to make this focused on refactoring to use the model cache

Hey @jarolrod, I honestly don't think that separating that easy reviewable fix would make any difference.

Would rather friendly ping/dm some possible reviewers as this set of changes will be quite beneficial for large wallets.

shaavan:
Thanks for the explanation @furszy! 

I understand now how setBalance is being called twice.

Also, after some thinking, I agree with your rationale for adding the `get` prefix. If small enough would make it easier to understand the use-case of this function, then I am OK with that!

Just one last nit suggestion.

> On line 308, of src/qt/overviewpage.cpp, the m_balances variable should be renamed to balances as the m_ prefix is reserved for global variables.

furszy:
Thanks @shaavan

> On line 308, of src/qt/overviewpage.cpp, the m_balances variable should be renamed to balances as the m_ prefix is reserved for global variables.

nit pushed in 42059fb 👍🏼

shaavan (approved):
ACK 42059fbc5cf4256c4f9e0b380164d34d962ae59a

Changes since my last review:
- Renamed `m_balances` -> `balances`

Apologize for the late reply. I missed your notification earlier when the PR was updated. Thanks for your patience!

w0xlt (approved):
ACK https://github.com/bitcoin-core/gui/pull/598/commits/42059fbc5cf4256c4f9e0b380164d34d962ae59a

furszy:
@hebasto maybe merge?

src/qt/walletmodel.cpp
```diff
@@ -599,3 +610,7 @@ uint256 WalletModel::getLastBlockProcessed() const
 {
     return m_client_model ? m_client_model->getBestBlockHash() : uint256{};
 }
+
+CAmount WalletModel::getAvailableBalance(const CCoinControl* control) {
```
hebasto, line 614: ed0a32ec66db47b99d7d89e8c6df94b3f405f8f0, style nit:
```suggestion
CAmount WalletModel::getAvailableBalance(const CCoinControl* control)
{
```

src/qt/sendcoinsdialog.cpp
```diff
@@ -23,7 +23,6 @@
 #include <key_io.h>
 #include <node/ui_interface.h>
 #include <policy/fees.h>
-#include <txmempool.h>
```
hebasto, line 26: ed0a32ec66db47b99d7d89e8c6df94b3f405f8f0

This removal looks unrelated to this PR, doesn't it?

hebasto (approved):
ACK 42059fbc5cf4256c4f9e0b380164d34d962ae59a, I have reviewed the code and it looks OK.

src/qt/sendcoinsdialog.cpp
furszy: Yeah, it's a manual removal of an unused include.
(This is an old PR, prior me knowing about IWYU existence.. sorry for that, I forgot to update it)

Do you want me to revert it or we just go ahead with it?

hebasto: I'll really appreciate addressing of comments before merging :)

furszy: pushed

src/qt/walletmodel.cpp
furszy: done

furszy:
nits tackled.

hebasto (approved):
re-ACK 4584d300a40bfd84517072f7a6eee114fb7cab08, only suggested changes and commit message formatting since my [recent](https://github.com/bitcoin-core/gui/pull/598#pullrequestreview-1071268192) review.

hebasto:
@achow101 @jarolrod @shaavan @w0xlt 
Mind having another (final?) look at this PR?

jarolrod:
ACK 4584d300a40bfd84517072f7a6eee114fb7cab08

I tested this with a script i wrote to generate a bunch of blocks on regtest and generate 10K transactions for a wallet. I tested the wallet with this PR and Master. I didn't visually perceive a difference between master and this PR; probably due to the power of my desktop, but obviously the code shows that this is an improvement.

furszy:
> ACK [4584d30](https://github.com/bitcoin-core/gui/commit/4584d300a40bfd84517072f7a6eee114fb7cab08)
> 
> I tested this with a script i wrote to generate a bunch of blocks on regtest and generate 10K transactions for a wallet. I tested the wallet with this PR and Master. I didn't visually perceive a difference between master and this PR; probably due to the power of my desktop, but obviously the code shows that this is an improvement.

10k txes isn't that much (otherwise the wallet would be pretty unusable). 100k should be enough to notice it even on powerful machines. Will see a longer loading time for the wallet at startup (in the worker thread).

Still.. have to say that this is just the tip of the iceberg, and we do have other stuff that can be improved to decrease the wallet startup time.

PR #600, opened 2022-05-19
refactor: Add OptionsModel getOption/setOption methods

ryanofsky:
This is a trivial change which is needed as part of #602 to get bitcoind and bitcoin-qt to use the same settings instead of different settings. It is split off from #602 because it causes a lot of rebase conflicts (any time there is a GUI options change).

This PR is very small and easy to review ignoring whitespace: https://github.com/bitcoin-core/gui/pull/600/files?w=1

### DIFF
--- a/src/qt/optionsmodel.cpp
+++ b/src/qt/optionsmodel.cpp
@@ -335,81 +335,7 @@ QVariant OptionsModel::data(const QModelIndex & index, int role) const
 {
     if(role == Qt::EditRole)
     {
-        QSettings settings;
-        switch(index.row())
-        {
-        case StartAtStartup:
-            return GUIUtil::GetStartOnSystemStartup();
-        case ShowTrayIcon:
-            return m_show_tray_icon;
-        case MinimizeToTray:
-            return fMinimizeToTray;
-        case MapPortUPnP:
-#ifdef USE_UPNP
-            return settings.value("fUseUPnP");
-#else
-            return false;
-#endif // USE_UPNP
-        case MapPortNatpmp:
-#ifdef USE_NATPMP
-            return settings.value("fUseNatpmp");
-#else
-            return false;
-#endif // USE_NATPMP
-        case MinimizeOnClose:
-            return fMinimizeOnClose;
-
-        // default proxy
-        case ProxyUse:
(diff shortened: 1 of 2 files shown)


vasild (approved):
ACK a63b60f02bf7987d0a430496abe524de94f3c8cb

src/qt/optionsmodel.cpp
```diff
@@ -335,260 +335,272 @@ QVariant OptionsModel::data(const QModelIndex & index, int role) const
...
+// write QSettings values
+bool OptionsModel::setData(const QModelIndex & index, const QVariant & value, int role)
+{
+    bool successful = true; /* set to false on parse error */
+    if(role == Qt::EditRole)
+    {
+        successful = setOption(OptionID(index.row()), value);
+    }
+
+    Q_EMIT dataChanged(index, index);
```
furszy, line 352: Not for this PR but we actually don't need to call `dataChanged` if nothing changed inside `setOption` (which can happen if the new value is the same as the stored one or if something bad happened internally, like a bad parsing or a impossibility to store the new value).

furszy (approved):
Code review ACK a63b60f0

src/qt/optionsmodel.cpp
```diff
@@ -335,260 +335,272 @@ QVariant OptionsModel::data(const QModelIndex & index, int role) const
...
+        }
+        break;
+    case Server:
+        if (settings.value("server") != value) {
+            settings.setValue("server", value);
+            setRestartRequired(true);
+        }
+        break;
+    default:
+        break;
```
promag, line 601: I wonder if it should set `successful = false`? Or maybe just drop the `default` cases and let the compiler complain about unhandled cases?

src/qt/optionsmodel.cpp
promag: I think it's fine as is.

promag (commented):
Code review ACK a63b60f02bf7987d0a430496abe524de94f3c8cb.

jarolrod (commented):
post merge ACK https://github.com/bitcoin-core/gui/commit/a63b60f02bf7987d0a430496abe524de94f3c8cb

PR #601, opened 2022-05-19
refactor: Pass interfaces::Node references to OptionsModel constructor

ryanofsky:
Giving OptionsModel access to the node interface is needed as part of #602 to get bitcoind and bitcoin-qt to use the same settings instead of different settings.

It has been split off from #602 to simplify that PR. Previously these commits were part of bitcoin/bitcoin#15936 and also had some review discussion there.

### DIFF
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -261,7 +261,7 @@ void BitcoinApplication::createPaymentServer()
 
 void BitcoinApplication::createOptionsModel(bool resetSettings)
 {
-    optionsModel = new OptionsModel(this, resetSettings);
+    optionsModel = new OptionsModel(node(), this, resetSettings);
 }
 
 void BitcoinApplication::createWindow(const NetworkStyle *networkStyle)
@@ -292,7 +292,6 @@ void BitcoinApplication::createNode(interfaces::Init& init)
 {
     assert(!m_node);
     m_node = init.makeNode();
-    if (optionsModel) optionsModel->setNode(*m_node);
     if (m_splash) m_splash->setNode(*m_node);
 }
 
@@ -633,6 +632,12 @@ int GuiMain(int argc, char* argv[])
     // Allow parameter interaction before we create the options model
     app.parameterSetup();
     GUIUtil::LogQtInfo();
+
+    if (gArgs.GetBoolArg("-splash", DEFAULT_SPLASHSCREEN) && !gArgs.GetBoolArg("-min", false))
+        app.createSplashScreen(networkStyle.data());
+
+    app.createNode(*init);
+
     // Load GUI settings from QSettings
     app.createOptionsModel(gArgs.GetBoolArg("-resetguisettings", false));
 
@@ -640,11 +645,6 @@ int GuiMain(int argc, char* argv[])
         // Store intro dialog settings other than datadir (network specific)
         app.InitPruneSetting(prune_MiB);
     }
-
-    if (gArgs.GetBoolArg("-splash", DEFAULT_SPLASHSCREEN) && !gArgs.GetBoolArg("-min", false))
-        app.createSplashScreen(networkStyle.data());
-
-    app.createNode(*init);
 
     int rv = EXIT_SUCCESS;
     try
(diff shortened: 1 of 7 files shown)


vasild (approved):
ACK 945e4676dd3c92b353e13c78c8db673d8b64fe2d

src/qt/splashscreen.cpp
```diff
@@ -130,25 +130,19 @@ SplashScreen::SplashScreen(const NetworkStyle* networkStyle)
...
+    subscribeToCoreSignals();
 }
 
-void SplashScreen::setNode(interfaces::Node& node)
+SplashScreen::~SplashScreen()
 {
-    assert(!m_node);
-    m_node = &node;
-    subscribeToCoreSignals();
-    if (m_shutdown) m_node->startShutdown();
```
furszy, line 145: Do you know if there was any reason for having this `startShutdown` here?

furszy (approved):
Code review ACK 945e4676

ryanofsky (commented):
Updated 945e4676dd3c92b353e13c78c8db673d8b64fe2d -> 31122aa979c4c9a40e276cfc44243420c367ba4f ([`pr/qtnode.1`](https://github.com/ryanofsky/gui/commits/pr/qtnode.1) -> [`pr/qtnode.2`](https://github.com/ryanofsky/gui/commits/pr/qtnode.2), [compare](https://github.com/ryanofsky/gui/compare/pr/qtnode.1..pr/qtnode.2)) removing code that was changing space screen init, now initializing OptionsModel after the splash screen, instead of before.

chetooo40 (commented):
Till me

src/qt/bitcoin.cpp
```diff
@@ -261,7 +261,7 @@ void BitcoinApplication::createPaymentServer()
 
 void BitcoinApplication::createOptionsModel(bool resetSettings)
 {
-    optionsModel = new OptionsModel(this, resetSettings);
+    optionsModel = new OptionsModel(node(), this, resetSettings);
```
promag, line 264: Just noting that `node()` asserts for `m_node`.

src/qt/bitcoin.cpp
```diff
@@ -633,6 +632,12 @@ int GuiMain(int argc, char* argv[])
     // Allow parameter interaction before we create the options model
     app.parameterSetup();
     GUIUtil::LogQtInfo();
+
```
promag, line 635: This hunk was moved because now `createOptionsModel` needs `m_node` instanced.

promag (commented):
Code review ACK 31122aa979c4c9a40e276cfc44243420c367ba4f.

furszy (approved):
Code ACK 31122aa9

With a small note for:
> now initializing OptionsModel after the splash screen, instead of before.

We can probably initialize the SplashScreen after the `OptionsModel`, so at presentation time (`show()`) the widget can get access to the user selected language from there and not fallback to the system's language.
Still.. probably a big meh as we aren't having anything to translate in that screen atm.

ryanofsky:
> We can probably initialize the SplashScreen after the `OptionsModel`, so at presentation time (`show()`) the widget can get access to the user selected language from there and not fallback to the system's language

I think it actually shouldn't need OptionsModel for this, should just use gArgs.GetArg("-lang"). The early-init gui code uses ArgsManager and QSettings directly to get started, then it switches to interfaces::Node when the node is running. Probably unless the splash screen gets really fancy, it shouldn't need an OptionsModel pointer, and if it does it could be passed after it is started.

src/qt/optionsmodel.h
```diff
@@ -41,7 +41,7 @@ class OptionsModel : public QAbstractListModel
     Q_OBJECT
 
 public:
-    explicit OptionsModel(QObject *parent = nullptr, bool resetSettings = false);
+    explicit OptionsModel(interfaces::Node& node, QObject *parent = nullptr, bool resetSettings = false);
```
jarolrod, line 44: It seems to not be strictly necessary to have to initialize `OptionsModel` with a reference to `interfaces::Node` and instead could just be set at a later point which is similar to how we do it now. Wondering what is the justification to add this to the constructor?

jarolrod: nevermind, I see how this is needed in consideration of future changes :D

jarolrod (approved):
ACK 31122aa979c4c9a40e276cfc44243420c367ba4f

Reviewed the code and agree it can be merged. Tested this with future changes in #602.

Issue #596, opened 2022-04-29
RFC: Is it OK to reset unused prune and proxy settings to default values after restarting?


ryanofsky:
I'm opening this issue to get feedback and figure out how to resolve a disagreement that came up in https://github.com/bitcoin/bitcoin/pull/15936#discussion_r850568749.

### Background

PR https://github.com/bitcoin/bitcoin/pull/15936 unifies Bitcoin-Qt and Bitcoind settings so they are stored in `<datadir>/settings.json` instead of externally (in windows registry or external config directories via QSettings). It fixes current behavior where Bitcoind and Bitcoin-Qt can be started from the same datadir but may have inconsistent settings applied. It should make it easier to see what settings are in effect, because they will live in the datadir instead of being merged from outside sources. And it can enable new features like a `bitcoin-cli config` command that can update the unified settings dynamically.

The problem is the `<datadir>/settings.json` file does not currently store unused prune and proxy values when prune and proxy settings are disabled. So if you check the "Prune block storage" box and change the default "2GB" value to something else, then uncheck the box, then restart, then recheck the box, the previous value you typed will be lost, and the default "2GB" will be shown again. The same thing happens with the "Connect through SOCKS5 proxy" box. If you change the default tor proxy address "localhost:9050" to something else, and then disable the proxy, and restart, then recheck the box, the proxy address shown will be the default proxy address, not whatever proxy address you typed previously.

It is possible to change this behavior by storing unused proxy and prune settings somewhere. I don't personally think storing unused proxy or pruning settings values is a good thing or that it justifies extra complexity. But it would be possible to implement as a followup to https://github.com/bitcoin/bitcoin/pull/15936. One approach I implemented in https://github.com/ryanofsky/bitcoin/commit/4e86ab5dfc9bf40502aca0f9e3a4690f6f5dfc19 does this by adding `prune-prev` and `proxy-prev` settings that hold unused values when corresponding `prune` and `proxy` settings are disabled. Other approaches like storing unused prune and proxy settings in QSettings, or completely changing the representation of prune and proxy settings so they have "enabled" fields separate from the value fields would also be possible.

### Questions

1. What is ideal behavior when checking pruning/proxy boxes, typing pruning/proxy values, unchecking the boxes, restarting Bitcoin-Qt, checking the boxes again? Is it ideal to show default values (2GB and localhost:9050) when these boxes are checked? Or is ideal to show previous values? Or is some other behavior better?

[...]

PR #602, opened 2022-05-19
Unify bitcoin-qt and bitcoind persistent settings

ryanofsky:
If a setting like pruning, port mapping, or a network proxy is enabled in the GUI, it will now be stored in the bitcoin persistent setting file in the datadir and shared with bitcoind, instead of being stored as Qt settings which end up in the the windows registry or platform specific config files and are ignored by bitcoind.

This PR has been split off from bitcoin/bitcoin#15936 so some review of these commits previously took place in that PR.

### DIFF
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -259,9 +259,26 @@ void BitcoinApplication::createPaymentServer()
 }
 #endif
 
-void BitcoinApplication::createOptionsModel(bool resetSettings)
+bool BitcoinApplication::createOptionsModel(bool resetSettings)
 {
-    optionsModel = new OptionsModel(node(), this, resetSettings);
+    optionsModel = new OptionsModel(node(), this);
+    if (resetSettings) {
+        optionsModel->Reset();
+    }
+    bilingual_str error;
+    if (!optionsModel->Init(error)) {
+        fs::path settings_path;
+        if (gArgs.GetSettingsPath(&settings_path)) {
+            error += Untranslated("\n");
+            std::string quoted_path = strprintf("%s", fs::quoted(fs::PathToString(settings_path)));
+            error.original += strprintf("Settings file %s might be corrupt or invalid.", quoted_path);
+            error.translated += tr("Settings file %1 might be corrupt or invalid.").arg(QString::fromStdString(quoted_path)).toStdString();
+        }
+        InitError(error);
+        QMessageBox::critical(nullptr, PACKAGE_NAME, QString::fromStdString(error.translated));
+        return false;
+    }
+    return true;
 }
 
 void BitcoinApplication::createWindow(const NetworkStyle *networkStyle)
@@ -641,7 +658,9 @@ int GuiMain(int argc, char* argv[])
     app.createNode(*init);
 
     // Load GUI settings from QSettings
-    app.createOptionsModel(gArgs.GetBoolArg("-resetguisettings", false));
+    if (!app.createOptionsModel(gArgs.GetBoolArg("-resetguisettings", false))) {
+        return EXIT_FAILURE;
+    }
 
     if (did_show_intro) {
         // Store intro dialog settings other than datadir (network specific)
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -344,7 +344,7 @@ void BitcoinApplication::parameterSetup()
 
 void BitcoinApplication::InitPruneSetting(int64_t prune_MiB)
 {
-    optionsModel->SetPruneTargetGB(PruneMiBtoGB(prune_MiB), true);
+    optionsModel->SetPruneTargetGB(PruneMiBtoGB(prune_MiB));
 }
 
 void BitcoinApplication::requestInitialize()
--- a/src/qt/bitcoin.h
+++ b/src/qt/bitcoin.h
@@ -47,7 +47,7 @@ public:
     /// parameter interaction/setup based on rules
     void parameterSetup();
     /// Create options model
-    void createOptionsModel(bool resetSettings);
+    [[nodiscard]] bool createOptionsModel(bool resetSettings);
     /// Initialize prune setting
     void InitPruneSetting(int64_t prune_MiB);
     /// Create main window
--- a/src/qt/optionsmodel.cpp
+++ b/src/qt/optionsmodel.cpp
@@ -45,6 +45,12 @@ static const char* SettingName(OptionsModel::OptionID option)
     case OptionsModel::MapPortNatpmp: return "natpmp";
     case OptionsModel::Listen: return "listen";
     case OptionsModel::Server: return "server";
+    case OptionsModel::ProxyIP: return "proxy";
+    case OptionsModel::ProxyPort: return "proxy";
+    case OptionsModel::ProxyUse: return "proxy";
+    case OptionsModel::ProxyIPTor: return "onion";
+    case OptionsModel::ProxyPortTor: return "onion";
+    case OptionsModel::ProxyUseTor: return "onion";
     default: throw std::logic_error(strprintf("GUI option %i has no corresponding node setting.", option));
     }
 }
@@ -68,6 +74,14 @@ static void UpdateRwSetting(interfaces::Node& node, OptionsModel::OptionID optio
     }
 }
 
+struct ProxySetting {
+    bool is_set;
+    QString ip;
+    QString port;
+};
+static ProxySetting ParseProxyString(const std::string& proxy);
+static std::string ProxyString(bool is_set, QString ip, QString port);
+
 OptionsModel::OptionsModel(interfaces::Node& node, QObject *parent) :
     QAbstractListModel(parent), m_node{node}
 {
@@ -81,6 +95,14 @@ void OptionsModel::addOverriddenOption(const std::string &option)
 // Writes all missing QSettings with their default values
 bool OptionsModel::Init(bilingual_str& error)
 {
+    // Initialize display settings from stored settings.
+    ProxySetting proxy = ParseProxyString(SettingToString(node().getPersistentSetting("proxy"), GetDefaultProxyAddress().toStdString()));
+    m_proxy_ip = proxy.ip;
+    m_proxy_port = proxy.port;
+    ProxySetting onion = ParseProxyString(SettingToString(node().getPersistentSetting("onion"), GetDefaultProxyAddress().toStdString()));
+    m_onion_ip = onion.ip;
+    m_onion_port = onion.port;
+
     checkAndMigrate();
 
     QSettings settings;
@@ -133,7 +155,7 @@ bool OptionsModel::Init(bilingual_str& error)
     // These are shared with the core or have a command-line parameter
     // and we want command-line parameters to overwrite the GUI settings.
     for (OptionID option : {DatabaseCache, ThreadsScriptVerif, SpendZeroConfChange, ExternalSignerPath, MapPortUPnP,
-                            MapPortNatpmp, Listen, Server}) {
+                            MapPortNatpmp, Listen, Server, ProxyUse, ProxyUseTor}) {
         std::string setting = SettingName(option);
         if (node().isSettingIgnored(setting)) addOverriddenOption("-" + setting);
         try {
@@ -168,28 +190,6 @@ bool OptionsModel::Init(bilingual_str& error)
     }
     m_sub_fee_from_amount = settings.value("SubFeeFromAmount", false).toBool();
 #endif
-
-    // Network
-
-    if (!settings.contains("fUseProxy"))
-        settings.setValue("fUseProxy", false);
-    if (!settings.contains("addrProxy"))
-        settings.setValue("addrProxy", GetDefaultProxyAddress());
-    // Only try to set -proxy, if user has enabled fUseProxy
-    if ((settings.value("fUseProxy").toBool() && !gArgs.SoftSetArg("-proxy", settings.value("addrProxy").toString().toStdString())))
-        addOverriddenOption("-proxy");
-    else if(!settings.value("fUseProxy").toBool() && !gArgs.GetArg("-proxy", "").empty())
-        addOverriddenOption("-proxy");
-
-    if (!settings.contains("fUseSeparateProxyTor"))
-        settings.setValue("fUseSeparateProxyTor", false);
-    if (!settings.contains("addrSeparateProxyTor"))
-        settings.setValue("addrSeparateProxyTor", GetDefaultProxyAddress());
-    // Only try to set -onion, if user has enabled fUseSeparateProxyTor
-    if ((settings.value("fUseSeparateProxyTor").toBool() && !gArgs.SoftSetArg("-onion", settings.value("addrSeparateProxyTor").toString().toStdString())))
-        addOverriddenOption("-onion");
-    else if(!settings.value("fUseSeparateProxyTor").toBool() && !gArgs.GetArg("-onion", "").empty())
-        addOverriddenOption("-onion");
 
     // Display
     if (!settings.contains("language"))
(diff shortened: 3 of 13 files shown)


ryanofsky:
Rebased db64d21fa932fcf2b6d82117adb86a1059e0c920 -> 60a9db7f58e4ece13bfb607ce914f94e518a2635 ([`pr/qtsopt.1`](https://github.com/ryanofsky/gui/commits/pr/qtsopt.1) -> [`pr/qtsopt.2`](https://github.com/ryanofsky/gui/commits/pr/qtsopt.2), [compare](https://github.com/ryanofsky/gui/compare/pr/qtsopt.1-rebase..pr/qtsopt.2)) after bitcoin/bitcoin#24830 with no other changes.

bitcoin/bitcoin#24830 doesn't interact with this PR directly, but @Rspigler mentioned retesting this https://github.com/bitcoin/bitcoin/pull/15936#issuecomment-1132422382, and testing this PR is annoying without bitcoin/bitcoin#24830

vasild (approved):
ACK 60a9db7f58e4ece13bfb607ce914f94e518a2635

I feel strongly that the [vanishing disabled values after restarting](https://github.com/bitcoin/bitcoin/pull/15936#discussion_r852998379) should be resolved, but that can be done as a followup, before 24.x.

ryanofsky (commented):
Thanks for reviewing!

Rebased 60a9db7f58e4ece13bfb607ce914f94e518a2635 -> 9ae2f6ff0f780e34b40768a5d6097704859a7928 ([`pr/qtsopt.2`](https://github.com/ryanofsky/gui/commits/pr/qtsopt.2) -> [`pr/qtsopt.3`](https://github.com/ryanofsky/gui/commits/pr/qtsopt.3), [compare](https://github.com/ryanofsky/gui/compare/pr/qtsopt.2-rebase..pr/qtsopt.3)) with suggested fixes on top of updated base PRs

ryanofsky:
re: https://github.com/bitcoin-core/gui/pull/602#pullrequestreview-979858810

> I feel strongly that the [vanishing disabled values after restarting](https://github.com/bitcoin/bitcoin/pull/15936#discussion_r852998379) should be resolved, but that can be done as a followup, before 24.x.

Since we talked about multiple solutions for this in #596, I posted #603 as one possible solution. I really do like simplicity of not keeping unused values across restarts. If I set a temporary ssh proxy value one day while I'm using bad wifi, I don't want that value sticking forever after I disabled the proxy, and confusing me months later or making it more difficult to set up tor. But #596 discusses this issue, #603 is one possible solution, and others solutions should be possible too.

Rspigler:
tACK 9ae2f6ff0f780e34b40768a5d6097704859a7928

Now the GUI changes `settings.json` (instead of just `settings.json` affecting the GUI; it's a two way street).

One thing I noticed is, setting `prune=x,000` always results in x+1,000.  For example, prune=2 -> prune=3 GB.

`bitcoin.conf` does also now change the GUI, with `settings.json` taking precedence.  `bitcoin.conf` is never written to by the GUI or `settings.json`, as discussed [here](https://github.com/bitcoin/bitcoin/pull/15936#issuecomment-1013142512).

Also, main settings dialog has changed appropriately: "Options set in this dialog are overridden by the command line".  (It used to say ..."or also in the configuration file").  (It is not overridden by the config file, since it should be matching the config file).

One thing I noticed:

If `settings.json` is empty, and `bitcoin.conf` is filled with settings; I open bitcoin-qt and the settings are reflected there.  I disable and re-enable those settings, close and open bitcoin-qt, shouldn't those settings be written in `settings.json`? They are not.

promag (commented):
Concept ACK.

Rspigler:
Was having issues with tests on my end, but now everything is passing. (tACK https://github.com/bitcoin-core/gui/commit/9ae2f6ff0f780e34b40768a5d6097704859a7928)

ryanofsky:
re: https://github.com/bitcoin-core/gui/pull/602#issuecomment-1133545038

> One thing I noticed is, setting `prune=x,000` always results in x+1,000. For example, prune=2 -> prune=3 GB.

This can happen because `-prune` setting is specified in MiB (2^20 bytes) while GUI setting is specified in GB (10^9 bytes). So in various cases rounded values will be displayed. Like if you have `prune=600` in bitcoin.conf, the gui will display pruning size of 1GB. But as soon as you change settings in the GUI, it will add `"prune": 953` (1GB) or `1907` (2GB) or `2861` (3GB), etc to the `settings.json` file and whatever value is in `bitcoin.conf` will be ignored. If you edit the `settings.json` file manually, whatever value you specify will be used, but the GUI options dialog will convert it to GB for display. You should let me know if you see anything more strange than this, but hopefully behavior should be sensible in every case.

> If `settings.json` is empty, and `bitcoin.conf` is filled with settings; I open bitcoin-qt and the settings are reflected there. I disable and re-enable those settings, close and open bitcoin-qt, shouldn't those settings be written in `settings.json`? They are not.

This is intentional. Only settings that a user actually changes will be written to `settings.json`. If all settings were copied to `settings.json`, then `bitcoin.conf` would stop having any effect after `settings.json` was written. Intention is for `bitcoin.conf` to be a source of settings from a packager or sysadmin, and `settings.json` to contain customizations from the user. User settings have precedence, but system settings take effect otherwise.

ryanofsky:
Rebased 9ae2f6ff0f780e34b40768a5d6097704859a7928 -> 37682ee2fe70da9bc8c41713508b1f884e4c6215 ([`pr/qtsopt.3`](https://github.com/ryanofsky/gui/commits/pr/qtsopt.3) -> [`pr/qtsopt.4`](https://github.com/ryanofsky/gui/commits/pr/qtsopt.4), [compare](https://github.com/ryanofsky/gui/compare/pr/qtsopt.3-rebase..pr/qtsopt.4)) after #600 merge

ryanofsky:
Rebased 37682ee2fe70da9bc8c41713508b1f884e4c6215 -> 689dd6588c102ad4dc9e06e7465bd8d3acec273a ([`pr/qtsopt.4`](https://github.com/ryanofsky/gui/commits/pr/qtsopt.4) -> [`pr/qtsopt.5`](https://github.com/ryanofsky/gui/commits/pr/qtsopt.5), [compare](https://github.com/ryanofsky/gui/compare/pr/qtsopt.4-rebase..pr/qtsopt.5)) after base PR's #601 and https://github.com/bitcoin/bitcoin/pull/15936 were merged (no other changes or conflicts)

vasild (approved):
ACK 689dd6588c102ad4dc9e06e7465bd8d3acec273a

src/qt/bitcoin.cpp
```diff
@@ -259,9 +259,24 @@ void BitcoinApplication::createPaymentServer()
...
+    optionsModel = new OptionsModel(node(), this);
+    if (resetSettings) {
+        optionsModel->Reset();
+    }
+    bilingual_str error;
+    if (!optionsModel->Init(error)) {
+        fs::path settings_path;
+        if (gArgs.GetSettingsPath(&settings_path)) {
+            error += Untranslated("\n");
+            error += strprintf(_("Settings file %s might be corrupt or invalid."), fs::quoted(fs::PathToString(settings_path)));
```
hebasto, line 273: Could `_()` be avoided in the GUI code?

See #454.

ryanofsky: > Could `_()` be avoided in the GUI code?
> 
> See #454.

I'm looking at approach taken in #454, but it doesn't seem to deal with template strings that have parameters. If I need to generalize that approach and use `QObject::tr()` method to format `"Settings file %1 might be corrupt"` template for the GUI, and `strprintf` to format `Settings file %s might be corrupt" template for `debug.log`, that will be make the code more repetitive and make the translated/original code paths diverge more.

I wonder why `_()` should be avoided in GUI code if handles bilingual strings well and we want bilingual strings for an init error? Is there a technical reason why `_()` does not work? Would be be difficult to fix?

hebasto: > I wonder why `_()` should be avoided in GUI code if handles bilingual strings well and we want bilingual strings for an init error? Is there a technical reason why `_()` does not work? Would be be difficult to fix?

The reason is that the `qt` sub-directory is not parsed by the `share/qt/extract_strings_qt.py` script, and it results with missed translatable strings.

Besides #454, there were an alternative approach in bitcoin/bitcoin#22764.

ryanofsky (commented):
Updated 689dd6588c102ad4dc9e06e7465bd8d3acec273a -> ac5fb0107489f5b04559627bbb4eeee698c7e987 ([`pr/qtsopt.5`](https://github.com/ryanofsky/gui/commits/pr/qtsopt.5) -> [`pr/qtsopt.6`](https://github.com/ryanofsky/gui/commits/pr/qtsopt.6), [compare](https://github.com/ryanofsky/gui/compare/pr/qtsopt.5..pr/qtsopt.6)) removing `_()` from qt code as suggested https://github.com/bitcoin-core/gui/pull/602#discussion_r882835185. Maybe will change it back depending how discussion goes in https://github.com/bitcoin/bitcoin/pull/22764#issuecomment-1142564272

vasild (approved):
ACK ac5fb0107489f5b04559627bbb4eeee698c7e987

src/qt/bitcoin.h
```diff
@@ -47,7 +47,7 @@ class BitcoinApplication: public QApplication
     /// parameter interaction/setup based on rules
     void parameterSetup();
     /// Create options model
-    void createOptionsModel(bool resetSettings);
+    bool createOptionsModel(bool resetSettings);
```
hebasto, line 50: nit
```suggestion
    [[nodiscard]] bool createOptionsModel(bool resetSettings);
```

with this patch:
```diff
--- a/src/qt/test/apptests.cpp
+++ b/src/qt/test/apptests.cpp
@@ -72,7 +72,7 @@ void AppTests::appTests()
 
     qRegisterMetaType<interfaces::BlockAndHeaderTipInfo>("interfaces::BlockAndHeaderTipInfo");
     m_app.parameterSetup();
-    m_app.createOptionsModel(true /* reset settings */);
+    QVERIFY(m_app.createOptionsModel(true /* reset settings */));
     QScopedPointer<const NetworkStyle> style(NetworkStyle::instantiate(Params().NetworkIDString()));
     m_app.setupPlatformStyle();
     m_app.createWindow(style.data());
```

vasild (approved):
ACK bc6a56217b6d799c124cd4ec9881f572545fb0e7

src/qt/bitcoin.cpp
```diff
@@ -259,9 +259,26 @@ void BitcoinApplication::createPaymentServer()
...
+    if (!optionsModel->Init(error)) {
+        fs::path settings_path;
+        if (gArgs.GetSettingsPath(&settings_path)) {
+            error += Untranslated("\n");
+            std::string quoted_path = strprintf("%s", fs::quoted(fs::PathToString(settings_path)));
+            error.original += strprintf("Settings file %s might be corrupt or invalid.", quoted_path);
+            error.translated += tr("Settings file %1 might be corrupt or invalid.").arg(QString::fromStdString(quoted_path)).toStdString();
+        }
+        InitError(error);
+        QMessageBox::critical(nullptr, PACKAGE_NAME, QString::fromStdString(error.translated));
```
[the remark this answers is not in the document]
ryanofsky: re: https://github.com/bitcoin-core/gui/pull/602#discussion_r892764135

This wouldn't be intended, but I don't think this problem happens in practice, or at least I can't reproduce it. I can trigger the error with `settings.json` file:

```json
{
    "dbcache": [
    ]
}
```

In this case, InitError just logs an error without creating a message box because `app.createWindow` has not been called yet. After that the `QMessageBox::critical` call actually shows a message box.

There is a bunch of other in code the `src/qt/bitcoin.cpp` file following the pattern of calling `InitError` then `QMessageBox::critical`. It might be something that can be simplified in the future, but I think it is all correct and not buggy.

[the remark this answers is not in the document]
furszy: ah ok, that was a bit misleading when I read it. Make sense, thanks 👌🏼

furszy (commented):
Code reviewed bc6a5621 (except the prune commit).
Left few comments, nothing big, not blocking.

Plus a small point about 01f51429:
The commit message talks about `node.args = nullptr` removal but that is not happening there.

src/qt/optionsmodel.cpp
```diff
@@ -19,6 +19,7 @@
 #include <txdb.h>       // for -dbcache defaults
 #include <util/string.h>
 #include <validation.h> // For DEFAULT_SCRIPTCHECK_THREADS
+#include <wallet/wallet.h> // For DEFAULT_SPEND_ZEROCONF_CHANGE
```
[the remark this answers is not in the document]
ryanofsky: re: https://github.com/bitcoin-core/gui/pull/602#discussion_r893508525

> Not for this PR but I would rather make the options model call getter functions that return these default values from the node and wallet interfaces instead of accessing them through direct includes.
> 
> I think that it's going against the multi-processes separation goal.

I think the PR helps with this goal by at least getting node and gui to read settings from the same sources, but reading settings still requires dealing with unset settings and knowing default values. If settings API were better it would know about default values, but at least having a set of shared constants prevents values from getting out of sync.

ryanofsky (commented):
Thanks for the review!

Updated bc6a56217b6d799c124cd4ec9881f572545fb0e7 -> e47c6c76561807d30cff3c2e5372ea83c91a3677 ([`pr/qtsopt.7`](https://github.com/ryanofsky/gui/commits/pr/qtsopt.7) -> [`pr/qtsopt.8`](https://github.com/ryanofsky/gui/commits/pr/qtsopt.8), [compare](https://github.com/ryanofsky/gui/compare/pr/qtsopt.7..pr/qtsopt.8)) with suggested changes

re: https://github.com/bitcoin-core/gui/pull/602#pullrequestreview-1001310497

> The commit message talks about `node.args = nullptr` removal but that is not happening there.

Thanks, this was obsolete, dropped it

src/qt/optionsmodel.cpp
```diff
@@ -610,14 +581,14 @@ bool OptionsModel::setOption(OptionID option, const QVariant& value)
...
             setRestartRequired(true);
         }
         break;
     case Server:
-        if (settings.value("server") != value) {
-            settings.setValue("server", value);
+        if (changed()) {
+            update(value.toBool());
             setRestartRequired(true);
         }
```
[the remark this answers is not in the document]
furszy: yep, sounds good

src/qt/optionsmodel.cpp
```diff
@@ -288,29 +318,27 @@ static const QString GetDefaultProxyAddress()
...
-    }
-}
-
-void OptionsModel::SetPruneTargetGB(int prune_target_gb, bool force)
-{
-    const bool prune = prune_target_gb > 0;
-    if (prune) {
-        QSettings settings;
-        settings.setValue("nPruneSize", prune_target_gb);
+    const util::SettingsValue cur_value = node().getPersistentSetting("prune");
```
furszy, line 318: In 9a016a3c:

tiny nit:
Better to try to not use hardcoded strings, we could now do:
`auto prune_name = SettingName(Prune);`

(same for line 331)

furszy (approved):
Code review ACK e47c6c76

With two, non-blocking, points for e47c6c76:

1) When the user presses "Reset Options", the triggered dialog is not telling the user that a backup file will be created.

2) Maybe the backup file should have a different name: `settings1.json.bak`, `settings2.json.bak`, `settings3.json.bak` etc. If there is an existing backup? (otherwise the previous file is being overwritten)

hebasto (approved):
ACK e47c6c76561807d30cff3c2e5372ea83c91a3677

Going to merge this PR and leave nice @furszy's suggestions for follow ups.

Issue #604, opened 2022-05-20
SplashScreen segfault when wallet cannot be loaded


achow101:
Asa reported in https://github.com/bitcoin/bitcoin/issues/25146, when configured to load with a wallet that cannot be loaded, there is a segfault during the shutdown procedure.

I have been able to trace this segfault to the `SplashScreen` where the it appears that it is executing an `interfaces::Handler` that is accessing an object that has already been deleted. This seems to be related to this [comment](https://github.com/bitcoin-core/gui/blob/master/src/qt/bitcoin.cpp#L364L371) about cleaning up `WalletController`. Deleting `m_splash` at the same place appears to resolve the problem. However I am not sure if that is a viable long term fix.

Note that the segfault does not always occur for me. I believe this is due to deletion order which is not always the same. So sometimes `m_splash` is deleted before contexts are deleted, and sometimes after those contexts are deleted. I believe the segfault is due to accessing those contexts after deletion.

hebasto:
> I believe the segfault is due to accessing those contexts after deletion.

I can confirm that segfault is due the `context` object has been deleted before the `context.wallet_load_fns.erase(it);` call: https://github.com/bitcoin-core/gui/blob/640eb772e55671c5dab29843cebe42ec35cb703f/src/wallet/wallet.cpp#L161-L166

PR #605, opened 2022-05-22
Delete splash screen widget early

hebasto:
Fixes bitcoin-core/gui#604.
Fixes bitcoin/bitcoin#25146.
Fixes bitcoin/bitcoin#26340.

`SplashScreen::deleteLater()` [does not guarantee](https://doc.qt.io/qt-5/qobject.html#deleteLater) deletion of the `m_splash` object prior to the wallet context deletion. If the latter happens first, the [segfault](https://github.com/bitcoin-core/gui/issues/604#issuecomment-1133907013) follows.

### DIFF
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -312,7 +312,6 @@ void BitcoinApplication::createSplashScreen(const NetworkStyle *networkStyle)
     assert(!m_splash);
     m_splash = new SplashScreen(networkStyle);
     m_splash->show();
-    connect(this, &BitcoinApplication::splashFinished, m_splash, &SplashScreen::finish);
 }
 
 void BitcoinApplication::createNode(interfaces::Init& init)
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -311,11 +311,8 @@ void BitcoinApplication::createSplashScreen(const NetworkStyle *networkStyle)
 {
     assert(!m_splash);
     m_splash = new SplashScreen(networkStyle);
-    // We don't hold a direct pointer to the splash screen after creation, but the splash
-    // screen will take care of deleting itself when finish() happens.
     m_splash->show();
     connect(this, &BitcoinApplication::splashFinished, m_splash, &SplashScreen::finish);
-    connect(this, &BitcoinApplication::requestedShutdown, m_splash, &QWidget::close);
 }
 
 void BitcoinApplication::createNode(interfaces::Init& init)
@@ -373,6 +370,9 @@ void BitcoinApplication::requestShutdown()
         w->hide();
     }
 
+    delete m_splash;
+    m_splash = nullptr;
+
     // Show a simple window indicating shutdown status
     // Do this first as some of the steps may take some time below,
     // for example the RPC console may still be executing a command.
@@ -412,10 +412,13 @@ void BitcoinApplication::requestShutdown()
 void BitcoinApplication::initializeResult(bool success, interfaces::BlockAndHeaderTipInfo tip_info)
 {
     qDebug() << __func__ << ": Initialization result: " << success;
+
     // Set exit result.
     returnValue = success ? EXIT_SUCCESS : EXIT_FAILURE;
-    if(success)
-    {
+    if(success) {
+        delete m_splash;
+        m_splash = nullptr;
+
         // Log this only after AppInitMain finishes, as then logging setup is guaranteed complete
         qInfo() << "Platform customization:" << platformStyle->getName();
         clientModel = new ClientModel(node(), optionsModel);
@@ -438,7 +441,6 @@ void BitcoinApplication::initializeResult(bool success, interfaces::BlockAndHead
         } else {
             window->showMinimized();
         }
-        Q_EMIT splashFinished();
         Q_EMIT windowShown(window);
 
 #ifdef ENABLE_WALLET
@@ -455,7 +457,6 @@ void BitcoinApplication::initializeResult(bool success, interfaces::BlockAndHead
 #endif
         pollShutdownTimer->start(SHUTDOWN_POLLING_DELAY);
     } else {
-        Q_EMIT splashFinished(); // Make sure splash screen doesn't stick around during shutdown
         requestShutdown();
     }
 }
--- a/src/qt/bitcoin.h
+++ b/src/qt/bitcoin.h
@@ -89,7 +89,6 @@ public Q_SLOTS:
 Q_SIGNALS:
     void requestedInitialize();
     void requestedShutdown();
-    void splashFinished();
     void windowShown(BitcoinGUI* window);
 
 protected:
--- a/src/qt/splashscreen.cpp
+++ b/src/qt/splashscreen.cpp
@@ -161,16 +161,6 @@ bool SplashScreen::eventFilter(QObject * obj, QEvent * ev) {
     return QObject::eventFilter(obj, ev);
 }
 
-void SplashScreen::finish()
-{
-    /* If the window is minimized, hide() will be ignored. */
-    /* Make sure we de-minimize the splashscreen window before hiding */
-    if (isMinimized())
-        showNormal();
-    hide();
-    deleteLater(); // No more need for this
-}
-
 static void InitMessage(SplashScreen *splash, const std::string &message)
 {
     bool invoked = QMetaObject::invokeMethod(splash, "showMessage",
--- a/src/qt/splashscreen.h
+++ b/src/qt/splashscreen.h
@@ -37,9 +37,6 @@ protected:
     void closeEvent(QCloseEvent *event) override;
 
 public Q_SLOTS:
-    /** Hide the splash screen window and schedule the splash screen object for deletion */
-    void finish();
-
     /** Show message and progress */
     void showMessage(const QString &message, int alignment, const QColor &color);
 


luke-jr:
Why is this so much more complicated than https://github.com/bitcoin/bitcoin/issues/25146#issuecomment-1129356954 ?

hebasto:
> Why is this so much more complicated than https://github.com/bitcoin/bitcoin/issues/25146#issuecomment-1129356954 ?

Because this change makes the resulted code more simpler.

src/qt/bitcoin.cpp
```diff
@@ -381,6 +378,16 @@ void BitcoinApplication::requestShutdown()
 void BitcoinApplication::initializeResult(bool success, interfaces::BlockAndHeaderTipInfo tip_info)
 {
     qDebug() << __func__ << ": Initialization result: " << success;
+
+    if (m_splash) {
+        // If success, m_splash is no longer needed.
```
promag, line 382: 0370fee74b2d665851214f15c0ba2d28e544a40b#

according to the comment here, the condition above should be `if (success)` ?

hebasto: The `if (m_splash)` condition actually guards the case when `bitcoin-qt` is run with `-nosplash`.

In the comment, I was going to emphasize on the fact the we are going to delete `m_splash` _regardless_ of `success`.
But motivation of such deletion depends on `success`.

hebasto: Unneeded condition has been removed.

hebasto:
Updated 66646b6f4d63a6bad09ac6bca8247a2549f002d0 -> ec8d6aacd97b8a7d98de3943c626cf49fe4f1ac8 ([pr605.01](https://github.com/hebasto/gui/commits/pr605.01) -> [pr605.02](https://github.com/hebasto/gui/commits/pr605.02), [diff](https://github.com/hebasto/gui/compare/pr605.01..pr605.02)):

- drooped unneeded condition

ryanofsky:
I think I have a similar question to luke. With this PR Now there are two different manual deletions for wallet handler in two different places. The `BitcoinApplication::requestShutdown()` method has:

```c++
    // Delete wallet controller here manually, instead of relying on Qt object
    // tracking (https://doc.qt.io/qt-5/objecttrees.html). This makes sure
    // walletmodel m_handle_* notification handlers are deleted before wallets
    // are unloaded, which can simplify wallet implementations. It also avoids
    // these notifications having to be handled while GUI objects are being
    // destroyed, making GUI code less fragile as well.
    delete m_wallet_controller;
    m_wallet_controller = nullptr;
```

And the `BitcoinApplication::initializeResult` method has:

```c++
    // If success, m_splash is no longer needed.
    // Otherwise, it must be deleted prior to requesting shutdown
    // to make sure the `SplashScreen::m_connected_wallet_handlers`
    // is deleted before the wallet context is.
    delete m_splash;
    m_splash = nullptr;
```

Shouldn't both wallet handler deletions be consolidated together? Or is there something that makes them different from each other?

hebasto:
> With this PR Now there are two different manual deletions for wallet handler in two different places.

Correct. But indirectly, via deleting of objects, members of which are wallet handlers.

> Shouldn't both wallet handler deletions be consolidated together? Or is there something that makes them different from each other?

Those wallet handlers belong to very different objects, `m_wallet_controller` and `m_splash`, with different roles and life cycles.

OTOH, the suggested approach eliminates duplication of `Q_EMIT splashFinished()`, removes a signal, a slot, and two connections, and makes code less [spaghetti-like](https://www.erisian.com.au/bitcoin-core-dev/log-2022-05-16.html#l-246).

ryanofsky:
I agree this the approach in this PR is clean and simplifies code. I think I'm just wondering if it is sufficient to only delete the splash screen object in initializeResult, or if it the splash screen should also be deleted in requestShutdown? Will initializeResult always happen before requestShutdown? Also wondering why requestShutdown is the right place to delete the wallet controller.

The comments do a good job of explaining why it is important to delete these objects, but it's unclear why the objects are deleted in these places. It's possible there aren't concrete reasons, but I'm asking to learn if there are, and maybe improve the comments.

furszy:
Wouldn't make more sense to unsubscribe from the core signals before the `deleteLater` call so the object can be freed whenever QT wants without any fear of receiving a post-deletion signal?

 In other words, calling `SplashScreen::unsubscribeFromCoreSignals()` right before the `deleteLater()` call inside `SplashScreen::finish()`

hebasto:
> I agree this the approach in this PR is clean and simplifies code. I think I'm just wondering if it is sufficient to only delete the splash screen object in initializeResult, or if it the splash screen should also be deleted in requestShutdown? Will initializeResult always happen before requestShutdown? Also wondering why requestShutdown is the right place to delete the wallet controller.
> 
> The comments do a good job of explaining why it is important to delete these objects, but it's unclear why the objects are deleted in these places. It's possible there aren't concrete reasons, but I'm asking to learn if there are, and maybe improve the comments.

The universal entry point of the shutdown process is `BitcoinApplication::requestShutdown()` which handles both internal an OS-initiated cases, and it always runs in the "main" GUI thread.

ryanofsky:
re: https://github.com/bitcoin-core/gui/pull/605#issuecomment-1137436764

> OTOH, BitcoinApplication::initializeResult() is called via Qt::QueuedConnection from the "qt-init"/"shutoff" thread. Therefore, there are no guarantees that it always happens before BitcoinApplication::requestShutdown() (for example, in case when the latter is triggered by OS's QEvent::Quit).

So if `QEvent::Quit` happens, and `BitcoinApplication::requestShutdown()` runs before `BitcoinApplication::initializeResult()`, is there another bug still after this PR where the wallet is deleted before the splash screen, and there is a crash because the splash screen is still registered for wallet notifications?

If so, it seems like maybe we need to add `delete m_splash; m_splash = nullptr;` lines to **both** of the `requestShutdown()` and `initializeResult()` methods, not just one of the methods. Or is this not the case?

re: https://github.com/bitcoin-core/gui/pull/605#issuecomment-1136396327

> Wouldn't make sense to unsubscribe from the core signals before the `deleteLater` call so the object can be freed whenever QT wants without any fear of receiving a post-controller-deletion signal?

Unclear if it would make things more complicated to partially shut down wallet controller and splash screen objects by disconnecting them from signals, and let Qt delete them later, than to simply delete these objects ourselves when we know they are no longer needed. But maybe it would be more targeted and not actually more complicated. Qt does seem to not really care whether we delete the objects or it will. Both approaches seem like they would be ok.

hebasto:
> So if `QEvent::Quit` happens, and `BitcoinApplication::requestShutdown()` runs before `BitcoinApplication::initializeResult()`, is there another bug still after this PR where the wallet is deleted before the splash screen, and there is a crash because the splash screen is still registered for wallet notifications?

Correct, although this scenario is unlikely. Nevertheless, the code is flawed.

> If so, it seems like maybe we need to add `delete m_splash; m_splash = nullptr;` lines to **both** of the `requestShutdown()` and `initializeResult()` methods, not just one of the methods. Or is this not the case?

Would it cleaner to introduce a dedicated `SplashScreen::releaseWalletHandler()` member function, and leave `m_splash` object lifetime management to Qt framework? The same approach could be used for wallet controller.

hebasto:
> So if `QEvent::Quit` happens, and `BitcoinApplication::requestShutdown()` runs before `BitcoinApplication::initializeResult()`, is there another bug still after this PR where the wallet is deleted before the splash screen, and there is a crash because the splash screen is still registered for wallet notifications?

Added a commit to prevent such a scenario.

ryanofsky (approved):
Code review ACK d4227f7449d080f1a30e97847ef8d73d18bf872d, but I tend to think it would be good to keep the first three commits, drop the fourth commit "Ignore QEvent::Quit until full initialization", and add a new commit with achow's suggestion also deleting the splash screen in requestShutdown() so requestShutdown() is less fragile. (See comments below)

> Would it cleaner to introduce a dedicated `SplashScreen::releaseWalletHandler()` member function, and leave `m_splash` object lifetime management to Qt framework? The same approach could be used for wallet controller.

IMO it is cleanest just to delete the splash screen as soon as you don't need the splash screen, rather than keep the splash screen alive but remove the wallet handler underneath it. If it were actually helpful to the user to keep showing the splash screen while unloading, maybe a more complicated approach like this could be good, but otherwise it seems like it wouldn't improve things.

hebasto:
Updated d4227f7449d080f1a30e97847ef8d73d18bf872d -> d2cfc7d6afad634d1debf7cdf5488f4953a975d9 ([pr605.03](https://github.com/hebasto/gui/commits/pr605.03) -> [pr605.04](https://github.com/hebasto/gui/commits/pr605.04)):

- rebased
- addressed @ryanofsky's comments

> drop the fourth commit "Ignore QEvent::Quit until full initialization", and add a new commit with achow's suggestion also deleting the splash screen in requestShutdown() so requestShutdown() is less fragile

Done.

cc @achow101 @dooglus @furszy @luke-jr @promag @ryanofsky

src/qt/bitcoin.cpp
```diff
@@ -460,7 +461,7 @@ WId BitcoinApplication::getMainWinId() const
 
 bool BitcoinApplication::event(QEvent* e)
 {
-    if (e->type() == QEvent::Quit) {
+    if (m_initialized && e->type() == QEvent::Quit) {
```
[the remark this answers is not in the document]
hebasto: [Updated](https://github.com/bitcoin-core/gui/pull/605#issuecomment-1288767332).

achow101:
Code Review ACK d2cfc7d6afad634d1debf7cdf5488f4953a975d9

jarolrod (approved):
tACK d2cfc7d6afad634d1debf7cdf5488f4953a975d9

I have reviewed the code and I agree it can be merged. Additionally did testing where I quit in various scenarios where the splash screen was still loading and quiting afterwards as well. 

In relation to https://github.com/bitcoin/bitcoin/issues/26340, I still couldn't recreate the issue. But since the splash screen would be the only core window showing at this time, it's reasonable to think it could be related.

src/qt/bitcoin.cpp
```diff
@@ -388,14 +388,16 @@ void BitcoinApplication::requestShutdown()
...
+    // Delete splash screen and wallet controller here manually, instead of relying on Qt object
     // tracking (https://doc.qt.io/qt-5/objecttrees.html). This makes sure
     // walletmodel m_handle_* notification handlers are deleted before wallets
     // are unloaded, which can simplify wallet implementations. It also avoids
     // these notifications having to be handled while GUI objects are being
     // destroyed, making GUI code less fragile as well.
     delete m_wallet_controller;
     m_wallet_controller = nullptr;
+    delete m_splash;
+    m_splash = nullptr;
```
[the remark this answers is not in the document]
hebasto: > the object will not be deleted if a shutdown is requested during load.

It will, via Qt a parent-child relation.

Early widget deletion is crucial when wallet code is active only.

furszy (approved):
ACK 1b228497
The failing CI task is pure bad luck.

dooglus:
This fix causes a crash elsewhere: https://github.com/bitcoin/bitcoin/issues/26340#issuecomment-1290740422

john-moffett (approved):
ACK 1b228497fa729c512a15bdfa80f61a610abfe8a5

PR #607, opened 2022-05-26
refactor: Skip manual signal disconnection in descructors

hebasto:
Internally, `interfaces::HandlerImpl` keeps a member of the `boost::signals2::scoped_connection` type https://github.com/bitcoin-core/gui/blob/192d639a6b1bd0feaa52e6ea4e63e33982704c32/src/interfaces/handler.cpp#L21  [which](https://www.boost.org/doc/libs/1_64_0/doc/html/boost/signals2/scoped_connection.html) automatically disconnects on destruction.

Therefore, manual disconnections are redundant in class destructors.

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -228,9 +228,6 @@ BitcoinGUI::BitcoinGUI(interfaces::Node& node, const PlatformStyle *_platformSty
 
 BitcoinGUI::~BitcoinGUI()
 {
-    // Unsubscribe from notifications from core
-    unsubscribeFromCoreSignals();
-
     QSettings settings;
     settings.setValue("MainWindowGeometry", saveGeometry());
     if(trayIcon) // Hide tray icon, as deleting will let it linger until quit (on Ubuntu)
--- a/src/qt/clientmodel.cpp
+++ b/src/qt/clientmodel.cpp
@@ -69,8 +69,6 @@ ClientModel::ClientModel(interfaces::Node& node, OptionsModel *_optionsModel, QO
 
 ClientModel::~ClientModel()
 {
-    unsubscribeFromCoreSignals();
-
     m_thread->quit();
     m_thread->wait();
 }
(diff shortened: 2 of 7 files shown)


promag (commented):
Code review ACK 512072f9c969b414a71cec56e348ef6136a31385.

furszy (approved):
Code review ACK 512072f9

To not forget about this, would be nice to add a doc comment at the top of the `HandlerImpl` class and `MakeHandler(boost::signals2::connection)` explaining that the returned `Handler` disconnects the signal when the object is destructed.

jarolrod (commented):
ACK 512072f9c969b414a71cec56e348ef6136a31385

Confirmed PR description, and compiled and ran locally.

hebasto:
@ryanofsky 

Would you mind having a look into this PR?

ryanofsky (approved):
Tentative code review ACK 512072f9c969b414a71cec56e348ef6136a31385. For all the commits except the first commit, the destructors are non-empty, so this PR changes behavior by disconnecting signals after the other code in the destructor instead of before the other code in the constructor. In theory this could be a problem if a new signal is sent while the object is being destroyed. Previously the signal would be ignored, now it could call methods on a half destroyed object.

It would be good if commit messages acknowledged the delayed disconnection of signals, and perhaps explained why it was safe to delay disconnecting signals.

ryanofsky:
> > ... now it could call methods on a half destroyed object.
> 
> During destruction, only one thread has access to the object, right?

The lambdas passed to handleNotification methods are capturing `[this]` so they can reference the object when notifications are sent. That is why it might be important to disconnect the notifications earlier instead of later. But I don't know whether this is important, which is why I raised the question. Also even if notifications were single threaded (I don't think they are) it might still matter if they are disconnected later during destruction instead of earlier, because maybe later code in the destructor could trigger them.

Overall, I would guess this PR is probably safe, I just think it could use a better explanation of how it is safe.

hebasto:
Thank all of reviewers for your reviews.

Going to close this PR to do not raise safety concerns.

PR #609, opened 2022-05-29
wallet, refactor: Drop unused `WalletModel::PaymentRequestExpired`

hebasto:
The `PaymentRequestExpired` value in the `WalletModel::StatusCode` enumeration has been unused since bitcoin/bitcoin#17165.

### DIFF
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -757,10 +757,6 @@ void SendCoinsDialog::processSendCoinsReturn(const WalletModel::SendCoinsReturn 
     case WalletModel::AbsurdFee:
         msgParams.first = tr("A fee higher than %1 is considered an absurdly high fee.").arg(BitcoinUnits::formatWithUnit(model->getOptionsModel()->getDisplayUnit(), model->wallet().getDefaultMaxTxFee()));
         break;
-    case WalletModel::PaymentRequestExpired:
-        msgParams.first = tr("Payment request expired.");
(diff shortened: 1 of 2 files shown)


furszy (approved):
ACK 151009cf, no usage for it.

kristapsk (approved):
cr ACK 151009cf76f3f1adc17630d5370bf019be127373, checked that `PaymentRequestExpired` is not referenced anywhere else.

src/qt/sendcoinsdialog.cpp
```diff
@@ -757,10 +757,6 @@ void SendCoinsDialog::processSendCoinsReturn(const WalletModel::SendCoinsReturn
     case WalletModel::AbsurdFee:
         msgParams.first = tr("A fee higher than %1 is considered an absurdly high fee.").arg(BitcoinUnits::formatWithUnit(model->getOptionsModel()->getDisplayUnit(), model->wallet().getDefaultMaxTxFee()));
         break;
-    case WalletModel::PaymentRequestExpired:
-        msgParams.first = tr("Payment request expired.");
```
flack, line 761: Should the corresponding entries from the locae files also be removed, or is that done in Transifex somehow?

hebasto: That is responsibility of the [translation process](https://github.com/bitcoin/bitcoin/blob/master/doc/translation_process.md).

jarolrod (commented):
post merge ACK https://github.com/bitcoin-core/gui/commit/151009cf76f3f1adc17630d5370bf019be127373

PR #610, opened 2022-05-29
wallet: Detect `DuplicateAddress` earlier

hebasto:
This PR makes check for `DuplicateAddress` error easier and earlier.

### DIFF
--- a/src/qt/walletmodel.cpp
+++ b/src/qt/walletmodel.cpp
@@ -28,12 +28,13 @@
 #include <wallet/coincontrol.h>
 #include <wallet/wallet.h> // for CRecipient
 
-#include <stdint.h>
 #include <functional>
+#include <set>
+#include <stdint.h>
+#include <utility>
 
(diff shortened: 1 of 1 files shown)


furszy (commented):
Could move it even earlier in the flow and remove the code from the model (which requires the wallet to be locked):

Inside `PrepareSendText`, before calling `prepareTransaction`, we are looping over all the entries to load up the recipients list --> Line 257 of `SendCoinsDialog`.

hebasto:
> which requires the wallet to be locked

Do you mean "unlocking" using `WalletModel::UnlockContext`?

> Inside `SendCoinsDialog ::PrepareSendText`, before calling `WalletModel::prepareTransaction`, we are looping over all the entries to load up the recipients list --> Line 257 of `SendCoinsDialog`.

Would it make logic a bit messier?

hebasto:
Going to leave this PR up for grabs.

cc @furszy

PR #118, opened 2020-10-27
Remove BDB version from the Information tab

hebasto:
Master (67d4643a1a763b893ee69f6bbbf59ef1f2cb0d51):
DeepinScreenshot_select-area_20201027161350

This PR:
DeepinScreenshot_select-area_20201027161449

PR #613, opened 2022-06-02
Remove unnecessary wallet includes from rpcconsole.cpp

laanwj:
Fixes bitcoin/bitcoin#25266

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -27,14 +27,6 @@
 
 #include <univalue.h>
 
-#ifdef ENABLE_WALLET
-#ifdef USE_BDB
-#include <wallet/bdb.h>
-#endif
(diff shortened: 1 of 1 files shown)


MarcoFalke:
cr ACK 09942736495502cedcef4b097dde6abd9327fbdb

Should be correct if it compiles?

laanwj:
I think so! I also did a cursory check for direct use of wallet classes but couldn't find any. Everything goes through `WalletModel` etc.

hebasto (approved):
ACK 09942736495502cedcef4b097dde6abd9327fbdb.

It has become redundant since bitcoin-core/gui#118.

jarolrod (commented):
post merge ACK https://github.com/bitcoin-core/gui/commit/09942736495502cedcef4b097dde6abd9327fbdb

PR #615, opened 2022-06-08
If -prune=0 is set, Uncheck Prune on Intro page

jadijadi:
If the bitcoin-qt is started with -prune=0 arg, On the Intro page,
the Prune Checkbox will be unchecked too, to prevent confusions.

refs: https://github.com/bitcoin/bitcoin/issues/25052

### DIFF
--- a/src/qt/intro.cpp
+++ b/src/qt/intro.cpp
@@ -287,7 +287,7 @@ void Intro::setStatus(int status, const QString &message, quint64 bytesAvailable
         ui->freeSpace->setText("");
     } else {
         m_bytes_available = bytesAvailable;
-        if (ui->prune->isEnabled()) {
+        if (ui->prune->isEnabled() && !(gArgs.IsArgSet("-prune") && gArgs.GetIntArg("-prune", 0) == 0)) {
             ui->prune->setChecked(m_bytes_available < (m_blockchain_size_gb + m_chain_state_size_gb + 10) * GB_BYTES);
         }
         UpdateFreeSpaceLabel();


ryanofsky:
This is a probably a good-enough fix and better than the status quo, but it could be improved in the futue:

- If -prune=_size_ setting is given, size could be shown in intro dialog bog
- If data directory path is changed, and path contains bitcoin.conf or settings.json files with a prune value, that value could be shown.

But this fix is probably good enough to fix problem at hand

jadijadi:
Thanks for the valuable comments. The first one is already working as intended. I took a note about the 2nd one and will work on it separately. 
Thanks.

jarolrod (commented):
tACK 329c96626b17e1b4dd093de19c9695f5eddd4699

shaavan (approved):
Code Review ACK 329c96626b17e1b4dd093de19c9695f5eddd4699

- I find this to be a concise and straightforward fix to the issue it targets.

w0xlt (approved):
Code Review ACK https://github.com/bitcoin-core/gui/pull/615/commits/329c96626b17e1b4dd093de19c9695f5eddd4699

src/qt/intro.cpp
```diff
@@ -287,7 +287,7 @@ void Intro::setStatus(int status, const QString &message, quint64 bytesAvailable
         ui->freeSpace->setText("");
     } else {
         m_bytes_available = bytesAvailable;
-        if (ui->prune->isEnabled()) {
+        if (ui->prune->isEnabled() && gArgs.GetIntArg("-prune", -1) != 0) { // prune checkbox is enabled and not manually set to 0
```
hebasto, line 290: Could comment be dropped as it says exactly what code does?

Also I doubt a bit about using `-1` as the default value, because `GetIntArg("-prune", 0)` is used across the whole code base. It could be worked out as
```suggestion
        if (ui->prune->isEnabled() && !(gArgs.IsArgSet("-prune") && gArgs.GetIntArg("-prune", 0) == 0)) {
```

hebasto (commented):
When passing `-prune=` >1 to the command line, it forces the prune checkbox to be checked and disables it.

With this PR, when passing `-prune=0` to the command line, it forces the prune checkbox to be unchecked but does not disable it.

Should behavior be consistent?

src/qt/intro.cpp
jadijadi: Thanks for the improvements. You are right.

ryanofsky:
> When passing `-prune=` >1 to the command line, it forces the prune checkbox to be checked and disables it.
> 
> With this PR, when passing `-prune=0` to the command line, it forces the prune checkbox to be unchecked but does not disable it.
> 
> Should behavior be consistent?

This is a good point. It would be better for behavior to be consistent and it would be better to disable the setting if leaving it enabled gives user the misleading impression that setting they are chosing in the checkbox will be applied. For example, if bitcoin is started with -prune=0 and checkbox is enabled and user checks, they could have misleading impression that pruning will be applied when it won't be.

Even so, the PR in it's current form is still an improvement over the status quo. Disabling the checkbox would be an even bigger improvement, but just having checkbox show correct value is an improvement by itself.

hebasto (approved):
ACK f4b0c0c5c7149ffc6392d5becba0ec885fe6296b

> Even so, the PR in it's current form is still an improvement over the status quo. Disabling the checkbox would be an even bigger improvement, but just having checkbox show correct value is an improvement by itself.

Agree. Could leave a bigger improvement for an follow up.

@jadijadi Please [squash](https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#squashing-commits) all commits into one before merging.

jadijadi:
> ACK [f4b0c0c](https://github.com/bitcoin-core/gui/commit/f4b0c0c5c7149ffc6392d5becba0ec885fe6296b)
> 
> > Even so, the PR in it's current form is still an improvement over the status quo. Disabling the checkbox would be an even bigger improvement, but just having checkbox show correct value is an improvement by itself.
> 
> Agree. Could leave a bigger improvement for an follow up.
> 
> @jadijadi Please [squash](https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#squashing-commits) all commits into one before merging.

Thanks @hebasto , I squashed the commits into one.

hebasto (approved):
re-ACK 40566e21c02ddec6458cd83fa76fea081c5469dc

PR #618, opened 2022-06-14
refactor: Add `transactionoverviewwidget.cpp` source file

hebasto:
The `TransactionOverviewWidget` class was added in bitcoin-core/gui#176 as a header-only one.

Apparently, in upcoming [CMake project](https://github.com/hebasto/bitcoin/pull/3), CMake [AUTOMOC](https://cmake.org/cmake/help/latest/prop_tgt/AUTOMOC.html) could be integrated better/simpler, if `QObject`-derived class implementation been placed into a source file.

### DIFF
--- a/src/qt/transactionoverviewwidget.cpp
+++ b/src/qt/transactionoverviewwidget.cpp
@@ -1,0 +1,27 @@
+// Copyright (c) 2021 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include <qt/transactionoverviewwidget.h>
+
+#include <qt/transactiontablemodel.h>
+
+#include <QListView>
+#include <QSize>
+#include <QSizePolicy>
+
+TransactionOverviewWidget::TransactionOverviewWidget(QWidget* parent)
+    : QListView(parent) {}
+
+QSize TransactionOverviewWidget::sizeHint() const
+{
+    return {sizeHintForColumn(TransactionTableModel::ToAddress), QListView::sizeHint().height()};
+}
+
+void TransactionOverviewWidget::showEvent(QShowEvent* event)
+{
+    Q_UNUSED(event);
+    QSizePolicy sp = sizePolicy();
+    sp.setHorizontalPolicy(QSizePolicy::Minimum);
+    setSizePolicy(sp);
+}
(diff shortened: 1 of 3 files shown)


Sjors:
tACK a50e0b1bcb120e097a2b23dbfb71533b2f9c4d6c

Using a cpp file is consistent with other widgets we have.

shaavan (approved):
ACK a50e0b1bcb120e097a2b23dbfb71533b2f9c4d6c

- I agree with the idea of transforming `transactionoverviewwidget` from a **header-only** file to having a separate cpp file to implement functions. 
- The code change and cpp implementation are cleanly implemented, with proper formatting.
- I successfully ran the PR on Ubuntu 22.04 with `Qt version 5.15.3`.
- Observed no functional difference between the master and this PR, showing that this is a pure refactor change.

src/qt/transactionoverviewwidget.h
```diff
@@ -5,11 +5,8 @@
 #ifndef BITCOIN_QT_TRANSACTIONOVERVIEWWIDGET_H
 #define BITCOIN_QT_TRANSACTIONOVERVIEWWIDGET_H
 
-#include <qt/transactiontablemodel.h>
-
 #include <QListView>
```
furszy, line 8: `QListView` is being included in both files now. Could remove it from here.

src/qt/transactionoverviewwidget.cpp
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2021 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include <qt/transactionoverviewwidget.h>
+
+#include <qt/transactiontablemodel.h>
+
+#include <QListView>
+#include <QSize>
```
furszy, line 10: `QSize` is included in the header file, could delete this line.

src/qt/transactionoverviewwidget.h
hebasto: From our [Developer Notes](https://github.com/bitcoin/bitcoin/blob/master/doc/developer-notes.md#source-code-organization):
> Every `.cpp` and `.h` file should `#include` every header file it directly uses classes, functions or other definitions from, even if those headers are already included indirectly through other headers.

chetoo40 (approved):

jarolrod (commented):
post-merge ACK https://github.com/bitcoin-core/gui/commit/a50e0b1bcb120e097a2b23dbfb71533b2f9c4d6c

PR #623, opened 2022-06-21
Getting ready to Qt 6 (9/n). Apply Qt 6 specific changes

hebasto:
One more step in migration to Qt 6.

Could be tested with hebasto/bitcoin#3 or bitcoin/bitcoin#24798.

No behavior change when compiling with Qt 5.

### DIFF
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -139,12 +139,17 @@ static void initTranslations(QTranslator &qtTranslatorBase, QTranslator &qtTrans
     // - First load the translator for the base language, without territory
     // - Then load the more specific locale translator
 
+#if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
+    const QString translation_path{QLibraryInfo::location(QLibraryInfo::TranslationsPath)};
+#else
+    const QString translation_path{QLibraryInfo::path(QLibraryInfo::TranslationsPath)};
+#endif
     // Load e.g. qt_de.qm
-    if (qtTranslatorBase.load("qt_" + lang, QLibraryInfo::location(QLibraryInfo::TranslationsPath)))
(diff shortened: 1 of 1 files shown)


jarolrod (commented):
Concept ACK, need to finish testing 

- 3f51d0b8b2b2fabf5e305bd71226a8e3e5f3ad2d
  - I have confirmed with Qt Docs that the commit description is correct and we do in fact need to remove the `qRegisterMetaTypeStreamOperator` when building with Qt6. The usage of `qRegisterMetaType` is the appropriate call for Qt6.

- ad73447dc2bd04120c4eb7f5757613ad35a35cf5
  - Confirmed with Qt docs that for Qt6 we should use `QLibraryInfo::path` instead of `QLibraryInfo::location`. The new code here is correct as we use `QLibraryInfoPath` and we pass in a valid `::LibraryPath` enum in `::TranslationsPath`

- d8d99d041a063a1719fc1d43ecd4a6365e0657ca
  - Confirming with [Qtdocs](https://doc.qt.io/qt-6/highdpi.html#configuring-x11) that we don't need to set hidpi flags with Qt6: "... Qt 5 behavior assumes that AA_EnableHighDpiScaling has been set (this flag is not needed on Qt 6)"

jarolrod (approved):
ACK d8d99d041a063a1719fc1d43ecd4a6365e0657ca

Tested with Qt6 and the experimental Cmake branch and confirmed the window size and components look fine on a HIDPI and low dpi display. Also tested that I can still choose and switch through translations. Additionally, I performed a sanity check building with Qt5.

laanwj:
Code review ACK d8d99d041a063a1719fc1d43ecd4a6365e0657ca
(and lightly tested on Qt5 only, the GUI starts, doesn't look weird, translations still work)

PR #626, opened 2022-06-29
gui: Showing Local Addresses in Node Window

jadijadi:
This change adds one line to the Node Window (debugwindow.ui)
under the Network section and shows the LocalAddresses there.

fixed#564


src/qt/rpcconsole.cpp
```diff
@@ -958,6 +959,18 @@ void RPCConsole::updateNetworkState()
     }
 
     ui->numberOfConnections->setText(connections);
+
+    // localAddressess
+    std::string localAddresses="";
```
jarolrod, line 964: you could just make this a QString, then `item.first.ToString()` -> `QString::fromStdString(item.first.ToString())`

jarolrod: Should rename from `localAddresses` -> `local_addresses` to keep in line with our coding style

src/qt/rpcconsole.cpp
```diff
@@ -958,6 +959,18 @@ void RPCConsole::updateNetworkState()
...
+    LOCK(g_maplocalhost_mutex);
+    for (const std::pair<const CNetAddr, LocalServiceInfo> &item : mapLocalHost)
+        localAddresses += item.first.ToString() + ":" + std::to_string(item.second.nPort) + ", ";
```
jarolrod, line 967: to fix the linter issue arrising from `std::to_string(item.second.nPort), and keeping in mind that `locale_addresses` is now a QString: 

```suggestion
        local_addresses += QString::fromStdString(item.first.ToString()) + ":" + QString::number(item.second.nPort) + ", ";
```

`QString::number` fixes the linter issue, from [Qt Docs](https://doc.qt.io/qt-5/qstring.html#number): "The formatting always uses [QLocale::C](https://doc.qt.io/qt-5/qlocale.html#Language-enum), i.e., English/UnitedStates"

jarolrod (commented):
concept ack

full diff: 

``` diff
index 01989b5a0..080f9b979 100644
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -961,15 +961,16 @@ void RPCConsole::updateNetworkState()
     ui->numberOfConnections->setText(connections);

     // localAddressess
-    std::string localAddresses="";
+    QString local_addresses;
     LOCK(g_maplocalhost_mutex);
     for (const std::pair<const CNetAddr, LocalServiceInfo> &item : mapLocalHost)
-        localAddresses += item.first.ToString() + ":" + std::to_string(item.second.nPort) + ", ";
-    if (localAddresses.length() > 2)
-        localAddresses[localAddresses.length()-2] = 0;
+        local_addresses += QString::fromStdString(item.first.ToString()) + ":" + QString::number(item.second.nPort) + ", ";
+    if (local_addresses.length() > 2)
+        local_addresses[local_addresses.length()-2] = 0;
     else
-        localAddresses = "None";
-    ui->localAddresses->setText(QString::fromStdString(localAddresses));
+        //: Signals to the user that they do not have any local addresses.
+        local_addresses = tr("None");
+    ui->localAddresses->setText(local_addresses);

 }
```

src/qt/rpcconsole.cpp
jadijadi: Thank you very much. Greatly appreciate. Will do the changes & will commit.

src/qt/rpcconsole.cpp
```diff
@@ -49,6 +49,7 @@
 
 #include <chrono>
 
+
```
furszy, line 52: Can remove this extra added space

furszy (commented):
Aside from the code comments, there is an important topic to mention:

Let's not access fields from the net layer in a GUI widget/view, the sources architecture is widget -> model -> interface -> backend.

Better to create the proper flow to access the information that you are looking for:
rpcconsole -> clientModel -> node interface -> net layer.

In this way, the field access is encapsulated, layers responsibilities are respected and the same function can be used, and tested, from other user facing end points like the RPC server equally.

src/qt/rpcconsole.cpp
```diff
@@ -958,6 +959,18 @@ void RPCConsole::updateNetworkState()
...
+    // localAddressess
+    std::string localAddresses="";
+    LOCK(g_maplocalhost_mutex);
+    for (const std::pair<const CNetAddr, LocalServiceInfo> &item : mapLocalHost)
+        localAddresses += item.first.ToString() + ":" + std::to_string(item.second.nPort) + ", ";
+    if (localAddresses.length() > 2)
+        localAddresses[localAddresses.length()-2] = 0;
+    else
+        localAddresses = "None";
```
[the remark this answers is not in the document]
jarolrod: This is a review on an earlier state of the PR, the following would be the appropriate diff to express your idea:

``` diff
index 386b9373c..a249d7895 100644
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -964,13 +964,12 @@ void RPCConsole::updateNetworkState()
     // localAddressess
     QString local_addresses;
     LOCK(g_maplocalhost_mutex);
-    for (const std::pair<const CNetAddr, LocalServiceInfo> &item : mapLocalHost)
-        local_addresses += QString::fromStdString(item.first.ToString()) + ":" + QString::number(item.second.nPort) + ", ";
-    if (local_addresses.length() > 2)
-        local_addresses[local_addresses.length()-2] = 0;
-    else
-        // Signals to the user that they do not have any local addresses.
-        local_addresses = tr("None");
+    for (auto it = mapLocalHost.begin(); it != mapLocalHost.end();) {
+        local_addresses += QString::fromStdString(it->first.ToString()) + ":" + QString::number(it->second.nPort) + ", ";
+        if (++it != mapLocalHost.end()) local_addresses += ", ";
+    }
+    //: Signals to the user that they do not have any local addresses.
+    if (local_addresses.isEmpty()) local_addresses = tr("None");
     ui->localAddresses->setText(local_addresses);
 }
```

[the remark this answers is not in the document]
furszy: I pointed to the previous state on purpose. My main suggestion was to encapsulate the mutex and `mapLocalHost` access in the backend, which is Qt/GUI independent. So cannot use QString from there. A plain std::string that will need to be converted once it pass through the model class or directly in the widget.

src/qt/forms/debugwindow.ui
```diff
@@ -249,6 +249,29 @@
...
+         </property>
+        </widget>
+       </item>
+       <item row="9" column="1" colspan="2">
+        <widget class="QLabel" name="localAddresses">
+         <property name="cursor">
+          <cursorShape>IBeamCursor</cursorShape>
+         </property>
+         <property name="text">
+          <string>N/A</string>
```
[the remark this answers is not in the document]
jadijadi: thanks. fixed this. will push.

src/qt/rpcconsole.cpp
```diff
@@ -949,6 +949,7 @@ void RPCConsole::message(int category, const QString &message, bool html)
 
 void RPCConsole::updateNetworkState()
 {
+    // numberOfConnections
```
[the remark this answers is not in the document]
jadijadi: You are right. Variable names are self explanatory. will remove.

src/qt/rpcconsole.cpp
jadijadi: done. will push.

src/qt/forms/debugwindow.ui
```diff
@@ -249,6 +249,29 @@
         </widget>
        </item>
        <item row="9" column="0">
+        <widget class="QLabel" name="label_14">
```
shaavan, line 252: nit:

```suggestion
        <widget class="QLabel" name="labelLocalAddresses">
```

shaavan (commented):
Concept ACK 

- I agree with showing the **Local Address** on the **Information Tab**.
- I successfully compiled the PR on `Ubuntu 22.04` with `Qt version 5.15.3`.
- I was able to verify that the **Local Address** option is displayed correctly on the **Information** Tab.

**Screenshot:**

**Screenshot from 2022-07-05 17-09-17**

Currently, I am figuring out how to add local addresses to check if they are displayed correctly. It would be awesome if someone could point towards documentation on how to do so.

src/qt/forms/debugwindow.ui
jadijadi: thanks for the checks & ACK @shaavan . I kept the label_14 in accordance in most of other fields which had the same pattern.

regarding the listening port, you can test with something like (for testnet):

```
testnet=1
blocksonly=1
[test]
#prune=1000 # this has no effect during syncing
externalip=100.200.100.10
rpcbind=127.0.0.1
rpcallowip=0.0.0.0/0
rpcuser=user
rpcpassword=pass
```

as your bitcoin.conf to force a listening IP.

src/qt/rpcconsole.cpp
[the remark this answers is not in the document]
furszy: > Dear @furszy thanks for the valuable input. Applied most of them except this encapsulation part. I understand your rational and I think it's a better design. But since the number of connections are also calculated here, I did the same with Local Address.

The connections aren't calculated here (on the widget as you are doing for the external addresses). It's encapsulated inside the connections manager. The flow is the following one widget -> model -> node -> conman:

`updateNetworkState` calls `clientModel->getNumConnections` which calls the interface `node.getNodeCount`, which calls the connection manager.

shaavan (approved):
ACK c47f01bf25e1353b51c24b9b31260a539a394b1a

Continuing on my [previous review](https://github.com/bitcoin-core/gui/pull/626#pullrequestreview-1028560104):

I was able to add local addresses successfully and tested that those were displayed perfectly on the GUI under the **Local Address** option.

**Method of testing:**

I added the following options in the `bitcoin.conf` file.

```
[signet]
externalip=100.200.100.11
externalip=100.200.100.10
externalip=100.200.100.12
```

The IP addresses, along with the port number, appeared in the GUI.

**Screenshot:**
Screenshot from 2022-07-09 19-36-14

Note: The `IP:port` appears in the sorted order irrespective of the order provided in the `bitcoin.conf` file. This behavior is the same as the `-netinfo` RPC command.

**Screenshot:**
image

Thanks, @jadijadi, for the help on adding the listening ports!

src/qt/forms/debugwindow.ui
```diff
@@ -249,6 +249,29 @@
...
+         <property name="text">
+          <string>Local Address</string>
```
jarolrod, line 254: we could add a translator comment here by adding in the `extracomment` field:

```suggestion
          <string extracomment="nice translator comment">Local Address</string>
```

jarolrod: While here, if there's more than one local address; I guess this should take the plural form, no?

jarolrod (commented):
Code Review ACK c47f01bf25e1353b51c24b9b31260a539a394b1a

Some nits

pablomartin4btc (commented):
Concept ACK and tested c47f01bf25e1353b51c24b9b31260a539a394b1a

image

Thanks @furszy for emphasizing the importance of a good code design and architectural integrity. These principles are crucial for the long-term maintainability and scalability of our project. I totally agree with his code change recommendation in terms of the flow/ pattern to follow.

In terms of the UI itself, I'd use the plural form "Local Addresses", as @jarolrod also pointed it out, to be consistent with the `getnetworkinfo` rpc command output.

image

src/qt/forms/debugwindow.ui
kristapsk: Or could be just changed to "Local Address(es)" maybe?

kristapsk:
Concept ACK, kinda works.

Minor nit noticed - could avoid adding ":0" as a port for I2P addresses.

kristapsk:
Another thing - list is currently comma separated, without wrapping. Wouldn't it be too long if somebody has IPv4, IPv6, DJDNS, Tor and I2P addresses at the same time?

hebasto:
This PR still have a couple of unaddressed comments: https://github.com/bitcoin-core/gui/pull/626#discussion_r915334998 and https://github.com/bitcoin-core/gui/pull/626#discussion_r1274077350.

@jadijadi Are you still working on this?

jadijadi:
> This PR still have a couple of unaddressed comments: [#626 (comment)](https://github.com/bitcoin-core/gui/pull/626#discussion_r915334998) and [#626 (comment)](https://github.com/bitcoin-core/gui/pull/626#discussion_r1274077350).
> 
> @jadijadi Are you still working on this?

Will check and comment soon

jadijadi:
> Concept ACK, kinda works.
> 
> Minor nit noticed - could avoid adding ":0" as a port for I2P addresses.

thanks for the suggestion, did this.

jadijadi:
> Another thing - list is currently comma separated, without wrapping. Wouldn't it be too long if somebody has IPv4, IPv6, CJDNS, Tor and I2P addresses at the same time?

Thanks for this point, it is fixed and the widget worswraps and extends when needed.

src/qt/rpcconsole.cpp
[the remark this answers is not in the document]
jadijadi: Thanks. Did the flow as you said. Now we are querying layer by layer. Please comment if it still needs improvement. I decided to pass the `std::map<CNetAddr, LocalServiceInfo>` to keep it flexible enough for different use cases; for example in this specific usecase (showing IPs in GUI), i was able to do check for "IsI2P" on CNetAddrs.

src/qt/rpcconsole.cpp
```diff
@@ -958,6 +958,17 @@ void RPCConsole::updateNetworkState()
     }
 
     ui->numberOfConnections->setText(connections);
+
+    QString local_addresses;
+    LOCK(g_maplocalhost_mutex);
+    for (auto it = mapLocalHost.begin(); it != mapLocalHost.end();) {
+        local_addresses += QString::fromStdString(it->first.ToString()) + ":" + QString::number(it->second.nPort);
+        if (++it != mapLocalHost.end()) local_addresses += ", ";
```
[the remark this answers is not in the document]
jadijadi: changed to a normal loop and checking for .begin().

jadijadi:
> This PR still have a couple of unaddressed comments: [#626 (comment)](https://github.com/bitcoin-core/gui/pull/626#discussion_r915334998) and [#626 (comment)](https://github.com/bitcoin-core/gui/pull/626#discussion_r1274077350).

Dear @hebasto , addressed all issues, updated the code & replied to queries. Thanks for the follow up.

src/qt/rpcconsole.cpp
```diff
@@ -975,6 +975,20 @@ void RPCConsole::updateNetworkState()
...
+
+    QString local_addresses;
+    std::map<CNetAddr, LocalServiceInfo> mapHosts = clientModel->getLocalAddresses();
+    for (auto it = mapHosts.begin(); it != mapHosts.end(); it++) {
+        if (it != mapHosts.begin()) local_addresses += ", ";
+        if (it->first.IsI2P())
+            local_addresses += QString::fromStdString(it->first.ToStringAddr());
+        else
+            local_addresses += QString::fromStdString(it->first.ToStringAddr()) + ":" + QString::number(it->second.nPort);
+    }
```
furszy, line 987: Maybe?
```c++
QString local_addresses;
std::map<CNetAddr, LocalServiceInfo> hosts = clientModel->getLocalAddresses();
for (const auto& [addr, info] : hosts) {
    local_addresses += QString::fromStdString(it->first.ToStringAddr());
    if (!addr.IsPrivacyNet()) local_addresses += ":" + QString::number(it->second.nPort);
    local_addresses += ", ";
}
local_addresses.chop(2); // remove last ", "
```

furszy (commented):
thanks for the encapsulation jadijadi.

src/node/interfaces.cpp
```diff
@@ -281,6 +281,13 @@ class NodeImpl : public Node
         }
         return false;
     }
+    std::map<CNetAddr, LocalServiceInfo> getLocalAddresses() override
+    {
+        std::map<CNetAddr, LocalServiceInfo> empty;
+
+        return m_context->connman ? m_context->connman->getLocalAddresses() : empty;
```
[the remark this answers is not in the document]
jadijadi: Thanks for all the support and patient. Applying them. The `{}` was also my first try but I got:
`error: initializer list cannot be used on the right hand side of operator ':'`
Now I changed it to 
```
        if (m_context->connman)
            return m_context->connman->getLocalAddresses();
        else
            return {};
```
to prevent that ugly `empty` variable of mine.

src/qt/rpcconsole.cpp
jadijadi: This is more beautiful and understandable. Replaced the `it->first` & `it->second` with `addr` & `info`:

```
    for (const auto& [addr, info] : hosts) {
        local_addresses += QString::fromStdString(addr.ToStringAddr());
        if (!addr.IsPrivacyNet()) local_addresses += ":" + QString::number(info.nPort);
        local_addresses += ", ";
    }
    local_addresses.chop(2); // remove last ", "
```

src/qt/rpcconsole.cpp
```diff
@@ -975,6 +975,18 @@ void RPCConsole::updateNetworkState()
     }
 
     ui->numberOfConnections->setText(connections);
+
+    QString local_addresses;
+    std::map<CNetAddr, LocalServiceInfo> hosts = clientModel->getLocalAddresses();
+    for (const auto& [addr, info] : hosts) {
+        local_addresses += QString::fromStdString(addr.ToStringAddr());
+        if (!addr.IsPrivacyNet()) local_addresses += ":" + QString::number(info.nPort);
```
luke-jr, line 983: AFAIK Tor hidden services do still have port numbers

luke-jr:
Also, please split the interfaces changes into a different commit than the GUI changes.

luke-jr:
Might be a good improvement to cherry-pick [e4fddc9f395819d0bdf74003676aa563929756fa](https://github.com/luke-jr/bitcoin-core-gui/commit/e4fddc9f395819d0bdf74003676aa563929756fa) and use it to update; in testing, I found myself watching for None to change, but it never did because the string isn't updated.

jadijadi:
> Also, please split the interfaces changes into a different commit than the GUI changes.

did that. cleaner and atomic. thanks.

luke-jr:
>did that. cleaner and atomic. thanks.

Hmm, it's not that way in your new push tho

luke-jr:
Somehow you introduced a typo in the commit message >_<

jadijadi:
> Somehow you introduced a typo in the commit message >_<

Oh.. sorry. fixed and added a spellchecker.

jadijadi:
@luke-jr May I kindly ask for a review on this PR?

hebasto:
cc @furszy

src/qt/clientmodel.h
```diff
@@ -13,6 +13,9 @@
 #include <sync.h>
 #include <uint256.h>
 
+#include <net.h>
```
furszy, line 16: Could remove the need for adding this dependency here: 
```diff
diff --git a/src/qt/clientmodel.h b/src/qt/clientmodel.h
--- a/src/qt/clientmodel.h	(revision ea576b65592c23895f80c9116f881ff063253ff8)
+++ b/src/qt/clientmodel.h	(date 1722363555212)
@@ -13,7 +13,6 @@
 #include <sync.h>
 #include <uint256.h>
 
-#include <net.h>
 #include <netaddress.h>
 
 class BanTableModel;
@@ -22,6 +21,7 @@
 class PeerTableModel;
 class PeerTableSortProxy;
 enum class SynchronizationState;
+struct LocalServiceInfo;
 
 namespace interfaces {
 class Handler;
```

This will avoid including the backend `net.h` file in all the gui files that are already including `clientmodel.h`.

Ideally, `net.h` structs/classes shouldn't be accesible by the GUI widgets.  But no prob, it seems we are already including it in `rpcconsole.h` too.

jadijadi: Thanks @furszy , this looks better. Applied it and pushed the new version with not `net.h` and with the `struct` only; as you suggested.

furszy:
utACK

Maybe add a QLabel hint describing what "local addresses" are for GUI users? (and/or use another term for them).

jadijadi:
> utACK
> 
> Maybe add a QLabel hint describing what "local addresses" are for GUI users? (and/or use another term for them).

Any suggestion? English is not my first language. Something like "Addresses where Bitcoin Core binds to"? On *DataDir* we have the tooltip on the value column. I'll also keep it like that.

pablomartin4btc:
> > Maybe add a QLabel hint describing what "local addresses" are for GUI users? (and/or use another term for them).
> 
> Any suggestion? English is not my first language. Something like "Addresses where Bitcoin Core binds to"? On _DataDir_ we have the tooltip on the value column. I'll also keep it like that.

Perhaps any of these (or a combination):

- Each local address broadcast by this node.

- IP addresses of your local machine or network interface that the node uses to connect to the Bitcoin network

- Network addresses that your Bitcoin node is currently using to communicate with other nodes.

pablomartin4btc (commented):
<details>
<summary>re tACK bb5bb2001cf805301f56aa1ec15ed1035febab4b</summary>

image

</details>

src/qt/clientmodel.h
luke-jr: You changed the net.h stuff in the second commit, after adding it in the first

jadijadi:
> * Network addresses that your Bitcoin node is currently using to communicate with other nodes.

Thanks @pablomartin4btc . Added this as the tooltip & thanks for the tACK

src/qt/clientmodel.h
jadijadi: thanks! rebased.

furszy (commented):
utACK 189c987

PR #631, opened 2022-07-15
Disallow encryption of watchonly wallets

achow101:
Watchonly wallets do not have any private keys to encrypt. It does not make sense to encrypt such wallets, so disable the option to encrypt them.

This avoids an assertion that can be hit when encrypting watchonly descriptor wallets.

As our current behavior allows for encrypting watchonly wallets (no crash with legacy, crash, but still encrypted with descriptors), the new `NoKeys` status is only returned for unencrypted watchonly wallets. This allows any watchonly wallets that were previously encrypted to show the correct encryption status (they have encryption keys, and so should be indicated as being encrypted).

### DIFF
--- a/src/qt/walletmodel.h
+++ b/src/qt/walletmodel.h
@@ -71,6 +71,7 @@ public:
 
     enum EncryptionStatus
     {
+        NoKeys,       // wallet->IsWalletFlagSet(WALLET_FLAG_DISABLE_PRIVATE_KEYS)
         Unencrypted,  // !wallet->IsCrypted()
         Locked,       // wallet->IsCrypted() && wallet->IsLocked()
         Unlocked      // wallet->IsCrypted() && !wallet->IsLocked()
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -1350,6 +1350,12 @@ void BitcoinGUI::setEncryptionStatus(int status)
 {
     switch(status)
     {
+    case WalletModel::NoKeys:
+        labelWalletEncryptionIcon->hide();
+        encryptWalletAction->setChecked(false);
+        changePassphraseAction->setEnabled(false);
+        encryptWalletAction->setEnabled(false);
+        break;
     case WalletModel::Unencrypted:
         labelWalletEncryptionIcon->hide();
         encryptWalletAction->setChecked(false);
--- a/src/qt/walletmodel.cpp
+++ b/src/qt/walletmodel.cpp
@@ -306,6 +306,11 @@ WalletModel::EncryptionStatus WalletModel::getEncryptionStatus() const
 {
     if(!m_wallet->isCrypted())
     {
+        // A previous bug allowed for watchonly wallets to be encrypted (encryption keys set, but nothing is actually encrypted).
+        // To avoid misrepresenting the encryption status of such wallets, we only return NoKeys for watchonly wallets that are unencrypted.
+        if (m_wallet->privateKeysDisabled()) {
+            return NoKeys;
+        }
         return Unencrypted;
     }
     else if(m_wallet->isLocked())


w0xlt (approved):
tACK https://github.com/bitcoin-core/gui/pull/631/commits/4c495413e138ec1dd6874e41b44e689f0c15e0e3

I couldn't reproduce the crash mentioned on IRC. On my machine (Ubuntu 21.10 VM) nothing happened after encrypting a wallet with private keys disabled.
But this option should be disabled for these wallets anyway.

However, this change does not prevent the user from creating a blank wallet (without disabling the private keys) and encrypting an empty wallet.

achow101:
> I couldn't reproduce the crash mentioned on IRC. On my machine (Ubuntu 21.10 VM) nothing happened after encrypting a wallet with private keys disabled.

The crash is for descriptor wallet specifically. Legacy wallets will "encrypt" successfully.

> However, this change does not prevent the user from creating a blank wallet (without disabling the private keys) and encrypting an empty wallet.

Yes, blank wallets are not watchonly because they can have private keys imported.

katesalazar:
concept ACK

kristapsk:
Concept ACK

src/qt/walletmodel.h
```diff
@@ -71,6 +71,7 @@ class WalletModel : public QObject
 
     enum EncryptionStatus
     {
+        NoKeys,       // wallet->IsWalletFlagSet(WALLET_FLAG_DISABLE_PRIVATE_KEYS)
```
luke-jr, line 74: nit: `Unsupported` seems like it would fit better here

luke-jr (approved):
tACK, 1 nit

luke-jr:
Note: This could be rebased as far back as the v0.5.0 tag (or at the latest for 0.21 compatibility, 831675c8dccfa6525ffe751da3cc60709c380953) to cleanly merge to older branches.

hebasto:
The issue reported on [IRC](https://bitcoin-irc.chaincode.com/bitcoin-core-dev/2022-07-15#834241;):
> [12:01](https://bitcoin-irc.chaincode.com/bitcoin-core-dev/2022-07-15#834241) \<stevenroose> bitcoin-qt: wallet/scriptpubkeyman.cpp:1850: bool DescriptorScriptPubKeyMan::AddDescriptorKeyWithDB(WalletBatch&, const CKey&, const CPubKey&): Assertion `!m_storage.IsWalletFlagSet(WALLET_FLAG_DISABLE_PRIVATE_KEYS)' failed
[12:01](https://bitcoin-irc.chaincode.com/bitcoin-core-dev/2022-07-15#834243) \<stevenroose> qt crashes when trying to encrypt a watchonly wallet
[12:02](https://bitcoin-irc.chaincode.com/bitcoin-core-dev/2022-07-15#834244) \<stevenroose> it prompts for the passphrase and then hangs about 5 seconds and crashes

hebasto (commented):
Code changes look sane. But I cannot reproduce the initially reported issue to make sure it has been fixed.

FWIW, here is the `getwalletinfo` RPC output for my test watchonly descriptor wallet which I can encrypt with no crash:
```
{
  "walletname": "220718d-v22-DisablePK.1",
  "walletversion": 169900,
  "format": "sqlite",
  "balance": 0.00100000,
  "unconfirmed_balance": 0.00000000,
  "immature_balance": 0.00000000,
  "txcount": 1,
  "keypoolsize": 1000,
  "keypoolsize_hd_internal": 0,
  "paytxfee": 0.00000000,
  "private_keys_enabled": false,
  "avoid_reuse": false,
  "scanning": false,
  "descriptors": true,
  "external_signer": false
}
```

@achow101 
> The crash is for descriptor wallet specifically.

Could you provide steps to reproduce the issue reliably?

shaavan (commented):
Concept ACK

- I agree with the concept of this PR.
    - Since encrypting a wallet ⇒ encrypting its private keys. And since a watch-only wallet has none, encrypting it doesn’t make sense.
- I was able to verify the code and confirm that encrypting a watch-only wallet in the GUI is disabled by this PR.

**Screenshot:**
Screenshot from 2022-07-18 19-05-59


- However, I was also unable to reproduce the error on the master branch.

achow101:
> Could you provide steps to reproduce the issue reliably?

Create a new descriptor wallet with only private keys diasbled. It must not be marked as a blank wallet. This requires using the RPC to create the wallet as the GUI will always set the blank wallet flag. Then encrypt that wallet using the GUI and it will crash.

hebasto:
Backported to the `23.x` branch in https://github.com/bitcoin/bitcoin/pull/25828.

hebasto:
Backported to the `22.x` branch in https://github.com/bitcoin/bitcoin/pull/26521.

PR #129, opened 2020-11-05
Fix Shortcut Ambiguities, Clean up text

jarolrod:
Removes shortcut ambiguities found in the Settings -> Options menu. Builds upon this [rationale](https://github.com/bitcoin-core/gui/issues/126#issuecomment-720837550).

#### Before/Master
master

#### After/PR
pr

#### Main Tab
1. `Start Bitcoin Core on system login`
   - Moved shortcut from [**S**]tart to -> [**B**]itcoin in order to reserve S for `Size of database cache`
2. `Prune block storage to`
   - Moved shortcut from [**B**]lock to -> [**P**]rune in order to reserve B for `Start Bitcoin Core on system login`
   - Added a colon at the end since this text denotes a data field
3. `Size of database cache`
   - Moved shortcut from [**D**]atabase to -> [**S**]ize in order to fix the conflict with the `Display` tab shortcut
   - Added a colon at the end since this text denotes a data field
4. `Number of script verfication threads`
   - Added a colon at the end since this text denotes a data field

#### Wallet Tab
1. `Enable coin control features`
   - Moved shortcut from [**C**]oin to -> [**E**]nable in order to fix conflict with the `Cancel` button

#### Network Tab
1. `Connect through SOCKS5 proxy (default proxy):
   - Moved shortcut from [**C**]onnect to -> SOCKS[**5**] in order to fix conlict with the `Cancel` button

#### Window Tab
1. `Minimize to the tray instead of the taskbar`
   - Moved shortcut from [**M**]inimize to -> [**T**]ray in order to fix conflict with the `Main` tab

Closes #126

PR #633, opened 2022-07-21
Fix shortcut ambiguities

### LINKED ISSUE #126: Shortcut ambiguities
An example of shortcut ambiguity:
DeepinScreenshot_select-area_20201031160227

Pressing `Alt+D` a user could expect switching to the "Display" tab, but actually the "Size of database cache" widget is activated. The following `Alt+D` routes to the "Display" tab.

Other tabs of the "Options" window also have such ambiguities.

Localized versions of the GUI have such ambiguities too.

shaavan:
Picks up #129 
fixes #126 

- This PR addresses and fixes the shortcut collisions.
- This PR takes reference from [this](https://github.com/bitcoin-core/gui/issues/126#issuecomment-720837550) commit.

Changes done in this PR:

|Option whose shortcut is changed|From -> To|To avoid collision with|
|---------------------------------------------|---------|-----------------------------|
|Size of database cache|`D` -> `I`|**D**isplay tab|
|Enable coin control|`C` -> `a`|**C**ancel option|
|Connect through SOCKS5 proxy|`C` -> `S`|**C**ancel option|
|Minimize to tray instead of taskbar|`M` -> `T`|**M**ain tab|

Also, this PR address [this](https://github.com/bitcoin-core/gui/pull/129#issuecomment-725613606) comment in #129, which was suggested to add with the original PR.

> "Third party transaction URLs" string could be ended with `:`

### DIFF
--- a/src/qt/forms/optionsdialog.ui
+++ b/src/qt/forms/optionsdialog.ui
@@ -752,7 +752,7 @@
             <string>Third-party URLs (e.g. a block explorer) that appear in the transactions tab as context menu items. %s in the URL is replaced by transaction hash. Multiple URLs are separated by vertical bar |.</string>
            </property>
            <property name="text">
-            <string>&amp;Third-party transaction URLs</string>
+            <string>&amp;Third-party transaction URLs:</string>
            </property>
            <property name="buddy">
             <cstring>thirdPartyTxUrls</cstring>
--- a/src/qt/forms/optionsdialog.ui
+++ b/src/qt/forms/optionsdialog.ui
@@ -108,7 +108,7 @@
             <string extracomment="Tooltip text for Options window setting that sets the size of the database cache. Explains the corresponding effects of increasing/decreasing this value.">Maximum database cache size. A larger cache can contribute to faster sync, after which the benefit is less pronounced for most use cases. Lowering the cache size will reduce memory usage. Unused mempool memory is shared for this cache.</string>
            </property>
            <property name="text">
-            <string>Size of &amp;database cache</string>
+            <string>S&amp;ize of database cache</string>
            </property>
            <property name="textFormat">
             <enum>Qt::PlainText</enum>
@@ -238,7 +238,7 @@
              <string>Whether to show coin control features or not.</string>
             </property>
             <property name="text">
-             <string>Enable coin &amp;control features</string>
+             <string>Enable co&amp;in control features</string>
             </property>
            </widget>
           </item>
@@ -351,7 +351,7 @@
           <string>Connect to the Bitcoin network through a SOCKS5 proxy.</string>
          </property>
          <property name="text">
-          <string>&amp;Connect through SOCKS5 proxy (default proxy):</string>
+          <string>Connect through &amp;SOCKS5 proxy (default proxy):</string>
          </property>
         </widget>
        </item>
@@ -662,7 +662,7 @@
           <string>Show only a tray icon after minimizing the window.</string>
          </property>
          <property name="text">
-          <string>&amp;Minimize to the tray instead of the taskbar</string>
+          <string>Minimize to the &amp;tray instead of the taskbar</string>
          </property>
         </widget>
        </item>


shaavan:
Updated from f77451c8c11b9d41f5b6fa8659d947b0c99051e7 -> dd536e97b19743898c531909731b7c40977e2cc9 ([pr633.01](https://github.com/shaavan/gui/commits/pr633.01) -> [pr633.02](https://github.com/shaavan/gui/commits/pr633.02), [diff](https://github.com/shaavan/gui/compare/pr633.01..pr633.02))

Changes:

- Added another commit that changes the shortcut for the **Close** option from `l` -> `C` without introducing shortcut collisions.

The reasoning for the change:

- Close action is marked with shortcut `c` everywhere else in the GUI.
- `c` is a more intuitive shortcut for **Close** action than `l`.
- There is no shortcut conflict introduced with this change.

src/qt/forms/optionsdialog.ui
```diff
@@ -238,7 +238,7 @@
              <string>Whether to show coin control features or not.</string>
             </property>
             <property name="text">
-             <string>Enable coin &amp;control features</string>
+             <string>Enable co&amp;in control features</string>
```
jarolrod, line 241: This makes the shortcut hard to see, it might as well not be there if you can't reasonably see it. Additionally, the letter `i` seems off for representing `Enable coin control features`. Maybe consider if there is a letter here that can make the shortcut more visible and feel more appropriate for the feature?

shaavan: I think the visibility is slightly affected, so I agree with your suggestion of using another letter.

However, all the first intuitive first choices for this option, `E`, `c`, `o`, `f`, are used by other options.

I think `a` would be a good choice for this feature. Though it is not the most intuitive letter, it will help with the visibility as this would at the starting of the option name.

`En&amp;able coin control features`

src/qt/forms/optionsdialog.ui
```diff
@@ -238,7 +238,7 @@
              <string>Whether to show coin control features or not.</string>
             </property>
             <property name="text">
-             <string>Enable coin &amp;control features</string>
+             <string>En&amp;able coin control features</string>
```
jarolrod, line 241: you've introduced a new ambiguity within this file, now this option collides with the shortcut to switch to the `Window` tab within the `Options Window`. If you test this new shortcut, the `Enable Coin Control features` shortcut will not actuate.

jarolrod (commented):
It should be noted that this is picking up #129, and should look at the discussion had there especially the feedback. 

Additionally, there's an ambiguity not fixed within this PR that can be found in the `Receiving addresses` window. The shortcuts for the `Copy` button and the `Close` button collide.

I have gone through all of the shortcuts and can confirm there are no more besides the one mentioned above.

shaavan:
Updated from cfe407c81d66ea4b14737208eb568dfd36265a73 to 5fde8fbe0850218403c54d0cee04ecb2f36d79a0 ([pr633.03](https://github.com/shaavan/gui/commits/pr633.03) -> [pr633.04](https://github.com/shaavan/gui/commits/pr633.04), [diff](https://github.com/shaavan/gui/compare/pr633.03..pr633.04))

Addressed @jarolrod comment

- Reverted to pr633.01. Corresponding changes include:
    - Changed Enable coin control features shortcut from `a` to `i`.
    - Changes shortcut of close option from `C` to `l`.
- Added a commit to addressing [this](https://github.com/bitcoin-core/gui/pull/129#issuecomment-725613606) comment in PR #129 from which this PR is picked. 

Thank you very much, Jarol, for your thorough review!

src/qt/forms/optionsdialog.ui
shaavan: Thanks, Jarol, for catching this! 

I have reverted to the original change done for this option in this PR.
Addressed in 5fde8fb.

luke-jr (approved):
utACK 5fde8fbe0850218403c54d0cee04ecb2f36d79a0, but PR description needs to be updated

pablomartin4btc (approved):
Tested ACK.

I've reproduced the issue and verified the collisions on the Options window (from main menu "Settings" item "Options") between the shorcuts (alt+?) for the 4 elements across the different tabs and the shorcuts to navigate through the different tabs and/ or the "Cancel" button at right side at the bottom.

After compiling the fix 5fde8fbe0850218403c54d0cee04ecb2f36d79a0, I've verified that there's no collisions and the shortcuts work as expected.

image

I've pasted the screenshot above just as a reference and to show some of the shortcuts.

I've also added a "Where" column at the table that was provided at the top description to make it clearer.

<html>
<body>


Option whose shortcut is changed | Where | From -> To | To avoid collision with
-- | -- | -- | --
Size of database cache | Main tab | D -> I | Display tab
Enable coin control | Wallet tab | C -> a | Cancel option
Connect through SOCKS5 proxy | Network tab | C -> S | Cancel option
Minimize to tray instead of taskbar | Window tab | M -> T | Main tab


</body>
</html>

pablomartin4btc:
Since we are here, shouldn't we add a shortcut for the "Options" window under the "Settings" menu? Like Ctrl+O or Ctrl+Shift+O?

image

Another kind of unrelated comment but it's related somehow with one of the options ("Minimize on close" on Window tab) and the [Ctrl+W](https://doc.qt.io/qt-5/qkeysequence.html) shortcut. After playing a bit with some shortcuts and accidentally closing the app with Ctrl+W (Ctrl+Q is the shortcut to close the app specified on the File menu), I found this was worked out on [15768](https://github.com/bitcoin/bitcoin/pull/15768), shouldn't be this "Minimize on close" be ticked by default?

hebasto:
Maybe start from catching [`QShortcut::activatedAmbiguously()`](https://doc.qt.io/qt-5/qshortcut.html#activatedAmbiguously) signals? It would help to test this PR, and translated GUI and future changes as well.

hebasto:
@shaavan Still working on this PR?

hebasto:
Closing for now due to lack of progress. Let us know if this should be reopened.

Labeling "Up for grabs" as well.

PR #634, opened 2022-07-22
[WIP/Draft] Add Address tab

w0xlt:
This PR proposes a new tab called "Addresses" which shows some wallet addresses and also adds the `listaddress` RPC.
This is a common feature that many wallets have, but not Bitcoin Core clients (GUI and bitcoin-cli).
Most of commits were picked from https://github.com/bitcoin/bitcoin/pull/23019, but they were considerably modified.

As stated in the title, it is still a draft. I would like to understand if it is a feature desired by the reviewers before proceeding (since Bitcoin Core has a different security model and architecture than most wallets).

Some caveats:

. This PR, at moment (aa6d6a20c6bc8d0ff9577ad41010de6f234157f0), seems to work fine for descriptor wallets as `m_map_script_pub_keys` keeps all the keys, even after they have been generated (for receiving or change).

. But for legacy wallet, it has been showing inconsistent behavior. `setInternalKeyPool` and `setExternalKeyPool` remove the generated keys. I retrieved them from `m_address_book` but apparently this has only the receiving keys.

. To extract the index of a key in the HD path, this PR uses a regex. Although it seems reliable, may have performance impact.

. The Address table shows `m_hd_chain.m_next_{external,internal}_index + 5` addresses for legacy wallets and `m_wallet_descriptor.next_index + 5` for descriptor. It means it shows up to the last generated address + 5.

. Just for curiosity. I tested loading all addresses available in keypool (by default, 1000 keys for each keypool / descriptor) and this has a significant negative impact on wallet loading. I wonder if it would be possible to do this with good performance.

. pre-HD wallets have not been tested.

Feedback and suggestions on how to handle the issues mentioned above are welcome.

image


sipa:
I thought this was deliberate, because addresses are intended to be single-use.

jarolrod (commented):
Before getting to the performance improvements, this may have to undergo some soul searching.

This should not be added as a primary tab in the GUI; it's just not a core utility or function. Having this as one of the five main actions would confuse many users. Additionally, presuming a more "advanced" user is interested in this, you shouldn't be so interested in your addresses that a case is made to have this as a main function. 

If someone is interested in all the addresses like this, it would be better to keep it associated with how user's currently look into their address book. 

There should be a way to integrate this into the address book window. That should create more of a reason for introducing this change within the GUI.

luke-jr:
Concept NACK. This is just a more complicated/confusing address book.

Also, addresses do not have balances. They only receive, never send.

hebasto:

PR #638, opened 2022-08-03
Minor Formatting Fixes

2140data:
**Update receivecoinsdialog.ui** - Line 73 - Adding a whitespace character (&#160;) on the QT Receive Coins form instructions in order to show a space after the period in the application.

**Update sendcoinsdialog.ui** - Line 120 - Adding Sentence case to the label to maintain consistency

### DIFF
--- a/src/qt/forms/receivecoinsdialog.ui
+++ b/src/qt/forms/receivecoinsdialog.ui
@@ -70,7 +70,7 @@
         <item row="2" column="2">
          <widget class="QLabel" name="label_5">
           <property name="text">
-           <string>Use this form to request payments. All fields are &lt;b&gt;optional&lt;/b&gt;.</string>
+           <string>Use this form to request payments.&#160;All fields are &lt;b&gt;optional&lt;/b&gt;.</string>
           </property>
          </widget>
         </item>
--- a/src/qt/forms/sendcoinsdialog.ui
+++ b/src/qt/forms/sendcoinsdialog.ui
@@ -117,7 +117,7 @@
           <item>
            <widget class="QLabel" name="labelCoinControlAutomaticallySelected">
             <property name="text">
-             <string>automatically selected</string>
+             <string>Automatically selected</string>
             </property>
             <property name="margin">
              <number>5</number>


jarolrod (commented):
This is a bug in Qt where it sets the advance or width of the space character very low if it is preceded by a comma or period when running on macOS. It doesn't affect Linux or Windows. Additionally it only affects users on macOS with Qt less than 5.15.3 who are compiling the gui on their own. See: https://bugreports.qt.io/browse/QTBUG-88495

This has been fixed in Qt 5.15.3, we build our releases with Qt 5.15.5 now, and so this issue won't be seen with the release binary. I have confirmed this myself now by guix building this PR and master and testing it on macOS, the master branch built with Qt 5.15.5 does not show the space issue. 

I think it's fine to change the spaces that are preceeded by a comma or period to "&#160;", this would be fixing the issue specifically for users who are compiling the gui on their own and on macOS; we would want to do it everywhere else this formatting bug exists.

It should be noted this can be a can of worms if we are changing translatable strings to include these seemingly foreign string of characters. It can trip up translators and they could remove it from their translations. Unless transifex automatically excludes these characters (cc @hebasto).

luke-jr (commented):
Concept NACK.

The *best* place for a break (if needed) is between sentences. So keeping a normal space there is preferable. We don't need to workaround bugs in other software for something so minor that doesn't even effect the current version, especially when the workaround is semantically incorrect.

As for the "automatically selected", I think it actually looks better lowercased.

2140data:
Thank you. I see your points. No space is apparent between the sentences on the Mac, but it's not a big deal in the end (as noted). Maybe `\u0020` would be more correct if such a change was desired. Closing.

2140data:
Example of the issue, this time on the Welcome screen.

example

hebasto:
@2140data 
> Example of the issue, this time on the Welcome screen.

Please provide more details about `bitcoin-qt` binary you run. Version (if downloaded) or commit (if self-compiled)?

2140data:
Thank you. Apologies for not adding these details sooner.

- Mac OS Monterrey 12.4
- Bitcoin Core download Version 22.0.0

hebasto:
Bitcoin Core v22.0 GUI uses Qt [5.12.11](https://github.com/bitcoin/bitcoin/blob/v22.0/depends/packages/qt.mk#L2). It indeed has a [bug](https://bugreports.qt.io/browse/QTBUG-88495).

You could use Bitcoin Core v23.0 uses Qt [5.15.2](https://github.com/bitcoin/bitcoin/blob/v23.0/depends/packages/qt.mk#L2). But the bug has been fixed only in Qt 5.15.3.

Probably, we should bump Qt for v23.1.

---

Also see bitcoin/bitcoin#24939.

2140data:
Thank you everyone for your time and review.

PR #653, opened 2022-08-14
Show watchonly balance only for Legacy wallets

achow101:
Descriptor wallets do not have a watchonly balance as wallets are designated watchonly or not. Thus we should not be displaying the empty watchonly balance for descriptor wallets.

The result is that instead of the send page showing "Watch-only balance: 0.00000000 BTC" for watchonly descriptor wallets, we see the actual balance as "Balance: 10.00000000 BTC"

### DIFF
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -703,7 +703,7 @@ void SendCoinsDialog::setBalance(const interfaces::WalletBalances& balances)
         CAmount balance = balances.balance;
         if (model->wallet().hasExternalSigner()) {
             ui->labelBalanceName->setText(tr("External balance:"));
-        } else if (model->wallet().privateKeysDisabled()) {
+        } else if (model->wallet().isLegacy() && model->wallet().privateKeysDisabled()) {
             balance = balances.watch_only_balance;
             ui->labelBalanceName->setText(tr("Watch-only balance:"));
         }


luke-jr:
Hmm, maybe we should hide the "Balance" for watch-only descriptor wallets, and show the right thing as watch-only?

luke-jr:
eg

```diff
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -702,7 +702,9 @@ void SendCoinsDialog::setBalance(const interfaces::WalletBalances& balances)
         if (model->wallet().hasExternalSigner()) {
             ui->labelBalanceName->setText(tr("External balance:"));
         } else if (model->wallet().privateKeysDisabled()) {
-            balance = balances.watch_only_balance;
+            if (model->wallet().isLegacy()) {
+                balance = balances.watch_only_balance;
+            }
             ui->labelBalanceName->setText(tr("Watch-only balance:"));
         }
         ui->labelBalance->setText(BitcoinUnits::formatWithUnit(model->getOptionsModel()->getDisplayUnit(), balance));
```

hernanmarino (commented):
tACK.

This might be slightly off-topic, but shouldn't the whole "Receive" dialog be disabled when the wallet is unable to generate addresses ? Currently, only the receive button is disabled.

pablomartin4btc (approved):
Tested ACK (using signet).

Reproduced the issue (watchonly descritor wallets with a balance show watch-only balance 0 on the Send window): 

image

image

I've verified that it has been fixed with this change (now it shows balance with the total amount on the Send window matching the one shown on the overview window):

image

Checked also on a descriptor watch-only blank wallet and using importdescriptors function to have some balance within.

As @luke-jr, we could be still showing "watch-only balance" for watch-only descriptor wallets with the balance amount as shown in the overview window instead of the current solution, reafirming that we are on a watch-only wallet.

On @hernanmarino's comment, I think it makes sense to grey out the Receive button when the wallets can't generate addresses avoiding displaying the whole thing:

image

johnny9 (approved):
tACK fdb8dc8a5a10c39927dc4b8ddc5e0038a633c50e

Tested on signet by creating a no-private-key descriptor wallet and then using the importdescriptor command to add an address with funds on it.

on master (65de8eeeca29e71378aa34602b287ab921b040e4):
Send dialog shows invalid watch-only balance


Screenshot from 2022-12-31 20-15-03

PR (fdb8dc8a5a10c39927dc4b8ddc5e0038a633c50e):
Send dialog shows correct balance

Screenshot from 2022-12-31 20-38-22

pablomartin4btc (commented):
@achow101 are you planning to move ahead with this? any stoppers?

achow101:
> @achow101 are you planning to move ahead with this? any stoppers?

Just waiting for reviewers.

furszy (approved):
ACK fdb8dc8a

Inlines the balance presented in the send screen with the balance presented in the overview screen.

Not for this PR but would be good to have a visual distinction between wallets with private keys enabled vs wallets without private keys enabled (The previous "watch-only" wording wasn't that bad to fulfill the purpose).

hebasto (approved):
ACK fdb8dc8a5a10c39927dc4b8ddc5e0038a633c50e

Comments about further UX improving could be addressed in follow ups.

PR #654, opened 2022-08-15
Update source translations for string freeze

laanwj:
Make sure transifex source translations match the ones in the source code after today's string freeze.
It looks like only one message was added: "Too many external signers
found".


hebasto (approved):
ACK fcc0c8448b514c874f570f0ee517e4abc05c2574, `make -C src translate` produces zero diff.

jarolrod (approved):
ACK fcc0c8448b514c874f570f0ee517e4abc05c2574

performed my own update and have a zero-diff

Q: one of the updates occurred on the bitcoin repo, and now an update here. We should probably keep opening of these updates to one repo.

Issue #652, opened 2022-08-12
Persist "mask values" to prevent shoulder surfing


fanquake:
Current Behavior: "Mask values" mode is off each launch and must be manually re-enabled.
If a wallet loads on startup from settings.json this leaves user vulnerable to shoulder surfing.

Solution: This mode should persist across restarts of bitcoin-qt by being saved in Bitcoin-Qt.conf.
Exactly how nDisplayUnit=0 will persist the unit to show amounts in.

Expected Behavior: If I shutdown Bitcoin core with it on, I'd expect it to be on when I restart and vice versa.

Moved from: https://github.com/bitcoin/bitcoin/issues/25826.

pablomartin4btc:
This has been worked on #701 and I've worked on a follow-up masking the transactions view on #708

hernanmarino:
Since #701 was merged, i believe this issue can be closed, right ?  @fanquake

PR #655, opened 2022-08-25
Persist "mask values" in gui

jadijadi:
fixes https://github.com/bitcoin-core/gui/issues/652

### DIFF
--- a/src/qt/optionsmodel.cpp
+++ b/src/qt/optionsmodel.cpp
@@ -54,6 +54,7 @@ static const char* SettingName(OptionsModel::OptionID option)
     case OptionsModel::ProxyPortTor: return "onion";
     case OptionsModel::ProxyUseTor: return "onion";
     case OptionsModel::Language: return "lang";
+    case OptionsModel::Privacy: return "privacy";
     default: throw std::logic_error(strprintf("GUI option %i has no corresponding node setting.", option));
     }
 }
@@ -188,6 +189,10 @@ bool OptionsModel::Init(bilingual_str& error)
         settings.setValue("enable_psbt_controls", false);
     }
     m_enable_psbt_controls = settings.value("enable_psbt_controls", false).toBool();
+
+    if (!settings.contains("privacy")) {
+        settings.setValue("privacy", false);
+    }
 
     // These are shared with the core or have a command-line parameter
     // and we want command-line parameters to overwrite the GUI settings.
--- a/src/qt/overviewpage.cpp
+++ b/src/qt/overviewpage.cpp
@@ -16,6 +16,8 @@
 #include <qt/transactiontablemodel.h>
 #include <qt/walletmodel.h>
 
+#include <QSettings>
+
 #include <QAbstractItemDelegate>
 #include <QApplication>
 #include <QDateTime>
@@ -176,6 +178,9 @@ void OverviewPage::handleTransactionClicked(const QModelIndex &index)
 
 void OverviewPage::setPrivacy(bool privacy)
 {
+    QSettings settings;
+    settings.setValue("privacy", privacy);
+
     m_privacy = privacy;
     if (m_balances.balance != -1) {
         setBalance(m_balances);
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -246,6 +246,8 @@ BitcoinGUI::~BitcoinGUI()
 
 void BitcoinGUI::createActions()
 {
+    QSettings settings;
+
     QActionGroup *tabGroup = new QActionGroup(this);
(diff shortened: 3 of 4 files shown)


hernanmarino (approved):
tested ACK

furszy (changes requested):
Need to add an option inside the `OptionsModel` and use `setOption` to store/update it. Not call directly the backend function as it break the layers division.

Essentially, `OptionsModel` class is the interface from the GUI to the node's configuration data structure.
In other words, GUI widgets/models uses the `OptionsModel` to access/modifiy the node settings.

jadijadi:
> Need to add an option inside the `OptionsModel` and use `setOption` to store/update it. Not call directly the backend function, it breaks the layers division.

Thanks for the important note. I'm traveling this week with limited access to my dev machine / internet but will try to provide a fix per your suggestion soon. Thanks again.

ryanofsky:
> Need to add an option inside the `OptionsModel` and use `setOption` to store/update it. Not call directly the backend function, it breaks the layers division.
> 
> Essentially, `OptionsModel` class is the interface from the GUI to the node's configuration data structure. In other words, GUI widgets/models uses the `OptionsModel` to access/modify the node settings.

Also, instead of using `Node::getPersistentSetting` and `Node::updateRwSetting` methods, this should probably use `QSettings::value` and `QSettings::setValue` methods to access the setting. If you look at `OptionsModel` you can see that it uses `QSettings` to store GUI preferences. The `Node` methods are only used to store settings which are `bitcoind` command line or configuration file options (and `-privacy` is not a `bitcoind` option).

jadijadi:
> tested ACK

Thanks for the tests & ACK. I've done some changes based on the comments from @ryanofsky and @furszy . Can you please have another look? Thank you in advance.

jadijadi:
> > Need to add an option inside the `OptionsModel` and use `setOption` to store/update it. Not call directly the backend function, it breaks the layers division.
> 
> Also, instead of using `Node::getPersistentSetting` and `Node::updateRwSetting` methods, this should probably use `QSettings::value` and `QSettings::setValue` methods to access the setting. If you look at `OptionsModel` you can see that it uses `QSettings` to store GUI preferences. The `Node` methods are only used to store settings which are `bitcoind` command line or configuration file options (and `-privacy` is not a `bitcoind` option).

thanks @ryanofsky & @furszy . I've updated the PR  based your comment; via OptionsModel.

src/qt/optionsmodel.cpp
```diff
@@ -54,6 +54,7 @@ static const char* SettingName(OptionsModel::OptionID option)
     case OptionsModel::ProxyPortTor: return "onion";
     case OptionsModel::ProxyUseTor: return "onion";
     case OptionsModel::Language: return "lang";
+    case OptionsModel::Privacy: return "privacy";
```
ryanofsky, line 57: In commit "Persist "mask values" in gui" (9fde0bae916f73b73fcdbeff6469dba72c4be212)

This change to `SettingName` function doesn't do anything and should be reverted. `SettingName` maps `OptionModel` enum values to `bitcoin.conf` / command line option names, and there is no "-privacy" option, so this wouldn't do the right thing even if the case was ever used.

If you want to use `OptionsModel` to read and write this setting, instead of changing the `SettingName` function, you should change the `OptionsModel::getOption()` function by adding:

```c++
    case Privacy:
        return settings.value("privacy");
```

And the `OptionsModel::setOption()` function by adding:

```c++
    case Privacy:
        settings.setValue("privacy", value.toBool());
        break;
```

You could grep `CoinControlFeatures` or `EnablePSBTControls` to see how another similar GUI settings implemented in `OptionsModel`.

src/qt/overviewpage.cpp
```diff
@@ -176,6 +178,9 @@ void OverviewPage::handleTransactionClicked(const QModelIndex &index)
 
 void OverviewPage::setPrivacy(bool privacy)
 {
+    QSettings settings;
+    settings.setValue("privacy", privacy);
```
ryanofsky, line 182: In commit "Persist "mask values" in gui" (9fde0bae916f73b73fcdbeff6469dba72c4be212)

I don't think this is too important, but I'm not sure if I agree with furszy's comment that this setting needs to be part of `OptionsModel`. It seems nice to make it part of `OptionsModel` but also reasonable to not make it part of `OptionsModel` and just read/write it with `QSettings` directly. 

But probably it would be better to pick one approach or the other. Either don't add the setting to OptionsModel and use QSettings to read and write. Or do add the setting to OptionsModel and use OptionsModel to read and write. Otherwise you wind up causing inconsistent access to the setting across layers that furszy was trying to avoid https://github.com/bitcoin-core/gui/pull/655#pullrequestreview-1103143874

hebasto:
It looks like #701 is being actively reviewed recently. Let's combine all discussions there.

PR #662, opened 2022-09-06
Fix transaction view/table

luke-jr:
#204 reverted a necessary bugfix, and #205 introduced regressions since `setModel` resets column widths. Note that you need to delete your saved GUI config to see the fix, otherwise the prior widths are restored.

Before regressions:

Screenshot_20220905_232835 branch-21

After regressions / current master:

Screenshot_20220905_233054 branch-22

With this PR:

Screenshot_20220906_134210

### DIFF
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -499,6 +499,120 @@ bool LabelOutOfFocusEventFilter::eventFilter(QObject* watched, QEvent* event)
     return QObject::eventFilter(watched, event);
 }
 
+void TableViewLastColumnResizingFixer::connectViewHeadersSignals()
+{
+    connect(tableView->horizontalHeader(), &QHeaderView::sectionResized, this, &TableViewLastColumnResizingFixer::on_sectionResized);
+    connect(tableView->horizontalHeader(), &QHeaderView::geometriesChanged, this, &TableViewLastColumnResizingFixer::on_geometriesChanged);
+}
+
+// We need to disconnect these while handling the resize events, otherwise we can enter infinite loops.
+void TableViewLastColumnResizingFixer::disconnectViewHeadersSignals()
+{
+    disconnect(tableView->horizontalHeader(), &QHeaderView::sectionResized, this, &TableViewLastColumnResizingFixer::on_sectionResized);
+    disconnect(tableView->horizontalHeader(), &QHeaderView::geometriesChanged, this, &TableViewLastColumnResizingFixer::on_geometriesChanged);
+}
+
+// Setup the resize mode, handles compatibility for Qt5 and below as the method signatures changed.
+// Refactored here for readability.
+void TableViewLastColumnResizingFixer::setViewHeaderResizeMode(int logicalIndex, QHeaderView::ResizeMode resizeMode)
+{
+    tableView->horizontalHeader()->setSectionResizeMode(logicalIndex, resizeMode);
+}
+
+void TableViewLastColumnResizingFixer::resizeColumn(int nColumnIndex, int width)
+{
+    tableView->setColumnWidth(nColumnIndex, width);
+    tableView->horizontalHeader()->resizeSection(nColumnIndex, width);
+}
+
+int TableViewLastColumnResizingFixer::getColumnsWidth()
+{
+    int nColumnsWidthSum = 0;
+    for (int i = 0; i < columnCount; i++)
+    {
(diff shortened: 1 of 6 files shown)


jarolrod (commented):
if we revert the PR's aren't we still left with a buggy transaction table? Additionally, it's possible that the class mentioned can lead to a segfault when loading in a very large wallet:[ ](https://github.com/bitcoin/bitcoin/issues/20776)

luke-jr:
>if we revert the PR's aren't we still left with a buggy transaction table?

I'm not aware of any documented bugs. With the PRs reverted, it works perfect for me.

> Additionally, it's possible that the class mentioned can lead to a segfault when loading in a very large wallet:

These bugs seem outside the scope of this PR, but if you can explain, I can try to take a shot at fixing them?

hebasto:
From PR description it is not clear what actual bug this PR is supposed to fix.

> #204 reverted a necessary bugfix,

Wrong. See #204 PR description.

> and https://github.com/bitcoin-core/gui/pull/205 introduced regressions since `setModel` resets column widths.

Wrong. Column widths belong to a "View" layer. A "Model" layer should do nothing with a visual representation. It is up to user to set the "View" layer, including column widths.

luke-jr:
Not wrong. See screenshots.

hebasto:

achow101:
Reopening was requested.

luke-jr:
Re-confirmed bug still exists, and rebased this fix.

pablomartin4btc (commented):
Tested 8e428cc1188568aaeeaeb8f1f4498ae5b738a5e2 on Ubuntu 22.04.

Deleted `TransactionViewHeaderState` row manually from `qt` config file as requested in the description by the author (`.config/Bitcoin/Bitcoin-Qt-regtest.conf`), instead of using `-resetguisettings` as I didn't want to lose the rest.

As @hebasto posted above, "Amount" column goes out of scope when I resize to the right and then to the left (and the same happens with the "Requested payments history" in the receivecoinsdialog).
Peek 2023-09-16 14-39

Something I noticed when I resize the Label column enough to fit the most of the address (when are not labeled) is that with this PR changes, when the  "Amount" column is correctly resized and gets fit in the table view, it's changing the Label column size that I set manually, so that's a bit  contradictory, we let the user change something and then we correct the user:
Peek 2023-09-16 15-31

So, in terms of the 2 issues detected and explained above I prefer the current state of the table view, in `master` we also have the horizontal scrollbar at the bottom which has disappear with the changes of this PR.

Going thru all the PRs and beyond, is the actual issue that the "last column itself cannot be resized" (in the case of the transaction view/ table would be the "Amount" column)? In this case I see the issue is still in `master` (user can make it wider but not shrink it - this is because of `setStretchLastSection(true)`) and this PR doesn't fix it, the PR deals more with the resizing of the table but user can't make the last column shorter or larger (last column is set as "interactive" but it's fixed for the user perspective). I see where this is achieved is on the Peers table (`setStretchLastSection` is not called so it's `false` by default), user can set the last column "User Agent" width freely.

I can verify that there's another issue in `master`, and this PR solves, which is the fact that the default widths are not set properly the first time `qt` runs on a machine (or settings of the transaction view are missing from the qt config file) and user can see the 2nd. screenshot from the top description.

Perhaps the description of the PR could be clearer in terms of what the proper issue is and just mention the history of tries as a reference. This is original issue in the bitcoin repo: [#2862](https://github.com/bitcoin/bitcoin/issues/2862).

src/qt/guiutil.cpp
```diff
@@ -499,6 +499,120 @@ bool LabelOutOfFocusEventFilter::eventFilter(QObject* watched, QEvent* event)
...
+    tableView(table),
+    lastColumnMinimumWidth(lastColMinimumWidth),
+    allColumnsMinimumWidth(allColsMinimumWidth)
+{
+    columnCount = tableView->horizontalHeader()->count();
+    lastColumnIndex = columnCount - 1;
+    secondToLastColumnIndex = columnCount - 2;
+    tableView->horizontalHeader()->setMinimumSectionSize(allColumnsMinimumWidth);
+    setViewHeaderResizeMode(secondToLastColumnIndex, QHeaderView::Interactive);
+    setViewHeaderResizeMode(lastColumnIndex, QHeaderView::Interactive);
```
pablomartin4btc, line 613: This line doesn't make any difference, user still can't change the last column size manually.
```suggestion
```

hebasto:
Closing due to lack of support from the PR author (unaddressed comments, the failed CI).

PR #629, opened 2022-07-11
Fix translator comment for Restore Wallet `QInputDialog`

w0xlt:
Fix translator comment for Restore Wallet `QInputDialog`, as suggested in https://github.com/bitcoin-core/gui/pull/471#discussion_r917437779.

This also changes the window title name from `Restore Name` to `Restore Wallet` as it seems clearer.

PR #663, opened 2022-09-07
Cleanup translation comment

hebasto:
An unneeded character slipped in bitcoin-core/gui#629.

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -431,7 +431,7 @@ void BitcoinGUI::createActions()
 
             bool wallet_name_ok;
             /*: Title of pop-up window shown when the user is attempting to
-+                restore a wallet. */
+                restore a wallet. */
             QString title = tr("Restore Wallet");
             //: Label of the input field where the name of the wallet is entered.
             QString label = tr("Wallet Name");


jarolrod (approved):
ACK 5f28fc81600736d499391c4cb6dc34ece5d4229c

translation source file diff shows's a + sign where the extra space is, this PR removes it:
``` diff
diff --git a/src/qt/locale/bitcoin_en.ts b/src/qt/locale/bitcoin_en.ts
index 883d3fe24..5a974826d 100644
--- a/src/qt/locale/bitcoin_en.ts
+++ b/src/qt/locale/bitcoin_en.ts
@@ -751,7 +751,7 @@ Signing is only possible with addresses of the type &apos;legacy&apos;.</source>
     <message>
         <location line="+8"/>
         <source>Restore Wallet</source>
-        <extracomment>Title of pop-up window shown when the user is attempting to + restore a wallet.</extracomment>
+        <extracomment>Title of pop-up window shown when the user is attempting to restore a wallet.</extracomment>
         <translation type="unfinished"></translation>
     </message>
     <message>
```

jonatack (commented):
utACK 5f28fc81600736d499391c4cb6dc34ece5d4229c

PR #471, opened 2021-11-14
Add Wallet Restore in the GUI

w0xlt:
This PR adds a menu item to restore a wallet from a backup file in the GUI.
Currently this option exists only in RPC interface.

Motivation: It makes easier for non-technical users to restore backups.

Master | PR |
--- | --- 
master | pr |

PR #665, opened 2022-09-08
Switch to the selected wallet after loading

w0xlt:
Currently, the user loads a wallet and the screen does not switch to the selected wallet after loading (File -> Open Wallet -> wallet name).

This PR changes that by making the `OpenWalletActivity::opened` signal connection a `Qt::QueuedConnection` type.

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -410,7 +410,7 @@ void BitcoinGUI::createActions()
 
                 connect(action, &QAction::triggered, [this, path] {
                     auto activity = new OpenWalletActivity(m_wallet_controller, this);
-                    connect(activity, &OpenWalletActivity::opened, this, &BitcoinGUI::setCurrentWallet);
+                    connect(activity, &OpenWalletActivity::opened, this, &BitcoinGUI::setCurrentWallet, Qt::QueuedConnection);
                     activity->open(path);
                 });
             }


hebasto:
Maybe amend a PR title/commit message to mention "switch to the selected wallet after loading"?

src/qt/bitcoingui.cpp
```diff
@@ -410,7 +410,7 @@ void BitcoinGUI::createActions()
 
                 connect(action, &QAction::triggered, [this, path] {
                     auto activity = new OpenWalletActivity(m_wallet_controller, this);
-                    connect(activity, &OpenWalletActivity::opened, this, &BitcoinGUI::setCurrentWallet);
+                    connect(activity, &OpenWalletActivity::opened, this, &BitcoinGUI::setCurrentWallet, Qt::QueuedConnection);
```
hebasto, line 413: Why this solution works?

w0xlt: There's a good explanation here: https://github.com/bitcoin-core/gui/pull/471#pullrequestreview-989631968 as this is the same case found in Restore Wallet.

As the default value for this parameter is `Qt::AutoConnection`, it appears to be using `Qt::DirectConnection`, in which the slot is invoked immediately. 

https://doc.qt.io/qt-6/qt.html#ConnectionType-enum

luke-jr: Feels like a race that would be better being explicitly handled in the correct order?

jarolrod (commented):
ACK b8b59ff9fea69f4f25d98005e3ac9172a7a11c12

Can confirm that this fixes the issue. It appears safe to explicitly specify what kind of `ConnectionType` we want, there isn't an issue with the parameters given to the meta system and queuing this. We probably still want to continue to flesh this out and determine if we can refactor to remove explicitly specifying the `ConnectionType`

src/qt/bitcoingui.cpp
w0xlt: I think `Qt::DirectConnection` guarantees the correct order in this case, as  `BitcoinGUI::setCurrentWallet` will only run after control returns to the event loop of the receiver's thread.

pablomartin4btc:
Tested ACK on ubuntu 20.04.

Loading a wallet now sets it on the "Wallet:" combobox at the top right hand corner.

image

So far I couldn't find if there's an alternative to avoid specifying the [ConnectionType](https://doc.qt.io/qt-5/threads-qobject.html) on the params. I see the different types will depend on whether the sender and receiver belong to the same thread or not.

pablomartin4btc (commented):
There's a kind of similar issue when there's a selected wallet on the main window and opening the rpc console sets a different one. I didn't check yet (rpcconsole.cpp?) if there's a call to select a wallet and it's not working properly.

image

Sjors:
Concept ACK. It'd be nice if Console switches too; wouldn't be the first time I import descriptors into the wrong wallet :-)

PR #677, opened 2022-11-04
Bugfix: Address broken things around Peers detail view

luke-jr:
* Ensure data for previously-selected peer doesn't remain when a new peer is selected (reset all fields to N/A before loading data for a new peer)
* Load user-agent along with peer version (so a blank UA actually shows as blank)
* Display actually having no permissions as "None" rather than "N/A"

### DIFF
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -3224,6 +3224,10 @@ void PeerManagerImpl::ProcessMessage(CNode& pfrom, const std::string& msg_type, 
         // Change version
         const int greatest_common_version = std::min(nVersion, PROTOCOL_VERSION);
         pfrom.SetCommonVersion(greatest_common_version);
+        {
+            LOCK(pfrom.m_subver_mutex);
+            pfrom.cleanSubVer = cleanSubVer;
+        }
         pfrom.nVersion = nVersion;
 
         const CNetMsgMaker msg_maker(greatest_common_version);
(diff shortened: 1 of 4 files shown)


jonatack:
Concept ACK. This also cleanly implements https://github.com/bitcoin-core/gui/pull/673#discussion_r1008328740.

src/net_processing.cpp
```diff
@@ -3224,6 +3224,10 @@ void PeerManagerImpl::ProcessMessage(CNode& pfrom, const std::string& msg_type,
         // Change version
         const int greatest_common_version = std::min(nVersion, PROTOCOL_VERSION);
         pfrom.SetCommonVersion(greatest_common_version);
+        {
```
jarolrod, line 3227: we really shouldn't have this change being performed from this repo :)

jarolrod (commented):
Concept ACK

src/net_processing.cpp
hebasto: ping @luke-jr

luke-jr: @hebasto ping seen. nothing to do here though.

hebasto: > ping seen. nothing to do here though.

This PR modifies `src/net_processing.cpp`, and we review such changes in the main repo, generally.

Mind moving this PR or the first commit from this repo into the main one?

luke-jr: Doesn't seem worth splitting up such a minor change IMO

pablomartin4btc (commented):
Approach ACK

hebasto:
Why the first commit "net: Ensure CNode.cleanSubVer is always assigned before nVersion" is needed?

PR #680, opened 2022-11-15
Fixes MacOS 13 segfault by preventing certain notifications after main window is destroyed

john-moffett:
This is a PR to address https://github.com/bitcoin/bitcoin/issues/26490

The menu bar currently subscribes to window focus change notifications to enable or disable certain menu options in response to the window status. 

Notifications are automatically unsubscribed (disconnected in Qt parlance) if the sender is deleted -- in this case, the sender is the QTApplication object (`qApp`). However, MacOS 13 sends a window focus change notification *after* the main window has been destroyed but *before* `qApp` has been fully destroyed. 

Since the menu bar is deleted in the main window's destructor, it no longer exists when it receives these notifications (in two different places via lambda expressions). The solution is to pass the main window (`this`) as context when subscribing to the notifications. In this [overloaded version](https://doc.qt.io/qt-5/qobject.html#connect-1) of `connect`, Qt automatically unsubscribes to notifications if the sender OR context (here the main window object) is destroyed. Since the spurious notifications are sent after the main window object is destroyed, this change prevents them from being sent.

Tested on Mac OS 13 and 12 only.

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -512,7 +512,7 @@ void BitcoinGUI::createMenuBar()
     connect(minimize_action, &QAction::triggered, [] {
         QApplication::activeWindow()->showMinimized();
     });
-    connect(qApp, &QApplication::focusWindowChanged, [minimize_action] (QWindow* window) {
+    connect(qApp, &QApplication::focusWindowChanged, this, [minimize_action] (QWindow* window) {
         minimize_action->setEnabled(window != nullptr && (window->flags() & Qt::Dialog) != Qt::Dialog && window->windowState() != Qt::WindowMinimized);
     });
 
@@ -527,7 +527,7 @@ void BitcoinGUI::createMenuBar()
         }
     });
 
-    connect(qApp, &QApplication::focusWindowChanged, [zoom_action] (QWindow* window) {
+    connect(qApp, &QApplication::focusWindowChanged, this, [zoom_action] (QWindow* window) {
         zoom_action->setEnabled(window != nullptr);
     });
 #endif


jarolrod:
I have confirmed that the issue is isolated to macOS 13 and that this PR fixes the issue. 

Normally we don't reference an issue in the commit title, and I believe this can lead to some issues and unwanted notifications. A better commit title would be: `qt: pass main window as context when subscribing to notifications`. Then in the commit description you can add at the bottom of what you've already written: `closes bitcoin#26490`

hebasto:
Concept ACK.

hebasto:
@john-moffett 

Thank you for your such a nice first contribution!

Tested 8a5014cd8a05b3ab86ae34a47653a82ce11bdf17 on macOS Ventura (M1). It fixes https://github.com/bitcoin/bitcoin/issues/26490 indeed.

> The solution is to pass the main window (`this`) as context when subscribing to the notifications. In this [overloaded version](https://doc.qt.io/qt-5/qobject.html#connect-1) of `connect`, Qt automatically unsubscribes to notifications if the sender OR context (here the main window object) is destroyed.

I believe you was going to point at https://doc.qt.io/qt-5/qobject.html#connect-5 :)

Indeed,
> The connection will automatically disconnect if the sender or the context is destroyed. 

Therefore, I think in our code we should avoid connections without specifying a context explicitly.

hebasto (approved):
ACK 8a5014cd8a05b3ab86ae34a47653a82ce11bdf17

hebasto:
Backported to:
- 22.x in https://github.com/bitcoin/bitcoin/pull/26521
- 23.x in https://github.com/bitcoin/bitcoin/pull/26522
- 24.x in https://github.com/bitcoin/bitcoin/pull/26523

john-moffett:
Thank you, @hebasto ! I did indeed link to the wrong overload :grimacing:. Thank you for the correction. I agree that context ought to be passed to connections that use lambda expressions. I'll take a look to see if there are any other potentially vulnerable places in the code.

PR #681, opened 2022-11-17
Fix Transaction Relay tooltip text in Peers details window

jonatack:
as a value of N/A could occur due to a lock or a disconnection race but not during connection setup -- see https://github.com/bitcoin/bitcoin/pull/26457#pullrequestreview-1181641835.  Credit to Martin Zumsande for finding this.

### DIFF
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -1185,7 +1185,7 @@
                <item row="6" column="0">
                 <widget class="QLabel" name="peerRelayTxesLabel">
                  <property name="toolTip">
-                  <string>Whether we relay transactions to this peer (not available while the peer connection is being set up).</string>
+                  <string>Whether we relay transactions to this peer.</string>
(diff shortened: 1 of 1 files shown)


jarolrod (approved):
ACK 7a53033303f25301675fb2d7c7b7032166807910

The extra note here is redundant anyway at a UI level; it's not as if you're going to have a script that is dependent on the value the GUI is showing for this when setting up a peer.

One note: it's always nice to add a translation comment to strings as we are updating them. For strings in the `.ui` files, you can add an `extracomment` property to the string tag like so:

```html
<string extracomment="nice translator comment">
```

| master | pr |
| ------ | -- |
| Screen Shot 2022-11-17 at 1 52 24 AM | Screen Shot 2022-11-17 at 1 43 30 AM |

jonatack:
Thanks, @jarolrod! I'll try to keep your translation comment suggestion in mind.

Issue #635, opened 2022-07-25
Abandoning a transaction causes a desktop notification for sending/receiving it


hebasto:
When a transaction is abandoned in the GUI, a desktop notification pops up as if it was just created, saying either we just sent it, or we just received it.

Moved from bitcoin/bitcoin#8659.

katesalazar:
What this necromancy? Are you gonna fix?

PR #682, opened 2022-11-28
Don't directly delete abandoned txes from GUI

john-moffett:
This fully closes bitcoin/bitcoin#12179. Currently, when a user abandons a transaction by clicking "Abandon Transaction" in the context menu, a call is made to remove it from the GUI view:

`model->getTransactionTableModel()->updateTransaction(hashQStr, CT_UPDATED, false);`

(The `false` parameter is for `bool showTransaction`)

This behavior is probably unwanted, as the transaction is not actually removed from the wallet and would show up again if the node is restarted.

However, the previous line, `model->wallet().abandonTransaction(hash);`, changes the underlying model and calls `NotifyTransactionChanged(wtx.GetHash(), CT_UPDATED);`, which queues a signal that eventually calls back to `updateTransaction`, this time with `showTransaction` set to `true`. This runs on a separate thread, so it gets called *after* the 'subsequent' `updateTransaction`. The transaction gets removed from the GUI and immediately added back.

In a nutshell, `updateTransaction` gets called twice. The first (direct) call deletes the transaction from the GUI. The second (sent via a queued signal) brings it back to the GUI. The first direct call is redundant and unwanted. Worse, if the `abandonTransaction` call fails for any reason, the transaction still gets removed from the GUI. (This is what caused bitcoin#12179. It can still be triggered if, eg., a user clicks "Abandon Transaction" the moment after a new block is found.)

There are no conditions (to my knowledge) where an abandoned transaction should be directly removed from the GUI. If the underlying model changes, the deletion should be reflected anyway by the queued signal to `updateTransaction`.

The behavior is borne out by the QT logs. To reproduce, send a transaction with RBF enabled, then bump the fee, then 'abandon transaction' on the first transaction. The logs will show something like this:

```
2022-11-28T14:48:00Z [qt] GUI: "NotifyTransactionChanged: 2c5811484f1adec92a739a5e70b453b03eaed0f7cc0538fbd0ee1589e586b951 status= 1"
2022-11-28T14:48:00Z [qt] GUI: "TransactionTablePriv::updateWallet: 2c5811484f1adec92a739a5e70b453b03eaed0f7cc0538fbd0ee1589e586b951 1"
2022-11-28T14:48:00Z [qt] GUI: "    inModel=1 Index=381-382 showTransaction=0 derivedStatus=2"
2022-11-28T14:48:00Z [qt] GUI: "TransactionTablePriv::updateWallet: 2c5811484f1adec92a739a5e70b453b03eaed0f7cc0538fbd0ee1589e586b951 1"
2022-11-28T14:48:00Z [qt] GUI: "    inModel=0 Index=381-381 showTransaction=1 derivedStatus=0" 
```

Notice the duplicate `updateWallet` calls with different `showTransaction` values.

### DIFF
--- a/src/qt/transactionview.cpp
+++ b/src/qt/transactionview.cpp
@@ -421,9 +421,6 @@ void TransactionView::abandonTx()
 
     // Abandon the wallet transaction over the walletModel
     model->wallet().abandonTransaction(hash);
-
-    // Update the table
-    model->getTransactionTableModel()->updateTransaction(hashQStr, CT_UPDATED, false);
 }
 
 void TransactionView::bumpFee([[maybe_unused]] bool checked)


hebasto:
@john-moffett Thank you for your contribution!

> This fully closes [bitcoin/bitcoin#12179](https://github.com/bitcoin/bitcoin/issues/12179).

That issue looks outdated as it describes behavior no longer possible on the master branch, for example, a user's ability to abandon of a conflicted transaction (see https://github.com/bitcoin/bitcoin/pull/12282).

> Currently, when a user abandons a transaction by clicking "Abandon Transaction" in the context menu, a call is made to remove it from the GUI view

Is there a specific wrong or unexpected behavior from a user's point of view?

john-moffett:
@hebasto Happy to help.

> That issue looks outdated as it describes behavior no longer possible on the master branch, for example, a user's ability to abandon of a conflicted transaction (see https://github.com/bitcoin/bitcoin/pull/12282).

Yes, the main issue has already been fixed -- that a conflicted tx could still be "abandoned". However, an issue that still isn't fully fixed in master is that a row may be removed from the GUI when it shouldn't be. This is what was referred to in [this comment](https://github.com/bitcoin/bitcoin/issues/12179#issuecomment-366288150). It typically won't occur, but still can in edge cases. For example, the following steps will reproduce the issue:

1. Send an RBF-enabled transaction.
2. Bump the fee. 
3. Right click the original transaction in the GUI. Notice "Abandon Transaction" is clickable (as it should be).
image

5. A new block is added before you actually click "Abandon Transaction". (To simulate this, I started Qt in regtest with `-server` and issued `sleep 10 && src/bitcoin-cli generatetoaddress 1 bcrt....`)

6. Click "Abandon Transaction" (at this point it should be a no-op since it's now conflicted). However, the transaction is removed from the GUI:

image

7. Restart Bitcoin-Qt and the transaction reappears:
image

hebasto (approved):
ACK e75d2276324d54a01971afdf531df91748275bd5

I'm curios whether another `TransactionTableModel::updateTransaction()` call is still required: https://github.com/bitcoin-core/gui/blob/e75d2276324d54a01971afdf531df91748275bd5/src/qt/transactionview.cpp#L442 ?

john-moffett:
> I'm curious whether another TransactionTableModel::updateTransaction() call is still required:

I believe it is unnecessary, as `bumpfee` will (on success) have committed a new transaction and the old transaction(s) get updated here:

https://github.com/bitcoin-core/gui/blob/0596aa40f77a630d8a21035856fa5fd6838b292e/src/wallet/wallet.cpp#L2301

Indeed, bumping the fee via bitcoin-cli results in an updated GUI.

john-moffett:
This _may_ also close https://github.com/bitcoin-core/gui/issues/635. I haven't been able to replicate that, but given the behavior of delete-then-add-back in `master`, it would make sense.

PR #683, opened 2022-12-02
doc: Drop no longer relevant comment

hebasto:
The comment was introduced in 4cf3411056f6a59fc5fe07784b6b6a512d76b046, and since 7e4bd19785ff9120b7242ff9f15231868aaf7db4 it has been no longer relevant.


jarolrod (approved):
ACK 5d332da2cfdc67d69a165119c6ff598e29b28afa

while obviously it would still stand that we wouldn't want to delete the line right after this comment, that can be deduced by the usage of this variable in the next lines. What isn't relevant any longer is this comment in general but specifically referencing Application and QCoreApplication.

Scottmiller666 (commented):
I

PR #684, opened 2022-12-05
Improve 'Requested Payments History' Multiselect

john-moffett:
The "Requested payments history" list has somewhat broken functionality. You can only select contiguous rows. Sorting the list unexpectedly modifies any selection you made:

_Initially sorted by date and multiple items selected:_

image

_Now sort by label ->_ 

image

The context menu appears when multiple rows are selected despite the actions only affecting the first in the list:

image

These issues are all corrected.

First, a `QSortFilterProxyModel` is inserted to take care of sorting instead of sorting the model manually. (This is the same approach taken for all other sortable lists in the GUI.) This also preserves any selections the user had made prior to sorting. Second, the selection model is changed to `ExtendedSelection` to allow for non-contiguous multi-select. ~~Third, the context menu is disabled when multiple rows are selected, as none of the context menu options operate on multiple selected items.~~ Update: now the context menu operates on multiple rows. It will copy the data to the clipboard separated by newlines.

image

### DIFF
--- a/src/qt/receivecoinsdialog.cpp
+++ b/src/qt/receivecoinsdialog.cpp
@@ -2,6 +2,7 @@
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
+#include <reverse_iterator.h>
 #include <wallet/wallet.h>
 
 #include <qt/receivecoinsdialog.h>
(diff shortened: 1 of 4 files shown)


hebasto:
> First, a `QSortFilterProxyModel` is inserted to take care of sorting instead of sorting the model manually.

Concept ACK on it.

> Second, the selection model is changed to `ExtendedSelection` to allow for non-contiguous multi-select.

Why? There are no use cases for multi-select, aren't there?

> Third, the context menu is disabled when multiple rows are selected, as none of the context menu options operate on multiple selected items.

Not sure if such a behavior is a user-friendly one.

john-moffett:
Updated to now show the context menu with multiple selections. The data is copied to the clipboard separated by newlines.

Example:

image

Results in the following copied to the clipboard:

```
bitcoin:TB1QETGUVSJCM80365PQ2GFT8NASVYYPMCG8FEWN3W?message=Message%20with%20no%20label%20or%20amount
bitcoin:mt2nX2QuRkWkRd2ewbE1mTZ1zaKUeBDEZs?amount=10.00000000&label=Legacy%20Address&message=For%20More%20Testing
bitcoin:2N1Qu4pSpTbCud8ToT9NJWW2PiDm7V97TSq?label=P2SH%20Segwit%20Address&message=For%20testing
```

The individual context actions are enabled only if all selected items have data for that category. For example, if only two of the three selected items have a "label", then the "Copy labels" context menu action is disabled. The "Copy URIs" and "Copy addresses" actions will always be active.

luke-jr:
Would prefer the refactoring split into a different commit

luke-jr:
Propose these changes: https://github.com/john-moffett/bitcoin-gui/compare/2022_12_FixReceiveCoinsMultiselect...luke-jr:bitcoin-core-gui:qt_reqs_multiselect_pr684

pablomartin4btc (commented):
Concept ACK and tACK

image

I agree with the improvement/ fix of non-contiguous multi-select behavior, it's something that a user have in most apps. 

The grayed-out of the context menu items behavior is something that's already in `master` and it's not part of this PR, just to avoid any confusions.
> The individual context actions are enabled only if all selected items have data for that category. For example, if only two of the three selected items have a "label", then the "Copy labels" context menu action is disabled.

For this specific case, I'd add another improvement/ fix which would be a tool-tip indicating why the action is disabled:
image


Same for the following statement, it's currently in `master`, since those fields are not optional when a user creates a payment/ new address.
> The "Copy URIs" and "Copy addresses" actions will always be active.


I agree with @luke-jr in the split of the code into x commits, I took a quick look at his proposal. I like more the fact that the text of the action changes to its plural form when multi-select is performed as the author of the PR propose here.

I'd also add another feature, if possible, or maybe on a follow-up. that would be to add another context menu action to "Select All", and maybe just "Copy" with a tool-tip indicating that "All fields of the selected row will be copied". But as @hebasto mentioned above, perhaps there's no real use case for what I'm proposing here.

hebasto:
@john-moffett Are you still working on this PR?

hebasto:
Closing due to author's inactivity. Labeled "Up for grabs".

PR #687, opened 2022-12-18
Load PSBTs using istreambuf_iterator rather than istream_iterator

achow101:
`istream_iterator` eats whitespace charactesr which causes parsing failures for PSBTs that contain the bytes corresponding to those characters. `istreambuf_iterator` is the correct thing to use here.

This is a regression in 24.0. https://github.com/bitcoin/bitcoin/pull/25001 accidentally changed the original `istreambuf_iterator` to `istream_iterator`.

### DIFF
--- a/src/qt/walletframe.cpp
+++ b/src/qt/walletframe.cpp
@@ -212,7 +212,7 @@ void WalletFrame::gotoLoadPSBT(bool from_clipboard)
             return;
         }
         std::ifstream in{filename.toLocal8Bit().data(), std::ios::binary};
-        data.assign(std::istream_iterator<unsigned char>{in}, {});
+        data.assign(std::istreambuf_iterator<char>{in}, {});
 
         // Some psbt files may be base64 strings in the file rather than binary data
         std::string b64_str{data.begin(), data.end()};


furszy (approved):
Tested ACK bb5ea1d9

Not needed but wouldn't hurt to decouple the file read functionality into a util function, so this bug can be covered in a unit test.

src/qt/walletframe.cpp
```diff
@@ -212,7 +212,7 @@ void WalletFrame::gotoLoadPSBT(bool from_clipboard)
             return;
         }
         std::ifstream in{filename.toLocal8Bit().data(), std::ios::binary};
-        data.assign(std::istream_iterator<unsigned char>{in}, {});
+        data.assign(std::istreambuf_iterator<char>{in}, {});
```
hebasto, line 215: This change reverts https://github.com/bitcoin/bitcoin/commit/a65931e3ce66d87b8f83d67ecdbb46f137e6a670 partially, in particular:
- `std::istream_iterator` --> `std::istreambuf_iterato`
- `unsigned char` --> `char`

But the local `std::vector data` is still parameterized by `unsigned char`: https://github.com/bitcoin-core/gui/blob/bb5ea1d9a954b7b9f443ee8fbbb04549cd0b08a7/src/qt/walletframe.cpp#L195 

Is it on purpose?

hebasto (commented):
> This is a regression in 24.0.

It would be nice to add a test to prevent such a regression in the future.

@sipa @MarcoFalke 

Perhaps you could be interested in reviewing of this PR as an author/a committer of https://github.com/bitcoin/bitcoin/commit/a65931e3ce66d87b8f83d67ecdbb46f137e6a670 in https://github.com/bitcoin/bitcoin/pull/25001.

src/qt/walletframe.cpp
achow101: Yes, that was intentional. Using `unsigned char` does not appear to work (fails to compile). But I don't think that data type makes a difference here.

MarcoFalke: Yeah, `std::ifstream` is `std::basic_ifstream<char>`, and passing that to `std::istreambuf_iterator` can only be done if it is templated `<char>` as well. C++ (17?) derives this automatically, so you can also write:


```cpp
data.assign(std::istreambuf_iterator{in}, {});

**Review by MarcoFalke — APPROVED** [review]
Docs: https://en.cppreference.com/w/cpp/iterator/istream_iterator#Notes

review ACK bb5ea1d9a954b7b9f443ee8fbbb04549cd0b08a7   🍇

<details><summary>Show signature</summary>

Signature:

```
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA512

review ACK bb5ea1d9a954b7b9f443ee8fbbb04549cd0b08a7   🍇
-----BEGIN PGP SIGNATURE-----

iQGzBAEBCgAdFiEE+rVPoUahrI9sLGYTzit1aX5ppUgFAlwqrYAACgkQzit1aX5p
pUg34wv+NFe6qKxea+Z5Jx0PTNpbWXv+nTGMQYsOek6a+pJIe00MuJGO4XIuDJS0
Mffqqi8fhC83It02w8Z6q9zLSXB094GabCYgFanN5g0AxDdwgvjrEM5nQ9cZj3Vo
hJHDzVpdgag5i311J+05yeHig9EOnRjxBRgf1p3pv+/Tg3yS/M7UIRKdpmqAxfOA
HdPMjRQr+Tjm7+/VHwJJpEnE0fPLUZX5NjblKUmG4wDIwlcqz0qYUK8JkAtznYz9
JJvXxOayoGKNk3IF/UfZs0/cmGlPh+5AJ3asnSIizAITiwIOFaNAIUfprgc4d86V
fo66d2avigfwHVpKsLsbQep52I0deYo1dItSGvhcYxhtunB/D8eBsMi8b5kk489j
kV6i1kP/VtUVu2v2Pkgl632aupAcWM62gY3YCl50IfqNfpMgf0uG58NEScZFHmbp
920c3zvXD9n7SDx3GVQPljssLPBG7WMUfBYmwYP0NtWx28qv250dThWYnOUUoour
1GDZDw4r
=QuKV
-----END PGP SIGNATURE-----
```


</details>

fanquake:
Removing label as this was backported: https://github.com/bitcoin/bitcoin/commit/0662105e884dce3eae9d3eb0b0d49600c1260b90.

Issue #55, opened 2020-08-14
Add generate method


fanquake:
Moved from https://github.com/bitcoin/bitcoin/issues/16000.

The `generate` method has been removed from Bitcoin Core, but it would be nice if the GUI had a `generate` method in the debug console. That method would effectively expand to `generatetoaddress(getnewaddress())` (just like the `def generate` in our test node python class)

Sjors:
Concept ACK.

hernanmarino:
Hi, I made a PR for this functionality reviewers are welcome to PR https://github.com/bitcoin-core/gui/pull/692

PR #692, opened 2022-12-28
GUI: Debug Console implementation of generate method

hernanmarino:
This is an implementation of a (gui console only) generate method, as described in issue https://github.com/bitcoin-core/gui/issues/55 .
It mimics the behaviour and return data from `bitcoin-cli`'s  `-generate` argument as implemented in https://github.com/bitcoin/bitcoin/pull/19133


pablomartin4btc (commented):
Code Review ACK. It's good to make the `-generate` command available at the rpc console level in `bitcoin-qt` as it's already in `bitcoin-cli`.

src/qt/rpcconsole.cpp
```diff
@@ -414,7 +415,33 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
     {
         std::string result;
         std::string executableCommand = command.toStdString() + "\n";
-
+        std::regex validRegex("generate\\s*[\\(]?\\s*(\\d+)?\\s*[,]?\\s*(\\d+)?\\s*[\\)]?\\s*");
+        std::regex invalidRegex("generate[\\s\\(]+.*");
+        std::smatch match;
+        if (std::regex_match(executableCommand, match, validRegex)) {
+            const std::string nblocks{match.str(1) == "" ? "1" : match.str(1)};
+            const std::string maxtries{match.str(2) == "" ? "1000000" : match.str(2)};
```
pablomartin4btc, line 423: Sorry, just realised, perhaps you don't need to set `maxtries `variable with the default value as in `mining.h`, `DEFAULT_MAX_TRIES`, it's already set if maxtries is empty in the `generatetoaddress` command at `mining.cpp`, so:

```suggestion
            const std::string maxtries = match.str[2];
```

pablomartin4btc (commented):
b7782a4177977d62a2e2860c3a99b919b8e5b0a2 ACK.

RandyMcMillan:
ConceptACK

Additional logic to detect whether `-generate` is or is not followed by an integer would be great.

generate-gui-console

hernanmarino:
Hi @RandyMcMillan . Thank you for taking the time to look a my PR. The changes I implemented are actually on `bitcoin-qt` and not on `bitcoin-cli` . I updated the PR description to make it clear. I believe the issues you describe are not present here, but you are welcome to test them and let me know if you find anything.

src/qt/rpcconsole.cpp
hernanmarino: Thanks, I'll test it and perhaps update the code in the next commit.

src/qt/rpcconsole.cpp
```diff
@@ -414,7 +415,32 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
     {
         std::string result;
         std::string executableCommand = command.toStdString() + "\n";
-
+        std::regex validRegex("generate\\s*[\\(]?\\s*(\\d+)?\\s*[,]?\\s*(\\d+)?\\s*[\\)]?\\s*");
```
andrewtoth, line 418: This is pretty complicated regex. Would it be easier to split `executableCommand` into tokens separated by whitespace?

src/qt/rpcconsole.cpp
```diff
@@ -414,7 +415,32 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
...
+        std::regex invalidRegex("generate[\\s\\(]+.*");
+        std::smatch match;
+        if (std::regex_match(executableCommand, match, validRegex)) {
+            const std::string nblocks{match.str(1) == "" ? "1" : match.str(1)};
+            const std::string maxtries = match.str(2);
```
andrewtoth, line 423:
```suggestion
            const std::string maxtries{match.str(2)};
```

src/qt/rpcconsole.cpp
```diff
@@ -414,7 +415,32 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
...
+            if (!RPCConsole::RPCExecuteCommandLine(m_node, result, "getnewaddress\n", nullptr, wallet_model)) {
+                Q_EMIT reply(RPCConsole::CMD_ERROR, QString("Parse error: unbalanced ' or \""));
+            }
+            else {
+                std::string address = result;
```
andrewtoth, line 428:
```suggestion
                const std::string address{result};
```

src/qt/rpcconsole.cpp
andrewtoth: See https://github.com/bitcoin/bitcoin/blob/master/src/bitcoin-cli.cpp#L679-L684.

src/qt/rpcconsole.cpp
```diff
@@ -414,7 +415,32 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
...
+                if (!RPCConsole::RPCExecuteCommandLine(m_node, result, "generatetoaddress " + nblocks + " " + address + " " + maxtries + "\n", nullptr, wallet_model)) {
+                    Q_EMIT reply(RPCConsole::CMD_ERROR, QString("Parse error: unbalanced ' or \""));
+                } else {
+                    std::string answer = "{\n  \"address\": \"" + address + "\",\n  \"blocks\": "+ result + "\n}";
```
andrewtoth, line 432: Should we use `UniValue` to parse the result and response? See https://github.com/bitcoin/bitcoin/blob/master/src/bitcoin-cli.cpp#L686-L692

pablomartin4btc (commented):
Tested ACK 9b0b762fd774207d127213c1c9da354aef38e739.
Checked output with different params (amount of blocks, tries), compared with outputs from `bitcoin-cli` as well, all good!

furszy (commented):
Concept ACK

Could re-write, without the regex stuff, a bit simpler:

```c++
std::vector<std::string> split_command = SplitString(executableCommand, " ");
if (!split_command.empty() && split_command[0] == "generate") {
    // Remove last "\n"
    std::string last_param = split_command[split_command.size()-1];
    split_command[split_command.size()-1] = last_param.substr(0, last_param.size() - 1);

    // Generate address
    std::string address_result;
    if (!RPCConsole::RPCExecuteCommandLine(m_node, address_result, "getnewaddress\n", nullptr, wallet_model)) {
        Q_EMIT reply(RPCConsole::CMD_ERROR, QString("Error: cannot get new address"));
         return;
    }

    // Craft command
    const std::string blocks_num = split_command.size() > 1 ? split_command[1] : "1";
    const std::string max_tries = split_command.size() > 2 ? split_command[2] : "";
    executableCommand = "generatetoaddress " + blocks_num + " \"" + address_result + "\" " + max_tries;
}
```

hernanmarino:
Thanks for all the insights. In the first days of the next week I'll try to incorporate the feedback and squash the commits.

hebasto:
Concept ACK.

@hernanmarino You could also [link](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) the initial issue in the PR description.

rserranon:
> Concept ACK
> 
> Could re-write it, without the regex stuff, a bit simpler:
> 
> ```c++
> std::vector<std::string> split_command = SplitString(executableCommand, " ");
> if (!split_command.empty() && split_command[0] == "generate") {
>     // Remove last "\n"
>     std::string last_param = split_command[split_command.size()-1];
>     split_command[split_command.size()-1] = last_param.substr(0, last_param.size() - 1);
> 
>     // Generate address
>     std::string address_result;
>     if (!RPCConsole::RPCExecuteCommandLine(m_node, address_result, "getnewaddress\n", nullptr, wallet_model)) {
>         Q_EMIT reply(RPCConsole::CMD_ERROR, QString("Error: cannot get new address"));
>          return;
>     }
> 
>     // Craft command
>     const std::string blocks_num = split_command.size() > 1 ? split_command[1] : "1";
>     const std::string max_tries = split_command.size() > 2 ? split_command[2] : "";
>     executableCommand = "generatetoaddress " + blocks_num + " \"" + address_result + "\" " + max_tries;
> }
> ```

Does it make sense to include parsing logic, generate a new address, and then craft the command to call the `generatetoaddress`, inside the `RPCExecutor::request` function, or it should be done on a separate function?

furszy:
> Does it make sense to include parsing logic, generate a new address, and then craft the command to call the generatetoaddress, inside the RPCExecutor::request function, or it should be done on a separate function?

A separate function returning the command string would be better.

RandyMcMillan:
rebased on top of a62231bca629e945349255a1d331dd5c7a86ddd1

tACK 9b0b762fd774207d127213c1c9da354aef38e739

macOS: Arm64

Screen Shot 2023-01-24 at 7 12 54 PM

RandyMcMillan:
tested on top of https://github.com/bitcoin-core/gui/commit/a62231bca629e945349255a1d331dd5c7a86ddd1

tACK https://github.com/bitcoin-core/gui/commit/9b0b762fd774207d127213c1c9da354aef38e739

macOS: x86_64

Screen Shot 2023-01-24 at 7 31 04 PM

RandyMcMillan:
@hernanmarino - @rserranon 's suggestion is compelling... 

[https://github.com/RandyMcMillan/gui/commit/612a0c14173d681a04021581851422ba832236ee
](https://github.com/RandyMcMillan/gui/commit/612a0c14173d681a04021581851422ba832236ee
)

Applied as a patch on top of this PR:
https://github.com/RandyMcMillan/gui/blob/612a0c14173d681a04021581851422ba832236ee/rserranon.patch

hernanmarino:
Thanks for the feedback and testing. I ' m already working on an update consider this and all feedback received, will update soon.

> @hernanmarino - @rserranon 's suggestion is compelling...
> 
> [RandyMcMillan/gui@`612a0c1`](https://github.com/RandyMcMillan/gui/tree/612a0c14173d681a04021581851422ba832236ee)
> 
> Applied as a patch on top of this PR: [RandyMcMillan/gui@`612a0c1`/rserranon.patch](https://github.com/RandyMcMillan/gui/blob/612a0c14173d681a04021581851422ba832236ee/rserranon.patch)
> 
> I haven't fully tested the patch but works on MacOS x86_64

src/qt/rpcconsole.cpp
hernanmarino: Hi @andrewtoth . I don't think it's really necesary , `bitcoin-qt `seems to take a different approach than `bitcoin-cli`, using strings and QStrings for communicating results. And although the RPC logic for actual commands use `Univalue` , these are commands intercepted and executed before that logic. Since the inputs (both address and block hashes) and the output needed are already both strings , using an `Univalue` as an intermediate object that is not used elsewhere, seems unnecesary. Unless you can observe something else that I'm missing, I prefer to leave this as a regular string.

hernanmarino:
Summary of changes just pushed:

- Changed the regex for a simpler string splitter, while still mantaining the syntax and several argument separators (beyond whitespace) supported by the console. Code now runs 180X faster (thanks @furszy and @andrewtoth for the suggestion)
- Moved the implementation to a separate method that deals with these console-only commands (@rserranon)Also moved the already implemented "help-console" command handling to this method.
- Added "help generate" code 
- Cleaned some code formatting issues according to clang-format ( thanks @jarolrod  for the last two :)  )
- Squashed all changes on a single commit

pablomartin4btc (commented):
re ack 3f8fe6fd698380d659e5c88b0795f8d6ea3a2be0, cleaner than before.

pablomartin4btc (approved):

furszy (commented):
Noticed that we are calling cli-only commands with the "-" prefix. Would be good to follow the same convention here too.
The renaming will let us detect them in a simpler manner by checking the first character without requiring to split the string etc.


Aside from that, now talking about the current changes:
What do you think about structuring the changes a bit differently so your patch set is focused to the introduced feature, and we can continue expanding the cli-only options in the future:

```c++
if (IsCliOnlyCommand(executableCommand)) { // --> here we check if string starts with "-". 
    // Execute cli-only commands
    if (!ExecuteCliCommand(executableCommand, result, etc)) { // --> This function will dispatch cli-only commands ("-generate" should be placed into its own standalone function).
         // emit any error here or inside the function
         return;
    }
} else {
   // Execute RPC commands
   if (!RPCConsole::RPCExecuteCommandLine(m_node, result, executableCommand, nullptr, wallet_model)) {
        // etc...
}
```

rserranon:
> ```c++
> if (IsCliOnlyCommand(executableCommand)) { // --> here we check if string starts with "-". 
>     // Execute cli-only commands
>     if (!ExecuteCliCommand(executableCommand, result, etc)) { // --> This function will dispatch cli-only commands ("-generate" should be placed into its own standalone function).
>          // emit any error here or inside the function
>          return;
>     }
> } else {
>    // Execute RPC commands
>    if (!RPCConsole::RPCExecuteCommandLine(m_node, result, executableCommand, nullptr, wallet_model)) {
>         // etc...
> }
> ```

I agree with furszy's idea of introducing the functions `IsCliOnlyCommand ` and `ExecuteCliCommand `, to implement the `generate` cli-only command, for a cleaner implementation and with the advantage of leaving the structure ready for potential expansion of additional cli-only commands.

pablomartin4btc (commented):
I wouldn't change the naming convention of the `generate` command in the "RPC Console" to `-generate` in order to match cli-commands. If we start to implement such commands into here maybe I'd think about it but this is the "RPC Console" it has a different purpose even the final objective is to communicate with the node.

Also, as soon we do this, code-wise, we would need to remove the `RPCHelpMan` in [`src/rpc/mining.cpp`](https://github.com/bitcoin/bitcoin/blob/2c1fe27bf3c1e504864987cbe80f9c24897d3cb1/src/rpc/mining.cpp#L237-L242):

```
static RPCHelpMan generate()
{
    return RPCHelpMan{"generate", "has been replaced by the -generate cli option. Refer to -help for more information.", {}, {}, RPCExamples{""}, [&](const RPCHelpMan& self, const JSONRPCRequest& request) -> UniValue {
        throw JSONRPCError(RPC_METHOD_NOT_FOUND, self.ToString());
    }};
}
```

Otherwise we will get this message error:

image

And if we do that, any other app trying to send the 'generate` command via RPC would receive the `method not found` error:

image

I believe the code is good as it is so far and as a first and safe approach to resolve the issue.

hernanmarino:
> Noticed that we are calling cli-only commands with the "-" prefix. Would be good to follow the same convention here too. The renaming will let us detect them in a simpler manner by checking the first character without requiring to split the string etc.

This seems like something "nice to have" but there are other issues apart from what Pablo mentioned that makes me think that we shouldn´t do that: 

- There are other console-only commands that already existed, like  `help-console`
- This PR not only introduces the new behaviour for `generate` but also `help generate` as @jarolrod  suggested
- Future console-only commands that might be introduced later will also require their own `help new_command`

None of this commands could be easily renamed to start with "-" without breaking "protocol"  and backwards compatibility, and if we don't do that, I believe the approach implemented in this PR is good enough. If in any case, if there's further interest to discuss alternatives, perhaps we can take that discussion to a follow-up PR.


> Aside from that, now talking about the current changes: What do you think about structuring the changes a bit differently so your patch set is focused to the introduced feature, and we can continue expanding the cli-only options in the future:
> 
> ```c++
> if (IsCliOnlyCommand(executableCommand)) { // --> here we check if string starts with "-". 
>     // Execute cli-only commands
>     if (!ExecuteCliCommand(executableCommand, result, etc)) { // --> This function will dispatch cli-only commands ("-generate" should be placed into its own standalone function).
>          // emit any error here or inside the function
>          return;
>     }
> } else {
>    // Execute RPC commands
>    if (!RPCConsole::RPCExecuteCommandLine(m_node, result, executableCommand, nullptr, wallet_model)) {
>         // etc...
> }
> ```

If we abandon the idea of renaming commands to start with "-" the suggested code is very similar to this PR's code, only that the function `ExecuteCliCommand` is called `executeConsoleOnlyCommand`.

hernanmarino:
> I agree with furszy's idea of introducing the functions `IsCliOnlyCommand ` and `ExecuteCliCommand `, to implement the `generate` cli-only command, for a cleaner implementation and with the advantage of leaving the structure ready for potential expansion of additional cli-only commands.

The provided method `executeConsoleOnlyCommand` is almost identical to the suggested `ExecuteCliCommand` and is also ready for potential future new console only commands, the only difference is that it detects ( like `IsCliOnlyCommand` ) and executes commands in a single method.

rserranon:
testing ACK

src/qt/rpcconsole.cpp
```diff
@@ -464,6 +446,101 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
...
+ *
+ * @param[in]    strCommand  Command line to parse
+ *
+ * @return a vector of strings with command and parameters
+ */
+
+std::vector<std::string> RPCExecutor::parseHelper(const std::string& strCommand)
+{
+    std::vector<std::string> vec = SplitString(strCommand, " (),\n");
+    auto should_remove = [](const std::string& str) { return str.empty(); };
```
LarryRuane, line 461:
```suggestion
    std::vector<std::string> vec{SplitString(strCommand, " (),\n")};
    auto should_remove{[](const std::string& str) { return str.empty(); }};
```

src/qt/rpcconsole.cpp
```diff
@@ -464,6 +446,101 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
...
+
+    // Catch the console-only generate command with 2 or less parameters before RPC call is executed .
+    if (!parsedCommand.empty() && parsedCommand[0] == "generate" && parsedCommand.size() <= 3) {
+        // Default number of blocks is 1 if missing
+        const std::string nblocks{parsedCommand.size() > 1 ? parsedCommand[1] : "1"};
+        // An empty maxtries parameters will be modified by generatetoaddress' default value
+        const std::string maxtries{parsedCommand.size() > 2 ? parsedCommand[2] : ""};
+
+        // Generate address
+        if (!RPCConsole::RPCExecuteCommandLine(m_node, address, "getnewaddress\n", nullptr, wallet_model)) {
```
LarryRuane, line 490:
```suggestion
        if (!RPCConsole::RPCExecuteCommandLine(m_node, address, "getnewaddress\n", /*pstrFilteredOut=*/nullptr, wallet_model)) {
```

src/qt/rpcconsole.cpp
```diff
@@ -464,6 +446,101 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
...
+            Q_EMIT reply(RPCConsole::CMD_ERROR, QString("Error: could not generate new address"));
+        } else {
+            executableCommand = "generatetoaddress " + nblocks + " \"" + address + "\" " + maxtries;
+            if (!RPCConsole::RPCExecuteCommandLine(m_node, result, "generatetoaddress " + nblocks + " " + address + " " + maxtries + "\n", nullptr, wallet_model)) {
```
LarryRuane, line 494:
```suggestion
            auto executableCommand = "generatetoaddress " + nblocks + " " + address + " " + maxtries;
            if (!RPCConsole::RPCExecuteCommandLine(m_node, result, executableCommand + "\n", nullptr, wallet_model)) {
```
Also remove the declaration of `executableCommand` near the top of the function; better to declare variables only where and when needed. You could do similarly with `result` and `address`.

src/qt/rpcconsole.cpp
```diff
@@ -464,6 +446,101 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
...
+                    Q_EMIT reply(RPCConsole::CMD_ERROR, QString("Error: could not generate blocks"));
```
LarryRuane, line 495: This line is indented one tab stop too much, and same for the two lines just below (after the `else`).

src/qt/rpcconsole.cpp
```diff
@@ -464,6 +446,101 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
...
+    }
+    // Catch the console-only "help generate" command when requested, or when "generate is called with wrong parameters.
+    if ((parsedCommand.size() >= 2 && parsedCommand[0] == "help" && parsedCommand[1] == "generate") || (parsedCommand.size() > 3 && parsedCommand[0] == "generate")) {
+        Q_EMIT reply(RPCConsole::CMD_REPLY, QString(("\n"
+                                                     "Generate blocks, equivalent to RPC getnewaddress followed by RPC generatetoaddress.\n"
+                                                     "Optional integer arguments are number of blocks to generate and maximum iterations to try.\n"
+                                                     "Equivalent to RPC generatetoaddress nblocks and maxtries arguments.\n"
+                                                     "   example:    generate\n"
+                                                     "   example:    generate 4\n"
+                                                     "   example:    generate 3 6000\n\n")));
```
LarryRuane, line 511: Formatting-only suggestion (personal preference to avoid long lines), and the argument to `QString()` had extra parentheses. Could make a similar change below.
```suggestion
        if ((parsedCommand.size() > 1 && parsedCommand[0] == "help" && parsedCommand[1] == "generate") ||
            (parsedCommand.size() > 3 && parsedCommand[0] == "generate")) {
                auto message{
                    "\n"
                    "Generate blocks, equivalent to RPC getnewaddress followed by RPC generatetoaddress.\n"
                    "Optional integer arguments are number of blocks to generate and maximum iterations to try.\n"
                    "Equivalent to RPC generatetoaddress nblocks and maxtries arguments.\n"
                    "   example:    generate\n"
                    "   example:    generate 4\n"
                    "   example:    generate 3 6000\n\n"};
                Q_EMIT reply(RPCConsole::CMD_REPLY, QString(message));
```

LarryRuane (commented):
Concept, approach, and tested ACK, this is a nice improvement. It might be good to prepend "test:" to the title of this PR (and this might pick up the "test" label, which would be nice). I suggest making that change to the commit title too ("test: qt: add generate ..."). If you'd like to make some or all of these suggested changes, I'll re-review.

furszy:
All good about the none usage of the "-" prefix on the GUI 👍🏼 .

 > The provided method executeConsoleOnlyCommand is almost identical to the suggested ExecuteCliCommand and is also ready for potential future new console only commands, the only difference is that it detects ( like IsCliOnlyCommand ) and executes commands in a single method.

I'm not fan of the current method mostly because of all the `!parsedCommand.empty()` and `parsedCommand.size() > something` and `parsedCommand.size() < something_else`.
Those are specific checks that should be done inside each command function, and not at the top level dispatcher function.

A slightly modified version of the code that shared before so you can see where I am going:
```c++
std::string cli_command;
if (IsCliOnlyCommand(executableCommand, cli_command)) {  // Check against a list of known cli-only commands and return the command inside "cli_command".
    // Craft Cli Command
    std::string res_cli_only_command;
    if (!CraftCliCommand(cli_command, executableCommand, res_cli_only_command)) {
         return; // errors could be emitted inside the function or here if it returns an error.
    }
    // Update command
    executableCommand = res_cli_only_command;
}

// Execute command
if (!RPCConsole::RPCExecuteCommandLine(m_node, result, executableCommand, nullptr, wallet_model)) {
    // etc...
```

Then you could do something like:

```c++
bool CraftCliCommand(const std::string& cli_command, const std::string& executableCommand, std::string& command_ret)
{
    switch(cli_command) {
        case GENERATE_COMMAND:
            command_ret  = craftGenerateCommand(executableCommand);
            return true;
        case OTHER_CLI_COMMAND:
            command_ret  = craftOtherCliCommand(executableCommand);
            return true;
        // etc..
    }

    // command not implemented.
    return false;
}

```

In this way, each command is contained within a their own function and new checks/changes can be added or removed without requiring to modify the top level function.

src/qt/rpcconsole.cpp
hernanmarino: I ' ll actually remove the variable `executableCommand` since it's only used once, I prefer avoiding memory overhead and better speed execution (althoug minimal). Okey with the other 2 variables

src/qt/rpcconsole.cpp
hernanmarino: Indentation was introduced after running the script suggested by @ jarolrod and are apparently ok according to clang format standard.

src/qt/rpcconsole.cpp
hernanmarino: Removed extra parenthesis, and I agree with extra long lines, but as I said before, indentation was introduced after running the script suggested by @ jarolrod and are apparently ok according to clang format standard.

src/qt/rpcconsole.cpp
```diff
@@ -415,35 +417,15 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
...
-
-                "Results without keys can be queried with an integer in brackets using the parenthesized syntax.\n"
-                "   example:    getblock(getblockhash(0),1)[tx][0]\n\n")));
-            return;
-        }
-        if (!RPCConsole::RPCExecuteCommandLine(m_node, result, executableCommand, nullptr, wallet_model)) {
-            Q_EMIT reply(RPCConsole::CMD_ERROR, QString("Parse error: unbalanced ' or \""));
-            return;
+        // Attempt to execute console-only commands
+        if (!RPCExecutor::executeConsoleOnlyCommand(executableCommand, wallet_model)) {
```
LarryRuane, line 421:
```suggestion
        if (RPCExecutor::executeConsoleOnlyCommand(executableCommand, wallet_model)) {
            return;
        }
```
(or could put `return` on the same line)

src/qt/rpcconsole.cpp
```diff
@@ -464,6 +446,101 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
...
+ * @param[in]    command  Command line to execute
+ * @param[in]    wallet_model  Wallet model to use
+ * @return  true if command was handled by this method (even on errors), false otherwise
+ *
+ */
+
+bool RPCExecutor::executeConsoleOnlyCommand(const std::string& command, const WalletModel* wallet_model)
+{
+    std::string result;
+    std::string address;
```
LarryRuane, line 478: these can be removed

src/qt/rpcconsole.cpp
```diff
@@ -464,6 +446,101 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
...
+    std::vector<std::string> parsedCommand{parseHelper(command)};
```
LarryRuane, line 479:
```suggestion
    const std::vector<std::string> parsedCommand{parseHelper(command)};
```

src/qt/rpcconsole.cpp
LarryRuane: That's strange, when I run `clang-format-diff`, it moves this left, where they should be:

<details>
<summary>click to expand</summary>

```
$ git show -U0 | ./contrib/devtools/clang-format-diff.py -p1 -i -v
Formatting src/qt/rpcconsole.cpp
$ git diff
index 1408b289f..63820464b 100644
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -419,12 +419,12 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
 
         // Attempt to execute console-only commands
         if (!RPCExecutor::executeConsoleOnlyCommand(executableCommand, wallet_model)) {
-                // Send to the RPC command parser if not console-only
-                if (!RPCConsole::RPCExecuteCommandLine(m_node, result, executableCommand, nullptr, wallet_model)) {
-                    Q_EMIT reply(RPCConsole::CMD_ERROR, QString("Parse error: unbalanced ' or \""));
-                    return;
-                }
-                Q_EMIT reply(RPCConsole::CMD_REPLY, QString::fromStdString(result));
+            // Send to the RPC command parser if not console-only
+            if (!RPCConsole::RPCExecuteCommandLine(m_node, result, executableCommand, nullptr, wallet_model)) {
+                Q_EMIT reply(RPCConsole::CMD_ERROR, QString("Parse error: unbalanced ' or \""));
+                return;
+            }
+            Q_EMIT reply(RPCConsole::CMD_REPLY, QString::fromStdString(result));
         }
     }
     catch (UniValue& objError)
@@ -492,10 +492,10 @@ bool RPCExecutor::executeConsoleOnlyCommand(const std::string& command, const Wa
             Q_EMIT reply(RPCConsole::CMD_ERROR, QString("Error: could not generate new address"));
         } else {
             if (!RPCConsole::RPCExecuteCommandLine(m_node, result, "generatetoaddress " + nblocks + " " + address + " " + maxtries + "\n", /*pstrFilteredOut=*/nullptr, wallet_model)) {
-                    Q_EMIT reply(RPCConsole::CMD_ERROR, QString("Error: could not generate blocks"));
+                Q_EMIT reply(RPCConsole::CMD_ERROR, QString("Error: could not generate blocks"));
             } else {
-                    std::string answer = "{\n  \"address\": \"" + address + "\",\n  \"blocks\": " + result + "\n}";
-                    Q_EMIT reply(RPCConsole::CMD_REPLY, QString::fromStdString("\n" + answer + "\n\n"));
+                std::string answer = "{\n  \"address\": \"" + address + "\",\n  \"blocks\": " + result + "\n}";
+                Q_EMIT reply(RPCConsole::CMD_REPLY, QString::fromStdString("\n" + answer + "\n\n"));
             }
         }
         return true;
```
</details>

LarryRuane (commented):
tested 037e8b01386d5f30f295eec53eb939204b855cb4

src/qt/rpcconsole.cpp
pablomartin4btc: I'll try from my side and give it a go later this week.

hebasto (commented):
Could this PR be separated in two commits:
- refactoring and introducing the `RPCExecutor::executeConsoleOnlyCommand()` function
- actually adding support for the `generate` command`
?

hernanmarino:
> Could this PR be separated in two commits:
> 
> * refactoring and introducing the `RPCExecutor::executeConsoleOnlyCommand()` function
> * actually adding support for the `generate` command`
> 
> ?

Yes, will do ASAP.

hernanmarino:
As requested by @hebasto i splitted this PR in 2 commits : 
ad5642ae91beb522b6ae806f28cb015a759d1d19 refactors the code to prepare for the new functionality implemented in the following commit, as well as for future console-only commands to be added.

411b1da407f78f2f973f90d48676ce1ae26734a7  adds the new "generate" command, as well as the updated "help generate"

hebasto:
The `help generate` command returns:
```
generate

has been replaced by the -generate cli option. Refer to -help for more information.
```

pablomartin4btc (approved):
Liked the refactoring and the 2 commits split as @hebasto requested.
Tested ACK both commits separately, also checked also other RPC commands, it works as expected.

src/qt/rpcconsole.cpp
```diff
@@ -464,6 +454,96 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
...
+ *
+ */
+bool RPCExecutor::executeConsoleOnlyCommand(const std::string& command, const WalletModel* wallet_model)
+{
+    // Parse command line into a vector of strings
+    const std::vector<std::string> parsed_command{parseHelper(command)};
+    if (parsed_command.empty()) {
+        return false;
+    } else {
+        // Iterate over m_method_map and execute the associated method if the beggining of parsed_command matches the key
```
furszy, line 603: In ad5642ae:

I think that would be cleaner to change the if/else with a:

```c++
const std::vector<std::string> parsed_command{parseHelper(command)};
if (parsed_command.empty()) return false;

// Iterate over m_method_map and execute the associated method if the beggining of parsed_command matches the key
etc..
```

src/qt/rpcconsole.cpp
```diff
@@ -487,6 +491,69 @@ bool RPCExecutor::commandMatches(const std::vector<std::string>& parsed_command,
...
+/**
+ * @brief  Executes the console-only command "help generate".
+ * @param parsed_command  A vector of strings with command and parameters, usually generated by RPCExecutor::parseHelper
+ * @param wallet_model  WalletModel to use for the command
+ * @return  True if the command was executed, false otherwise.
+ */
+bool RPCExecutor::executeConsoleHelpGenerate(const std::vector<std::string>& parsed_command, const WalletModel* wallet_model)
+{
+    // Catch the console-only "help generate" command when requested, or when "generate is called with wrong parameters.
+    if ((parsed_command.size() >= 2 && parsed_command[0] == "help" && parsed_command[1] == "generate") || (parsed_command.size() > 3 && parsed_command[0] == "generate")) {
```
furszy, line 542: In 411b1da4:

Similar to the one above, The `parsed_command[0] == "help" && parsed_command[1] == "generate"` statement is not needed. The code wouldn't reach this point if the command wouldn't be a `help generate` command.

furszy (commented):
Looking much better, thanks for applying the suggestions 👌🏼.

Left few comments, it's getting closer.

src/qt/rpcconsole.cpp
```diff
@@ -464,6 +454,96 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
...
+
+/**
+ * Catches console-only command before a RPC call is executed
+ *
+ * @param[in]    command  Command line to execute
+ * @param[in]    wallet_model  Wallet model to use
+ * @return  true if command was handled by this method (even on errors), false otherwise
+ *
+ */
+bool RPCExecutor::executeConsoleOnlyCommand(const std::string& command, const WalletModel* wallet_model)
```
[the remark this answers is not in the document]
pablomartin4btc: I agree.

src/qt/rpcconsole.cpp
```diff
@@ -487,6 +491,69 @@ bool RPCExecutor::commandMatches(const std::vector<std::string>& parsed_command,
...
+/**
+ * @brief Executes the console-only command "generate".
+ * @param parsed_command  A vector of strings with command and parameters, usually generated by RPCExecutor::parseHelper
+ * @param wallet_model  WalletModel to use for the command
+ * @return True if the command was executed, false otherwise.
+ */
+bool RPCExecutor::executeConsoleGenerate(const std::vector<std::string>& parsed_command, const WalletModel* wallet_model)
+{
+    // Catch the console-only generate command with 2 or less parameters before RPC call is executed .
+    if (!parsed_command.empty() && parsed_command[0] == "generate" && parsed_command.size() <= 3) {
```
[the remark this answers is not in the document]
pablomartin4btc: I agree, I thought before that could be a pseudo-implementation of a handler pattern but the evaluation is being done checking the key of the map is being the specific command in the command-line.

src/qt/rpcconsole.cpp
pablomartin4btc: It would if -> Default to "help generate" when wrong number of parameters are given (line 524), but it's the same case anyway, it's the condition after the`or`operator,

pablomartin4btc (commented):
I think you'd need to check some comments.

hernanmarino:
> @hernanmarino Are you still working on this PR?

Yes, will update ASAP.

src/qt/rpcconsole.cpp
```diff
@@ -464,6 +454,96 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
...
+ * Splits a command line string into a vector with command at position 0, and parameters after
+ *
+ * @param[in]    strCommand  Command line to parse
+ *
+ * @return a vector of strings with command and parameters
+ */
+std::vector<std::string> RPCExecutor::parseHelper(const std::string& strCommand)
+{
+    // Split while recognizing the several characters that can be used as separators in the GUI console
+    std::vector<std::string> vec{SplitString(strCommand, " (),\n")};
```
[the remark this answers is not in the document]
hernanmarino: Done

src/qt/rpcconsole.cpp
```diff
@@ -96,6 +96,16 @@ public Q_SLOTS:
 
 private:
     interfaces::Node& m_node;
+    static std::vector<std::string> parseHelper(const std::string& strCommand);
+    static bool commandMatches(const std::vector<std::string>& parsed_command, const std::string& key);
```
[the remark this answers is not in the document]
hernanmarino: Done. Since the compilar doesn´t actually allow what you suggest, i removed the declarations from the class and implemented this 2 method as independent functions.

src/qt/rpcconsole.cpp
```diff
@@ -464,6 +454,96 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
...
+/**
+ * @brief  Executes the console-only command "help-console".
+ * @param  parsed_command  A vector of strings with command and parameters, usually generated by RPCExecutor::parseHelper
+ * @param wallet_model  WalletModel to use for the command
+ * @return  True if the command was executed, false otherwise.
+ */
+bool RPCExecutor::executeConsoleHelpConsole(const std::vector<std::string>& parsed_command, const WalletModel* wallet_model)
+{
+    // Catch the console-only-help command before RPC call is executed and reply with help text as-if a RPC reply.
+    if (!parsed_command.empty() && parsed_command[0] == "help-console") {
```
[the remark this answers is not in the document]
hernanmarino: Modified

src/qt/rpcconsole.cpp
hernanmarino: Agree ! Changed

src/qt/rpcconsole.cpp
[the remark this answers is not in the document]
hernanmarino: I usually agree with you @furszy , but in this case I'm  just passing around a pointer i didn´t create but received from a chain of method calls that already use pointers. Also, the pointer will be needed further down the chain of calls by other methods. I prefer to not modify this and keep it this way, if you don't mind :slightly_smiling_face:

src/qt/rpcconsole.cpp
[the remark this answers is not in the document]
hernanmarino: Changed.

src/qt/rpcconsole.cpp
hernanmarino: Modified.

hernanmarino:
I changed the code to address @furszy's suggestions. The new commit hashes are:
[`aa898d6`](https://github.com/bitcoin-core/gui/pull/692/commits/aa898d6a72643542c08f061086ec7eaeb9e16e78) refactors the code to prepare for the new functionality implemented in the following commit, as well as for future console-only commands to be added.

[`1db3da2`](https://github.com/bitcoin-core/gui/pull/692/commits/1db3da2f8b30221f0a94b5881577e2704c1e6da7) adds the new "generate" command, as well as the updated "help generate"

pablomartin4btc (commented):
re-ACK, verified all code changes, nice job.
Tested the following commands in the RPC Console window in qt using `regtest`: `generate` (also with invalid arguments), `help`, `help-console`, `help generate`, `getblockhash 0`, `getindexinfo`.

furszy (commented):
Left a quick review, not really blocking. Mostly performance and code improvements.
Will finish testing it later today and ACK it if all goes well.

furszy (commented):
Two small scenarios to fix:

1) Should guard `generate` from using negative numbers. Currently, can call `generate -20` and the process will return an empty blocks array instead of failing.
2) The `generatetoaddress` result has a bracket without the correct indentation. See 
screenshot

luke-jr:
Approach NACK. If we're going to go this route, we can just add back the RPC method... :/

src/qt/rpcconsole.cpp
```diff
@@ -485,6 +489,60 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
...
+        std::string result;
+        std::string address;
+        if (!RPCConsole::RPCExecuteCommandLine(m_node, address, "getnewaddress\n", /*pstrFilteredOut=*/nullptr, wallet_model)) {
+            Q_EMIT reply(RPCConsole::CMD_ERROR, QString("Error: could not generate new address"));
+        } else {
+            if (!RPCConsole::RPCExecuteCommandLine(m_node, result, "generatetoaddress " + nblocks + " " + address + " " + maxtries + "\n", /*pstrFilteredOut=*/nullptr, wallet_model)) {
+                    Q_EMIT reply(RPCConsole::CMD_ERROR, QString("Error: could not generate blocks"));
+            } else {
+                    std::string answer = "{\n  \"address\": \"" + address + "\",\n  \"blocks\": " + result + "\n}";
+                    Q_EMIT reply(RPCConsole::CMD_REPLY, QString::fromStdString("\n" + answer + "\n\n"));
```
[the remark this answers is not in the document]
hernanmarino: Modified

src/qt/rpcconsole.cpp
```diff
@@ -485,6 +489,60 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
...
+        const std::string maxtries{parsed_command.size() > 2 ? parsed_command[2] : ""};
+
+        // Generate address
+        std::string result;
+        std::string address;
+        if (!RPCConsole::RPCExecuteCommandLine(m_node, address, "getnewaddress\n", /*pstrFilteredOut=*/nullptr, wallet_model)) {
+            Q_EMIT reply(RPCConsole::CMD_ERROR, QString("Error: could not generate new address"));
+        } else {
+            if (!RPCConsole::RPCExecuteCommandLine(m_node, result, "generatetoaddress " + nblocks + " " + address + " " + maxtries + "\n", /*pstrFilteredOut=*/nullptr, wallet_model)) {
+                    Q_EMIT reply(RPCConsole::CMD_ERROR, QString("Error: could not generate blocks"));
```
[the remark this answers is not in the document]
hernanmarino: Modified

src/qt/rpcconsole.cpp
```diff
@@ -485,6 +489,60 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
...
 
+/**
+ * @brief Executes the console-only command "generate".
+ * @param parsed_command  A vector of strings with command and parameters, usually generated by RPCExecutor::parseHelper
+ * @param wallet_model  WalletModel to use for the command
+ * @return True if the command was executed, false otherwise.
+ */
+bool RPCExecutor::executeConsoleGenerate(const std::vector<std::string>& parsed_command, const WalletModel* wallet_model)
+{
+    // Catch the console-only generate command with 2 or less parameters before RPC call is executed .
```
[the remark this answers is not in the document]
hernanmarino: Done

src/qt/rpcconsole.cpp
```diff
@@ -96,8 +96,49 @@ public Q_SLOTS:
...
+ * Splits a command line string into a vector with command at position 0, and parameters after
+ *
+ * @param[in]    strCommand  Command line to parse
+ *
+ * @return a vector of strings with command and parameters
+ */
+std::vector<std::string> parseHelper(const std::string& strCommand)
+{
+    // Split while recognizing the several characters that can be used as separators in the GUI console
+    std::vector<std::string> vec{SplitString(strCommand, " (),")};
```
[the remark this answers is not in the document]
hernanmarino: Yes, agree. This will have to be refactored if such thing is ever needed.

src/qt/rpcconsole.cpp
```diff
@@ -464,6 +485,58 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
...
+    const std::vector<std::string> parsed_command{parseHelper(command)};
+
+    if (parsed_command.empty()) return false;
+
+    // Iterate over m_method_map and execute the associated method if the beggining of parsed_command matches the key
+    for (auto const& [key, method] : m_method_map) {
+        if (commandMatches(parsed_command, key)) {
+            return (this->*method)(parsed_command, wallet_model);
+        }
+    }
```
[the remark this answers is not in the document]
hernanmarino: Implemented

hernanmarino:
Updated with all the changes suggested by @furszy in his latest review.

pablomartin4btc (approved):
re tACK e6e0dacc92953f1a190a18bf64d18fd7572354d7

Tested the [2 scenarios](https://github.com/bitcoin-core/gui/pull/692#pullrequestreview-1511346984) detected by @furszy above, both fixed.

pablomartin4btc:
> Approach NACK. If we're going to go this route, we can just add back the RPC method... :/

I think we can't bring the RPC command back as it was removed to decouple the wallet dependency from the node ([#14468](https://github.com/bitcoin/bitcoin/pull/14468)) simplifying also its interface ([#10973](https://github.com/bitcoin/bitcoin/pull/10973)).

Also as `generate` is only used for testing, and it's already implemented in `cli`, personally makes sense to have it in `qt`, as for me is even more practical to use it from the `rpc console` rather than configure `qt` to enable `rpc` (need to add `"server": true,` in `settings.json`) and send commands from a terminal with `cli`.

Code-wise, the first commit, which refactors the current console-side commands (eg `help-console`) allows adding new console commands easily (if we ever need to).

hernanmarino:
> Approach NACK. If we're going to go this route, we can just add back the RPC method... :/

I understand your point of view, but the reasons mentioned in the issues ( https://github.com/bitcoin-core/gui/issues/55 https://github.com/bitcoin/bitcoin/issues/16000 ), some of them mentioned by Pablo,  encouraged me to work on this. Also, it evolved to (a litlle bit) more than bringing back the RPC call.

hernanmarino:
@maflcko @jonatack @Sjors since you opened or ACK'ed the original issues I kindly request your review on this PR.

Sjors:
> Approach NACK. If we're going to go this route, we can just add back the RPC method... :/

IIUC the reason the RPC method was removed (and replaced with `bitcoin-cli -generate`) is because it had a wallet dependency. The GUI has wallet dependencies all over the place, so I don't see a problem adding this.

src/qt/rpcconsole.cpp
```diff
@@ -468,6 +471,73 @@ void RPCExecutor::request(const QString &command, const WalletModel* wallet_mode
...
+    const std::string maxtries{parsed_command.size() > 2 ? parsed_command[2] : "1000000"};
+
+    // Handle some special cases...
+    // Default to console help generate if more than 3 parameters or if "help generate" was called
+    if (parsed_command.size() > 3 || exec_help) {
+        executeConsoleHelpGenerate();
+        return true;
+    }
+    // Fail if we are on mainnet, to avoid generating addresses for blocks that will not be generated
+    if (Params().GetChainType() == ChainType::MAIN) {
```
Sjors, line 493: Better to fail on all networks expect regtest. It might be possible to mine on a custom signet too, but anyone who knows how to make that can use the RPC.

hebasto:
Related:
- https://github.com/bitcoin/bitcoin/issues/28387
- https://bitcoin.stackexchange.com/questions/119491/generate-regtest-blocks-with-bitcoin-qt/119492

src/qt/rpcconsole.cpp
hernanmarino: Agree, I'll change it soon

hernanmarino:
Rebased on top of master to fix CI errors.

furszy (commented):
Executing `generate 1` on a node without a wallet leads to a crash. And left some other review notes.

Will read the historical context for the `generate` command deletion before proceed.

PR #693, opened 2023-01-02
Fix segfault when shutdown during wallet open

### LINKED ISSUE #689: Segmentation fault when closing while loading wallets
Steps to reproduce:

* Exit or CTRL+C while the "Loading wallets" *pop-up* is visible.

```
Thread 1 "bitcoin-qt" received signal SIGSEGV, Segmentation fault.


Thread 1 (Thread 0x7ffff5935080 (LWP 52927) "bitcoin-qt"):
#0  0x00007ffff78cf1ae in QObjectPrivate::addConnection(int, QObjectPrivate::Connection*) () at /lib64/libQt5Core.so.5
#1  0x00007ffff78d5067 in QObjectPrivate::connectImpl(QObject const*, int, QObject const*, void**, QtPrivate::QSlotObjectBase*, Qt::ConnectionType, int const*, QMetaObject const*) () at /lib64/libQt5Core.so.5
#2  0x00007ffff78d54d5 in QObject::connectImpl(QObject const*, void**, QObject const*, void**, QtPrivate::QSlotObjectBase*, Qt::ConnectionType, int const*, QMetaObject const*) () at /lib64/libQt5Core.so.5
#3  0x00005555556c700d in QObject::connect<void (WalletModel::*)(interfaces::WalletBalances const&), void (SendCoinsDialog::*)(interfaces::WalletBalances const&)>(QtPrivate::FunctionPointer<void (WalletModel::*)(interfaces::WalletBalances const&)>::Object const*, void (WalletModel::*)(interfaces::WalletBalances const&), QtPrivate::FunctionPointer<void (SendCoinsDialog::*)(interfaces::WalletBalances const&)>::Object const*, void (SendCoinsDialog::*)(interfaces::WalletBalances const&), Qt::ConnectionType) (sender=0x7fff28001900, signal=(void (WalletModel::*)(WalletModel * const, const interfaces::WalletBalances &)) 0x5555556c83e0 <SendCoinsDialog::setBalance(interfaces::WalletBalances const&)>, receiver=0x555556f20860, slot=(void (SendCoinsDialog::*)(SendCoinsDialog * const, const interfaces::WalletBalances &)) 0x555556e4c2c0, this adjustment 93825023925952, type=Qt::AutoConnection) at /usr/include/qt5/QtCore/qobject.h:264
#4  SendCoinsDialog::setModel(WalletModel*) (this=0x555556f20860, _model=0x7fff28001900) at qt/sendcoinsdialog.cpp:167
#5  0x000055555568726a in WalletView::WalletView(WalletModel*, PlatformStyle const*, QWidget*) (this=0x555556650360, wallet_model=<optimized out>, _platformStyle=<optimized out>, parent=<optimized out>) at qt/walletview.cpp:65
#6  0x000055555560321a in BitcoinGUI::addWallet(WalletModel*) (this=0x555556fde0e0, walletModel=0x7fff28001900) at qt/bitcoingui.cpp:697
#7  0x0000555555609c22 in _ZZN7GUIUtil20ExceptionSafeConnectIP16WalletControllerMS1_FvP11WalletModelEP10BitcoinGUIMS7_FvS4_EEEDaT_T0_T1_T2_N2Qt14ConnectionTypeEENKUlDpOT_E_clIJRS4_EEEDaSJ_ (this=0x5555566cebd0, args=<optimized out>) at ./qt/guiutil.h:400
#8  0x0000555555609bc0 in _ZN9QtPrivate11FunctorCallINS_11IndexesListIJLi0EEEENS_4ListIJP11WalletModelEEEvZN7GUIUtil20ExceptionSafeConnectIP16WalletControllerMS9_FvS5_EP10BitcoinGUIMSD_FvS5_EEEDaT_T0_T1_T2_N2Qt14ConnectionTypeEEUlDpOT_E_E4callERSQ_PPv (f=..., arg=0x5) at /usr/include/qt5/QtCore/qobjectdefs_impl.h:146
#9  _ZN9QtPrivate7FunctorIZN7GUIUtil20ExceptionSafeConnectIP16WalletControllerMS3_FvP11WalletModelEP10BitcoinGUIMS9_FvS6_EEEDaT_T0_T1_T2_N2Qt14ConnectionTypeEEUlDpOT_E_Li1EE4callINS_4ListIJS6_EEEvEEvRSM_PvPSS_ (f=..., arg=0x5) at /usr/include/qt5/QtCore/qobjectdefs_impl.h:256
#10 _ZN9QtPrivate18QFunctorSlotObjectIZN7GUIUtil20ExceptionSafeConnectIP16WalletControllerMS3_FvP11WalletModelEP10BitcoinGUIMS9_FvS6_EEEDaT_T0_T1_T2_N2Qt14ConnectionTypeEEUlDpOT_E_Li1ENS_4ListIJS6_EEEvE4implEiPNS_15QSlotObjectBaseEP7QObjectPPvPb (which=<optimized out>, this_=<optimized out>, r=0x555557728670, a=0x5, ret=0x7fffffffcd18) at /usr/include/qt5/QtCore/qobjectdefs_impl.h:443
#11 0x00007ffff78d2ec4 in QObject::event(QEvent*) () at /lib64/libQt5Core.so.5
#12 0x00007ffff6faed12 in QApplicationPrivate::notify_helper(QObject*, QEvent*) () at /lib64/libQt5Widgets.so.5
#13 0x00007ffff78a8278 in QCoreApplication::notifyInternal2(QObject*, QEvent*) () at /lib64/libQt5Core.so.5
#14 0x00007ffff78ab5e4 in QCoreApplicationPrivate::sendPostedEvents(QObject*, int, QThreadData*) () at /lib64/libQt5Core.so.5
#15 0x00007ffff78f9897 in postEventSourceDispatch(_GSource*, int (*)(void*), void*) () at /lib64/libQt5Core.so.5
#16 0x00007ffff5266cbf in g_main_context_dispatch () at /lib64/libglib-2.0.so.0
#17 0x00007ffff52bc598 in g_main_context_iterate.constprop () at /lib64/libglib-2.0.so.0
#18 0x00007ffff5263f40 in g_main_context_iteration () at /lib64/libglib-2.0.so.0
#19 0x00007ffff78f938a in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) () at /lib64/libQt5Core.so.5
#20 0x00007ffff78a6cca in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) () at /lib64/libQt5Core.so.5
#21 0x00007ffff78aed92 in QCoreApplication::exec() () at /lib64/libQt5Core.so.5
#22 0x00005555555f25fd in GuiMain(int, char**) (argc=<optimized out>, argv=<optimized out>) at qt/bitcoin.cpp:697
#23 0x00005555555ecc4a in main(int, char**) (argc=-67945086, argv=0x3) at qt/main.cpp:24

### PR DESCRIPTION
**Pull request by john-moffett**
Fixes #689 

## Summary

If you open a wallet and send a shutdown signal during that process, you'll get a segfault when the wallet finishes opening. That's because the `WalletController` object gets deleted manually in bitcoin.cpp during shutdown, but copies of the pointer (and pointers to child objects) are dangling in various places and are accessed in queued events after the deletion.

## Details

The issue in #689 is caused by the following sequence of events:

1. Wallet open modal dialog is shown and worker thread does the actual work.
2. Every 200ms, the main event loop checks to see if a shutdown has been requested, but only if a modal is not being shown.
3. Request a shutdown while the modal window is shown.
4. The wallet open process completes, the modal window is dismissed, and various `finish` signals are sent.
5. During handling of one of the `finish` signals, `qApp->processEvents()` is [called](https://github.com/bitcoin-core/gui/blob/e9262ea32a6e1d364fb7974844fadc36f931f8c6/src/qt/sendcoinsdialog.cpp#L603), which causes the main event loop to detect the shutdown (now that the modal window has been dismissed). The `WalletController` and all the `WalletModel`s are [deleted](https://github.com/bitcoin-core/gui/blob/65de8eeeca29e71378aa34602b287ab921b040e4/src/qt/bitcoin.cpp#L394-L401).
6. Control returns to the `finish` method, which eventually tries to send a [signal](https://github.com/bitcoin-core/gui/blob/e9262ea32a6e1d364fb7974844fadc36f931f8c6/src/qt/sendcoinsdialog.cpp#L167) from a wallet model, but it's been deleted already (and the signal is sent from a now-[dangling](https://github.com/bitcoin-core/gui/blob/d8bdee0fc889def7c5f5076da13db4fce0a3728a/src/qt/walletview.cpp#L65) pointer).

The simplest fix for that is to change the `qApp->processEvents()` into a `QueuedConnection` call. (The `qApp->processEvents() was a [workaround](https://github.com/bitcoin/bitcoin/pull/593#issuecomment-3050699) to get the GUI to scroll to the last item in a list that just got added, and this is just a safer way of doing that).

However, once that segfault is fixed, another segfault occurs due to some queued wallet events happening after the wallet controller object is deleted here:

https://github.com/bitcoin-core/gui/blob/65de8eeeca29e71378aa34602b287ab921b040e4/src/qt/bitcoin.cpp#L394-L401

Since `m_wallet_controller` is a copy of that pointer in `bitcoingui.cpp`, it's now dangling and `if(null)` checks won't work correctly. For instance, this line:

https://github.com/bitcoin-core/gui/blob/65de8eeeca29e71378aa34602b287ab921b040e4/src/qt/bitcoingui.cpp#L413

sets up a `QueuedConnection` to `setCurrentWallet`, but by the time control reaches that method (one event cycle after shutdown deleted `m_wallet_controller` in `bitcoin.cpp`), the underlying objects have been destroyed (but the pointers are still dangling).

Ideally, we'd use a `QPointer` or `std::shared_ptr / std::weak_ptr`s for these, but the changes would be more involved.

This is a minimal fix for the issues. Just set `m_wallet_controller` to `nullptr` in `bitcoingui.cpp`, check its value in a couple places, and avoid a use of `qApp->processEvents`.

### DIFF
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -600,10 +600,15 @@ SendCoinsEntry *SendCoinsDialog::addEntry()
     entry->clear();
     entry->setFocus();
     ui->scrollAreaWidgetContents->resize(ui->scrollAreaWidgetContents->sizeHint());
-    qApp->processEvents();
-    QScrollBar* bar = ui->scrollArea->verticalScrollBar();
-    if(bar)
-        bar->setSliderPosition(bar->maximum());
+
+    // Scroll to the newly added entry on a QueuedConnection because Qt doesn't
+    // adjust the scroll area and scrollbar immediately when the widget is added.
+    // Invoking on a DirectConnection will only scroll to the second-to-last entry.
+    QMetaObject::invokeMethod(ui->scrollArea, [this] {
+        if (ui->scrollArea->verticalScrollBar()) {
+            ui->scrollArea->verticalScrollBar()->setValue(ui->scrollArea->verticalScrollBar()->maximum());
+        }
+    }, Qt::QueuedConnection);
 
     updateTabsAndLabels();
     return entry;
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -680,6 +680,10 @@ void BitcoinGUI::setWalletController(WalletController* wallet_controller)
 
     GUIUtil::ExceptionSafeConnect(wallet_controller, &WalletController::walletAdded, this, &BitcoinGUI::addWallet);
     connect(wallet_controller, &WalletController::walletRemoved, this, &BitcoinGUI::removeWallet);
+    connect(wallet_controller, &WalletController::destroyed, this, [this] {
+        // wallet_controller gets destroyed manually, but it leaves our member copy dangling
+        m_wallet_controller = nullptr;
+    });
 
     auto activity = new LoadWalletsActivity(m_wallet_controller, this);
     activity->load();
@@ -692,7 +696,7 @@ WalletController* BitcoinGUI::getWalletController()
 
 void BitcoinGUI::addWallet(WalletModel* walletModel)
 {
-    if (!walletFrame) return;
+    if (!walletFrame || !m_wallet_controller) return;
 
     WalletView* wallet_view = new WalletView(walletModel, platformStyle, walletFrame);
     if (!walletFrame->addView(wallet_view)) return;
@@ -742,7 +746,7 @@ void BitcoinGUI::removeWallet(WalletModel* walletModel)
 
 void BitcoinGUI::setCurrentWallet(WalletModel* wallet_model)
 {
-    if (!walletFrame) return;
+    if (!walletFrame || !m_wallet_controller) return;
     walletFrame->setCurrentWallet(wallet_model);
     for (int index = 0; index < m_wallet_selector->count(); ++index) {
         if (m_wallet_selector->itemData(index).value<WalletModel*>() == wallet_model) {
--- a/src/qt/test/wallettests.cpp
+++ b/src/qt/test/wallettests.cpp
@@ -212,6 +212,8 @@ void TestGUI(interfaces::Node& node)
     QCOMPARE(transactionTableModel->rowCount({}), 105);
     uint256 txid1 = SendCoins(*wallet.get(), sendCoinsDialog, PKHash(), 5 * COIN, /*rbf=*/false);
     uint256 txid2 = SendCoins(*wallet.get(), sendCoinsDialog, PKHash(), 10 * COIN, /*rbf=*/true);
+    // Transaction table model updates on a QueuedConnection, so process events to ensure it's updated.
+    qApp->processEvents();
     QCOMPARE(transactionTableModel->rowCount({}), 107);
     QVERIFY(FindTx(*transactionTableModel, txid1).isValid());
     QVERIFY(FindTx(*transactionTableModel, txid2).isValid());

### CODE REVIEW

**Comment by john-moffett** [comment]
Pushed an update to address the failing tests. 

They failed because the `TransactionTableModel` gets updated on a `QueuedConnection`:

https://github.com/bitcoin-core/gui/blob/d8bdee0fc889def7c5f5076da13db4fce0a3728a/src/qt/transactiontablemodel.cpp#L73-L76

The reason given is `// queue notifications to show a non freezing progress dialog e.g. for rescan`.

The `qApp->processEvents()` workaround in `SendCoinsDialog` had the side effect of making it act like a `DirectConnection` when sent from the GUI. 

The fix for that is either to change the connection in the `TransactionTableModel` into `AutoConnection` (which goes against the comment) or do a `qApp->processEvents()` in the test so that the model updates in the next event loop. 

I chose the latter method.

File: src/qt/sendcoinsdialog.cpp [hunk 1]
```diff
@@ -600,10 +600,11 @@ SendCoinsEntry *SendCoinsDialog::addEntry()
...
     ui->scrollAreaWidgetContents->resize(ui->scrollAreaWidgetContents->sizeHint());
-    qApp->processEvents();
-    QScrollBar* bar = ui->scrollArea->verticalScrollBar();
-    if(bar)
-        bar->setSliderPosition(bar->maximum());
+    QMetaObject::invokeMethod(ui->scrollArea, [this] {
+        if (ui->scrollArea->verticalScrollBar()) {
+            ui->scrollArea->verticalScrollBar()->setValue(ui->scrollArea->verticalScrollBar()->maximum());
+        }
+    }, Qt::QueuedConnection);
```
furszy, line 607: This is most likely because we are creating the widget in a separate thread.

Try this:
```diff
diff --git a/src/qt/sendcoinsdialog.cpp b/src/qt/sendcoinsdialog.cpp
--- a/src/qt/sendcoinsdialog.cpp	(revision 114d738830368dc155f27a72128adf91b4fffd37)
+++ b/src/qt/sendcoinsdialog.cpp	(date 1672923728012)
@@ -600,11 +600,9 @@
     entry->clear();
     entry->setFocus();
     ui->scrollAreaWidgetContents->resize(ui->scrollAreaWidgetContents->sizeHint());
-    QMetaObject::invokeMethod(ui->scrollArea, [this] {
-        if (ui->scrollArea->verticalScrollBar()) {
-            ui->scrollArea->verticalScrollBar()->setValue(ui->scrollArea->verticalScrollBar()->maximum());
-        }
-    }, Qt::QueuedConnection);
+    if (ui->scrollArea->verticalScrollBar()) {
+        ui->scrollArea->verticalScrollBar()->setValue(ui->scrollArea->verticalScrollBar()->maximum());
+    }
 
     updateTabsAndLabels();
     return entry;
@@ -697,6 +695,13 @@
     return true;
 }
 
+void SendCoinsDialog::showEvent(QShowEvent* event)
+{
+    if (ui->scrollArea->verticalScrollBar()) {
+        ui->scrollArea->verticalScrollBar()->setValue(ui->scrollArea->verticalScrollBar()->maximum());
+    }
+}
+
 void SendCoinsDialog::setBalance(const interfaces::WalletBalances& balances)
 {
     if(model && model->getOptionsModel())
Index: src/qt/sendcoinsdialog.h
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/qt/sendcoinsdialog.h b/src/qt/sendcoinsdialog.h
--- a/src/qt/sendcoinsdialog.h	(revision 114d738830368dc155f27a72128adf91b4fffd37)
+++ b/src/qt/sendcoinsdialog.h	(date 1672923728023)
@@ -49,6 +49,8 @@
     void pasteEntry(const SendCoinsRecipient &rv);
     bool handlePaymentRequest(const SendCoinsRecipient &recipient);
 
+    void showEvent(QShowEvent* event) override;
+
 public Q_SLOTS:
     void clear();
     void reject() override;
```

john-moffett: Thank you for the suggestion!

While this (also) fixes the first segfault, it doesn't achieve the intended scrolling behavior (at least on my macOS setup). 

Before clicking 'Add Recipient':

image

After:

image

Whereas the original workaround (and my workaround of the workaround) gives this "After":

image

It seems to be a fairly [common](https://stackoverflow.com/questions/65087088/how-to-move-to-end-of-scrollbar-after-adding-qwidget-into-the-qscrollarea) issue [without](https://stackoverflow.com/questions/63884331/scroll-to-the-bottom-when-adding-an-element-to-a-qscrollarea) an [ideal](https://stackoverflow.com/questions/19836404/qscrollarea-does-not-scroll-to-maximum-after-widgets-have-been-added) solution. There are more involved solutions, like subscribing to `rangeChanged` signals, but those would require much more logic (eg - ignore those when manually resizing the viewport, etc.).

furszy: hmm true, ok. Sounds good to go with the simplest fix here.
Would be good to add a comment above so we don't forget the rationale behind it.

furszy (commented):
Code ACK 114d7388 with the nuance of not being able to reproduce the crash on master on my system.

src/qt/sendcoinsdialog.cpp
john-moffett: Good call. Adding.

furszy:
Thanks for the comment.
Should squash the commits, check https://github.com/bitcoin-core/gui/blob/master/CONTRIBUTING.md#squashing-commits

hebasto:
cc @promag

hebasto:
Concept ACK.

PR #696, opened 2023-01-10
Switch RPCConsole wallet selection to the one most recently opened/restored/created

john-moffett:
If a user opens multiple wallets in the GUI from the menu bar, the last one opened is the active one in the main window. However, For the RPC Console window, the  _first_ one opened is active. This can be confusing, as wallet RPC commands may be sent to a wallet the user didn't intend.

This PR makes the RPC Console switch to the wallet just opened / restored / created from the menu bar, which is how the main GUI now works.

Similar to https://github.com/bitcoin-core/gui/pull/665 and specifically requested [in a comment](https://github.com/bitcoin-core/gui/pull/665#issuecomment-1270003660).

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -798,6 +798,12 @@ void RPCConsole::removeWallet(WalletModel * const walletModel)
         ui->WalletSelectorLabel->setVisible(false);
     }
 }
+
+void RPCConsole::setCurrentWallet(WalletModel* const wallet_model)
+{
+    QVariant data = QVariant::fromValue(wallet_model);
+    ui->WalletSelector->setCurrentIndex(ui->WalletSelector->findData(data));
+}
 #endif
 
 static QString categoryClass(int category)
--- a/src/qt/rpcconsole.h
+++ b/src/qt/rpcconsole.h
@@ -136,6 +136,10 @@ public Q_SLOTS:
     void unbanSelectedNode();
     /** set which tab has the focus (is visible) */
     void setTabFocus(enum TabTypes tabType);
+#ifdef ENABLE_WALLET
+    /** Set the current (ie - active) wallet */
+    void setCurrentWallet(WalletModel* const wallet_model);
+#endif // ENABLE_WALLET
 
 private:
     struct TranslatedStrings {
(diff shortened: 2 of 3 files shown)


john-moffett:
Is there any guidance on what to do in cases of (almost certainly) spurious CI test failures? The test [failed](https://cirrus-ci.com/task/4566312526020608?logs=functional_tests#L39) here:

https://github.com/bitcoin-core/gui/blob/9887fc789886b356efc6818783cabaab2b14481f/test/functional/p2p_invalid_messages.py#L90-L92

It seems this behavior is expected in very rare cases. I imagine that asking for a manual CI restart is preferred over (say) force pushing an empty amended commit?

hebasto:
> I imagine that asking for a manual CI restart is preferred

Restarted.

hebasto:
Concept ACK.

src/qt/rpcconsole.h
```diff
@@ -58,6 +58,7 @@ class RPCConsole: public QWidget
 #ifdef ENABLE_WALLET
     void addWallet(WalletModel* const walletModel);
     void removeWallet(WalletModel* const walletModel);
+    void SetCurrentWallet(WalletModel* const wallet_model);
```
hebasto, line 61: 1. Should it be declared as a slot?

2. In the `qt` subdirectory we usually follow Qt's naming convention. Mind renaming please
```suggestion
    void setCurrentWallet(WalletModel* const wallet_model);
```
?

src/qt/rpcconsole.cpp
```diff
@@ -798,6 +798,12 @@ void RPCConsole::removeWallet(WalletModel * const walletModel)
         ui->WalletSelectorLabel->setVisible(false);
     }
 }
+
+void RPCConsole::SetCurrentWallet(WalletModel* const wallet_model)
```
hebasto, line 802: Why `const`? The parameter of all connected signals is `WalletModel* wallet_model`, right?

luke-jr: It's the pointer that's `const` here. Seems pretty reasonable, though it only has a net effect of preventing accidentally changing `wallet_model` in this method.

src/qt/rpcconsole.h
luke-jr: The `const` has no effect here.

luke-jr (changes requested):
utACK 5781f45ff32841a3b5d365f5683def34ff2c356e, after method renamed to `setCurrentWallet`

hebasto:
ping @john-moffett :)

john-moffett:
Thanks for pings, @hebasto and @jarolrod. Apologies for the delay. I've renamed the method and moved the declaration to the `Q_SLOTS` section.

luke-jr (approved):
utACK 99c0eb9701e71f16aa360a420b7e4851d5b92510

hebasto (approved):
ACK 99c0eb9701e71f16aa360a420b7e4851d5b92510, tested on Ubuntu 23.04.

PR #697, opened 2023-01-17
Remove reindex special case from the progress bar label

MarcoFalke:
The user knows which option they passed to the program, so it seems overly verbose to offer the user feedback whether or not they passed `-reindex`. Treat it as `DISK`, like all other cases that are treated as `DISK`:

* `-reindex-chainstate`
* `-loadblock`

### DIFF
--- a/src/interfaces/node.h
+++ b/src/interfaces/node.h
@@ -177,11 +177,8 @@ public:
     //! Is initial block download.
     virtual bool isInitialBlockDownload() = 0;
 
-    //! Get reindex.
-    virtual bool getReindex() = 0;
-
-    //! Get importing.
-    virtual bool getImporting() = 0;
+    //! Is loading blocks.
+    virtual bool isLoadingBlocks() = 0;
 
     //! Set network active.
     virtual void setNetworkActive(bool active) = 0;
--- a/src/node/interfaces.cpp
+++ b/src/node/interfaces.cpp
@@ -295,8 +295,7 @@ public:
     bool isInitialBlockDownload() override {
         return chainman().ActiveChainstate().IsInitialBlockDownload();
     }
-    bool getReindex() override { return node::fReindex; }
-    bool getImporting() override { return node::fImporting; }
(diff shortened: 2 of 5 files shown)


MarcoFalke:
For reference:

* `-reindex-chainstate` remains unchanged:
rcs

* `-loadblock` remains unchanged:
lblk

* `-reindex` *stage 2* remains unchanged:
reindex_same_2

* `-reindex` *stage 1* (master):
reindex_master


* `-reindex` *stage 1* (this pull):
reindex_1_pull

MarcoFalke:
So the only user-visible change should be changing the word "Reindex" to "Index"

hebasto:
Concept ACK.

As this PR touches code outside the `src/qt` directory, i.e., `src/interfaces/node.h` and `src/node/interfaces.cpp`, should it be moved to the main repo for a broader reviewing?

MarcoFalke:
It is only changing an interface that the gui uses, so I think it should be fine

hebasto:
cc @ryanofsky

LarryRuane (approved):
code-review, tested ACK fa200442946437f6de8aec95487ccbb541cc485b

I ran `src/qt/bitcoin-qt -noconnect -signet -reindex` with and without the PR, and the status message (lower-left corner) for stage 1 was as expected. I didn't try `reindex-chainstate`, `loadblock`, or `reindex` stage 2.

john-moffett (approved):
ACK fa200442946437f6de8aec95487ccbb541cc485b

Nit: Maybe remove the one remaining unnecessary `enum` use for `BlockSource`?

https://github.com/bitcoin-core/gui/blob/fa200442946437f6de8aec95487ccbb541cc485b/src/qt/bitcoingui.cpp#L1074

MarcoFalke:
Thanks, removed and rebased. Should be trivial to re-ACK with git range-diff

john-moffett:
Re-ACK faff2ba4f80fad5af63a31559fd4065e631a8166

LarryRuane:
post-merge re-ACK

PR #119, opened 2020-10-27
Replace send-to-self with dual send+receive entries

luke-jr:
Makes the GUI transaction list more like the RPC, and IMO clearer in general.

As a side effect, this also fixes the GUI entries when a transaction is a net profit to us, but some inputs were also from us.

Originally https://github.com/bitcoin/bitcoin/pull/15115

Has Concept ACKs from @*Empact @*jonasschnelli

PR #698, opened 2023-01-19
gui: Show labels for multiple SendToSelf addresses

john-moffett:
Currently, the transaction table will display an address label for self-payments only if there's one recipient (ie - if you've sent your entire balance to one address you own).

This PR changes it to show labels for any addresses that have labels in the address book. It's a follow-up to https://github.com/bitcoin/bitcoin/pull/15098.

Before:

image

After:

image

The addresses are [stored](https://github.com/bitcoin-core/gui/blob/8ae2808a4354e8dcc697f76bacc5e2f2befe9220/src/qt/transactionrecord.cpp#L96-L103) in `wtx->address` separated by `, `, so split the string and call `lookupAddress` for each one.

### DIFF
--- a/src/qt/transactiontablemodel.cpp
+++ b/src/qt/transactiontablemodel.cpp
@@ -18,6 +18,7 @@
 #include <core_io.h>
 #include <interfaces/handler.h>
 #include <uint256.h>
+#include <util/string.h>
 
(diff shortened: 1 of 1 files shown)


john-moffett:
Closing because I think #119 is superior.

PR #700, opened 2023-01-23
Dialog for allowing the user to choose the change output when bumping a tx

achow101:
Based on https://github.com/bitcoin/bitcoin/pull/26467

Implements a GUI dialog for allowing the user to choose the output to reduce when bumping a transaction. This adds the functionality that was added to the RPC.

### DIFF
--- a/src/qt/bumpfeechoosechangedialog.cpp
+++ b/src/qt/bumpfeechoosechangedialog.cpp
@@ -1,0 +1,75 @@
+// Copyright (c) 2022 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or https://www.opensource.org/licenses/mit-license.php.
+
+#if defined(HAVE_CONFIG_H)
+#include <config/bitcoin-config.h>
+#endif
+
+#include <qt/bumpfeechoosechangedialog.h>
+#include <qt/forms/ui_bumpfeechoosechangedialog.h>
+
+#include <addresstype.h>
+#include <key_io.h>
+#include <qt/bitcoinunits.h>
+#include <qt/guiutil.h>
+#include <qt/optionsmodel.h>
+#include <qt/walletmodel.h>
+
+#include <QHBoxLayout>
+#include <QLabel>
+#include <QRadioButton>
+#include <QVBoxLayout>
+
+BumpfeeChooseChangeDialog::BumpfeeChooseChangeDialog(WalletModel *model, QWidget *parent, const Txid& txid) :
+    QDialog(parent, GUIUtil::dialog_flags),
+    ui(new Ui::BumpfeeChooseChangeDialog),
+    model(model)
+{
+    ui->setupUi(this);
+
+    bool found_change = false;
+    CTransactionRef tx = model->wallet().getTx(txid);
+    for (size_t i = 0; i < tx->vout.size(); ++i) {
+        const CTxOut& txout = tx->vout.at(i);
+        QString address_info = tr("No address decoded");
+        CTxDestination dest;
+        if (ExtractDestination(txout.scriptPubKey, dest)) {
+            std::string address = EncodeDestination(dest);
+            std::string label;
+            if (model->wallet().getAddress(dest, &label, nullptr) && !label.empty()) {
+                address_info = QString::fromStdString(label) + QString(" (") + QString::fromStdString(address) + QString(")");
+            } else {
+                address_info = QString::fromStdString(address);
+            }
+        }
+        QString output_info = tr("%1: %2 to %3").arg(i).arg(BitcoinUnits::formatWithUnit(model->getOptionsModel()->getDisplayUnit(), txout.nValue)).arg(address_info);
+
+        QRadioButton *radio_button = new QRadioButton(output_info, nullptr);
+        radio_button->setObjectName(QString::number(i) + QString("_radioButton"));
+        ui->verticalLayout->addWidget(radio_button);
+
+        if (!found_change && model->wallet().isChange(txout)) {
+            radio_button->setChecked(true);
+            ui->none_radioButton->setChecked(false);
+            found_change = true;
+        }
+    }
+    GUIUtil::handleCloseWindowShortcut(this);
+}
+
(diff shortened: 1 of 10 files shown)


Sjors:
Can you add a screenshot to the PR description?

john-moffett (commented):
Approach ACK

I think I'd prefer if the user had some more information about the outputs in the dialog. Maybe something like this, but worded better:

image
 
Maybe the actual change _address_ isn't even needed, or could be revealed in a tooltip. 

Also, if there's a single change output with sufficient funds for feebumping, then selecting "none" has basically no effect compared to selecting that output, as the wallet will always(?) deduct from that existing change output. Maybe this common case can be simplified by removing the "None" option?

Sjors:
Can we be a bit more bold and just assume that a change address is change, and not show the dialog in that case? Ideally we don't show such a dialog more often than necessary.

john-moffett:
That would prevent the case where you wanted to deduct the extra fee from the recipient, such as when you initially selected "Subtract fee from amount". 

Ideally, you'd only show the dialog if that were selected previously, but I don't _think_ it's persisted anywhere. Eg -

https://github.com/bitcoin-core/gui/blob/30f553d45797847166109a161628dba3bf00bc94/src/qt/sendcoinsrecipient.h#L42

I could be wrong, though.

luke-jr:
Concept ACK. I'd just show "Change" for the change, though. End users shouldn't need to know about change addresses.

src/qt/bumpfeechoosechangedialog.cpp
```diff
@@ -0,0 +1,75 @@
...
+        if (ExtractDestination(txout.scriptPubKey, dest)) {
+            std::string address = EncodeDestination(dest);
+            std::string label;
+            if (model->wallet().getAddress(dest, &label, nullptr, nullptr) && !label.empty()) {
+                address_info = QString::fromStdString(label) + QString(" (") + QString::fromStdString(address) + QString(")");
+            } else {
+                address_info = QString::fromStdString(address);
+            }
+        }
+        QString output_info = QString::number(i) + QString(": ") + BitcoinUnits::formatWithUnit(model->getOptionsModel()->getDisplayUnit(), txout.nValue) + tr(" to ") + address_info;
```
hebasto, line 46: To make translators' life easier and to keep the context as large as possible, I suggest to use `QString::arg`. Also it will fix the `lint-qt-translation.py` warning.

hebasto (commented):
The first commit e1ea0ba0f2aa9b36df6b74b664087b236f174990 fails to compile:
```
  CXX      qt/libbitcoinqt_a-walletmodel.o
qt/walletmodel.cpp: In member function ‘bool WalletModel::bumpFee(uint256, uint256&)’:
qt/walletmodel.cpp:486:41: error: no matching function for call to ‘interfaces::Wallet::createBumpTransaction(uint256&, wallet::CCoinControl&, std::vector<bilingual_str>&, CAmount&, CAmount&, CMutableTransaction&)’
  486 |     if (!m_wallet->createBumpTransaction(hash, coin_control, errors, old_fee, new_fee, mtx)) {
      |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from ./qt/walletmodel.h:16,
                 from qt/walletmodel.cpp:9:
./interfaces/wallet.h:166:18: note: candidate: ‘virtual bool interfaces::Wallet::createBumpTransaction(const uint256&, const wallet::CCoinControl&, std::vector<bilingual_str>&, CAmount&, CAmount&, CMutableTransaction&, std::optional<unsigned int>)’
  166 |     virtual bool createBumpTransaction(const uint256& txid,
      |                  ^~~~~~~~~~~~~~~~~~~~~
./interfaces/wallet.h:166:18: note:   candidate expects 7 arguments, 6 provided
make: *** [Makefile:13550: qt/libbitcoinqt_a-walletmodel.o] Error 1
```

src/qt/bumpfeechoosechangedialog.cpp
achow101: Done

achow101:
Rebased and fixed the build issue in the first commit.

Aminzarbani (approved):

w0xlt (commented):
Approach ACK

hebasto:
@achow101 

`test_bitcoin-qt` still [fails](https://github.com/bitcoin-core/gui/actions/runs/25048051241/job/73368281475) in the CI on Windows.

achow101:
> `test_bitcoin-qt` still [fails](https://github.com/bitcoin-core/gui/actions/runs/25048051241/job/73368281475) in the CI on Windows.

Huh, it works when run directly but not in ctest.

achow101:
The CI failure has something to do with using `QDialog::exec` and trying to send button clicks from the test within a single thread. I cannot figure out how to resolve that problem. Given that this has not had any major review in the past 3 years, I'm going to close this as up for grabs. Hopefully someone who better understands Qt can figure out the problem, or maybe this can be implemented in the new QML GUI.

hebasto:
> ... or maybe this can be implemented in the new QML GUI.

Requested in https://github.com/bitcoin-core/gui-qml/issues/813.

pablomartin4btc (commented):
Tracked down the CI issue. The deadlock is in the test helpers: `ConfirmSend` and `ChooseBumpfeeOutput` used `while(true)` spin loops inside `QTimer::singleShot` callbacks to wait for dialogs to appear. Spinning inside the callback blocks the Qt event loop, so `QDialog::exec()` can never make the dialog visible.

The fix replaces the spin loops with a non-blocking retry: if the dialog is not yet visible, reschedule with a 10 ms delay so the next check fires in a later event loop iteration.

Branch with fix on top: [pablomartin4btc:pr700-rebased](https://github.com/pablomartin4btc/bitcoin-core-gui/tree/pr700-rebased) and CI green: [run 30593340908](https://github.com/pablomartin4btc/bitcoin-core-gui/actions/runs/30593340908).

I'll be opening a rebased PR with this fix integrated.

PR #701, opened 2023-01-24
Persist Mask Values option

achow101:
The mask values option is memory only. If a user has enabled this option, it's reasonable to expect that they would want to have it enabled on the next start.

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -647,6 +647,8 @@ void BitcoinGUI::setClientModel(ClientModel *_clientModel, interfaces::BlockAndH
             // initialize the disable state of the tray icon with the current value in the model.
             trayIcon->setVisible(optionsModel->getShowTrayIcon());
         }
+
+        m_mask_values_action->setChecked(_clientModel->getOptionsModel()->getOption(OptionsModel::OptionID::MaskValues).toBool());
     } else {
         if(trayIconMenu)
         {
--- a/src/qt/optionsmodel.cpp
+++ b/src/qt/optionsmodel.cpp
@@ -227,6 +227,8 @@ bool OptionsModel::Init(bilingual_str& error)
     m_use_embedded_monospaced_font = settings.value("UseEmbeddedMonospacedFont").toBool();
     Q_EMIT useEmbeddedMonospacedFontChanged(m_use_embedded_monospaced_font);
 
+    m_mask_values = settings.value("mask_values", false).toBool();
+
     return true;
 }
 
@@ -435,6 +437,8 @@ QVariant OptionsModel::getOption(OptionID option) const
         return SettingToBool(setting(), DEFAULT_LISTEN);
     case Server:
         return SettingToBool(setting(), false);
+    case MaskValues:
+        return m_mask_values;
     default:
         return QVariant();
     }
@@ -611,6 +615,10 @@ bool OptionsModel::setOption(OptionID option, const QVariant& value)
             update(value.toBool());
             setRestartRequired(true);
         }
+        break;
+    case MaskValues:
+        m_mask_values = value.toBool();
+        settings.setValue("mask_values", m_mask_values);
         break;
     default:
         break;
--- a/src/qt/optionsmodel.h
+++ b/src/qt/optionsmodel.h
@@ -72,6 +72,7 @@ public:
         Listen,                 // bool
         Server,                 // bool
         EnablePSBTControls,     // bool
+        MaskValues,             // bool
         OptionIDRowCount,
     };
 
@@ -120,6 +121,7 @@ private:
     bool fCoinControlFeatures;
     bool m_sub_fee_from_amount;
     bool m_enable_psbt_controls;
+    bool m_mask_values;
 
     //! In-memory settings for display. These are stored persistently by the
     //! bitcoin node but it's also nice to store them in memory to prevent them
--- a/src/qt/overviewpage.cpp
+++ b/src/qt/overviewpage.cpp
@@ -173,6 +173,7 @@ void OverviewPage::handleTransactionClicked(const QModelIndex &index)
 void OverviewPage::setPrivacy(bool privacy)
 {
     m_privacy = privacy;
+    clientModel->getOptionsModel()->setOption(OptionsModel::OptionID::MaskValues, privacy);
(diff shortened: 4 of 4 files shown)


src/qt/optionsmodel.cpp
```diff
@@ -612,6 +616,9 @@ bool OptionsModel::setOption(OptionID option, const QVariant& value)
             setRestartRequired(true);
         }
         break;
+    case MaskValues:
+        m_mask_values = value.toBool();
+        settings.setValue("mask_values", m_mask_values);
```
jarolrod, line 621: :) 
```suggestion
        settings.setValue("mask_values", m_mask_values);
        break;
```

instagibbs:
concept ACK, this is better for sure

src/qt/optionsmodel.cpp
achow101: Fixed.

jarolrod (approved):
tACK 4de02def844102c08b65bf1311a333e7aca482b9

Had been on my list to-do for a while. I agree with this, a follow-up could be to have a setting that would make it so that you'd have to unlock the wallet (if it's encrypted) in order to unmask values.

RandyMcMillan:
tACK https://github.com/bitcoin-core/gui/commit/4de02def844102c08b65bf1311a333e7aca482b9


Obfuscating the amounts in the Transactions panel based on this user preference may be a good follow up to this change. 
Screen Shot 2023-01-23 at 10 24 26 PM

hernanmarino (approved):
ACK , persisting this value seems like the natural option

hebasto:
> The mask values option is memory only. If a user has enabled this option, it's reasonable to expect that they would want to have it enabled on the next start.

It would be nice to have some words how this PR differs from https://github.com/bitcoin-core/gui/pull/655.

achow101:
> It would be nice to have some words about how this PR differs from #655.

Hmm, didn't realize that PR existed.

The difference is that this PR uses `OptionsModel` and follows the pattern used for all of our other settings rather than writing to `QSettings` directly.

jarolrod:
I agree with the direction this takes over the other pr's approach

pablomartin4btc (approved):
tested ACK 4de02def844102c08b65bf1311a333e7aca482b9.

Verified that file `Bitcoin-Qt.conf` gets written with line `mask_values=true`.

This PR fixed issue #652.

pablomartin4btc (commented):
Also, it set's the value to false in the config file when the check is unticked.

I agree with the follow-ups (@jarolrod  & @RandyMcMillan).

kristapsk:
Concept ACK

pablomartin4btc (commented):
> tACK [4de02de](https://github.com/bitcoin-core/gui/commit/4de02def844102c08b65bf1311a333e7aca482b9)
> 
> Obfuscating the amounts in the Transactions panel based on this user preference may be a good follow up to this change. 

I have done it on PR #708.

john-moffett (approved):
tACK 4de02def844102c08b65bf1311a333e7aca482b9

hebasto (commented):
Instead of raw `"mask_values"` literals, suggesting to update the `SettingName()` function:
```diff
--- a/src/qt/optionsmodel.cpp
+++ b/src/qt/optionsmodel.cpp
@@ -54,6 +54,7 @@ static const char* SettingName(OptionsModel::OptionID option)
     case OptionsModel::ProxyPortTor: return "onion";
     case OptionsModel::ProxyUseTor: return "onion";
     case OptionsModel::Language: return "lang";
+    case OptionsModel::MaskValues: return "mask_values";
     default: throw std::logic_error(strprintf("GUI option %i has no corresponding node setting.", option));
     }
 }
```
and use its return value.

src/qt/bitcoingui.cpp
```diff
@@ -647,6 +647,8 @@ void BitcoinGUI::setClientModel(ClientModel *_clientModel, interfaces::BlockAndH
             // initialize the disable state of the tray icon with the current value in the model.
             trayIcon->setVisible(optionsModel->getShowTrayIcon());
         }
+
+        m_mask_values_action->setChecked(_clientModel->getOptionsModel()->getOption(OptionsModel::OptionID::MaskValues).toBool());
```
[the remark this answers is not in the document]
pablomartin4btc: I think @ryanofsky was recommending to remove the case from the `SettingsName` due to its purpose of mapping `OptionsModel` to command line options but to add `OptionsModel::MaksValues` (or leave in this case as @achow101 already added them) to `setOption` and `getOption`.

The implementation of the `getter` looks simpler but then you would need to add the `setter` too to make it consistent and revert the above (remove the case from `setOption` and `getOption`).

Perhaps as a follow-up we could check the model consistent with either solution for the rest of settings (?).

pablomartin4btc, line 651: Also, you could just use `OptionsModel::MaskValues` without the `OptionID`, is this an issue?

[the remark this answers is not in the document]
pablomartin4btc: I think @ryanofsky was recommending to remove the case from the `SettingsName` due to its purpose of mapping `OptionsModel` to command line options but to add `OptionsModel::MaksValues` (or leave in this case as @achow101 already added them) to `setOption` and `getOption`.

The implementation of the getter looks simpler but then you would need to add the setter too to make it consistent and revert the above (remove the case from `setOption` and `getOption`).

Perhaps as a follow-up we could check the model consistent with either solution for the rest of settings (?).

achow101:
> Instead of raw `"mask_values"` literals, suggesting to update the `SettingName()` function:
> ...
> UPD. It seems @ryanofsky has a different [opinion](https://github.com/bitcoin-core/gui/pull/655/files#r979159654), so feel free to ignore this comment.

Originally I had done that, but was looking into how that function is used and decided that adding the option there doesn't make sense. It looks like `SettingName()` is used to map settings that also appear on the command line or in the bitcoin.conf, and the `mask_values` does not.

src/qt/bitcoingui.cpp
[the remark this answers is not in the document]
achow101: I think I'll leave this as is since it's consistent with most of the other settings which use the generic function with the enum.

achow101: Not an issue, will change if I need to retouch.

PR #704, opened 2023-01-26
Correctly limit overview transaction list

### LINKED ISSUE #703: assert: "last < rowCount(parent)" in qabstractitemmodel.cpp
Some but not all of my (descriptor) wallets trigger the following crash, immediately when loading them.

```
GUI: ASSERT: "last < rowCount(parent)" in file itemmodels/qabstractitemmodel.cpp, line 2815
```

Happens on master @ fcff639af137f2a6636ef454d0404b969ab7772e with a `DEBUG=1` depends build, but regardless of the `./configure` `--debug` flag). Intel macOS 13.1. Will try some other configurations.

Also tested against Homebrew qt 5.15.8 (depends is at 5.15.5), which doesn't produce the problem.

I tried bumping depends QT to 5.15.8 but this requires changing our patches, which I'm not sure how to do. Bumping QT is only a solution if the bug is in QT, not if it's on our end.

Interestingly I don't experience this bug in the v24.0.1 guix build, even though that uses the same QT version (since https://github.com/bitcoin/bitcoin/pull/25719).

Random other project that ran into the same assert: https://invent.kde.org/utilities/kfind/-/merge_requests/1

john-moffett:
Fixes #703 

The way the main overview page limits the number of transactions displayed (currently 5) is not an appropriate use of Qt. Our subclassed transaction sort/filter proxy model returns a maximum of `5` in `rowCount()`. However, the model itself actually may hold significantly more. While this has _worked_, it breaks the contract of `rowCount()`. 

If `bitcoin-qt` is run with a DEBUG build of Qt, it'll result in an assert-crash in certain relatively common situations (see #703 for details). Instead of artificially limiting the `rowCount()` in the subclassed filter, we can hide/unhide the rows in the displaying `QListView` upon any changes in the sorted proxy filter.

I loaded a wallet with 20,000 transactions and did not notice any performance differences between master and this branch.

For reference, this is the list I'm referring to:

image

### DIFF
--- a/src/qt/overviewpage.cpp
+++ b/src/qt/overviewpage.cpp
@@ -262,7 +262,6 @@ void OverviewPage::setWalletModel(WalletModel *model)
         // Set up transaction list
         filter.reset(new TransactionFilterProxy());
         filter->setSourceModel(model->getTransactionTableModel());
-        filter->setLimit(NUM_ITEMS);
         filter->setDynamicSortFilter(true);
         filter->setSortRole(Qt::EditRole);
         filter->setShowInactive(false);
@@ -271,6 +270,10 @@ void OverviewPage::setWalletModel(WalletModel *model)
         ui->listTransactions->setModel(filter.get());
         ui->listTransactions->setModelColumn(TransactionTableModel::ToAddress);
 
+        connect(filter.get(), &TransactionFilterProxy::rowsInserted, this, &OverviewPage::LimitTransactionRows);
+        connect(filter.get(), &TransactionFilterProxy::rowsRemoved, this, &OverviewPage::LimitTransactionRows);
+        connect(filter.get(), &TransactionFilterProxy::rowsMoved, this, &OverviewPage::LimitTransactionRows);
+        LimitTransactionRows();
         // Keep up to date with wallet
         setBalance(model->getCachedBalance());
         connect(model, &WalletModel::balanceChanged, this, &OverviewPage::setBalance);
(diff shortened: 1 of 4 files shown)


Sjors:
tACK 08209c039ff4ca5be4982da7a2ab7a624117ce1a

hebasto:
Concept ACK.

src/qt/overviewpage.cpp
```diff
@@ -299,6 +302,16 @@ void OverviewPage::changeEvent(QEvent* e)
     QWidget::changeEvent(e);
 }
 
+// Only show most recent NUM_ITEMS rows
+void OverviewPage::LimitTransactionRows()
+{
+    if (filter && ui->listTransactions && ui->listTransactions->model() && filter.get() == ui->listTransactions->model()) {
+        for (int i = 0; i < filter->rowCount(); ++i) {
+            ui->listTransactions->setRowHidden(i, i >= NUM_ITEMS);
+        }
```
hebasto, line 311: Could this cause any performance issues with 100K+ txs wallets?

hebasto (commented):
Tested 08209c039ff4ca5be4982da7a2ab7a624117ce1a ob Ubuntu 22.04. It fixes #703 for me.

Still reviewing.

src/qt/overviewpage.cpp
john-moffett: I loaded a wallet with 30K txs and ran a time profiler. This call took approximately 3 milliseconds:

image

The vast majority of time was consumed by re-running the filter and converting to `QDateTime` (both already existing behavior and unrelated to this PR):

image

Qualitatively, I didn't notice any difference at all.

I'm happy to generate a bunch more transactions overnight and report back tomorrow.

hebasto (approved):
ACK 08209c039ff4ca5be4982da7a2ab7a624117ce1a, tested on Ubuntu 22.04.

fanquake:
Added to https://github.com/bitcoin/bitcoin/pull/26878 for backporting to 24.x.

PR #705, opened 2023-01-27
doc: Fix comment about how wallet txs are sorted

john-moffett:
The wallet transactions in the node are not sorted by txid (or any hash) since https://github.com/bitcoin/bitcoin/pull/24699. 

This is how they're stored in memory now:

https://github.com/bitcoin-core/gui/blob/835212cd1d8f8fc7f19775f5ff8cc21c099122b2/src/wallet/wallet.h#L397-L399

### DIFF
--- a/src/qt/transactiontablemodel.cpp
+++ b/src/qt/transactiontablemodel.cpp
@@ -93,10 +93,7 @@ public:
 
     TransactionTableModel *parent;
 
-    /* Local cache of wallet.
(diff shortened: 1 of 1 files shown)


hebasto:
cc @achow101

achow101:
ACK c497a198db6f417d2612078a9fbc101e259fab33

jarolrod (approved):
ACK c497a198db6f417d2612078a9fbc101e259fab33

PR #716, opened 2023-03-01
Adjust plural forms for translations

hebasto:
These plural forms were re-added in https://github.com/bitcoin-core/gui/pull/557.

The following `make -C src translate` also caught a new last-minute [added](https://github.com/bitcoin/bitcoin/pull/27157) string.


hebasto:
cc @jarolrod @luke-jr

jarolrod:
concept ack

hebasto:
Closed in favour of #724.

PR #720, opened 2023-03-15
Correct poor grammar in wallet synchronization warning.

rebroad:
This pull request updates the wallet synchronization warning text to provide a clearer and more concise message to users. The original text was somewhat clumsy and suffered from grammatical issues, which could lead to confusion. The changes made are as follows:

- Replace the original warning text with a more concise and grammatically correct version.
- Clarify that transactions using bitcoin from unseen transactions due to ongoing synchronization will be rejected by the network.

These changes aim to improve user experience by providing clearer information about potential issues related to wallet synchronization.

Old text:
"Recent transactions may not yet be visible, and therefore your wallet's balance might be incorrect. This information will be correct once your wallet has finished synchronizing with the bitcoin network, as detailed below.
Attempting to spend bitcoins that are affected by not-yet-displayed transactions will not be accepted by the network."

New text:
"Wallet balance may be inaccurate during synchronization, as recent transactions may still be loading (details below).
Caution: Transactions using bitcoin from unseen transactions (due to ongoing synchronization) will be rejected by the network."

### DIFF
--- a/src/qt/forms/modaloverlay.ui
+++ b/src/qt/forms/modaloverlay.ui
@@ -130,7 +130,7 @@ QLabel { color: rgb(40,40,40);  }</string>
              <item>
               <widget class="QLabel" name="infoText">
                <property name="text">
-                <string>Recent transactions may not yet be visible, and therefore your wallet's balance might be incorrect. This information will be correct once your wallet has finished synchronizing with the bitcoin network, as detailed below.</string>
+                <string>Wallet balance may be inaccurate during synchronization, as recent transactions may still be loading (details below).</string>
                </property>
                <property name="textFormat">
                 <enum>Qt::RichText</enum>
@@ -149,7 +149,7 @@ QLabel { color: rgb(40,40,40);  }</string>
                 </font>
                </property>
                <property name="text">
-                <string>Attempting to spend bitcoins that are affected by not-yet-displayed transactions will not be accepted by the network.</string>
+                <string>Caution: Transactions using bitcoin from unseen transactions (due to ongoing synchronization) will be rejected by the network.</string>
                </property>
                <property name="textFormat">
                 <enum>Qt::RichText</enum>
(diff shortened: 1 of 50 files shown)


hebasto:
Translation files should be modified during the translation process only. Please drop those changes from this PR.

rebroad:
If these files aren't changed then the translations would break, wouldn't
they?

hebasto:
> If these files aren't changed then the translations would break, wouldn't they?

The `bitcon_<LANGUAGE>.ts` translation files are updated by [`bitcoin-maintainer-tools/update-translations.py`](https://github.com/bitcoin-core/bitcoin-maintainer-tools/blob/main/update-translations.py) tool. For example, see https://github.com/bitcoin/bitcoin/pull/27251.

Updating `<source>` strings without updating `<translation>` ones do break translations.

rebroad:
The meaning hasn't changed though, just the grammar, so the existing
translations in other languages are still good. Can the existing
translations be retained if I delete the translation file changes?

hebasto:
> Can the existing translations be retained if I delete the translation file changes?

If the suggested changes to the `src/qt/forms/modaloverlay.ui` file will be merged, during the next translation cycle translators will see the updated strings being marked as "untranslated". Therefore, they have to translate them again,  which is, actually, an additional burden. And `*.ts` translation files will be updated regardless of your manual update. Therefore, it looks just pointless.

rebroad:
So, how are we supposed to change just the English language in a change
without affecting other languages?

luke-jr:
>FYI: the new wording was generated by GPT-4, and educated on how bitcoin works as part of this process. 

That means it has a dubious copyright status and can't be used in the first place. Please don't do that.

rebroad:
> > FYI: the new wording was generated by GPT-4, and educated on how bitcoin works as part of this process.
> 
> That means it has a dubious copyright status and can't be used in the first place. Please don't do that.

@luke-jr  https://youtu.be/roiOLaXr-os - GPT-4 is actually less likely to breach copyright than an organic human is.

Issue #394, opened 2021-08-04
Add go back button to `Confirm wallet encryption` window, add cancel button to `wallet to be encrypted` window


jarolrod:
When i go to create a wallet, I have the option to encrypt my wallet. If i choose this option, the next step is to choose a password. 
ok 👍, I've chosen a password. Now, I click on the `Ok` button to continue to the next screen which is the following warning:

![warning](https://user-images.githubusercontent.com/23396902/128109376-a892312b-a93b-4cf9-a0fc-24f70d93cd4c.png)

Now i have been reminded with the life changing knowledge that if I lose this passphrase then I lose all of my Bitcoin. I now have to think about whether or not the wallet I have chosen is what i want it to be. If I am still content with my wallet, then I can continue on. If i am not content with my password, I cannot go back and change it. Instead, I have to click the `Cancel` button and start all over. The `Confirm wallet encryption` wallet should introduce a `go back` button that allows me to 'go back' to the `Encrypt Wallet` window. 

Additionally, the `Cancel` button in this window can be removed. If we go back we can cancel in the `Encrypt Wallet` window, if we go forward -> read on ...

If I was content with my wallet password I can continue on. The next window is another message reminding me that wallet is going to be encrypted and that this does not fully protect my Bitcoin from being stolen:

![wallet-to-be-encrypted](https://user-images.githubusercontent.com/23396902/128109928-84dfed23-bd3f-4527-bb68-183bc88fa4b1.png)

If I am content with this security assumption I can continue on, but at this stage there is no option to be able to cancel my creation of this wallet. The `Wallet to be encrypted` window should introduce a `Cancel` button that 'cancels' the creation of this wallet.

MarcoFalke:
See also https://github.com/bitcoin/bitcoin/issues/18085

hebasto:
A bit related https://github.com/bitcoin-core/gui/issues/151.

hebasto:
@jarolrod 

Mind reviewing https://github.com/bitcoin-core/gui/pull/722?

PR #722, opened 2023-03-16
wallet: Allow user to navigate options while encrypting at creation

hernanmarino:
This fixes https://github.com/bitcoin-core/gui/issues/394. 
It adds a  "Go back" button to the "Confirm wallet encryption" window, allowing the users to change the password if they want to. It also adds a Cancel button to the "Wallet to be encrypted" window.
Prior to this change users had no option to alter the password, and were forced to either go ahead with wallet creation or cancel the whole process. Also, at the final window, they were shown a warning but with no option to cancel.
The new workflow for wallet encryption and creation is similar to the following:

videoNavigation

### DIFF
--- a/src/qt/askpassphrasedialog.cpp
+++ b/src/qt/askpassphrasedialog.cpp
@@ -104,10 +104,14 @@ void AskPassphraseDialog::accept()
             // Cannot encrypt with empty passphrase
             break;
         }
-        QMessageBox::StandardButton retval = QMessageBox::question(this, tr("Confirm wallet encryption"),
-                 tr("Warning: If you encrypt your wallet and lose your passphrase, you will <b>LOSE ALL OF YOUR BITCOINS</b>!") + "<br><br>" + tr("Are you sure you wish to encrypt your wallet?"),
-                 QMessageBox::Yes|QMessageBox::Cancel,
-                 QMessageBox::Cancel);
+        QMessageBox msgBoxConfirm(QMessageBox::Question,
+                                  tr("Confirm wallet encryption"),
+                                  tr("Warning: If you encrypt your wallet and lose your passphrase, you will <b>LOSE ALL OF YOUR BITCOINS</b>!") + "<br><br>" + tr("Are you sure you wish to encrypt your wallet?"),
+                                  QMessageBox::Cancel | QMessageBox::Yes, this);
+        msgBoxConfirm.button(QMessageBox::Yes)->setText(tr("Continue"));
+        msgBoxConfirm.button(QMessageBox::Cancel)->setText(tr("Back"));
+        msgBoxConfirm.setDefaultButton(QMessageBox::Cancel);
+        QMessageBox::StandardButton retval = (QMessageBox::StandardButton)msgBoxConfirm.exec();
         if(retval == QMessageBox::Yes)
         {
             if(newpass1 == newpass2)
@@ -116,10 +120,19 @@ void AskPassphraseDialog::accept()
                 "your bitcoins from being stolen by malware infecting your computer.");
                 if (m_passphrase_out) {
                     m_passphrase_out->assign(newpass1);
-                    QMessageBox::warning(this, tr("Wallet to be encrypted"),
-                                         "<qt>" +
-                                         tr("Your wallet is about to be encrypted. ") + encryption_reminder +
-                                         "</b></qt>");
+                    QMessageBox msgBoxWarning(QMessageBox::Warning,
+                                              tr("Wallet to be encrypted"),
+                                              "<qt>" +
+                                                  tr("Your wallet is about to be encrypted. ") + encryption_reminder + " " +
+                                                  tr("Are you sure you wish to encrypt your wallet?") +
+                                                  "</b></qt>",
+                                              QMessageBox::Cancel | QMessageBox::Yes, this);
+                    msgBoxWarning.setDefaultButton(QMessageBox::Cancel);
+                    QMessageBox::StandardButton retval = (QMessageBox::StandardButton)msgBoxWarning.exec();
+                    if (retval == QMessageBox::Cancel) {
+                        QDialog::reject();
+                        return;
+                    }
                 } else {
                     assert(model != nullptr);
                     if (model->setWalletEncrypted(newpass1)) {
@@ -145,11 +158,7 @@ void AskPassphraseDialog::accept()
                                      tr("The supplied passphrases do not match."));
             }
         }
-        else
-        {
-            QDialog::reject(); // Cancelled
-        }
-        } break;
+    } break;
     case Unlock:
         try {
             if (!model->setWalletLocked(false, oldpass)) {


src/qt/askpassphrasedialog.cpp
```diff
@@ -104,10 +104,15 @@ void AskPassphraseDialog::accept()
...
             break;
         }
-        QMessageBox::StandardButton retval = QMessageBox::question(this, tr("Confirm wallet encryption"),
-                 tr("Warning: If you encrypt your wallet and lose your passphrase, you will <b>LOSE ALL OF YOUR BITCOINS</b>!") + "<br><br>" + tr("Are you sure you wish to encrypt your wallet?"),
-                 QMessageBox::Yes|QMessageBox::Cancel,
-                 QMessageBox::Cancel);
+        QMessageBox msgBoxConfirm(QMessageBox::Question,
+                                  tr("Confirm wallet encryption"),
+                                  tr("Warning: If you encrypt your wallet and lose your passphrase, you will <b>LOSE ALL OF YOUR BITCOINS</b>!<br><br>"
+                                     "Are you sure you wish to encrypt your wallet?"),
```
hebasto, line 110: Could we avoid changes in translatable strings in this PR?

src/qt/askpassphrasedialog.cpp
```diff
@@ -104,10 +104,15 @@ void AskPassphraseDialog::accept()
...
+                                  QMessageBox::Cancel | QMessageBox::Yes, this);
+        msgBoxConfirm.button(QMessageBox::Yes)->setText(tr("Continue"));
+        msgBoxConfirm.button(QMessageBox::Cancel)->setText(tr("Go back"));
```
hebasto, line 113: I'm not sure which label, "Go back" or just "Back", is more expected in user experience.

ping @GBKS

hebasto (commented):
Concept ACK.

src/qt/askpassphrasedialog.cpp
GBKS: I think both are fine. Personally, I would go for "Back", as the word "Go" doesn't add any additional information.

src/qt/askpassphrasedialog.cpp
hernanmarino: Yes, will do.

src/qt/askpassphrasedialog.cpp
hernanmarino: Changed to "Back"

src/qt/askpassphrasedialog.cpp
hernanmarino: I changed the code to reuse already existing strings, can you confirm it's okey in this new way ?

hernanmarino:
Updated the code as per @hebasto suggestions.

pablomartin4btc (approved):
re-ACK 78660e72001a2561c7ad3026367a69f65414dbd9. cc @jarolrod.

I see[`wallet_create_tx.py --descriptors`](https://cirrus-ci.com/task/6164029874372608?logs=functional_tests#L2752)test is failing and you are re-running it, I don't think it's related with your change but let's see.

jarolrod (approved):
ACK 78660e72001a2561c7ad3026367a69f65414dbd9

Tested for functionality and sanity checked the user flows between master and this PR.

src/qt/askpassphrasedialog.cpp
```diff
@@ -145,11 +158,7 @@ void AskPassphraseDialog::accept()
                                      tr("The supplied passphrases do not match."));
             }
         }
-        else
-        {
-            QDialog::reject(); // Cancelled
-        }
-        } break;
+    } break;
```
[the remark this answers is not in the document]
hernanmarino: When originally testing, the reject() seemed redundant, because the window was also closing without this line. Since the next thing to happen was the method finishing / returning I assumed at that time that the default behaviour of QDialog class was being triggered. But now I was having second thoughts, and brought back the original behavior of calling an explicit reject.

hernanmarino:
After @hebasto 's comment I decided to update the code to explicitly call reject() when cancelling the first dialog. This introduces one less change from the original code, and also makes the intention more explicit to developers / reviewers.

luke-jr:
Concept ACK

But shouldn't the final confirmation dialog also have a "Go back" instead of "Cancel"?

(Better yet, why don't we just show these warnings upfront in the password entry dialog?)

pablomartin4btc (commented):
re-Tested nACK. 

As I mentioned in the above response to @luke-jr, last change made the fix not to work anymore (see animated gif). Please consider my comments above if others also see them useful.

hernanmarino:
Rolled back to undo changes detected by @pablomartin4btc . Now changes are at commit 78660e72001a2561c7ad3026367a69f65414dbd9 , the last one ACK ed by @jarolrod and others.

pablomartin4btc:
re-ACK https://github.com/bitcoin-core/gui/commit/78660e72001a2561c7ad3026367a69f65414dbd9 as previously [here](https://github.com/bitcoin-core/gui/pull/722#pullrequestreview-1414865992).

BrandonOdiwuor (commented):
Concept ACK,

Allowing users to go back and update their passphrase before encrypting the wallet

BrandonOdiwuor (approved):
Tested ACK 78660e72001a2561c7ad3026367a69f65414dbd9

Screenshot from 2023-11-29 20-35-54

alfonsoromanz (approved):
Tested ACK https://github.com/bitcoin-core/gui/pull/722/commits/78660e72001a2561c7ad3026367a69f65414dbd9

hernanmarino:
Rebased for CI, no actual code changes in my code.

alfonsoromanz (approved):
Re-Tested ACK

test_wallet-encryption-navigation

BrandonOdiwuor (approved):
re-Tested ACK cccddc03f0c625daeac7158eb20c1508aea5df39

hebasto (approved):
ACK cccddc03f0c625daeac7158eb20c1508aea5df39, tested on Ubuntu 24.04.

hebasto:
> It looks like a "Back" button is added to the warning pop-up, but it will only take you back to enter a different password, not to disable the password selection in the create wallet dialog. Wouldn't it be better to add the "Back" button to the password dialog?
> 
> Also, wouldn't it be better if the text from the two warning pop-ups was added to the top of the password dialog? Otherwise, it seems odd to first ask for a password from the user, then print two warnings about it, when the warnings could have been provided in one go?

@GBKS 

What kind of a workflow will suit this part of the UI in your opinion?

GBKS:
Good points. I think it could work well to combine the two warnings to a single dialog. It could be placed between the initial `Create wallet` dialog, and the password input.

1. **Create wallet dialog:** The user checks `Encrypt wallet` and clicks `Continue`.
2. **Password warning dialog:** The application informs the user about the risks of using a password and offers `Back` and `Continue` options. The user considers, decides to move forward with the password, and clicks `Continue`.
3. **Password input dialog:** The user enters their password and clicks `Continue` (the other option would be `Back`). It could be an option to have a `I have stored this password in a safe place` check box here. Another option is a password difficulty indicator.

This avoids the stack of dialog-on-dialog by putting them in a linear sequence with consistent `Continue` and `Back` options.

When dialogs are stacked like this and you mix buttons like `Cancel` and `Back`, it can be hard to know what they reference. Do you cancel the whole wallet creation or just the password option, or just the dialog? What does `Back` refer to in a dialog that sits on top of another dialog? Keeping the sequence linear makes this more obvious.

Just some thoughts, not sure if you even want to revisit this or not right after working through this PR.

hebasto:
@hernanmarino 

Are you still working on this PR? If so, what do you think about GBKS's [suggestion](https://github.com/bitcoin-core/gui/pull/722#issuecomment-2124671596)?

hernanmarino:
> @hernanmarino
> 
> Are you still working on this PR? If so, what do you think about GBKS's [suggestion](https://github.com/bitcoin-core/gui/pull/722#issuecomment-2124671596)?

I like them. I can work on them and open a follow up PR soon.

hebasto:
> > @hernanmarino
> > Are you still working on this PR? If so, what do you think about GBKS's [suggestion](https://github.com/bitcoin-core/gui/pull/722#issuecomment-2124671596)?
> 
> I like them. I can work on them and open a follow up PR soon.

Why not implementing them in this PR?

Issue #679, opened 2022-11-13
Change "Open Wallet" popup text


mmikeww:
**Is your feature request related to a problem? Please describe.**

When I use the menu for `File -> Open Wallet`, I get a popup saying "Opening wallet xxx...." with a progress bar. There is no other information. The progress bar just loops endlessly. I can of course close this popup by clicking the X normally. If I try to use the menu item again, I get the same popup and the same looping progress bar.

![image](https://user-images.githubusercontent.com/1900684/201532617-65bb175a-51df-4a9f-991b-d55740384c09.png)

Initially I thought this was a bug, and was in the process of submitting a bug report, until I checked `debug.log`. I shouldn't have to check a log file to understand whats happening. Inside I found these lines:

```
2022-11-13T16:24:26Z [specter3c64f804925fff38\mymultisig] Still rescanning. At block 752699. Progress=0.977750
2022-11-13T16:25:26Z [specter3c64f804925fff38\mymultisig] Still rescanning. At block 753057. Progress=0.978535
2022-11-13T16:26:26Z [specter3c64f804925fff38\mymultisig] Still rescanning. At block 753399. Progress=0.979322
2022-11-13T16:27:26Z [specter3c64f804925fff38\mymultisig] Still rescanning. At block 753844. Progress=0.980183

```

Waiting until this process completed allowed the wallet to open.

**Describe the solution you'd like**

I think the text on that popup window should be changed to be more descriptive for the user. The current text of "Opening wallet....." with nothing but a looping progress bar, initially led me to believe that the wallet file was corrupted or something. That maybe Core couldn't open it, and was just hung or looped. 

The text should be changed to something like "Syncing wallet with chain..." or however you want to word it, additionally with a progress bar that show's the actual percentage like is shown inside the log file.

katesalazar:
+ a "more details" button that once clicked shows:
  * the "At block #i" log lines
  * **and** the AddToWallet lines

pablomartin4btc:
This is kind of duplicated with #651, but I appreciate the level of details, I agree with the change of the title of the window + updating the progress bar with the actual progress to enhance accuracy but to avoid confusion as well.

Issue #651, opened 2022-08-09
Progress for opening wallet


fanquake:
While there is progress for importing private keys and rescanning, there is good reason to also add progress for wallet opening.
Now you need to look into debug.log and trace it for progress, so adding a progress bar for opening wallet (it is rescan in the end) would be a very nice feature.

Moved from: https://github.com/bitcoin/bitcoin/issues/25531.

PR #738, opened 2023-06-13
Add menu option to migrate a wallet

achow101:
GUI users need to be able to migrate wallets without going to the RPC console.

### DIFF
--- a/src/qt/walletcontroller.cpp
+++ b/src/qt/walletcontroller.cpp
@@ -432,3 +432,67 @@ void RestoreWalletActivity::finish()
 
     Q_EMIT finished();
 }
+
+void MigrateWalletActivity::migrate(WalletModel* wallet_model)
+{
+    // Warn the user about migration
+    QMessageBox box(m_parent_widget);
+    box.setWindowTitle(tr("Migrate wallet"));
+    box.setText(tr("Are you sure you wish to migrate the wallet <i>%1</i>?").arg(GUIUtil::HtmlEscape(wallet_model->getDisplayName())));
+    box.setInformativeText(tr("Migrating the wallet will convert this wallet to one or more descriptor wallets. A new wallet backup will need to be made.\n"
+                "If this wallet contains any watchonly scripts, a new wallet will be created which contains those watchonly scripts.\n"
+                "If this wallet contains any solvable but not watched scripts, a different and new wallet will be created which contains those scripts.\n\n"
+                "The migration process will create a backup of the wallet before migrating. This backup file will be named "
+                "<wallet name>-<timestamp>.legacy.bak and can be found in the directory for this wallet. In the event of "
+                "an incorrect migration, the backup can be restored with the \"Restore Wallet\" functionality."));
+    box.setStandardButtons(QMessageBox::Yes|QMessageBox::Cancel);
+    box.setDefaultButton(QMessageBox::Yes);
+    if (box.exec() != QMessageBox::Yes) return;
+
+    // Get the passphrase if it is encrypted regardless of it is locked or unlocked. We need the passphrase itself.
+    SecureString passphrase;
+    WalletModel::EncryptionStatus enc_status = wallet_model->getEncryptionStatus();
+    if (enc_status == WalletModel::EncryptionStatus::Locked || enc_status == WalletModel::EncryptionStatus::Unlocked) {
+        AskPassphraseDialog dlg(AskPassphraseDialog::Unlock, m_parent_widget, &passphrase);
+        dlg.setModel(wallet_model);
+        dlg.exec();
+    }
+
+    // GUI needs to remove the wallet so that it can actually be unloaded by migration
+    const std::string name = wallet_model->wallet().getWalletName();
+    m_wallet_controller->removeAndDeleteWallet(wallet_model);
+
+    showProgressDialog(tr("Migrate Wallet"), tr("Migrating Wallet <b>%1</b>…").arg(GUIUtil::HtmlEscape(name)));
+
+    QTimer::singleShot(0, worker(), [this, name, passphrase] {
+        auto res{node().walletLoader().migrateWallet(name, passphrase)};
+
+        if (res) {
+            m_success_message = tr("The wallet '%1' was migrated successfully.").arg(GUIUtil::HtmlEscape(res->wallet->getWalletName()));
+            if (res->watchonly_wallet_name) {
+                m_success_message += tr(" Watchonly scripts have been migrated to a new wallet named '%1'.").arg(GUIUtil::HtmlEscape(res->watchonly_wallet_name.value()));
+            }
+            if (res->solvables_wallet_name) {
+                m_success_message += tr(" Solvable but not watched scripts have been migrated to a new wallet named '%1'.").arg(GUIUtil::HtmlEscape(res->solvables_wallet_name.value()));
+            }
+            m_wallet_model = m_wallet_controller->getOrCreateWallet(std::move(res->wallet));
+        } else {
+            m_error_message = util::ErrorString(res);
+        }
+
+        QTimer::singleShot(0, this, &MigrateWalletActivity::finish);
+    });
+}
+
+void MigrateWalletActivity::finish()
+{
+    if (!m_error_message.empty()) {
+        QMessageBox::critical(m_parent_widget, tr("Migration failed"), QString::fromStdString(m_error_message.translated));
+    } else {
+        QMessageBox::information(m_parent_widget, tr("Migration Successful"), m_success_message);
+    }
+
+    if (m_wallet_model) Q_EMIT migrated(m_wallet_model);
+
+    Q_EMIT finished();
+}
--- a/src/wallet/interfaces.cpp
+++ b/src/wallet/interfaces.cpp
@@ -42,6 +42,7 @@ using interfaces::Wallet;
 using interfaces::WalletAddress;
 using interfaces::WalletBalances;
 using interfaces::WalletLoader;
+using interfaces::WalletMigrationResult;
 using interfaces::WalletOrderForm;
 using interfaces::WalletTx;
 using interfaces::WalletTxOut;
@@ -630,6 +631,18 @@ public:
         } else {
             return util::Error{error};
         }
+    }
+    util::Result<WalletMigrationResult> migrateWallet(const std::string& name, const SecureString& passphrase) override
+    {
+        auto res = wallet::MigrateLegacyToDescriptor(name, passphrase, m_context);
+        if (!res) return util::Error{util::ErrorString(res)};
+        WalletMigrationResult out{
+            .wallet = MakeWallet(m_context, res->wallet),
+            .watchonly_wallet_name = res->watchonly_wallet ? std::make_optional(res->watchonly_wallet->GetName()) : std::nullopt,
+            .solvables_wallet_name = res->solvables_wallet ? std::make_optional(res->solvables_wallet->GetName()) : std::nullopt,
+            .backup_path = res->backup_path,
+        };
+        return {std::move(out)}; // std::move to work around clang bug
     }
     std::string getWalletDir() override
     {
--- a/src/interfaces/wallet.h
+++ b/src/interfaces/wallet.h
@@ -50,6 +50,7 @@ struct WalletBalances;
 struct WalletTx;
 struct WalletTxOut;
 struct WalletTxStatus;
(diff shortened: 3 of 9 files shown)


src/wallet/interfaces.cpp
```diff
@@ -631,6 +632,18 @@ class WalletLoaderImpl : public WalletLoader
...
+    {
+        auto res = wallet::MigrateLegacyToDescriptor(name, passphrase, m_context);
+        if (!res) return util::Error{util::ErrorString(res)};
+        WalletMigrationResult out{
+            .wallet = MakeWallet(m_context, res->wallet),
+            .watchonly_wallet_name = res->watchonly_wallet ? std::make_optional(res->watchonly_wallet->GetName()) : std::nullopt,
+            .solvables_wallet_name = res->solvables_wallet ? std::make_optional(res->solvables_wallet->GetName()) : std::nullopt,
+            .backup_path = res->backup_path,
+        };
+        return out;
```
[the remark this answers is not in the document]
achow101: Done as suggested

hebasto:
Usually, we submit changes to non-`qt` subdirectory, i.e., the first commit of this PR, to the main repo, no?

hebasto:
> > Usually, we submit changes to non-qt subdirectory, i.e., the first commit of this PR, to the main repo, no?
> 
> I don't think this rule applies to interface changes that only affect the gui. Otherwise, it just seems like extra work for reviewers and authors to artificially rip stuff apart for no good reason, with the extra risk of ending up merging just one ripped-off half, but not the other?

I agree. It would be better to just document this case for further references.

hebasto:
Concept ACK.

hebasto:
cc @furszy @ryanofsky @S3RK

src/qt/walletcontroller.cpp
```diff
@@ -432,3 +432,67 @@ void RestoreWalletActivity::finish()
...
     Q_EMIT finished();
 }
+
+void MigrateWalletActivity::migrate(WalletModel* wallet_model)
+{
+    // Warn the user about migration
+    QMessageBox box(m_parent_widget);
+    box.setWindowTitle(tr("Migrate wallet"));
+    box.setText(tr("Are you sure you wish to migrate the wallet <i>%1</i>?").arg(GUIUtil::HtmlEscape(wallet_model->getDisplayName())));
+    box.setInformativeText(tr("Migrating the wallet will convert this wallet to a descriptor wallet. A new wallet backup will need to be made.\n"
```
S3RK, line 442: .. convert this wallet to **one or more descriptor wallets**

S3RK:
Concept ACK

Not qualified to review GUI code, but agree it's a good idea to show migration option in GUI.

src/qt/walletcontroller.cpp
achow101: Done as suggested.

kristapsk:
Concept ACK

src/qt/walletcontroller.cpp
```diff
@@ -432,3 +432,67 @@ void RestoreWalletActivity::finish()
...
 }
+
+void MigrateWalletActivity::migrate(WalletModel* wallet_model)
+{
+    // Warn the user about migration
+    QMessageBox box(m_parent_widget);
+    box.setWindowTitle(tr("Migrate wallet"));
+    box.setText(tr("Are you sure you wish to migrate the wallet <i>%1</i>?").arg(GUIUtil::HtmlEscape(wallet_model->getDisplayName())));
+    box.setInformativeText(tr("Migrating the wallet will convert this wallet to one or more descriptor wallets. A new wallet backup will need to be made.\n"
+                "If this wallet contains any watchonly those scripts, a new wallet will be created which contains those watchonly scripts.\n"
```
jarolrod, line 443: I do think this is quite wordy, and maybe more information than is needed. But, no need to reduce the amount of information here.

I know that for some users, they continue to use the Bitcoin Core GUI because their wallet was generated with it in some early version, and they are scared to use anything else but that. So, migrating can be a scary task.

When implementing this functionality for the gui-qml we'd probably make this less intimidating, perhaps as a wizard, and with more interaction/feedback with the process. CC @GBKS @mouxdesign

Screen Shot 2023-06-22 at 6 55 11 PM

src/qt/walletcontroller.cpp
```diff
@@ -432,3 +432,67 @@ void RestoreWalletActivity::finish()
...
+                "If this wallet contains any solvable but not watched scripts, a different new wallet will be created which contains those scripts.\n\n"
+                "The migration process will create a backup of the wallet before migrating. This backup file will be named "
+                "<wallet name>-<timestamp>.legacy.bak and can be found in the directory for this wallet. In the event of "
+                "an incorrect migration, the backup can be restored with the Restore Wallet functionality."));
```
jarolrod, line 447: maybe? for legibility.
```suggestion
                "an incorrect migration, the backup can be restored with the \"Restore Wallet\" functionality."));
```

src/qt/walletcontroller.cpp
```diff
@@ -432,3 +432,67 @@ void RestoreWalletActivity::finish()
...
+    box.setStandardButtons(QMessageBox::Yes|QMessageBox::Cancel);
+    box.setDefaultButton(QMessageBox::Yes);
+    if (box.exec() != QMessageBox::Yes) return;
+
+    // Get the passphrase if it is encrypted regardless of it is locked or unlocked. We need the passphrase itself.
+    SecureString passphrase;
+    WalletModel::EncryptionStatus enc_status = wallet_model->getEncryptionStatus();
```
jarolrod, line 454: This seems to always fail and the wallet gets removed, or am I the only one?

cc @johnny9 @promag

jarolrod (commented):
Concept ACK, important feature

src/qt/walletcontroller.cpp
johnny9: I am unable to migrate an encrypted wallet as well. The error message that appears for me is "Error: Wallet decryption failed, the wallet passphrase was not provided or was incorrect"

johnny9 (commented):
Concept ACK. Very useful feature.

Wallet migration doesn't work for me if I have a legacy sqlite wallet. Unsure if the assumption is that this only works with a bdb legacy wallet.

src/qt/walletcontroller.cpp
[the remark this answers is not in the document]
achow101: Done

src/qt/walletcontroller.cpp
```diff
@@ -432,3 +432,67 @@ void RestoreWalletActivity::finish()
...
+
+void MigrateWalletActivity::migrate(WalletModel* wallet_model)
+{
+    // Warn the user about migration
+    QMessageBox box(m_parent_widget);
+    box.setWindowTitle(tr("Migrate wallet"));
+    box.setText(tr("Are you sure you wish to migrate the wallet <i>%1</i>?").arg(GUIUtil::HtmlEscape(wallet_model->getDisplayName())));
+    box.setInformativeText(tr("Migrating the wallet will convert this wallet to one or more descriptor wallets. A new wallet backup will need to be made.\n"
+                "If this wallet contains any watchonly those scripts, a new wallet will be created which contains those watchonly scripts.\n"
+                "If this wallet contains any solvable but not watched scripts, a different new wallet will be created which contains those scripts.\n\n"
```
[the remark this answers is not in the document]
achow101: Done

src/qt/walletcontroller.cpp
achow101: Done

src/qt/walletcontroller.cpp
achow101: Fixed. Apparently `AskpassphraseDialog` didn't set `m_passphrase_out` for `Unlock`. I've added a commit to do that and migration of encrypted wallets now works.

D33r-Gee: this works now tested ACK on WSL Ubuntu 22.04

src/qt/walletcontroller.cpp
```diff
@@ -432,3 +432,67 @@ void RestoreWalletActivity::finish()
...
     Q_EMIT finished();
 }
+
+void MigrateWalletActivity::migrate(WalletModel* wallet_model)
+{
+    // Warn the user about migration
+    QMessageBox box(m_parent_widget);
+    box.setWindowTitle(tr("Migrate wallet"));
+    box.setText(tr("Are you sure you wish to migrate the wallet <i>%1</i>?").arg(GUIUtil::HtmlEscape(wallet_model->getDisplayName())));
+    box.setInformativeText(tr("Migrating the wallet will convert this wallet to one or more descriptor wallets. A new wallet backup will need to be made.\n"
```
jarolrod, line 442: Confirmed that all the new strings introduced here are nicely transferred over to be translated after running `make translate`

While not a blocker; the longer I look at this, the more I think it can be slimmed down. Will have another look tomorrow.

At least this is a chunky boy to translate, and will need a good translator comment:
```
<message>
        <location line="+1"/>
        <source>Migrating the wallet will convert this wallet to one or more descriptor wallets. A new wallet backup will need to be made.
If this wallet contains any watchonly scripts, a new wallet will be created which contains those watchonly scripts.
If this wallet contains any solvable but not watched scripts, a different and new wallet will be created which contains those scripts.

The migration process will create a backup of the wallet before migrating. This backup file will be named &lt;wallet name&gt;-&lt;timestamp&gt;.legacy.bak and can be found in the directory for this wallet. In the event of an incorrect migration, the backup can be restored with the &quot;Restore Wallet&quot; functionality.</source>
        <translation type="unfinished"></translation>
</message>
```

D33r-Gee (commented):
tested ACK on WSL Ubuntu 22.04
Migrated a few test wallets using the GUI on regtest, both encrypted and not-encrypted wallet yelled successful results.

jarolrod (approved):
ACK 48aae2cffeb91add75a70ac4d5075c38054452fa

This introduces the functionality as promised, works well, and didn't have any issues with legacy wallets + watch only addrs. 

Improving the text and it's translatability to can occur in a follow-up.

src/qt/walletcontroller.cpp
```diff
@@ -432,3 +432,67 @@ void RestoreWalletActivity::finish()
...
+        auto res{node().walletLoader().migrateWallet(name, passphrase)};
+
+        if (res) {
+            m_success_message = tr("The wallet '%1' was migrated successfully.").arg(GUIUtil::HtmlEscape(res->wallet->getWalletName()));
+            if (res->watchonly_wallet_name) {
+                m_success_message += tr(" Watchonly scripts have been migrated to a new wallet named '%1'.").arg(GUIUtil::HtmlEscape(res->watchonly_wallet_name.value()));
+            }
+            if (res->solvables_wallet_name) {
+                m_success_message += tr(" Solvable but not watched scripts have been migrated to a new wallet named '%1'.").arg(GUIUtil::HtmlEscape(res->solvables_wallet_name.value()));
+            }
```
pablomartin4btc, line 477: Perhaps we can refactor this piece a bit, maybe extracting the message building to other functions?

pablomartin4btc (commented):
tACK 48aae2cffeb91add75a70ac4d5075c38054452fa

Very nice feature. 

The menu option is disabled correctly if the selected wallet is not a Legacy one:

image

I had to `./configure --with-dbd` in order to create a Legacy wallet first:

image
Not related with this PR: Do I need to use an older version to make it "portable"? It's referring to the backup of a legacy wallet or the copy of the files?

And ofc this is something we already saw in previous versions of `bitcoin-qt` before the "Descriptor Wallet" became checked by default and disabled so user is forced to use descriptor instead of legacy wallets:

image

Once a Legacy wallet is selected we can see the new menu option "Migrate Wallet":

image

Regarding the long message, perhaps can be restructured a bit with some bullets points?

image

Perhaps as a follow-up we can show a proper progress being updated (we have already these open issues related with that: #651, #679):

image

It works as expected and I was able to restore the migrated wallet successfully.

image

One more detail I noticed is that the migration process ends up saving a `.bak` extension file, when I wanted to try the restore of it, I had to change that extension to `.dat`, otherwise it won't be seen by the open/ restore file dialog as there's only one type of file possible in the bottom right combo-box:
 
image

Should we add another type there or perhaps set the migrated file extension as `.dat` too? The filename itself is automatically defined, should we give the user the option to name it instead?

hebasto (approved):
ACK 48aae2cffeb91add75a70ac4d5075c38054452fa

Issue #69, opened 2020-08-22
Confusion with Create Wallet menu options


Rspigler:
This is in regards to https://github.com/bitcoin/bitcoin/pull/15450

If you check 'Disable Private Keys', is that not initially a Blank Wallet, until you import an xpub?
I don't understand what the difference would be between checking just 'Disable Private Keys', and checking 'Disable Private Keys' and 'Make Blank Wallet'.

Should checking 'Disable Private Keys' automatically check 'Make Blank Wallet'? (But not vice versa - Making a Blank Wallet with/without Disabling Private Keys affects the future ability to import.)

But then I am concerned that this would be very confusing for the end user.  Enabling 'Encrypt Wallet' already disables 'Disable Private Keys'.

Rspigler:
I like that a lot! (I had some difficulty in opening this in a DispVM, maybe paste screenshots/code here for people who don't want to leave github to review?)

Some typos in the 'Blank Wallet' descriptions; example "A blank wallet has will generate..."

In the 1.2 modal; are you able to click more than one wallet option? Should selecting one, grey out the others?

Rspigler:
Referencing this here:
https://github.com/bitcoin/bitcoin/pull/17879#issuecomment-578533079

Bosch-0:
I'll upload a screenshot once I have made some changes - though there are many UI/UX changes being made other than just the initial menu options so ill likely open a new issue. 

Fixed that typo cheers, no spellcheck on Figma - usually clean up those details at the end ;)

prayank23:
> I've got a WIP design related to this issue that more clearly explains each option, you can view it here: https://www.figma.com/file/FJ02rY3m8V9ZCDvoXjW39W/Bitcoin-Core?node-id=281%3A0

 @Bosch-0 This looks good.

![screenshot](https://i.imgur.com/HV6zEiU.jpg)

PR #739, opened 2023-06-13
Disable and uncheck blank when private keys are disabled

achow101:
Unify the GUI's create wallet with the RPC createwallet so that the blank flag is not set when private keys are disabled.

### DIFF
--- a/src/qt/createwalletdialog.cpp
+++ b/src/qt/createwalletdialog.cpp
@@ -58,10 +58,7 @@ CreateWalletDialog::CreateWalletDialog(QWidget* parent) :
         ui->descriptor_checkbox->setChecked(checked);
         ui->encrypt_wallet_checkbox->setChecked(false);
         ui->disable_privkeys_checkbox->setChecked(checked);
-        // The blank check box is ambiguous. This flag is always true for a
-        // watch-only wallet, even though we immedidately fetch keys from the
-        // external signer.
-        ui->blank_wallet_checkbox->setChecked(checked);
+        ui->blank_wallet_checkbox->setChecked(false);
     });
 
     connect(ui->disable_privkeys_checkbox, &QCheckBox::toggled, [this](bool checked) {
@@ -69,9 +66,10 @@ CreateWalletDialog::CreateWalletDialog(QWidget* parent) :
         // set to true, enable it when isDisablePrivateKeysChecked is false.
         ui->encrypt_wallet_checkbox->setEnabled(!checked);
 
-        // Wallets without private keys start out blank
+        // Wallets without private keys cannot set blank
+        ui->blank_wallet_checkbox->setEnabled(!checked);
         if (checked) {
-            ui->blank_wallet_checkbox->setChecked(true);
+            ui->blank_wallet_checkbox->setChecked(false);
         }
 
         // When the encrypt_wallet_checkbox is disabled, uncheck it.
@@ -81,8 +79,11 @@ CreateWalletDialog::CreateWalletDialog(QWidget* parent) :
     });
 
     connect(ui->blank_wallet_checkbox, &QCheckBox::toggled, [this](bool checked) {
-        if (!checked) {
-          ui->disable_privkeys_checkbox->setChecked(false);
+        // Disable the disable_privkeys_checkbox when blank_wallet_checkbox is checked
+        // as blank-ness only pertains to wallets with private keys.
+        ui->disable_privkeys_checkbox->setEnabled(!checked);
+        if (checked) {
+            ui->disable_privkeys_checkbox->setChecked(false);
         }
     });
 


hebasto:
Concept ACK.

hebasto:
cc @furszy @ryanofsky @S3RK

src/qt/createwalletdialog.cpp
```diff
@@ -58,20 +58,18 @@ CreateWalletDialog::CreateWalletDialog(QWidget* parent) :
...
     });
 
     connect(ui->disable_privkeys_checkbox, &QCheckBox::toggled, [this](bool checked) {
         // Disable the encrypt_wallet_checkbox when isDisablePrivateKeysChecked is
         // set to true, enable it when isDisablePrivateKeysChecked is false.
         ui->encrypt_wallet_checkbox->setEnabled(!checked);
 
-        // Wallets without private keys start out blank
+        // Wallets without private keys cannot set blank
+        ui->blank_wallet_checkbox->setEnabled(!checked);
```
S3RK, line 70: If we disable `blank` checkbox when `disable_privkeys` is checked should we do the reverse?

S3RK:
> Unify the GUI's create wallet with the RPC createwallet so that the blank flag is not set when private keys are disabled.

Do I understand correctly that we want to make these two flags mutually exclusive?

Let me make sure I correctly understand the rational. I guess the state when both are `true` is not meaningful because `disable_privkeys` is more restrictive than `blank`.  There will be no keys generated and no keys could be imported, so there is no keys for the wallet software to manage and hence `blank` (aka manual) is not needed.

Assuming the above is correct, do we want also to make them exclusive in RPC? If I'm not mistaken one can set both today.

Also, IIUC existing wallets that set both are not a problem, because it's just redundant information and will not lead to unexpected behaviour. Am I correct in my understanding?

ryanofsky:
Concept ACK. This change seems like an improvement, and I think it is good to make it clear that when private keys are disabled, the "blank" option is not useful and effectively meaningless.

I think these options are still confusing, though, and instead of having checkboxes it would be better if you could choose what actual behavior you want:

1. Generate private keys
2. Do not generate private keys  (blank wallet)
3. Do not generate private keys and disallow them (watch-only wallet)

I notice there was a similar proposal in https://github.com/bitcoin-core/gui/issues/69#issuecomment-681179436:

[](https://github.com/bitcoin-core/gui/issues/69#issuecomment-681179436)

---

re: https://github.com/bitcoin-core/gui/pull/739#issuecomment-1600347050

> Do I understand correctly that we want to make these two flags mutually exclusive? [...]

Yes if the disable private keys option is checked the blank option is basically useless, so it's confusing to leave it as option that can be toggled on an off but effectively doesn't do anything.

> Assuming the above is correct, do we want also to make them exclusive in RPC? If I'm not mistaken one can set both today.

RPC is a lower level interface than the GUI and exposes both flags directly, so maybe it could warn when blank flag is set uselessly, but as an API I don't think it should make an error out of a configuration that makes sense but is a little redundant, or that it should break code that worked previously.

Also settings the flags independently through RPC can have some effect and may not be completely useless. For example, the blank flag was introduced in a later software release than the disable private keys flag, so setting the blank flag prevents the wallet from opening the wallet in older software. Also, IIUC, it is possible in RPC interface to disable the disable private keys option later, so maybe setting the blank flag could be useful if you anticipate doing that.


> 
> Also, IIUC existing wallets that set both are not a problem, because it's just redundant information and will not lead to unexpected behaviour. Am I correct in my understanding?

hernanmarino (commented):
Concept ACK

pablomartin4btc (commented):
Tested ACK.

I think there's a concise description on identifying these "disable private keys" and "blank wallet" concepts by @achow101 [here](https://github.com/bitcoin-core/gui/issues/69#issuecomment-678691104). 

I like the approach on replacing the checkboxes by radiobuttons proposed in #69 and bringing it here by [@ryanofsky](https://github.com/bitcoin-core/gui/pull/739#issuecomment-1600761925), I also like the fact having the description under each optiion, and more practical (removing the other 2 checkboxes that are disabled already: "external signer" in `false` and "descriptor wallet" in `true`) it seems clearer, wouldn't it be on a different PR?.

On the RPC side of this, I find this comment from @ryanofsky very useful:
> Also setting the flags independently through RPC can have some effect and may not be completely useless. For example, the blank flag was introduced in a later software release than the disable private keys flag, so setting the blank flag prevents the wallet from opening the wallet in older software. Also, IIUC, it is possible in RPC interface to disable the disable private keys option later, so maybe setting the blank flag could be useful if you anticipate doing that.

src/qt/createwalletdialog.cpp
achow101: Done

S3RK:
I'm nor a GUI code expert neither a UX (user experience) expert, so take all of this with a healthy skepticism.
The code changes look good to me, but I'm unsure about whether this is a better UX.
Code review ACK 9ea31eba04ff8dcb9d7d993ce28bb10731a35177 

I'm unsure if it's a clear win for UX because there's no hint to the user why the checkboxes change their state. I also like the idea with radio buttons and I'd be happy to review a new PR or re-review this one though.

Maybe change the description of the PR, it seems like we actually want to keep the behaviour of RPC and GUI somewhat different, so the unification part is confusing.

ryanofsky:
@S3RK do you think the UI changes here are actually worse, or just not an improvement? I agree the changes aren't ideal and aren't a clear win, but I think they don't make it worse necessarily. This PR disables the "make blank wallet" checkbox when "disable private keys" checkbox is enabled, instead of just automatically checking it. I think disabling the checkbox makes sense because it makes it clear than once private keys are disabled, the choice of whether to have a blank wallet is moot because the wallet will necessarily be empty.

On the other hand, maybe the current UI makes it clearer that disabling private keys implies having a blank wallet, because it automatically checks blank wallet when you check disable private keys, and automatically unchecks disable private keys when you uncheck blank wallet. I guess both UI's seem reasonable to me, even if neither is ideal.

I do agree that PR description could be clearer, and could describe what the UI change is, along with mentioning that the leaving the blank flag unset when the disable private keys flag is set makes GUI behavior match default RPC behavior.

S3RK:
@ryanofsky no no, I don't think this change is making it worse.

From my perspective, it's a in improvement for core developers, because it provides clarity on the wallet flags usage.
UX wise it's ~0 as the flags are implementation details. Both pre- and post-PR whether blank checkbox is disabled or automatically checked is equally not explained to the users.

jarolrod (commented):
ACK 9ea31eba04ff8dcb9d7d993ce28bb10731a35177

This is correct, and an improvement.

Visual and UX improvements on wallet creation can be handled for a follow-up here if someone wants to take it. The issue around UX of wallet creation will be tackled in https://github.com/bitcoin-core/gui-qml

src/qt/createwalletdialog.cpp
```diff
@@ -81,8 +79,11 @@ CreateWalletDialog::CreateWalletDialog(QWidget* parent) :
...
 
     connect(ui->blank_wallet_checkbox, &QCheckBox::toggled, [this](bool checked) {
-        if (!checked) {
-          ui->disable_privkeys_checkbox->setChecked(false);
+        // Disable the disable_privkeys_checkbox when blank_wallet_checkbox is checked
+        // as blank-ness only pertains to wallets with private keys.
+        ui->disable_privkeys_checkbox->setEnabled(!checked);
+        if (checked) {
+            ui->disable_privkeys_checkbox->setChecked(false);
         }
```
pablomartin4btc, line 87: nit:
```suggestion
        if (checked) ui->disable_privkeys_checkbox->setChecked(false);
```

src/qt/createwalletdialog.cpp
```diff
@@ -58,20 +58,18 @@ CreateWalletDialog::CreateWalletDialog(QWidget* parent) :
...
         // set to true, enable it when isDisablePrivateKeysChecked is false.
         ui->encrypt_wallet_checkbox->setEnabled(!checked);
 
-        // Wallets without private keys start out blank
+        // Wallets without private keys cannot set blank
+        ui->blank_wallet_checkbox->setEnabled(!checked);
         if (checked) {
-            ui->blank_wallet_checkbox->setChecked(true);
+            ui->blank_wallet_checkbox->setChecked(false);
         }
```
pablomartin4btc, line 73: nit:
```suggestion
        if (checked) ui->blank_wallet_checkbox->setChecked(false);
```

pablomartin4btc (approved):
tACK 9ea31eba04ff8dcb9d7d993ce28bb10731a35177

PR #740, opened 2023-06-21
Show own outputs on PSBT signing window

### LINKED ISSUE #732: Sign PSBT: Can't verify change output

### Current behaviour

When I load a PSBT from file to sign it, the dialog that appears looks like this:

Screenshot from 2023-05-19 10-02-22

There's no way for me to know that the second output belongs to me (it does). Without knowing this I'd feel really uncomfortable signing the PSBT.

### Expected behaviour

The change output should be marked as belonging to me.

### Steps to reproduce

Open a PSBT file that has one output for the payee (not yours) and one output to you (as change)

### Relevant log output

_No response_

Pre-built binaries

25.0rc2

Debian 11

### Machine specifications

_No response_

hernanmarino:
This fixes https://github.com/bitcoin-core/gui/issues/732 .
It allows you to identify your own addresses in the outputs of a transaction in the PSBT signing window. This enables easy identification of change outputs, and prevents certain attacks where someone (co-signers of a multisig, or others ) might trick you into signing a transaction while they are stealing the change, since prior to this modification there was no easy way of knowing this.

The identification of the output is similar to the way this is done in the transaction details window. 

A sample output is :

image

### DIFF
--- a/src/qt/psbtoperationsdialog.cpp
+++ b/src/qt/psbtoperationsdialog.cpp
@@ -182,6 +182,8 @@ std::string PSBTOperationsDialog::renderTransaction(const PartiallySignedTransac
         tx_description.append(tr(" * Sends %1 to %2")
             .arg(BitcoinUnits::formatWithUnit(BitcoinUnit::BTC, out.nValue))
             .arg(QString::fromStdString(EncodeDestination(address))));
+        // Check if the address is one of ours
+        if (m_wallet_model != nullptr && m_wallet_model->wallet().txoutIsMine(out)) tx_description.append(" (" + tr("own address") + ")");
         tx_description.append("<br>");
     }
 


pablomartin4btc (commented):
Tested ACK on signet, both, @kallerosenbaum's use case (where you have a watch-only wallet "online" and you sign the tx "offline", then you will need to broadcast it ofc) and a multisig wallet case mentioned by @hernanmarino in the description of this PR, work as expected.
Perhaps it would be more noticeable if the colour of the added text "(own address)" is changed?
Also, I couldn't find any test using `txoutIsMine()`, although there are for `IsMine()`.

edit: forgot to mention, for the multisig wallet use case, one should create it with the own private key and the public key/ s from the other signers, so when we select it and check the PBST it will be shown the "own address" text in the tx window.

bitcoinfinancier (approved):

jarolrod (commented):
ACK 4da243ba023f2987e97fc62886c6ebc70d6ee50a

Tested for functionality, and confirmed new translation added here shows up when building translations.

achow101:
ACK 4da243ba023f2987e97fc62886c6ebc70d6ee50a

PR #742, opened 2023-06-29
Exit and show error if unrecognized command line args are present

### LINKED ISSUE #741: Command line options after any non-hyphen arguments are silently ignored

### Current behaviour

For example, invoking `bitcoin-qt -server=1 foo -regtest` on a fresh install will run `mainnet` instead of the expected `regtest`.

### Expected behaviour

I would expect it to do one of the following:

1. Run `regtest`
2. Warn me that my arguments contained a "command" and any subsequent options were ignored
3. Exit with an error message

The problem is that `ArgsManager::ParseParameters` interprets any non-hypen ('-' or '--') arguments as "commands", with later arguments being assigned to that command.

This is not an issue with `bitcoind` since it has additional logic to detect "loose" commands and refuses to start:

https://github.com/bitcoin/bitcoin/blob/c6287faae4c0e705a9258a340dfcf548906f12af/src/bitcoind.cpp#L127-L132

### Steps to reproduce

Invoke `bitcoin-qt foo -regtest` on a fresh install

### Relevant log output

_No response_

Compiled from source

master@d6ee03507f39

MacOS Ventura 13.5

### Machine specifications

_No response_

john-moffett:
Fixes https://github.com/bitcoin-core/gui/issues/741 

Starting bitcoin-qt with non-hyphen ("-") arguments causes it to silently ignore any later valid options. For instance, invoking `bitcoin-qt -server=1 foo -regtest` on a fresh install will run `mainnet` instead of `regtest`.

This change makes the client exit with an error message if any such "loose" arguments are encountered. This mirrors how `bitcoind` handles it:

https://github.com/bitcoin/bitcoin/blob/c6287faae4c0e705a9258a340dfcf548906f12af/src/bitcoind.cpp#L127-L132

However, BIP-21 `bitcoin:` payment URIs are still allowed, but only if they're not followed by any additional options.

### DIFF
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -543,6 +543,34 @@ int GuiMain(int argc, char* argv[])
         return EXIT_FAILURE;
     }
 
+    // Error out when loose non-argument tokens are encountered on command line
+    // However, allow BIP-21 URIs only if no options follow
+    bool payment_server_token_seen = false;
+    for (int i = 1; i < argc; i++) {
+        QString arg(argv[i]);
+        bool invalid_token = !arg.startsWith("-");
+#ifdef ENABLE_WALLET
+        if (arg.startsWith(BITCOIN_IPC_PREFIX, Qt::CaseInsensitive)) {
+            invalid_token &= false;
+            payment_server_token_seen = true;
+        }
+#endif
+        if (payment_server_token_seen && arg.startsWith("-")) {
+            InitError(Untranslated(strprintf("Options ('%s') cannot follow a BIP-21 payment URI", argv[i])));
+            QMessageBox::critical(nullptr, PACKAGE_NAME,
+                                  // message cannot be translated because translations have not been initialized
+                                  QString::fromStdString("Options ('%1') cannot follow a BIP-21 payment URI").arg(QString::fromStdString(argv[i])));
+            return EXIT_FAILURE;
+        }
+        if (invalid_token) {
+            InitError(Untranslated(strprintf("Command line contains unexpected token '%s', see bitcoin-qt -h for a list of options.", argv[i])));
+            QMessageBox::critical(nullptr, PACKAGE_NAME,
+                                  // message cannot be translated because translations have not been initialized
+                                  QString::fromStdString("Command line contains unexpected token '%1', see bitcoin-qt -h for a list of options.").arg(QString::fromStdString(argv[i])));
+            return EXIT_FAILURE;
+        }
+    }
+
     // Now that the QApplication is setup and we have parsed our parameters, we can set the platform style
     app.setupPlatformStyle();
 
--- a/src/qt/paymentserver.h
+++ b/src/qt/paymentserver.h
@@ -54,6 +54,8 @@ class QLocalServer;
 class QUrl;
 QT_END_NAMESPACE
 
+extern const QString BITCOIN_IPC_PREFIX;
+
 class PaymentServer : public QObject
 {
     Q_OBJECT


john-moffett:
I'm not entirely sure why they weren't the same. In the beginning, `bitcoind` would handle RPC commands. Then QT got special-cased so that you [couldn't do that](https://github.com/theuni/bitcoin/commit/b2d1129f27a499fa00ab8f5b8f0e9f926ae66362) for `bitcoin-qt` -- though it wouldn't give any errors. It would just silently ignore the subsequent options.

Then the argument-handling code split entirely [here](https://github.com/bitcoin/bitcoin/pull/690). Now `bitcoind` and `bitcoin-qt` mostly did their own thing. 

Subsequently, the ability to call RPC methods from `bitcoind`'s command line was [removed](https://github.com/bitcoin/bitcoin/commit/b750cf1fb9b6529ab1761b22947677ca78a2c626), and "loose" (non-hyphen) arguments would result in errors.

I can't see any good reason for "loose" arguments to remain in `bitcoin-qt`. 

The "ignored options" issue was actually first raised [here](https://github.com/bitcoin/bitcoin/issues/1044). The [fix](https://github.com/bitcoin/bitcoin/pull/13112) didn't address the specific issue exactly, as it just caught unrecognized *options* (ie - arguments with a hyphen). But by then, as I [mentioned](https://github.com/bitcoin/bitcoin/commit/b750cf1fb9b6529ab1761b22947677ca78a2c626), the "loose" argument problem had already been fixed in `bitcoind`.

achow101:
I think this is needed to handle BIP 21 URIs. The "loose" argument is interpreted as the URI, although ignored if they could not be recognized as one. Anything that follows it is explicitly ignored to avoid argument injection, see https://achow101.com/2021/02/0.18-uri-vuln.

john-moffett:
Ah, thanks @achow101. I think this may still be salvageable by making an exception for just those arguments. 

https://github.com/bitcoin-core/gui/blob/561915f35f4f75365c78df939b68c9062d3d3581/src/qt/paymentserver.cpp#L83

Will revisit tomorrow.

hebasto:
Is it safer and simpler to just document that a payment URi must follow all command line options? In `bitcoin-qt -help`?

john-moffett:
I could modify this PR to enforce that behavior and add the documentation to `bitcoin-qt -help`. Right now, the PR only allows non-option arguments which must start with `bitcoin:` (case insensitive). It will still ignore any options after that.

I think leaving the current `master` behavior and just adding documentation to reflect it would be insufficient. My worry is that a user could be fooled by something like:

Scammer: *"Can you send me some testnet coins? Start your client with `bitcoin-qt select-chain -testnet`. You should have the same number of testnet coins as you have on the real network. Just send me some, please."* 

It's also just an annoyance. I spent an embarrassingly long time trying to figure out why `bitcoin-qt datadir=bitcoindatadir -regtest` kept opening mainnet before I realized I omitted the hyphen on `datadir`.

john-moffett:
Modified to follow @hebasto 's idea of only allowing `bitcoin:` URIs if no other options follow.

hernanmarino (approved):
tested ACK 51e4dc49f5335b5bae6c14606d1cc653a08a96b1

pablomartin4btc:
Tested ACK.

I like both features the error shown when invalid args are passed and the error for options that are passed following a BIP21.

For the last one and regarding the `-regtest` being ignored in your original PR description, I detected a similar scenario in `bitcoin-cli` a few months ago ([#26990](https://github.com/bitcoin/bitcoin/pull/26990) - still in progress) where `[options]` passed after a command that receives arguments are not detected by the `ArgsManager`, they are being interpreted as another arg of the last command.

Last thing, not part of the scope of this PR, I think we could validate the BIP21 URI [grammar](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki#user-content-ABNF_grammar) before opening `bitcoin-qt`.

pablomartin4btc (commented):
tACK 51e4dc49f5335b5bae6c14606d1cc653a08a96b1

I like both features the error shown when invalid args are passed and the error for options that are passed following a BIP21.

For the last one and regarding the -regtest being ignored in your original PR description, I detected a similar scenario in bitcoin-cli a few months ago (https://github.com/bitcoin/bitcoin/pull/26990 - still in progress) where [options] passed after a command that receives arguments are not detected by the ArgsManager, they are being interpreted as another arg of the last command.

Last thing, not part of the scope of this PR, I think we could validate the BIP21 URI [grammar](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki#user-content-ABNF_grammar) before opening bitcoin-qt.

src/qt/bitcoin.cpp
```diff
@@ -543,6 +543,34 @@ int GuiMain(int argc, char* argv[])
...
+#ifdef ENABLE_WALLET
+        if (arg.startsWith(BITCOIN_IPC_PREFIX, Qt::CaseInsensitive)) {
+            invalid_token &= false;
+            payment_server_token_seen = true;
+        }
+#endif
+        if (payment_server_token_seen && arg.startsWith("-")) {
+            InitError(Untranslated(strprintf("Options ('%s') cannot follow a BIP-21 payment URI", argv[i])));
+            QMessageBox::critical(nullptr, PACKAGE_NAME,
+                                  // message cannot be translated because translations have not been initialized
```
luke-jr, line 561: Should we just initialize them in this codepath?

src/qt/bitcoin.cpp
```diff
@@ -543,6 +543,34 @@ int GuiMain(int argc, char* argv[])
...
+        QString arg(argv[i]);
+        bool invalid_token = !arg.startsWith("-");
+#ifdef ENABLE_WALLET
+        if (arg.startsWith(BITCOIN_IPC_PREFIX, Qt::CaseInsensitive)) {
+            invalid_token &= false;
+            payment_server_token_seen = true;
+        }
+#endif
+        if (payment_server_token_seen && arg.startsWith("-")) {
+            InitError(Untranslated(strprintf("Options ('%s') cannot follow a BIP-21 payment URI", argv[i])));
```
luke-jr, line 559:
```suggestion
            InitError(Untranslated(strprintf("Options (eg, '%s') cannot follow a URI", argv[i])));
```

PR #744, opened 2023-06-30
Drop macOS ForceActivation workaround

fanquake:
Discussion in https://github.com/bitcoin/bitcoin/pull/16720 seems to point to this being no-longer needed after qt 5.6+. 
We now require 5.11.x+.

### DIFF
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -80,7 +80,6 @@
 
 #include <QProcess>
 
-void ForceActivation();
 #endif
 
 using namespace std::chrono_literals;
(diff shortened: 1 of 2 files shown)


hebasto:
Concept ACK.

> We now require 5.11.x+.

When building on macOS, Homebrew provides Qt [5.15.10](https://formulae.brew.sh/formula/qt@5). Our depends does 5.15.5.

cc @Sjors

hebasto:
I think we can drop `-DOBJC_OLD_DISPATCH_PROTOTYPES=0` now. At least, the code compiles for me natively on macOS Ventura (arm64) and when cross-building on Ubuntu 23.04.

Sjors:
Tested 362e98937356a74a5b70d65348957edf05b1138a but it seems the workaround is still needed.

This was introduced in https://github.com/bitcoin/bitcoin/pull/14123 as a fix to https://github.com/bitcoin/bitcoin/issues/13829.

With this PR, built on macOS 13.4.1 (Intel) with qt@5/5.15.10 via Homebrew, when I use `⌘ + H` to hide the window, and then e.g. click on "Receive" in the dock, the window doesn't appear. Without this PR the window _does_ appear.

Issue #748, opened 2023-08-13
`-min` does not minimize wallet loading dialog


achow101:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behaviour

A user on the Bitcoin stackexchange [reported](https://bitcoin.stackexchange.com/questions/119283/is-it-possible-to-run-bitcoin-core-gui-at-the-same-time-as-bitcoind) that starting `-min` still results in the `Loading wallets` dialog box appearing.

### Expected behaviour

`-min` should minimize all startup dialogs.

### Steps to reproduce

Start with `-min` with wallets automatically loaded (presumably, I haven't tried yet)

### How did you obtain Bitcoin Core

Pre-built binaries

### What version of Bitcoin Core are you using?

unknown

### Operating system and version

Some Linux with XFCE

hebasto:
Closed via 9561e497e9f6387b0173343ef686496efb515c51.

achow101:
Was able replicate in 26.0, needed to use the release built binaries. I guess something in the way we build qt is causing this?

hebasto:
Confirming that the issue still exists for the release/Guix built `bitcoin-qt`.

> I guess something in the way we build qt is causing this?

The static Qt in depends misses GTK+ support. However, I'm not sure if it is the cause of the issue.

---

Other issues with different behavior for release binaries:
- https://github.com/bitcoin-core/gui/issues/33
- https://github.com/bitcoin-core/gui/issues/639

PR #749, opened 2023-08-13
make '-min' minimize wallet loading dialog

furszy:
Simple fix for #748.

When '-min' is enabled, no loading dialog should
be presented on screen during startup.


pablomartin4btc (commented):
I've tested this change. Created several wallets in order to see the "Loading wallet..." dialog for a while.
Without your change in master when I run `bitcoin-qt -min` the dialog is not shown and the app opens as usual.
With your change when I run `bitcoin-qt -min`, the app gets minimized entirely, is that intentional?

hebasto (approved):
ACK 32db15450a9ef2a45de29b3b2ae60491a38edbd6, tested on Debian 11 + XFCE.

The bug did not reveal itself on my Ubuntu + GNOME.

PR #751, opened 2023-09-08
macOS, do not process actions during shutdown

furszy:
As the 'QMenuBar' is created without a parent window in MacOS, the app crashes when the user presses the shutdown button and, right after it, triggers any action in the menu bar.

This happens because the QMenuBar is manually deleted in the BitcoinGUI destructor but the events attached to it children actions are not disconnected, so QActions events such us the 'QMenu::aboutToShow' could try to access null pointers.

Instead of guarding every single QAction pointer inside the QMenu::aboutToShow slot, or manually disconnecting all registered events in the destructor, we can check if a shutdown was requested and discard the event.

The 'node' field is a ref whose memory is held by the main application class, so it is safe to use here. Events are disconnected prior destructing the main application object.

Furthermore, the 'MacDockIconHandler::dockIconClicked' signal can make the app crash during shutdown for the very same reason. The 'show()' call triggers the 'QApplication::focusWindowChanged' event, which is connected to the 'minimize_action' QAction, which is also part of the app menu bar, which could no longer exist.

Another cause of crashes stems from the shortcuts provided by the `appMenuBar` submenus during shutdown. For instance, executing actions like opening the information dialog (command + I) or the console dialog (command + T) lead to access null pointers. The second commit addresses and resolves these issues.
Basically, in the present setup, we create a parentless `appMenuBar` whose submenus `QActions` are connected to `qApp` events (the app's global instance). However, at the `BitcoinGUI` destructor, we manually destruct this object without properly disconnecting the events. This leaves `qApp` events, such as `focusWindowChanged`, tied to submenus' `QAction` pointers, which causes the application to crash when it attempts to access them.

Important Note:
This happened to me few times. The worst consequence was an inconsistent chain state during IBD. Which triggered a full "replay blocks" process on the next startup. Which was painfully slow.

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -864,6 +864,7 @@ void BitcoinGUI::createTrayIconMenu()
     // Note: On macOS, the Dock icon is used to provide the tray's functionality.
     MacDockIconHandler* dockIconHandler = MacDockIconHandler::instance();
     connect(dockIconHandler, &MacDockIconHandler::dockIconClicked, [this] {
+        if (m_node.shutdownRequested()) return; // nothing to show, node is shutting down.
         show();
         activateWindow();
     });
@@ -875,6 +876,8 @@ void BitcoinGUI::createTrayIconMenu()
         // See https://bugreports.qt.io/browse/QTBUG-91697
         trayIconMenu.get(), &QMenu::aboutToShow,
         [this, show_hide_action, send_action, receive_action, sign_action, verify_action, options_action, node_window_action, quit_action] {
+            if (m_node.shutdownRequested()) return; // nothing to do, node is shutting down.
+
             if (show_hide_action) show_hide_action->setText(
                 (!isHidden() && !isMinimized() && !GUIUtil::isObscured(this)) ?
                     tr("&Hide") :
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -239,7 +239,6 @@ BitcoinGUI::~BitcoinGUI()
         trayIcon->hide();
 #ifdef Q_OS_MACOS
     delete m_app_nap_inhibitor;
-    delete appMenuBar;
     MacDockIconHandler::cleanup();
 #endif
 
@@ -470,13 +469,7 @@ void BitcoinGUI::createActions()
 
 void BitcoinGUI::createMenuBar()
 {
-#ifdef Q_OS_MACOS
-    // Create a decoupled menu bar on Mac which stays even if the window is closed
-    appMenuBar = new QMenuBar();
-#else
-    // Get the main window's menu bar on other platforms
     appMenuBar = menuBar();
-#endif
 
     // Configure the menus
     QMenu *file = appMenuBar->addMenu(tr("&File"));


hernanmarino (approved):
Approach ACK . I cannot test this (not my OS) but changes seem straightforward

src/qt/bitcoingui.cpp
```diff
@@ -470,13 +469,8 @@ void BitcoinGUI::createActions()
 
 void BitcoinGUI::createMenuBar()
 {
-#ifdef Q_OS_MACOS
-    // Create a decoupled menu bar on Mac which stays even if the window is closed
-    appMenuBar = new QMenuBar();
-#else
     // Get the main window's menu bar on other platforms
```
luke-jr, line 472: This comment no longer makes sense

furszy: upps, thanks. Removed.

Aminkavoos (approved):

hebasto (approved):
ACK bae209e3879fa099302d3b211362c49bbbfbdd14.

I had some doubts regarding the last commit due to Qt [docs](https://doc.qt.io/qt-5/qmainwindow.html#menuBar):
> If you want all windows in a Mac application to share one menu bar, don't use this function to create it, because the menu bar created here will have this [QMainWindow](https://doc.qt.io/qt-5/qmainwindow.html) as its parent. Instead, you must create a menu bar that does not have a parent, which you can then share among all the Mac windows.

However, I tested this PR on macOS Monterey 12.6.9. It works flawlessly. The application menu is still available and works fine even the main window is closed and the "Node window" is opened.

hebasto:
cc @Sjors @jarolrod

RandyMcMillan:
utACK bae209e

Sjors:
Tested with bae209e3879fa099302d3b211362c49bbbfbdd14

Tested on (Intel) macOS Ventura 13.6, was able to reproduce various crashes.

I used the Load URL option in the File menu, which crashes without this PR. With this PR the dialog appears and then goes away when the app is done quitting, but no crash. Similiar result when opening the Information menu.

Another way to trigger a crash is to hover over the Open Wallet menu option. This PR doesn't fix that, so maybe it's a different bug?

The `m_node.shutdownRequested()` seems a simple enough fix.

I didn't study the second commit deeply. It might be good to point to the PR where it was introduced and retry whatever bug was being fixed at the time.

(in unrelated news, I'll be able to test Ventura for quite some time into the future, because my 2017 iMac doesn't support newer macOS versions)

furszy (commented):
> Another way to trigger a crash is to hover over the Open Wallet menu option. This PR doesn't fix that, so maybe it's a different bug?

Yeah. Good catch. Just to move forward here, I can fix that one in a quick follow-up.

> I didn't study the second commit deeply. It might be good to point to the PR where it was introduced and retry whatever bug was being fixed at the time.

The second commit actually fixes another set of crash causes.

Basically, can crash the app by running the shortcuts provided by the `appMenuBar` submenus during shutdown. For instance, executing actions like opening the information dialog (command + I) or the console dialog (command + T) lead to access null pointers.

This occurs because, In the present setup, we create a parentless `appMenuBar` whose submenus QActions are connected to qApp events (the app's global instance). However, at the BitcoinGUI destructor, we manually destruct this object without properly disconnecting the events. This leaves qApp events, such as focusWindowChanged, tied to submenus' QAction pointers, which causes the application to crash when it attempts to access them.

Just updated the PR description with this information too.

hebasto:
Backported to 25.1 in https://github.com/bitcoin/bitcoin/pull/28487.

PR #752, opened 2023-09-08
Modify command line help to show support for BIP21 URIs

hernanmarino:
While reviewing a different PR (see https://github.com/bitcoin-core/gui/pull/742 ) **hebasto** suggested that the help for bitcoin-qt should be updated to reflect the fact that bitcoin-qt supports an optional BIP21 URI parameter.

Since this reflects actual behaviour of bitcoin-qt and is independent of whether or not the other PR gets merged, I created this simple PR to fix the help message.

### DIFF
--- a/src/qt/utilitydialog.cpp
+++ b/src/qt/utilitydialog.cpp
@@ -58,7 +58,8 @@ HelpMessageDialog::HelpMessageDialog(QWidget *parent, bool about) :
         ui->helpMessage->setVisible(false);
     } else {
         setWindowTitle(tr("Command-line options"));
-        QString header = "Usage:  bitcoin-qt [command-line options]                     \n";
+        QString header = "Usage: bitcoin-qt [command-line options] [URI]\n\n"
+                         "Optional URI is a Bitcoin address in BIP21 URI format.\n";
         QTextCursor cursor(ui->helpMessage->document());
         cursor.insertText(version);
         cursor.insertBlock();


RandyMcMillan:
concept ACK

jarolrod (approved):
ACK b45658ad093b61e92e9b2edbfb7003bf544e52ab

src/qt/utilitydialog.cpp
```diff
@@ -58,7 +58,8 @@ HelpMessageDialog::HelpMessageDialog(QWidget *parent, bool about) :
         ui->helpMessage->setVisible(false);
     } else {
         setWindowTitle(tr("Command-line options"));
-        QString header = "Usage:  bitcoin-qt [command-line options]                     \n";
+        QString header = "Usage:  bitcoin-qt [command-line options] [BIP21 URI]                     \n\n"
```
luke-jr, line 61: What is the purpose of these extra spaces on the end? Do they need to be adjusted? Can't seem to figure it out from the git history either :/

luke-jr, line 61: Probably better to just leave it as `[URI]` here, and if BIP21 is mentioned, only do so in the explanation below.

pablomartin4btc: I think it can be removed, it doesn't make any difference. I couldn't find any other reference in the code of such a thing (extra spaces at the end of a line before an \n). I think it was some formatting logic left behind.

Regarding its historical origins, it seems that it was coming from the `bitcoind`/ `bitcoin-cli` helps and copied over ([#13872](https://github.com/bitcoin/bitcoin/pull/13872)).
You can still see it at some places (exactly the same 22 spaces):
image

pablomartin4btc: Why not the [complete syntax representation](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki#user-content-ABNF_grammar) of the BIP21 command?
```
 bitcoinurn     = "bitcoin:" bitcoinaddress [ "?" bitcoinparams ]
 bitcoinaddress = *base58
 bitcoinparams  = bitcoinparam [ "&" bitcoinparams ]
 bitcoinparam   = [ amountparam / labelparam / messageparam / otherparam / reqparam ]
 amountparam    = "amount=" *digit [ "." *digit ]
 labelparam     = "label=" *qchar
 messageparam   = "message=" *qchar
 otherparam     = qchar *qchar [ "=" *qchar ]
 reqparam       = "req-" qchar *qchar [ "=" *qchar ]

```

pablomartin4btc: Also, I'd remove 1 space after "Usage:", so the start of "bitcoind" matches the alignment of the options helps.
```suggestion
        QString header = "Usage: bitcoin-qt [command-line options] [BIP21 URI]                     \n\n"
```

pablomartin4btc (commented):
tested ACK.

image

Added some obs on @luke-jr's comments.

src/qt/utilitydialog.cpp
hernanmarino: @pablomartin4btc I think this is not the place to explain BIP21, and I'll follow @luke-jr 's advice on this one.

hernanmarino: I agree with @luke-jr and @pablomartin4btc , I ll remove the extra spaces at the end, I see no reason to leave them. Regarding the help2man tool mentioned in https://github.com/bitcoin/bitcoin/pull/13872 , there is no reason for extra spaces.

hernanmarino:
Updated to address @luke-jr and @pablomartin4btc

pablomartin4btc (approved):
re ACK

src/qt/utilitydialog.cpp
```diff
@@ -58,7 +58,8 @@ HelpMessageDialog::HelpMessageDialog(QWidget *parent, bool about) :
         ui->helpMessage->setVisible(false);
     } else {
         setWindowTitle(tr("Command-line options"));
-        QString header = "Usage:  bitcoin-qt [command-line options]                     \n";
+        QString header = "Usage: bitcoin-qt [command-line options] [URI]\n\n"
+                         "Optional [URI] is a bitcoin address in BIP21 URI format.\n";
```
luke-jr, line 62: Bitcoin should be capitalized here

kristapsk, line 62: Square brackets mean that URI option is optional. Either use "[URI]" or "Optional URI" below, not "Optional [URI]".
```suggestion
                         "Optional URI is a bitcoin address in BIP21 URI format.\n";
```

hernanmarino: Done

hernanmarino: Done

hernanmarino:
Fixed the 2 typos mentioned by @luke-jr and @kristapsk . Also rebased for CI.

hernanmarino:
Fixed the 2 typos mentioned by @luke-jr and @kristapsk . Also rebased for CI.

kristapsk (approved):
utACK ede5014c445dcb40ddcfdede2c51236bbfe85f5e

pablomartin4btc (commented):
lgtm, re ACK ede5014c445dcb40ddcfdede2c51236bbfe85f5e

hebasto (approved):
ACK ede5014c445dcb40ddcfdede2c51236bbfe85f5e.

PR #754, opened 2023-09-11
Add BIP324-specific labels to peer details

hebasto:
This PR adds BIP324-specific labels to the peer details widget:
-  a transport version
- a session id

See: https://github.com/bitcoin/bitcoin/pull/28331#issuecomment-1693239025.

image

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -12,6 +12,7 @@
 #include <chainparams.h>
 #include <common/system.h>
 #include <interfaces/node.h>
+#include <node/connection_types.h>
 #include <qt/bantablemodel.h>
 #include <qt/clientmodel.h>
 #include <qt/guiutil.h>
@@ -514,8 +515,17 @@ RPCConsole::RPCConsole(interfaces::Node& node, const PlatformStyle *_platformSty
         /*: Explanatory text for a short-lived outbound peer connection that is used
             to request addresses from a peer. */
         tr("Outbound Address Fetch: short-lived, for soliciting addresses")};
-    const QString list{"<ul><li>" + Join(CONNECTION_TYPE_DOC, QString("</li><li>")) + "</li></ul>"};
-    ui->peerConnectionTypeLabel->setToolTip(ui->peerConnectionTypeLabel->toolTip().arg(list));
+    const QString connection_types_list{"<ul><li>" + Join(CONNECTION_TYPE_DOC, QString("</li><li>")) + "</li></ul>"};
+    ui->peerConnectionTypeLabel->setToolTip(ui->peerConnectionTypeLabel->toolTip().arg(connection_types_list));
+    const std::vector<QString> TRANSPORT_TYPE_DOC{
+        //: Explanatory text for "detecting" transport type.
+        tr("detecting: peer could be v1 or v2"),
+        //: Explanatory text for v1 transport type.
+        tr("v1: unencrypted, plaintext transport protocol"),
+        //: Explanatory text for v2 transport type.
+        tr("v2: BIP324 encrypted transport protocol")};
+    const QString transport_types_list{"<ul><li>" + Join(TRANSPORT_TYPE_DOC, QString("</li><li>")) + "</li></ul>"};
+    ui->peerTransportTypeLabel->setToolTip(ui->peerTransportTypeLabel->toolTip().arg(transport_types_list));
     const QString hb_list{"<ul><li>\""
         + ts.to + "\" – " + tr("we selected the peer for high bandwidth relay") + "</li><li>\""
         + ts.from + "\" – " + tr("the peer selected us for high bandwidth relay") + "</li><li>\""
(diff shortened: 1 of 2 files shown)


hebasto:
> I don't think we should add a lock icon or something like that...

A screenshot has been added to the PR description.

jarolrod (commented):
Concept ACK

Sjors:
No strong feelings about the lock :-) But can you add the session id?

theStack (commented):
Concept ACK

Will review this PR right after https://github.com/bitcoin/bitcoin/pull/28331.

hebasto:
> But can you add the session id?

Done.

Sjors:
tACK 43268b0a2eaab02b8a23cddf3b4c6eae9bb36afe (tested by cherry-picking the GUI commits)

src/qt/rpcconsole.cpp
```diff
@@ -514,8 +515,17 @@ RPCConsole::RPCConsole(interfaces::Node& node, const PlatformStyle *_platformSty
         /*: Explanatory text for a short-lived outbound peer connection that is used
             to request addresses from a peer. */
         tr("Outbound Address Fetch: short-lived, for soliciting addresses")};
-    const QString list{"<ul><li>" + Join(CONNECTION_TYPE_DOC, QString("</li><li>")) + "</li></ul>"};
-    ui->peerConnectionTypeLabel->setToolTip(ui->peerConnectionTypeLabel->toolTip().arg(list));
+    const QString connection_types_list{"<ul><li>" + Join(CONNECTION_TYPE_DOC, QString("</li><li>")) + "</li></ul>"};
+    ui->peerConnectionTypeLabel->setToolTip(ui->peerConnectionTypeLabel->toolTip().arg(connection_types_list));
+    const std::vector<QString> TRANSPORT_TYPE_DOC{
+        //: Explanatory text for "detecting" transport type.
+        tr("detecting: incoming connection; peer could be v1 or v2"),
```
theStack, line 522: In the latest version of https://github.com/bitcoin/bitcoin/pull/28331 (since the change in https://github.com/bitcoin/bitcoin/pull/28331#issuecomment-1721167073 was applied), the "detecting" state can appear in both directions, i.e. the description should be adapted accordingly:
```suggestion
        tr("detecting: peer could be v1 or v2"),
```

theStack (commented):
Tested this PR with both in- and outbound connections, and it seems to work just fine :ok_hand: 

It might be worthwhile to show the transport type also as extra column directly in the table rather than only in the details (useful for e.g. quickly identifying all v2 peers by sorting by type), but that's not urgent can probably wait until the first release that has set `-v2transport` to default-on.

hebasto:
Updated 43268b0a2eaab02b8a23cddf3b4c6eae9bb36afe -> f73f0411cf8a76e4272bd7e5c4583e9110d97d5a ([pr754.03](https://github.com/hebasto/gui/commits/pr754.03) -> [pr754.04](https://github.com/hebasto/gui/commits/pr754.04), [diff](https://github.com/hebasto/gui/compare/pr754.03..pr754.04)):

- addressed @theStack's [comment](https://github.com/bitcoin-core/gui/pull/754#discussion_r1335072653)

src/qt/rpcconsole.cpp
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui/pull/754#issuecomment-1732614324).

hebasto:
> It might be worthwhile to show the transport type also as extra column directly in the table rather than only in the details (useful for e.g. quickly identifying all v2 peers by sorting by type), but that's not urgent and can probably wait until the first release that has set `-v2transport` to default-on.

Considering the total number of columns, I believe we should provide users with the ability to hide selected ones.

hebasto:
Rebased on top of just merged https://github.com/bitcoin/bitcoin/pull/28331 and undrafted.

cc @Sjors @theStack @jarolrod

Sjors:
ACK 805273d043055d8c93fa7571fd4845a4b026f803

Tested on macOS Ventura 13.6 and lightly reviewed the code.

theStack (approved):
Tested ACK 805273d043055d8c93fa7571fd4845a4b026f803

Ran this branch on OpenBSD 7.3 with Xfce Window Manager, using Qt version 5.15.8. Checked that the new BIP324 labels are correct with two v2 connections (1 inbound + 1 outbound). :ok:

MarnixCroes (commented):
maybe only display `Session id` property when Transport is v2?

nit, everywhere in the [BIP](https://github.com/bitcoin/bips/blob/master/bip-0324.mediawiki) they use `Session ID` (capital ID), maybe use the same

hebasto:
> maybe only display `Session id` property when Transport is v2?
> 
> nit, everywhere in the [BIP](https://github.com/bitcoin/bips/blob/master/bip-0324.mediawiki?rgh-link-date=2023-10-04T07%3A48%3A39Z) they use `Session ID` (capital ID), maybe use the same

Thanks! [Done](https://github.com/hebasto/gui/compare/pr754.05..pr754.06).

achow101:
ACK d9c4e344d70bbf31ccb7162d83d4bd25762bc678

theStack (approved):
Tested re-ACK d9c4e344d70bbf31ccb7162d83d4bd25762bc678

RandyMcMillan:
ACK d9c4e34

MarnixCroes (approved):
tACK d9c4e344d70bbf31ccb7162d83d4bd25762bc678
nice!

jarolrod (commented):
post merge ack d9c4e344d70bbf31ccb7162d83d4bd25762bc678

PR #765, opened 2023-10-06
Fix wallet list hover crash on shutdown

furszy:
Small follow-up to #751.

Fixes another crash cause during shutdown. Which occurs when the user hovers over the wallets list.

Future Note:
This surely happen in other places as well, we should re-work the way we connect signals. Register
lambas without any precaution can leave dangling pointers.

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -392,7 +392,7 @@ void BitcoinGUI::createActions()
         connect(usedSendingAddressesAction, &QAction::triggered, walletFrame, &WalletFrame::usedSendingAddresses);
         connect(usedReceivingAddressesAction, &QAction::triggered, walletFrame, &WalletFrame::usedReceivingAddresses);
         connect(openAction, &QAction::triggered, this, &BitcoinGUI::openClicked);
-        connect(m_open_wallet_menu, &QMenu::aboutToShow, [this] {
+        connect(m_open_wallet_menu, &QMenu::aboutToShow, m_wallet_controller, [this] {
             m_open_wallet_menu->clear();
             for (const std::pair<const std::string, bool>& i : m_wallet_controller->listWalletDir()) {
                 const std::string& path = i.first;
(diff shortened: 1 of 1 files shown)


hebasto:
> Future Note: This surely happen in other places as well, we should re-work the way we connect signals. Register lambas without any precaution can leave dangling pointers.

Wouldn't be enough to use an `QObject::connect` overload with the additional `context` parameter, i.e.
```
QObject::connect(const QObject* sender, PointerToMemberFunction signal, const QObject* context, Functor functor, Qt::ConnectionType type)
```
instead of
```
QObject::connect(const QObject* sender, PointerToMemberFunction signal, Functor functor)
```

The former [assumes](https://doc.qt.io/qt-5/qobject.html#connect-5) that:
> The connection will automatically disconnect if the sender or the context is destroyed. However, you should take care that any objects used within the functor are still alive when the signal is emitted.

Also see a discussion in https://github.com/bitcoin-core/gui/pull/680.

furszy (commented):
> > Future Note: This surely happen in other places as well, we should re-work the way we connect signals. Register lambas without any precaution can leave dangling pointers.
> 
> Wouldn't be enough to use an `QObject::connect` overload with the additional `context` parameter, i.e.
> 
> ```
> QObject::connect(const QObject* sender, PointerToMemberFunction signal, const QObject* context, Functor functor, Qt::ConnectionType type)
> ```
> 
> instead of
> 
> ```
> QObject::connect(const QObject* sender, PointerToMemberFunction signal, Functor functor)
> ```
> 
> The former [assumes](https://doc.qt.io/qt-5/qobject.html#connect-5) that:
> 
> > The connection will automatically disconnect if the sender or the context is destroyed. However, you should take care that any objects used within the functor are still alive when the signal is emitted.
> 
> Also see a discussion in #680.

Tackled. Thanks.
I was a bit doubtful initially because we are manually deleting the `wallet_controller` field instead of letting QT do it for us.. but, thanks to the `QObject::destroyed` signal, this works fine.

Other than that, while testing the changes, I saw that the GUI shutdown process was being frozen when the user hovers over the non-triggerable top bar menu actions. So, the second commit also fixes this by clearing the menu (same as we do with the tray icon menu).

src/qt/bitcoingui.cpp
```diff
@@ -650,7 +650,8 @@ void BitcoinGUI::setClientModel(ClientModel *_clientModel, interfaces::BlockAndH
 
         m_mask_values_action->setChecked(_clientModel->getOptionsModel()->getOption(OptionsModel::OptionID::MaskValues).toBool());
     } else {
-        if(trayIconMenu)
+        // Shutdown requested, disable menus
+        if (trayIconMenu)
         {
```
hebasto, line 655: style nit:
```suggestion
        if (trayIconMenu) {
```

hebasto (approved):
ACK 8b6470a90652fcffc45b8d7998af7c8ad6251332, I've tested each commit separately on macOS Sonoma 14.0 (Apple M1).

hebasto:
Unfortunately, it introduced a regression. See https://github.com/bitcoin-core/gui/pull/770.

PR #766, opened 2023-10-06
gui: fix coin control input size accounting for taproot spends

theStack:
If manual coin control is used in the GUI, the input size accounting for P2TR is currently overshooting, as it still assumes P2WPKH (segwitv0) spends which have a larger witness, as ECDSA signatures are longer and the pubkey also has to be provided. Fix that by adding sizes depending on the witness version. Note that the total accounting including outputs is still off and there is some weird logic involved depending on whether SFFO is used, but it's (hopefully) a first step into the right direction.

### DIFF
--- a/src/qt/coincontroldialog.cpp
+++ b/src/qt/coincontroldialog.cpp
@@ -421,7 +421,19 @@ void CoinControlDialog::updateLabels(CCoinControl& m_coin_control, WalletModel *
         std::vector<unsigned char> witnessprogram;
         if (out.txout.scriptPubKey.IsWitnessProgram(witnessversion, witnessprogram))
         {
-            nBytesInputs += (32 + 4 + 1 + (107 / WITNESS_SCALE_FACTOR) + 4);
+            // add input skeleton bytes (outpoint, scriptSig size, nSequence)
+            nBytesInputs += (32 + 4 + 1 + 4);
+
+            if (witnessversion == 0) { // P2WPKH
+                // 1 WU (witness item count) + 72 WU (ECDSA signature with len byte) + 34 WU (pubkey with len byte)
+                nBytesInputs += 107 / WITNESS_SCALE_FACTOR;
+            } else if (witnessversion == 1) { // P2TR key-path spend
+                // 1 WU (witness item count) + 65 WU (Schnorr signature with len byte)
+                nBytesInputs += 66 / WITNESS_SCALE_FACTOR;
+            } else {
+                // not supported, should be unreachable
+                throw std::runtime_error("Trying to spend future segwit version script");
+            }
             fWitness = true;
         }
         else if(ExtractDestination(out.txout.scriptPubKey, address))


hebasto:
cc @achow101 @furszy @Sjors

src/qt/coincontroldialog.cpp
```diff
@@ -421,7 +421,19 @@ void CoinControlDialog::updateLabels(CCoinControl& m_coin_control, WalletModel *
         std::vector<unsigned char> witnessprogram;
         if (out.txout.scriptPubKey.IsWitnessProgram(witnessversion, witnessprogram))
         {
-            nBytesInputs += (32 + 4 + 1 + (107 / WITNESS_SCALE_FACTOR) + 4);
+            // add input skeleton bytes (outpoint, scriptSig size, nSequence)
+            nBytesInputs += (32 + 4 + 1 + 4);
```
MarnixCroes, line 425: bit of a nit: the outpoint is 36 [right](https://bitcoinops.org/en/tools/calc-size/)? maybe good to consolidate 32 and 4 or breakdown the _outpoint_ comment (into txid and vout index number) to prevent confusion?

MarnixCroes (commented):
nice catch.
Tested a bit, works as expected.

> Note that the total accounting including outputs is still off and there is some weird logic involved depending on whether SFFO is used, but it's (hopefully) a first step into the right direction.

Thanks for mentioning, I was a bit confused.

I'm not 100% sure about the exact number myself,
so cACK

src/qt/coincontroldialog.cpp
```diff
@@ -421,7 +421,19 @@ void CoinControlDialog::updateLabels(CCoinControl& m_coin_control, WalletModel *
...
+            if (witnessversion == 0) { // P2WPKH
+                // 1 WU (witness item count) + 72 WU (ECDSA signature with len byte) + 34 WU (pubkey with len byte)
+                nBytesInputs += 107 / WITNESS_SCALE_FACTOR;
+            } else if (witnessversion == 1) { // P2TR key-path spend
+                // 1 WU (witness item count) + 65 WU (Schnorr signature with len byte)
+                nBytesInputs += 66 / WITNESS_SCALE_FACTOR;
+            } else {
+                // not supported, should be unreachable
+                throw std::runtime_error("Trying to spend future segwit version script");
+            }
```
furszy, line 436: I don't think that throwing exception in the GUI for backend incompatibilities between releases is good. This will crash the app.
Instead, would be better to disable the fee, bytes, etc calculations and show "no information" in the labels when the version is unknown.

furszy (commented):
utACK 00a52e6394

src/qt/coincontroldialog.cpp
furszy: > I don't think it is supported to use a gui version of a different commit than the node?

Doesn't necessarily need to be a different commit. The supported witness version numbers are hardcoded above. If we add a new version, and for some reason the wallet opens (which is not supported atm), this will crash the app.

But, my comment was more general. Ideally, all these hardcoded size and fee calculations shouldn't be part of the gui code. The backend should provide the information and the gui should either present it or show something else denoting the "not enough information to calculate size/fee". Never throw an exception that crashes the app, gui developers shouldn't be aware of the wallet internals.

jarolrod (commented):
post merge ACK

PR #770, opened 2023-10-23
Revert "gui: provide wallet controller context to wallet actions"

hebasto:
The commit 7066e8996d0ac090535cc97cdcb54a219986460f from https://github.com/bitcoin-core/gui/pull/765 breaks "Open Wallet", "Close Wallet" and "Close All Wallets" items in the File menu (at least on Ubuntu 23.10 + Wayland).

Reverting it to avoid this regression in the 26.0 release.

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -392,7 +392,7 @@ void BitcoinGUI::createActions()
         connect(usedSendingAddressesAction, &QAction::triggered, walletFrame, &WalletFrame::usedSendingAddresses);
         connect(usedReceivingAddressesAction, &QAction::triggered, walletFrame, &WalletFrame::usedReceivingAddresses);
         connect(openAction, &QAction::triggered, this, &BitcoinGUI::openClicked);
-        connect(m_open_wallet_menu, &QMenu::aboutToShow, m_wallet_controller, [this] {
+        connect(m_open_wallet_menu, &QMenu::aboutToShow, [this] {
             m_open_wallet_menu->clear();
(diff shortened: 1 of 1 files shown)


hebasto:
cc @furszy @jarolrod @Sjors

MarnixCroes (approved):
tack f09bfab4aff04a9cc1ea157b94bfdae19f3465b1

furszy (commented):
Upps.. it is because `m_wallet_controller` is null when the events are connected (at widget creation time). We need to move the connection to `setWalletController()`, otherwise there is also a possible crash that could arise when any of these events is triggered prior to the wallet controller is set.

fanquake:
> Upps.. it is because m_wallet_controller is null when the events are connected (at widget creation time). We need to move the connection to setWalletController(), otherwise there is also a possible crash that could arise when any of these events is triggered prior to the wallet controller is set.

So this isn't actually platform specific, it was just always broken?

furszy:
> > Upps.. it is because m_wallet_controller is null when the events are connected (at widget creation time). We need to move the connection to setWalletController(), otherwise there is also a possible crash that could arise when any of these events is triggered prior to the wallet controller is set.
> 
> So this isn't actually platform specific, it was just always broken?

Since #765 merge, last week. Yes. It tell us how much we need to improve the GUI test coverage.
We fixed an issue and broke something else.

furszy (approved):

jarolrod (approved):
ACK f09bfab4aff04a9cc1ea157b94bfdae19f3465b1

The specified menu options don't lead to crash on macOS, but are unusable. This was an oversight in review of https://github.com/bitcoin-core/gui/pull/765. The specified menu actions work again with this branch.

PR #771, opened 2023-10-23
Avoid error-prone leading whitespace in translatable strings

hebasto:
While working on the GUI translation via Transifex web interface, I found it error-prone to have leading whitespace in translatable strings. This is because it is very easy to unintentionally drop them in translations unnoticed.

Fixed all current cases. Added a linter to prevent similar cases in the future.


furszy (commented):
utACK 856325f

PR #773, opened 2023-10-24
Check for private keys disabled before attempting unlock

### LINKED ISSUE #772: When using an unencrypted read-only wallet, pressing "Create Unsigned", shows "This operation needs you wallet passphrase to unlock the wallet"

### Current behaviour

When using an unencrypted, read-only (descriptor) wallet, when trying to make a new transaction and pressing the "Create Unsigned" button, a dialog is shown with the message "This operation needs you wallet passphrase to unlock the wallet". There is no wallet passphrase as I have never created it. It's also not possible to enter the empty passphrase as the submit button is grayed out.

### Expected behaviour

"Create Unsigned" creates an unsigned PSBT.

### Steps to reproduce

> When using an unencrypted, read-only (descriptor) wallet, when trying to make a new transaction and pressing the "Create Unsigned" button

### Relevant log output

no log output without debugging enabled

Package manager

v25.1

Guix

### Machine specifications

_No response_

achow101:
Before trying to unlock a wallet, first check if it has private keys disabled. If so, there is no need to unlock.

Note that such wallets are not expected to occur in typical usage. However bugs in previous versions allowed such wallets to be created, and so we need to handle them.

Fixes #772

For some additional context, see #631

### DIFF
--- a/src/qt/walletmodel.cpp
+++ b/src/qt/walletmodel.cpp
@@ -449,6 +449,13 @@ void WalletModel::unsubscribeFromCoreSignals()
 // WalletModel::UnlockContext implementation
 WalletModel::UnlockContext WalletModel::requestUnlock()
 {
+    // Bugs in earlier versions may have resulted in wallets with private keys disabled to become "encrypted"
+    // (encryption keys are present, but not actually doing anything).
+    // To avoid issues with such wallets, check if the wallet has private keys disabled, and if so, return a context
+    // that indicates the wallet is not encrypted.
+    if (m_wallet->privateKeysDisabled()) {
+        return UnlockContext(this, /*valid=*/true, /*relock=*/false);
+    }
     bool was_locked = getEncryptionStatus() == Locked;
     if(was_locked)
     {


src/qt/walletmodel.cpp
```diff
@@ -449,6 +449,13 @@ void WalletModel::unsubscribeFromCoreSignals()
 // WalletModel::UnlockContext implementation
 WalletModel::UnlockContext WalletModel::requestUnlock()
 {
+    // Bugs in earlier versions may have resulted in wallets with private keys disabled to become "encrypted"
+    // (encryption keys are present, but not actually doing anything).
+    // To avoid issues with such wallets, check if the wallet has private keys disabled, and if so, return a context
+    // that indicates the wallet is not encrypted.
+    if (m_wallet->privateKeysDisabled()) {
```
hebasto, line 456: It's a pity that `getEncryptionStatus() == NoKeys` does not work here.

Is it worth to warn the user when they load a watch-only wallet with an encryption keys?

hebasto (approved):
ACK 517c7f9cba306292e12e166b9dbc6c0838f05b27, I have reviewed the code and it looks OK.

Do we want to backport this PR into the 26.x branch?

src/qt/walletmodel.cpp
achow101: > It's a pity that `getEncryptionStatus() == NoKeys` does not work here.

It could, but #631 states that `NoKeys` is not returned for encrypted watchonly so that other users of `getEncryptionStatus()` get this status correctly. Perhaps that isn't necessary, but I wanted to keep this change targeted for just this specific issue and changing `getEncryptionStatus()`'s behavior could have unanticipated side effects.

> Is it worth to warn the user when they load a watch-only wallet with an encryption keys?

https://github.com/bitcoin/bitcoin/pull/28724 proposes to just delete them for the user on loading.

hebasto:
cc @furszy

furszy (commented):
Doesn't mean that the migration process will also be blocked when this happen? --> [look at this line](https://github.com/bitcoin/bitcoin/blob/2a349f9ea5de9f0157cd117289996e8640473a6d/src/qt/walletcontroller.cpp#L457)
`getEncryptionStatus()` should be returning `Locked`, which here would trigger the askpassphrase dialog in the same way described on #772.

Also, what is the expectation here, merge and backport this PR without backporting #28724? Because, after #28724, this isn't a problem anymore.

achow101:
> Doesn't mean that the migration process will also be blocked when this happen? --> [look at this line](https://github.com/bitcoin/bitcoin/blob/2a349f9ea5de9f0157cd117289996e8640473a6d/src/qt/walletcontroller.cpp#L457) `getEncryptionStatus()` should be returning `Locked`, which here would trigger the askpassphrase dialog in the same way described on #772.

That's a good point. Perhaps `getEncryptionStatus()` should just move to returning `NoKeys` for all private key disabled wallets.

> Also, what is the expectation here, merge and backport this PR without backporting #28724? Because, after #28724, this isn't a problem anymore.

28724 is a much scarier change since it deletes things from the wallet database and I wasn't sure if people were comfortable with that. If we move forward with it, then that can supersede this PR.

hebasto:
> > Also, what is the expectation here, merge and backport this PR without backporting #28724? Because, after #28724, this isn't a problem anymore.
> 
> 28724 is a much scarier change since it deletes things from the wallet database and I wasn't sure if people were comfortable with that. If we move forward with it, then that can supersede this PR.

I'm going to merge this to get this bugfix into 27.0 as https://github.com/bitcoin/bitcoin/pull/28724 remains untagged.

Issue #645, opened 2022-08-08
Bitcoin Core GUI getrawtransaction implementation


fanquake:
Bitcoin Core offers sovereignty by people controlling their keys and validating those keys. But many users today hold their keys on hardware wallets or mobile wallets and it's a bit difficult to connect those wallets to a full node and validate them. 

The cli users have the option to "getrawtransaction _txid_ true" in order to find valuable information about a transaction that could be theirs. This is a very powerful api call that effectively eliminates the need of an Electrum server.

I think many users would find a lot of value in easily verifying their personal transaction directly from Bitcoin Core GUI! 

Something like Help -> Verify external txid. There would need to be an option to enable txindex also. This would greatly increase the use of Bitcoin Core for the ease of validating a transaction. Sure users could create a watch only wallet but these are extra steps that maybe many of the non technical users would get discouraged from doing.

Moved from: https://github.com/bitcoin/bitcoin/issues/19161.

hebasto:
@zndtoshi @BrandonOdiwuor

Are there any compelling reasons not using RPC in the "Console" window?

PR #777, opened 2023-11-14
GUI getrawtransaction implementation

BrandonOdiwuor:
Fixes https://github.com/bitcoin-core/gui/issues/645.
Add `getrawtransaction` RPC to GUI on `Help -> Verify external txid.` tab

Screenshot from 2023-11-13 20-26-10

verbose off: `getrawtransaction txid false blockhash`
Screenshot from 2023-11-13 21-12-04

verbose on:  `getrawtransaction txid true blockhash`
Screenshot from 2023-11-13 21-11-47


willcl-ark:
Cool, that's pretty nice!

Just thinking out loud, but it does make me wonder if perhaps a new `Verify` or `Utils` button, in line with "Send | Recieve | Transactions" could be even nicer? If we added it there, we might consider adding a few more utils to the GUI like "Decode raw transaction", "Decode Script", "Get Descriptor Info", "Verify Message" etc. which folks who use the gui generally use other software for today...

BrandonOdiwuor:
> Cool, that's pretty nice!
> 
> Just thinking out loud, but it does make me wonder if perhaps a new `Verify` or `Utils` button, in line with "Send | Recieve | Transactions" could be even nicer? If we added it there, we might consider adding a few more utils to the GUI like "Decode raw transaction", "Decode Script", "Get Descriptor Info", "Verify Message" etc. which folks who use the gui generally use other software for today...

@willcl-ark would love to hear more opinions on this, I also think ` Help -> Verify external txid` tab might not work well with adding more utils

willcl-ark:
Yeah I don't want to derail this pr, which I think looks good on its own!

If there was desire to add more utils we could always move this later; I think this looks nice.

luke-jr (changes requested):
Needs a lot of UX work. Definitely doesn't belong on the Help menu...

src/qt/utilitydialog.cpp
```diff
@@ -27,82 +29,117 @@
...
+            ui->aboutMessage->setText(version + "<br><br>" + licenseInfoHTML);
+            ui->aboutMessage->setWordWrap(true);
+            ui->helpMessage->setVisible(false);
+            ui->labelTxId->setVisible(false);
+            ui->txidEdit->setVisible(false);
+            ui->labelBlockHash->setVisible(false);
+            ui->blockHashEdit->setVisible(false);
+            ui->submitButton->setVisible(false);
+            ui->verboseCheckbox->setVisible(false);
+            break;
```
pablomartin4btc, line 72: Since you are here (refactoring commit 10b3cd129e1fa8ece229a255d6fff89f419751c0) I'd extract all these settings into a function like "`initializeAboutModeWindow`" or something like that, same for the other modes, making the code a bit more maintainable and easier to follow.

pablomartin4btc (commented):
Concept ACK

As @luke-jr perhaps we can create a separate menu item like "Tools" or "Utils" with this feature in it (and would be more suitable if we need to add for example other features as @willcl-ark [mentioned](https://github.com/bitcoin-core/gui/pull/777#issuecomment-1810068057)).

Light CR.

hernanmarino (approved):
Concept ACK.

BrandonOdiwuor:
I have updated the PR based on the feedback given above:

- Added `Tools` menu
- Moved `getrawtransaction RPC` implementation to `Tools > verify external txid`

cc: @willcl-ark @luke-jr @pablomartin4btc

src/qt/forms/toolsdialog.ui
```diff
@@ -0,0 +1,151 @@
...
+    <layout class="QVBoxLayout" name="verticalLayout_2">
+     <item>
+      <layout class="QFormLayout" name="formLayout">
+       <property name="bottomMargin">
+        <number>12</number>
+       </property>
+       <item row="0" column="0">
+        <widget class="QLabel" name="labelTxId">
+         <property name="text">
+          <string>transaction id: </string>
```
pablomartin4btc, line 24: nit:
```suggestion
          <string>Transaction id: </string>
```
or `ID`?

src/qt/forms/toolsdialog.ui
```diff
@@ -0,0 +1,151 @@
...
+         </property>
+         <property name="checked">
+          <bool>false</bool>
+         </property>
+        </widget>
+       </item>
+       <item row="2" column="0">
+        <widget class="QLabel" name="labelBlockHash">
+         <property name="text">
+          <string>blockhash (optional): </string>
```
pablomartin4btc, line 54: nit:
```suggestion
          <string>Blockhash (optional): </string>
```

src/qt/bitcoingui.cpp
```diff
@@ -368,12 +369,17 @@ void BitcoinGUI::createActions()
     m_mask_values_action->setStatusTip(tr("Mask the values in the Overview tab"));
     m_mask_values_action->setCheckable(true);
 
+    getRawTransactionAction = new QAction(tr("&Verify external txid"), this);
```
pablomartin4btc, line 372: nit:
```suggestion
    getRawTransactionAction = new QAction(tr("&Verify transaction"), this);
```

Some thoughts: perhaps we use a more generic naming here and in the form we indicate if it's a wallet transaction with a checkbox (and then we change the form removing the blockhash option and call `gettransaction`?). Or maybe just on a follow-up.

pablomartin4btc (commented):
utACK 9659890f7c9223d85dda3d066c2e3de42cabe7cf

I agree that the new place for this PR's feature looks better under the new "Tools" menu.

Also code-wise it's better to have a new separate dialog (`toolsdialog`) rather than re-using the existing help dialog (`HelpMessageDialog`). Thinking it loud, I wonder if we need to specify a more specific name instead of the current generic one (what other use cases would open this `ToolsDialog`? Other RPC commands?). Another thoughts: how this new feature (and future ones?) integrates with the current RPC console? Should we also have some kind of link on the form to open the console with the execution of the rpc command and its output? Should we create a new sub-menu under "Tools" like "RPC commands"? These are some questions that don't need to be answered right now nor stopping this PR to go ahead and get merged, just to keep the focus on what directions we want to give to the UI and follow it up.

kristapsk:
Concept ACK

alfonsoromanz (approved):
Tested ACK https://github.com/bitcoin-core/gui/pull/777/commits/9659890f7c9223d85dda3d066c2e3de42cabe7cf

screen-recording

PR #780, opened 2023-12-07
Fix: Ensure 'Transaction View' remains disabled if no wallet is selected

pablomartin4btc:
This PR addresses an issue where, with no wallet selected, ticking on "Settings -> Mask values" checkbox twice enables the transaction tab when the checkbox is unticked. 

<details>
<summary>Current behavior display on master</summary>

Peek 2023-12-06 19-18

</details>

<details>
<summary>Correction display from this branch</summary>

Peek 2023-12-07 13-07

</details>

Note for maintaners: this PR should be backported to both 25.x and 26.x.

---

Originally this PR was disabling the "Mask Values" checkbox when no wallet was selected but since a reviewer pointed out that a user might want to open a wallet already on "privacy mode" I rolled that change out. 

<details>
<summary>Original correction  display disabling "Mask Values" </summary>

Peek 2023-12-06 19-11

</details>

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -673,8 +673,10 @@ void BitcoinGUI::setClientModel(ClientModel *_clientModel, interfaces::BlockAndH
 #ifdef ENABLE_WALLET
 void BitcoinGUI::enableHistoryAction(bool privacy)
 {
-    historyAction->setEnabled(!privacy);
-    if (historyAction->isChecked()) gotoOverviewPage();
+    if (walletFrame->currentWalletModel()) {
+        historyAction->setEnabled(!privacy);
+        if (historyAction->isChecked()) gotoOverviewPage();
+    }
 }
 
 void BitcoinGUI::setWalletController(WalletController* wallet_controller, bool show_loading_minimized)


pablomartin4btc:
> I'm not sure if this is the right approach: user might want to enable `Mask values` before opening his wallet?

Interesting point of view, I agree with it... Not sure if other wallet apps do it but ok, I'll remove that "fix" and perhaps other reviewers give their opinions on it. Thanks for the review!

pablomartin4btc:
> I'm not sure if this is the right approach: user might want to enable `Mask values` before opening his wallet?

I was just giving it a second thought and if we follow the original approach a user could still have the desired behaviour as the following:

<details>
<summary>check display here</summary>

Peek 2023-12-07 13-42

</details>

First time the user opens or creates a wallet can tick the checkbox on "Mask Values", this is also persisted, so next time the wallet will be on privacy mode. The fix was not changing the current value of the "Mask Values", so if a user updates QT to the version containing the fix it won't impact on their setup.

luke-jr (approved):
utACK

MarnixCroes:
> > I'm not sure if this is the right approach: user might want to enable `Mask values` before opening his wallet?
> 
> I was just giving it a second thought and if we follow the original approach a user could still have the desired behaviour as the following:
> check display here
> 
> First time the user opens or creates a wallet can tick the checkbox on "Mask Values", this is also persisted, so next time the wallet is open or selected will be on privacy mode. The fix was not changing the current value of the "Mask Values", so if a user updates QT to the version containing the fix it won't impact on their setup.

@pablomartin4btc yes correct it remembers if `Mask Values` is enabled or not.

Being only able to change it requires a wallet to be open, to me would feel like a bug, no?
that's more my "concern"

pablomartin4btc:
> Being only able to change it requires a wallet to be open, to me would feel like a bug, no? that's more my "concern"

Fair enough, thanks for sharing your insights.

alfonsoromanz (approved):
Tested ACK https://github.com/bitcoin-core/gui/pull/780/commits/b2e531e70a88f5c9e1c055ae7341520a3128e15d

Screen Recording 2023-12-22 at 18 24 11

pablomartin4btc:
Thanks @MarnixCroes, @luke-jr and @alfonsoromanz for reviewing!

BrandonOdiwuor (commented):
Concept ACK

hebasto (approved):
ACK b2e531e70a88f5c9e1c055ae7341520a3128e15d, tested on Ubuntu 22.04.

Issue #259, opened 2021-03-26
When opening or autoloading wallets there should be clear messages about rescanning in progress and wallets' names.


streamofstars:
There are at least two ways to open a wallet.
1. File, Open wallet
2. or by manually appending settings.json file with wallet name.

When opening a wallet that has not been synchronized recently, in the first case there is a small window displayed with a message "opening wallet <walletname>". The node is rescanning at this time but nothing really indicates that, so unless user peeks in the debug.log file they may think bitcoin core is stuck.

<img width="680" alt="Screenshot 2021-03-24 at 18 21 13" src="https://user-images.githubusercontent.com/22457855/112633972-50d02e00-8e3a-11eb-92af-a9c3bd223c70.png">

In the second scenario, the initial window displays a message with just a single word "rescanning". This is better in a way that at least, as a user I am not confused about what is actually happening. However, there is a different problem: no indication about which wallet is being loaded. If a user has multiple wallets and don't keep them all opened all the time, but opens one, closes another and so on, upon (re)starting Bitcoin-qt they would not know for which wallet rescanning is happening.

<img width="592" alt="Screenshot 2021-03-25 at 01 13 54" src="https://user-images.githubusercontent.com/22457855/112633990-5463b500-8e3a-11eb-84ac-a6c678dbf7a1.png">

In my opinion, both windows should clearly state:
- what is going on right now exactly, i.e. rescanning
- names of wallets that are being loaded

streamofstars:
Keep in mind that there is also a case to cover in my second scenario where a json file is configured to load multiple wallets and more than one require rescanning. However, I am not familiar with how the rescan feature works under the hood specifically. Whether it rescans for each wallet one by one or rescans for all loaded wallets at once. If the latter is the case, is it really feasible to list all those wallets there? Two or three is fine but what if someone has dozens of them? (unless there is a limit to the number of wallets)

hebasto:
> Whether it rescans for each wallet one by one or rescans for all loaded wallets at once

IIRC, the former one.

khan201kp:
What do you mean by wallet name? Do you mean the platform name, repo name, or the name of individual, wallet owner,s name?

hebasto:
@khan201kp 
> What do you mean by wallet name?

A `Wallet::getWalletName()` return value: https://github.com/bitcoin-core/gui/blob/77a2f5d30c5ecb764b8a7c098492e1f5cdec90f0/src/interfaces/wallet.h#L82-L83

[...]

PR #786, opened 2024-01-10
FIX:When opening or autoloading wallets there should be clear messages about rescanning in progress and wallets' names.

xonx4l:
issue-: #259

The "Opening Wallet" popup would now show "Opening Wallet: " instead of just "Opening Wallet". This identifies the specific wallet.

The "Rescanning" popup would show "Rescanning: " or "Rescanning - 0%" if a progress bar is added. Again, this identifies the wallet.

The main window title would include the wallet name after being opened/loaded thanks to the change in WalletView::showNormalIfMinimized().

Other dialogs like Send/Receive coins would also show the wallet name in the title.

So in summary, by appending the wallet name retrieved from WalletModel::getWalletName() to titles and popup messages, it provides clearer feedback to users about which wallet is being loaded and rescanned. No more confusion about which wallet Bitcoin Core is working on.

### DIFF
--- a/src/qt/receivecoinsdialog.cpp
+++ b/src/qt/receivecoinsdialog.cpp
@@ -72,6 +72,9 @@ void ReceiveCoinsDialog::setModel(WalletModel *_model)
 {
     this->model = _model;
 
+        setWindowTitle(tr("Request payments - %1").arg(model->getWalletName()));
+
+
     if(_model && _model->getOptionsModel())
     {
         _model->getRecentRequestsTableModel()->sort(RecentRequestsTableModel::Date, Qt::DescendingOrder);
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -145,7 +145,9 @@ void SendCoinsDialog::setClientModel(ClientModel *_clientModel)
 void SendCoinsDialog::setModel(WalletModel *_model)
 {
     this->model = _model;
-
+    
+      setWindowTitle(tr("Send Coins - %1").arg(model->getWalletName()));
+  
     if(_model && _model->getOptionsModel())
     {
         for(int i = 0; i < ui->entries->count(); ++i)


src/qt/receivecoinsdialog.cpp
```diff
@@ -72,6 +72,9 @@ void ReceiveCoinsDialog::setModel(WalletModel *_model)
 {
     this->model = _model;
 
+        setWindowTitle(tr("Request payments - %1").arg(model->getWalletName()));
```
pablomartin4btc, line 75: CI lint is [failing](https://github.com/bitcoin-core/gui/pull/786/checks?check_run_id=20356215277) due to unnecessary white-spaces in the files (you can run the script `lint-whitespace.py` locally or even [all linters](https://github.com/bitcoin/bitcoin/blob/master/test/lint/README.md#running-locally))
```suggestion
    setWindowTitle(tr("Request payments - %1").arg(model->getWalletName()));
```

pablomartin4btc (commented):
Thanks @xonx4l for working on this.

I don't think this PR fixes #259 as, according to its description and what I understand, a fix should be add more clarification (percentage?) about the rescanning of a wallet when they get open in 2 situations:
1. During bitcoin-qt startup, as author of the PR specified, the user could add the wallets manually into the `settings.json` `qt` config file. In this case, perhaps the name of each wallet (if there are many, and may be also x/ total) should be added after the "rescanning..." text on the spash-screen (initial/ startup window).
2. When a user open a wallet for the first time thru the menu->File->Open Wallet. For this case, perhaps a "rescanning..." (and maybe a % or progress bar - check issue #679) text could be added at the bottom of the opening window.

@hebasto on his [comment](https://github.com/bitcoin-core/gui/issues/259#issuecomment-811297476) on the issue, mentioned to add the wallet name on the title of "Receiving addresses" & "Sending addresses" windows (which has been done and merged already on #757), not to the `receivecoins` or `sendcoins` dialogs which this PR changes, moreover since these dialogs are within the main window and titles won't be displayed:

image

Other than that, the title and description of the PR doesn't reflect the proper change that's making, please check the [contributor](https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md) and [developer](https://github.com/bitcoin/bitcoin/blob/master/doc/developer-notes.md) guidelines (e.g. commit name may start with prefix `gui:`, please check other PRs as reference).

Please don't get discouraged by these observations. Feel free to reach out if you have questions or need clarification on any points. I encourage you to continue working on this issue or any others. Thanks again!

hebasto:
Closing due to lack of support from the PR author (unaddressed comments, failed CI).

Feel free to re-open.

PR #787, opened 2024-01-12
Fix typos

GoodDaisy:
fix typo in doc/Doxyfile.in
fix typo in test/functional/wallet_backup.py


alfonsoromanz:
Concept ACK. Maybe it's better to squash both commits into a single one? Not sure what others think

hebasto:
These changes are not GUI-specific. Therefore, they should be submitted to the main repository, i.e., https://github.com/bitcoin/bitcoin.

PR #788, opened 2024-01-17
debugwindow: update session ID tooltip

MarnixCroes:
When you have a v2 connection, there is always a session ID.

the _if any_ is a leftover from https://github.com/bitcoin-core/gui/pull/754, where the session ID property initially would always be displayed (transport v1 and v2).
So the session ID could be empty when you have a v1 connection.

As now the _Session ID_ property only is displayed for v2 connection, there will always be a session ID.

master 

sessionIDifany

PR

sessionID

Session ID not shown when transport v1

transportv1

### DIFF
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -1113,7 +1113,7 @@
                <item row="3" column="0">
                 <widget class="QLabel" name="peerSessionIdLabel">
                  <property name="toolTip">
-                  <string>The BIP324 session ID string in hex, if any.</string>
+                  <string>The BIP324 session ID string in hex.</string>
                  </property>
                  <property name="text">
                   <string>Session ID</string>


MarnixCroes:
> Concept ACK
> 
> nit: @MarnixCroes, maybe you can add the screenshot for transport v1 and session ID is empty/ not shown, if possible, even this PR doesn't affect that logic.

done

kristapsk (approved):
ACK 3bf00e13609eefa6ddb11353519bb1aec2342513

BrandonOdiwuor (commented):
Concept ACK

vostrnad:
ACK 3bf00e13609eefa6ddb11353519bb1aec2342513

hernanmarino (approved):
Concept ACK

jarolrod (approved):
ACK 3bf00e13609eefa6ddb11353519bb1aec2342513

hebasto:
> When you have a v2 connection, there is always a session ID.

What if the connection type is still `DETECTING`?

The code reference:https://github.com/bitcoin-core/gui/blob/6737331c4ccc6da578bb44c809cc4e18f017ba51/src/node/connection_types.h#L83-L88

pablomartin4btc (approved):
tACK 3bf00e13609eefa6ddb11353519bb1aec2342513

<details>
<summary>Tested on Ubuntu 22.04 using <code>signet</code> (<code>./src/qt/bitcoin-qt -signet</code>) otherwise I couldn't find <code>v2</code> peers on <code>testnet</code>.</summary>

- `v2` transport with Session ID field displayed:

image

- `v1` transport without it:

image

</details>

@hebasto this PR only updates `peerSessionIdLabel` tooltip; this widget and the "Session ID" one, right below on the `ui` form, won't be shown on "`DETECTING`" `TransportProtocolType enum`:

https://github.com/bitcoin-core/gui/blob/d04324a7056a735c1127ba8ccdc720a16e7281a3/src/qt/rpcconsole.cpp#L1208-L1215

pablomartin4btc:
nit: you could expand a bit on the details of the reason of the removal of "if any" from the label in the commit body (i.e. as it's only shown in v2 transport which will always have a session id).

hebasto (approved):
ACK 3bf00e13609eefa6ddb11353519bb1aec2342513.

gatleas17 (commented):
Help!

PR #789, opened 2024-01-19
Avoid non-self-contained Windows header

hebasto:
Using the `windows.h` header guarantees correctness regardless of the content of other headers.

For more details, please refer to https://stackoverflow.com/questions/4845198/fatal-error-no-target-architecture-in-visual-studio

Fixes the MSVC build when using the upcoming CMake-based build system and Qt packages installed via the vcpkg package manager.

Related to https://github.com/hebasto/bitcoin/pull/77.

### DIFF
--- a/src/qt/winshutdownmonitor.h
+++ b/src/qt/winshutdownmonitor.h
@@ -10,7 +10,7 @@
 #include <QString>
 #include <functional>
 
-#include <windef.h> // for HWND
+#include <windows.h>
 
 #include <QAbstractNativeEventFilter>
 


hebasto:
Friendly ping @sipsorcery :)

sipsorcery:
utACK c95767aa68e9e0e71e76884dbc83d237afdd7a37.

theuni:
I don't disagree with using self-contained headers, but according to your link it should only be necessary if we don't define `WIN32_LEAN_AND_MEAN`, [which we do](https://github.com/bitcoin/bitcoin/blob/master/configure.ac#L694).

So, this makes me think there must be some compilation units where this somehow doesn't get defined?

hebasto:
> So, this makes me think there must be some compilation units where this somehow doesn't get defined?

I suppose, they might be the `qt*` stuff built by vcpkg.

hebasto:
@theuni 

You might be interested to consider the following example: https://godbolt.org/z/b8rns1e6x

src/qt/winshutdownmonitor.h
```diff
@@ -10,7 +10,7 @@
 #include <QString>
 #include <functional>
 
-#include <windef.h> // for HWND
+#include <windows.h> // for HWND
```
fanquake, line 13: Comment should be removed? We are removing these "for xyz" comments in any case, but this one would also now be incorrect.

hebasto: > Comment should be removed?

Thanks! Done.

theuni:
ACK 8023640a71a10ec54a6a8e6b95a29d07f7be218d. It's not completely clear to me why this currently works, but I don't think it's worth wasting more time on. `windows.h` seems more correct regardless.

PR #790, opened 2024-01-21
refactor: Readable widget naming for debug window

s1Sharp:
Fixed widget names in the debug window. Make it more developer friendly to avoid confusion inside Qt Designer.


s1Sharp:
Qt Designer view before changes:
image
Qt Designer view after changes:
image

And a few changes for other widget names that were not included in the screenshots

PR #806, opened 2024-03-21
refactor: Misc int sign change fixes

maflcko:
This is allowed by the language. However, the `integer` sanitizer complains about it. Thus, fix it, so that the `integer` sanitizer can be used in the future to catch unintended sign changes.

Fixes #805.

Issue #805, opened 2024-03-20
implicit-integer-sign-change wallet/interfaces.cpp:289:57


maflcko:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behaviour

```
wallet/interfaces.cpp:289:57: runtime error: implicit conversion from type 'int' of value -1 (32-bit, signed) to type 'unsigned int' changed the value to 4294967295 (32-bit, unsigned)
    #0 0x55dcca43aaea in wallet::(anonymous namespace)::WalletImpl::createTransaction(std::vector<wallet::CRecipient, std::allocator<wallet::CRecipient>> const&, wallet::CCoinControl const&, bool, int&, long&) src/wallet/interfaces.cpp:289:57
    #1 0x55dcc9439953 in WalletModel::prepareTransaction(WalletModelTransaction&, wallet::CCoinControl const&) src/qt/walletmodel.cpp:211:37
    #2 0x55dcc9624ce5 in SendCoinsDialog::PrepareSendText(QString&, QString&, QString&) src/qt/sendcoinsdialog.cpp:287:28
    #3 0x55dcc960c98b in SendCoinsDialog::sendButtonClicked(bool) src/qt/sendcoinsdialog.cpp:482:10
    #4 0x55dcc9665fdc in auto auto GUIUtil::ExceptionSafeConnect<QPushButton*, void (QAbstractButton::*)(bool), SendCoinsDialog*, void (SendCoinsDialog::*)(bool)>(QPushButton*, void (QAbstractButton::*)(bool), SendCoinsDialog*, void (SendCoinsDialog::*)(bool), Qt::ConnectionType)::'lambda'(auto&&...)::operator()<bool&>(auto&&...) const src/./qt/guiutil.h:400:21
```

### Expected behaviour

no sign change

### Steps to reproduce

* Compile with integer sanitizer
* Create transaction in the gui

### How did you obtain Bitcoin Core

Compiled from source

### What version of Bitcoin Core are you using?

master

### Operating system and version

Linux

furszy:
If the line in question is this one:
```
change_pos = txr.change_pos ? *txr.change_pos : -1;
```

I'm not seeing the issue. `change_pos` is an int and `txr.change_pos` is an `optional<unsigned int>`.

Yet, we could clean this code by making the `CreatedTransactionResult` struct reachable from the GUI code (just need to move it to another file).

furszy:
[...]

PR #807, opened 2024-03-22
refactor: interfaces, make 'createTransaction' less error-prone 

furszy:
Bundle all function's outputs inside the `util::Result` returned object.

Removals:
- The input-output 'change_pos' ref arg from `createTransaction`, which has been a source of bugs in the past.
- The 'fee' ref arg from `createTransaction`, which is currently only set when the transaction creation process succeeds.
- The currently unreachable `AmountWithFeeExceedsBalance` error (more info about its re-introduction at https://github.com/bitcoin/bitcoin/pull/25269).

Additionally, this PR moves the `CreatedTransactionResult` struct into its own file. This change is made to avoid further expanding the GUI dependencies on `wallet.h`. Structurally, the GUI should only access the model/interfaces and never the wallet directly.


pablomartin4btc (commented):
tACK dfbd145676255f82bad529976928dff1939c5e48

<details>
<summary>This PR solves the error described in #805, even #806 fixes other 2 related errors.</summary> 

```
wallet/interfaces.cpp:289:57: runtime error: implicit conversion from type 'int' of value -1 (32-bit, signed) to type 'unsigned int' changed the value to 4294967295 (32-bit, unsigned)
```
</details>

I agree with the reasoning behind the removal of some arguments and the unreachable exception.

Out of curiosity, could you please add some refs on "_The input-output 'change_pos' ref arg from createTransaction, which has been a source of bugs in the past._", if possible?

Are there other components/ functions that you think could be moved into the new `util_spend.h` file (perhaps as a follow-up)?

furszy (commented):
> Out of curiosity, could you please add some refs on "_The input-output 'change_pos' ref arg from createTransaction, which has been a source of bugs in the past._", if possible?

Sure, https://github.com/bitcoin/bitcoin/pull/29065.

> Are there other components/ functions that you think could be moved into the new util_spend.h file (perhaps as a follow-up)?

Not off the top of my head. Matter of continue breaking the GUI classes dependency on `wallet.h`.

src/interfaces/wallet.h
```diff
@@ -142,11 +143,10 @@ class Wallet
     virtual void listLockedCoins(std::vector<COutPoint>& outputs) = 0;
 
     //! Create transaction.
-    virtual util::Result<CTransactionRef> createTransaction(const std::vector<wallet::CRecipient>& recipients,
+    virtual util::Result<CreatedTransactionResult> createTransaction(const std::vector<wallet::CRecipient>& recipients,
```
pablomartin4btc, line 146: **nit**: the tx won't be created when an error is raised during the `createTransaction`'s call (eg no recipient) so the obj should represent the correct thing, unless I misunderstanding the use case here... ok I see now this is how `src/util/result.h` works, so ignore my correction, I find it a bit weird but perhaps cos I'm not used to it yet. As a ref this change could be useful https://github.com/bitcoin/bitcoin/pull/25665 
```suggestion
    virtual util::Result<CreateTransactionResult> createTransaction(const std::vector<wallet::CRecipient>& recipients,
```

pablomartin4btc (approved):
cr ACK dfbd145676255f82bad529976928dff1939c5e48

src/interfaces/wallet.h
pablomartin4btc: It's ok, I got confused at first glance but then read more about how `util::Result` is being used in all our code.

https://github.com/bitcoin-core/gui/blob/0d509bab45d292caeaf34600e57b5928757c6005/src/util/result.h#L19-L33

furszy (commented):
Rebased after #806 merge.

hebasto:
While introducing a new `wallet/util_spend.h` header, it seems reasonable to address all related IWYU warnings, no?

hebasto:
cc @achow101 @ryanofsky

src/wallet/util_spend.h
```diff
@@ -0,0 +1,24 @@
...
+
+#ifndef BITCOIN_WALLET_UTIL_SPEND_H
+#define BITCOIN_WALLET_UTIL_SPEND_H
+
+#include <policy/fees.h>
+#include <wallet/transaction.h>
+
+#include <optional>
+
+struct CreatedTransactionResult
```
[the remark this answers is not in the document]
furszy: > Would suggest maybe moving this struct to wallet/types.h instead of introducing a new header. That file is meant to hold wallet types that are used outside of the wallet library. (For context see https://github.com/bitcoin/bitcoin/pull/29415#discussion_r1543379055 which talks about wallet/types.h, node/types.h, and common/types.h files).

Nice, sure. I'm on the same boat, I just wrote this without thinking on a general convention.

furszy (commented):
updated per feedback. Thanks @ryanofsky!

pablomartin4btc (commented):
re-ACK 4d941b5ff43e6e7efa83913ded8cd5806ef92dbe

- Changes from my last review: `CreatedTransactionResult` struct moved as [suggested](https://github.com/bitcoin-core/gui/pull/807#discussion_r1598730244) by @ryanofsky [explaining](https://github.com/bitcoin/bitcoin/pull/29415#discussion_r1543379055) reasoning behind the convention. I like the idea of keeping the namespace, not only for [consistency](https://github.com/bitcoin/bitcoin/pull/29015#discussion_r1424337693) but for clarity and making the code more readable. I think it would be good to document this somewhere, perhaps in the [developer notes](https://github.com/bitcoin/bitcoin/blob/master/doc/developer-notes.md) to begin with (?).

nit: on 2nd. commit (95aaaa4f023c6a6b629c36cfb9d8abd4bab1cb66), you forgot to remove the change from the makefile as it's not longer needed.

furszy:
> nit: on 2nd. commit (https://github.com/bitcoin-core/gui/commit/95aaaa4f023c6a6b629c36cfb9d8abd4bab1cb66), you forgot to remove the change from the makefile as it's not longer needed.

Removed. Thanks.

furszy:
CI failure is unrelated.

src/qt/walletmodel.cpp
```diff
@@ -214,12 +214,7 @@ WalletModel::SendCoinsReturn WalletModel::prepareTransaction(WalletModelTransact
         if (fSubtractFeeFromAmount && newTx)
             transaction.reassignAmounts(nChangePosRet);
 
-        if(!newTx)
-        {
-            if(!fSubtractFeeFromAmount && (total + nFeeRequired) > nBalance)
-            {
-                return SendCoinsReturn(AmountWithFeeExceedsBalance);
```
ryanofsky, line 221: In commit "gui: remove unreachable AmountWithFeeExceedsBalance error" (355199c7a6c58eb812dd7ef8e204e6e068c142ae)

This is a nice catch noticing that this code doesn't work. But I think the fact that it doesn't work is a pretty unfortunate problem, and the current PR will make fixing it harder.

It seems like #20640 unintentionally introduced a bug where the more accurate error message "The total exceeds your balance when the %1 transaction fee is included" can never trigger, and instead only less informative "Insufficient funds" or "The amount exceeds your balance" errors will be shown.

It looks like there are few ways this could be fixed:

- An easy way to fix it would be to restore the `CAmount& fee` output parameter removed by #20640 so it is always returned and this code can work the way it was originally intended.

- A slightly more complicated but cleaner fix would be to use `CreatedTransactionResult` as a return type instead of `util::Result<CreatedTransactionResult>`  and add a `bilingual_str error` field to `CreatedTransactionResult` so fee information and an error string can be returned together.

- Another alternative fix could be to build on [bitcoin/bitcoin#25665](https://github.com/bitcoin/bitcoin/pull/25665) and keep using `util::Result`, but return the fee in the new `util::Result` failure field.

- And the best but probably most complicated fix would be to remove error message code from the GUI and instead generate detailed error messages in wallet code. This would require bigger changes in the GUI code and might require changing the interface to provide the wallet with more information about transactions.

Against that background, I think this PR (as of edc3f9a733aaa1a1cc2547013d09c27828151e8a) makes some nice cleanups, but I don't like the current changes because I think they make all of the fixes above except maybe the last one harder to implement.

So I don't think I would ACK this PR currently, I'd happily ACK it if it either (1) fixed the problem, maybe using one of the easier fixes suggested above, or (2) did not fix the problem, but at least did not make it harder to fix in the future. This could be implemented by keeping most of the changes in the PR, but not deleting this GUI code in this commit and not deleting the `CAmount& fee` output parameter from the wallet interface after this commit.

ryanofsky (commented):
Concept ACK edc3f9a733aaa1a1cc2547013d09c27828151e8a, but this has some changes I like and other changes I don't think are good. Left a more detailed comment below.

ryanofsky:
I just noticed https://github.com/bitcoin/bitcoin/pull/25269, which I hadn't seen when I posted my comment above. It looks like that PR was closed for inactivity but it does look like a good approach that solves the issue I didn't like in my comment (though maybe it is still a slight regression because it doesn't include the amount of the fee in the error message). I wonder if you'd consider reopening that PR and maybe including the changes in this PR or basing this PR on that one?

Or if you no longer think that PR is worth pursuing, I think you could separate the cleanups here from the issue that PR was trying to address by leaving the `CAmount& fee` output parameter and GUI code intact here, but still doing the other cleanups.

furszy (commented):
Thanks for the review ryanofsky!

> I just noticed [bitcoin/bitcoin#25269](https://github.com/bitcoin/bitcoin/pull/25269), which I hadn't seen when I posted my comment above. It looks like that PR was closed for inactivity but it does look like a good approach that solves the issue I didn't like in my comment. (Though maybe that PR is still a slight regression because it doesn't include the amount of the fee in the error message.) I wonder if you'd consider reopening that PR and maybe including the changes in this PR there, or basing this PR on that one?

Yeah sure, I closed it for lack of movement after 6 months.
I just reopen it, reworking part of it, and rebased this one on top. I think it make sense to re-introduce the functionality there and cleanup the GUI code here.

hebasto:
Concept ACK.

> Additionally, this PR moves the `CreatedTransactionResult` struct into its own file.

This part of the PR description seems no longer relevant.

> Thanks for the review ryanofsky!
> 
> > I just noticed [bitcoin/bitcoin#25269](https://github.com/bitcoin/bitcoin/pull/25269), which I hadn't seen when I posted my comment above. It looks like that PR was closed for inactivity but it does look like a good approach that solves the issue I didn't like in my comment. (Though maybe that PR is still a slight regression because it doesn't include the amount of the fee in the error message.) I wonder if you'd consider reopening that PR and maybe including the changes in this PR there, or basing this PR on that one?
> 
> Yeah sure, I closed it for lack of movement after 6 months. I just reopen it, reworking part of it, and rebased this one on top. I think it make sense to re-introduce the functionality there and cleanup the GUI code here.

I agree on focusing this PR on the code cleanup only.

furszy:
Drafted until https://github.com/bitcoin/bitcoin/pull/25269 gets merged. This is a continuation of that work.

vasild:
Concept ACK

src/qt/walletmodel.cpp
```diff
@@ -153,6 +154,8 @@ bool WalletModel::validateAddress(const QString& address) const
 
 WalletModel::SendCoinsReturn WalletModel::prepareTransaction(WalletModelTransaction &transaction, const CCoinControl& coinControl)
 {
+    transaction.getWtx() = nullptr; // reset tx output
```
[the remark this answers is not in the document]
furszy: > I am not sure why this is needed.

It is just a safety belt that resets the state in case the caller invokes the function with an existing `WalletModelTransaction`. This shouldn’t happen nowadays since we create a new pointer per call, but if it ever does, for example with a duplicate recipient, the function would return `StatusCode::OK` with a tx even though nothing was sent.
Overall, this function needs lot more love but this reset is quite simple and effective to prevent other issues for now.

PR #812, opened 2024-03-29
Fix create unsigned transaction fee bump

### LINKED ISSUE #810: Create unsigned when increasing fee asks for the passphrase

### Current behaviour

Trying to increase the fee on a pending transaction, creating a psbt is asking for the passphrase, though it shouldn't need to unlock the wallet to create a psbt (and didn't require when creating the initial transaction).

### Expected behaviour

Create the psbt without the need for unlocking the wallet

### Steps to reproduce

Locked wallet
Set PSBT actions to enabled in settings
create a transaction with a fee increase replacement option and send it
raise the fee
Choose create PSBT
Get prompt for wallet unlock

### Relevant log output

_No response_

Package manager

26.0.0.0

Ubuntu 22.04

### Machine specifications

_No response_

furszy:
Fixes #810.

Not much to explain; we were requiring the wallet to be unlocked for the unsigned transaction creation process.
Fix this by moving the unlock wallet request to the signed transaction creation process.

### DIFF
--- a/src/qt/walletmodel.cpp
+++ b/src/qt/walletmodel.cpp
@@ -529,12 +529,6 @@ bool WalletModel::bumpFee(uint256 hash, uint256& new_hash)
         return false;
     }
 
-    WalletModel::UnlockContext ctx(requestUnlock());
-    if(!ctx.isValid())
-    {
-        return false;
-    }
-
     // Short-circuit if we are returning a bumped transaction PSBT to clipboard
     if (retval == QMessageBox::Save) {
         // "Create Unsigned" clicked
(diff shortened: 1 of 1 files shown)


pablomartin4btc (commented):
Concept ACK

I'll test it soon.

MarnixCroes (approved):
tack 671b7a32516d62e1e79393ded4b45910bd08010a

`Send` prompts the  passphrase, `Create Unsigned` does not prompt the passphrase

src/qt/walletmodel.cpp
```diff
@@ -549,10 +543,15 @@ bool WalletModel::bumpFee(uint256 hash, uint256& new_hash)
         DataStream ssTx{};
         ssTx << psbtx;
         GUIUtil::setClipboard(EncodeBase64(ssTx.str()).c_str());
-        Q_EMIT message(tr("PSBT copied"), tr("Copied to clipboard", "Fee-bump PSBT saved"), CClientUIInterface::MSG_INFORMATION);
+        Q_EMIT message(tr("PSBT copied"), tr("Fee-bump PSBT copied to clipboard"), CClientUIInterface::MSG_INFORMATION | CClientUIInterface::MODAL);
```
pablomartin4btc, line 546: At least on Ubuntu 22.04, now after the fee-bumping unsigned a modal pop-up is displayed:

image

while before there was a message sent thru the OS system information (?):

image

I see in the code we usually use `CClientUIInterface::MSG_INFORMATION` (`transactionview.cpp`, `walletview.cpp`) and I'd prefer to keep this look & feel but at the same time I think having the modal window popping up it's making the user more aware of what just happened with the intention to accept it.

pablomartin4btc (approved):
tACK 671b7a32516d62e1e79393ded4b45910bd08010a

Tested and verified this PR fixes #810 and now creation of unsigned tx/ PSBT/ during fee-bumping doesn't ask for passphrase/ requires wallet unlock.

src/qt/walletmodel.cpp
furszy: The change was introduced because the OS system notification wasn't popping up on macOS. Either way is ok for me.

hebasto (approved):
ACK 671b7a32516d62e1e79393ded4b45910bd08010a, tested on Ubuntu 24.04.

fanquake:
Backported in https://github.com/bitcoin/bitcoin/pull/30092.

Issue #809, opened 2024-03-27
gui: node shutting down after incorrect proxy IP address input


nbit846:
### Issues, reports or feature requests related to the GUI should be opened directly on the GUI repo

- [X] I still think this issue should be opened here

### Report

My node was running fine until I decided to change the setting under the nerwork tab, I checked the box for "Use separate SOCKS5 proxy to reach peers via Tor onion services" and on Proxy IP box i put 127.0.0.1:9150, on Port box 9150. When I tried to restart the node it wont start anymore, it gives me "Error 127.0.0.1:9150 incorrect Proxy" and it shuts down immediately. What can I do to get the node to start again?  the GUI is not even starting. I just stared exploring with bitcoin nodes and Im not knowledgable at all, any help will be greatly appreciated. Im running the node on Debian Linux.

willcl-ark:
Hi @nbit846,
 
This issue tracker is supposed to be used to track technical issues related to the Bitcoin Core GUI code base.

General bitcoin questions and/or support requests are best directed to the [Bitcoin StackExchange](https://bitcoin.stackexchange.com/) or the #bitcoin IRC channel on the [Libera Chat](https://libera.chat/) network.

In your case I'd advise checking the bitcoin data directory (`~/.bitcoin`) while the node is not running. You should find a file named `settings.json` which may be persisting the erroneous setting. This can be removed and the node then restarted.

That said, I don't experience the same issue as you when specifying a bad port in that dialogue box, so I'm not totally convinced this is the root cause of your issue (nor that the above step will solve it). Are the any other errors printed in `~/.bitcoin/debug.log`?

nbit846:
I didnt see any other errors printed just this:
2024-03-27T20:18:53Z Loaded 64927 addresses from peers.dat  352ms
2024-03-27T20:18:53Z init message: Loading banlist…
2024-03-27T20:18:53Z SetNetworkActive: true
2024-03-27T20:18:53Z Error: Invalid -onion address or hostname: '127.0.0.1:9050:9050'
2024-03-27T20:19:06Z Shutdown: In progress...
2024-03-27T20:19:06Z scheduler thread exit
2024-03-27T20:19:06Z Flushed fee estimates to fee_estimates.dat.
2024-03-27T20:19:06Z Shutdown: done

willcl-ark:
Yeah. It should not allow port to be duplicated as per this line:

[...]

PR #813, opened 2024-04-03
Don't permit port in proxy IP option

willcl-ark:
Fixes: https://github.com/bitcoin-core/gui/issues/809

Previously it was possible through the GUI to enter an IP address:port into the "Proxy IP" configuration box. After the node was restarted the errant setting would prevent the node starting back up until manually removed from settings.json.

Prevent this with a simple check for ":" in the string. This is acceptable here in the GUI setting because we already fail on a hostname such as "http://x.x.x.x", so it won't cause false positives.

### DIFF
--- a/src/qt/optionsdialog.cpp
+++ b/src/qt/optionsdialog.cpp
@@ -20,6 +20,7 @@
 #include <node/chainstatemanager_args.h>
 #include <netbase.h>
 #include <txdb.h>
+#include <util/strencodings.h>
 
 #include <chrono>
 
@@ -482,7 +483,10 @@ QValidator(parent)
 QValidator::State ProxyAddressValidator::validate(QString &input, int &pos) const
 {
     Q_UNUSED(pos);
-    // Validate the proxy
+    uint16_t port{0};
+    std::string hostname;
+    if (!SplitHostPort(input.toStdString(), port, hostname) || port != 0) return QValidator::Invalid;
+
     CService serv(LookupNumeric(input.toStdString(), DEFAULT_GUI_PROXY_PORT));
     Proxy addrProxy = Proxy(serv, true);
     if (addrProxy.IsValid())


pablomartin4btc (changes requested):
Concept ACK 94173eac6e89f374f1a98b5d0e449a0a07ed832d

I'd suggest to use a regex validator for the IP address, it's much simpler I think and the user won't be allow to enter any other symbols or chars (typo?) that aren't digits (plus user can't type more than 3 subnets separated by dots where currently you could and any symbols):

```diff
diff --git a/src/qt/optionsdialog.cpp b/src/qt/optionsdialog.cpp
index dd654a7ab..1fabee7f2 100644
--- a/src/qt/optionsdialog.cpp
+++ b/src/qt/optionsdialog.cpp
@@ -33,6 +33,11 @@
 #include <QSystemTrayIcon>
 #include <QTimer>
 
+// For IP address validators
+QRegularExpression ipRegex("^([0-9]{1,3}\\.){3}[0-9]{1,3}$");
+// For Port number validators
+QRegularExpression portRegex("^(6553[0-5]|655[0-2]\\d|65[0-4]\\d{2}|6[0-4]\\d{3}|[1-5]\\d{4}|[1-9]\\d{0,3})$");
+
 int setFontChoice(QComboBox* cb, const OptionsModel::FontChoice& fc)
 {
     int i;
@@ -112,13 +117,18 @@ OptionsDialog::OptionsDialog(QWidget* parent, bool enableWallet)
     ui->mapPortNatpmp->setEnabled(false);
 #endif
 
+    QRegularExpressionValidator *ipValidator = new QRegularExpressionValidator(ipRegex, this);
+    QRegularExpressionValidator *portValidator = new QRegularExpressionValidator(portRegex, this);
+
     ui->proxyIp->setEnabled(false);
+    ui->proxyIp->setValidator(ipValidator);
+
     ui->proxyPort->setEnabled(false);
-    ui->proxyPort->setValidator(new QIntValidator(1, 65535, this));
+    ui->proxyPort->setValidator(portValidator);
 
     ui->proxyIpTor->setEnabled(false);
+   ui->proxyIpTor->setValidator(ipValidator);
     ui->proxyPortTor->setEnabled(false);
-    ui->proxyPortTor->setValidator(new QIntValidator(1, 65535, this));
+    ui->proxyPortTor->setValidator(portValidator);
 
     connect(ui->connectSocks, &QPushButton::toggled, ui->proxyIp, &QWidget::setEnabled);
     connect(ui->connectSocks, &QPushButton::toggled, ui->proxyPort, &QWidget::setEnabled);

```

I've also included in the above a "port number validator", the regex there is a bit more "elaborated" if you like but currently using the int validator, user can enter a dot and then the `CService` will return that the port is not valid.

Last one since we are here, I'd change the call to 'CService' (that performs currently the validation of the port number - eg as a mentioned above that user can enter a dot) to when the value is changed (`textChanged`?) - this is something that some [reviewer noticed](https://github.com/bitcoin-core/gui-qml/pull/391#issuecomment-2021250132) testing the [corresponding control/ widget](https://github.com/bitcoin-core/gui-qml/pull/391) in the [QML gui repo](https://github.com/bitcoin-core/gui-qml/) and I find it more practical - not as how currently behaves that the validation is done on focus changed (you can also see in the code that the tor inputs have different signal connections somehow, and the `textChanged` there doesn't work properly either) but this can be done also on a follow-up.

willcl-ark:
> I'd suggest to use a regex validator for the IP address, it's much simpler I think and the user won't be allow to enter any other symbols or chars (typo?) that aren't digits (plus user can't type more than 3 subnets separated by dots where currently you could and any symbols)

I'm not convinced that a large regex is "simpler" than checking if a ":" is in the string?

Thanks for the other suggestions, but I don't feel that interested in picking them here with the QML work going on, and them not being much of an issue IMO. In contrast this fix prevents a bug whereby an incorrect setting can stop the node starting up.

The only potential issue I saw with my own solution was that it would outright prevent setting the tor proxy IP to an IPv6 address, however from my (quick) research, it didn't seem that this was actually something anyone did (and it might not even be possible/practical to do). If I'm wrong on this, I'll be happy to update the solution further, but from what I can see even setting Tor ORPort to IPV6 is barely supported, and ControlPort is limited to an IPv4 address:port...

src/qt/optionsdialog.cpp
```diff
@@ -482,7 +482,10 @@ QValidator(parent)
 QValidator::State ProxyAddressValidator::validate(QString &input, int &pos) const
 {
     Q_UNUSED(pos);
-    // Validate the proxy
+    // "http://x.x.x.x" already prohibited, so safe to fail early here on a ":"
+    if (input.contains(':')) {
+        return QValidator::Invalid;
+    }
```
furszy, line 488: What about using `SplitHostPort()` from `#include <util/strencodings.h>` instead?
Would only need to verify that the port isn't set.

willcl-ark: That certainly looks like it could be what is needed here!

pablomartin4btc:
> I'm not convinced that a large regex is "simpler" than checking if a ":" is in the string?

The large one was for the port number, the one I referred as "simpler" was for the IP address: `QRegularExpression ipRegex("^([0-9]{1,3}\\.){3}[0-9]{1,3}$");`. What I meant is that we know in advance that the ":" would be invalid so we could restrict the user to even enter it. As current we could see this (not only ":"):

image

The regex would keep it cleaner, more user friendly and we don't have to do an extra validation, but I'll stop trying to convince you :smile: (btw, also we'd need to set the max length to 15/ follow up).

> I don't feel that interested in picking them here with the QML work going on, and them not being much of an issue IMO. 

QML is far for being at the stage where QT currently is, improving user experience could also be useful for both projects feedback.

> In contrast this fix prevents a bug whereby an incorrect setting can stop the node starting up.

I agree and there's a [workaround](https://github.com/bitcoin-core/gui/issues/809#issuecomment-2026525827) already which is just removing the IP line from the file.

> The only potential issue I saw with my own solution was that it would outright prevent setting the tor proxy IP to an IPv6 address, however from my (quick) research, it didn't seem that this was actually something anyone did (and it might not even be possible/practical to do). If I'm wrong on this, I'll be happy to update the solution further, but from what I can see even setting Tor ORPort to IPV6 is barely supported, and ControlPort is limited to an IPv4 address:port...

Is it even possible to enable those checkboxes?

image

furszy (commented):
utACK 10c5275ba45

luke-jr:
@jarolrod Isn't that a bug independent from the validation one?

hebasto (approved):
ACK 10c5275ba4532fb1bf54057d2f61fc35b51f1e85, tested on Ubuntu 24.04.

fanquake:
Backported in https://github.com/bitcoin/bitcoin/pull/30092.

Issue #718, opened 2023-03-11
wiadomośc


ponury1990:
https://github.com/bitcoin-core/gui/blob/9887fc789886b356efc6818783cabaab2b14481f/test/functional/p2p_invalid_messages.py#L90-L92

PR #815, opened 2024-04-09
Bugfix on TransactionsView - Disable if privacy mode is set during wallet selection

pablomartin4btc:
<details>
<summary>Currenlty on <code>master</code>, when the "mask values" checkbox is ticked if the user selects a different wallet, the history action is enable and if the user clicks on it can see all the transactions in the transaction view.</summary>

Peek 2024-04-09 17-37

</details>
<details>
<summary>This PR fixes it.</summary>

Peek 2024-04-09 17-45

</details>

Note for maintainers: this needs to be backported to 25.x and 26.x.


alfonsoromanz (approved):
Tested ACK d3da5025f61534d126fef12f198afb65f3a17897

pablomartin4btc:
> This feels like the wrong place to fix it. Why not inside `setWalletActionsEnabled` (or whatever is enabling it to begin with)?

Yeah, it makes more sense, I'll rework it. Thanks!

pablomartin4btc:
Updates:
- Addressed @luke-jr's [feedback](https://github.com/bitcoin-core/gui/pull/815#pullrequestreview-2013545536).

src/qt/bitcoingui.cpp
```diff
@@ -714,6 +714,8 @@ void BitcoinGUI::addWallet(WalletModel* walletModel)
     WalletView* wallet_view = new WalletView(walletModel, platformStyle, walletFrame);
     if (!walletFrame->addView(wallet_view)) return;
 
+    wallet_view->setPrivacy(isPrivacyModeActivated());
```
luke-jr, line 717: Why move this up top? Now it's after the signal is connected, so there's a tiny chance of a race...

src/qt/bitcoingui.cpp
```diff
@@ -769,6 +768,7 @@ void BitcoinGUI::setCurrentWallet(WalletModel* wallet_model)
             break;
         }
     }
+    setWalletActionsEnabled(true);
```
luke-jr, line 771: This shouldn't be necessary.

luke-jr (changes requested):
I'm pretty sure you can just modify the existing line in `setWalletActionsEnabled` to:

```c++
    historyAction->setEnabled(enabled && !isPrivacyModeActivated());
```

src/qt/bitcoingui.cpp
pablomartin4btc: I'll check, thanks.

src/qt/bitcoingui.cpp
pablomartin4btc: Since I moved `enableHistoryAction` within `setWalletActionsEnabled` I thought there was a evaluation of the `wallet_view` privacy there... taking that back down.

pablomartin4btc:
> ```c++
>     historyAction->setEnabled(enabled && !isPrivacyModeActivated());
> ```

Just replacing the operator with `||` (e.g. when the user closes all wallets, all tabs will be disabled except for `Transactions`).

pablomartin4btc:
Updates:

- Addressed @luke-jr's [feedback](https://github.com/bitcoin-core/gui/pull/815#pullrequestreview-2072577992).

src/qt/bitcoingui.cpp
pablomartin4btc: done.

luke-jr:
This one-line change works without anything else:

https://github.com/bitcoin-core/gui/pull/815#pullrequestreview-2072577992

pablomartin4btc:
> This one-line change works without anything else:
> 
> [#815 (review)](https://github.com/bitcoin-core/gui/pull/815#pullrequestreview-2072577992)

Sorry, I made another mistake, thanks for double checking.

To other reviewers: please hold on till next push, thanks.

src/qt/bitcoingui.cpp
pablomartin4btc: Sorry, that was because `setWalletActionsEnabled` was not working properly, and `setWalletActionsEnabled` was not working because I misread your suggestion and I was still calling the function `enableHistoryAction` incorrectly instead of just using directly `historyAction->setEnabled`. It fixed now in 260d6eb. Thanks!

pablomartin4btc:
Updates:
- Incorporated @luke-jr's [suggestion](https://github.com/bitcoin-core/gui/pull/815#issuecomment-2127782169), simplifying the fix.

hebasto:
@pablomartin4btc Did you consider https://github.com/bitcoinknots/bitcoin/issues/83?

hebasto:
@pablomartin4btc 
> ... was introduced in #718...

Sure about PR number? ('cause there is no such a number in this repo)

pablomartin4btc:
> @pablomartin4btc
> 
> > ... was introduced in #718...
> 
> Sure about PR number? ('cause there is no such a number in this repo)

oh! my bad.. a typo there... how lucky! I'll play the lottery with that one today... I meant #708, I think it's the only place related to the "mask value" where I switched to the overview tab when the mask value checkbox on the menu is ticked but still need to check it.

pablomartin4btc:
@hebasto, I couldn't reproduce the issue described in https://github.com/bitcoinknots/bitcoin/issues/83, trying both `master` and this PR, also checked the [fix in knots](https://github.com/bitcoinknots/bitcoin/releases/tag/v27.1.knots20240621) which is similar to this PR code change. It would be good to know what was the state of the "mask value" during the use case described in https://github.com/bitcoinknots/bitcoin/issues/83, the only thing I can think of it's that the issue is the same one described above on this PR but when that happens loading a second wallet would disable the Transactions view tab so it can't be selected. Perhaps @luke-jr can describe the steps to reproduce it.

pablomartin4btc:
@hebasto, the issue mentioned in https://github.com/bitcoinknots/bitcoin/issues/83 is fixed by this PR.
I've done a bit of investigation and the problem (in `master`, not on this PR) was not "switching" between wallets (selecting another wallet from the combo box won't show the overview page while the user is landing on the transactions tab), but when the user opens another wallet (which hasn't been loaded before) and the current wallet view is on the transactions/ history tab, then the system sets the overview page as a default view lets say (the problematic [call](https://github.com/bitcoin-core/gui/blob/4c62f4b53561bce5b2eb8639cdc24d284be537eb/src/qt/bitcoingui.cpp#L737) is removed on this PR).

pablomartin4btc:
Updates:
- Rebased (fixed previous CI failure - _ASan + LSan + UBSan..._ ).

PR #817, opened 2024-04-12
Bugfix - don't allow multiple dialogs for same tx in TransactionView

pablomartin4btc:
Limit to one the transaction details dialogs that a user can open. If a transaction details dialog is already open, bring it to the foreground when a user clicks on its row again - as [suggested](https://github.com/bitcoin-core/gui/pull/817#issuecomment-2051671919..) by flack.

<details>
<summary>Currently a user can open unlimited tx details dialogs for the same tx id.</summary>

Peek 2024-04-12 00-50

</details>
<details>
<summary>This PR fixes the issue.</summary>

Peek 2024-04-12 00-37

</details>

### DIFF
--- a/src/qt/transactionview.cpp
+++ b/src/qt/transactionview.cpp
@@ -35,6 +35,7 @@
 #include <QPoint>
 #include <QScrollBar>
 #include <QSettings>
+#include <QString>
 #include <QTableView>
 #include <QTimer>
 #include <QUrl>
@@ -528,13 +529,15 @@ void TransactionView::showDetails()
     QModelIndexList selection = transactionView->selectionModel()->selectedRows();
     if(!selection.isEmpty())
     {
-        TransactionDescDialog *dlg = new TransactionDescDialog(selection.at(0));
-        dlg->setAttribute(Qt::WA_DeleteOnClose);
-        m_opened_dialogs.append(dlg);
-        connect(dlg, &QObject::destroyed, [this, dlg] {
-            m_opened_dialogs.removeOne(dlg);
-        });
-        dlg->show();
+        if(!detailsAlreadyShown(selection.at(0))) {
+            TransactionDescDialog *dlg = new TransactionDescDialog(selection.at(0));
+            dlg->setAttribute(Qt::WA_DeleteOnClose);
+            m_opened_dialogs.append(dlg);
+            connect(dlg, &QObject::destroyed, [this, dlg] {
+                m_opened_dialogs.removeOne(dlg);
+            });
+            dlg->show();
+        }
     }
 }
 
@@ -662,4 +665,15 @@ void TransactionView::closeOpenedDialogs()
         dlg->close();
     }
     m_opened_dialogs.clear();
+}
+
+bool TransactionView::detailsAlreadyShown(const QModelIndex &idx)
+{
+    for (TransactionDescDialog* dlg : m_opened_dialogs) {
+        if (dlg->getTransactionId() == idx.data(TransactionTableModel::TxHashRole).toString()) {
+            GUIUtil::bringToFront(dlg);
+            return true;
+        }
+    }
+    return false;
 }
--- a/src/qt/transactiondescdialog.cpp
+++ b/src/qt/transactiondescdialog.cpp
@@ -15,7 +15,8 @@ TransactionDescDialog::TransactionDescDialog(const QModelIndex &idx, QWidget *pa
     ui(new Ui::TransactionDescDialog)
 {
     ui->setupUi(this);
-    setWindowTitle(tr("Details for %1").arg(idx.data(TransactionTableModel::TxHashRole).toString()));
+    m_transaction_id = idx.data(TransactionTableModel::TxHashRole).toString();
+    setWindowTitle(tr("Details for %1").arg(m_transaction_id));
     QString desc = idx.data(TransactionTableModel::LongDescriptionRole).toString();
     ui->detailText->setHtml(desc);
 
--- a/src/qt/transactiondescdialog.h
+++ b/src/qt/transactiondescdialog.h
@@ -6,6 +6,7 @@
 #define BITCOIN_QT_TRANSACTIONDESCDIALOG_H
 
 #include <QDialog>
+#include <QString>
 
 namespace Ui {
     class TransactionDescDialog;
@@ -24,8 +25,11 @@ public:
     explicit TransactionDescDialog(const QModelIndex &idx, QWidget *parent = nullptr);
     ~TransactionDescDialog();
 
+    QString getTransactionId() {return m_transaction_id;};
+
 private:
     Ui::TransactionDescDialog *ui;
+    QString m_transaction_id;
 };
 
 #endif // BITCOIN_QT_TRANSACTIONDESCDIALOG_H
--- a/src/qt/transactionview.h
+++ b/src/qt/transactionview.h
@@ -108,6 +108,7 @@ private Q_SLOTS:
     void updateWatchOnlyColumn(bool fHaveWatchOnly);
     void abandonTx();
     void bumpFee(bool checked);
+    bool detailsAlreadyShown(const QModelIndex &idx);
 
 Q_SIGNALS:
     void doubleClicked(const QModelIndex&);


flack:
would be nice if a click on a row where the dialog is already open would bring that dialog to the foreground

pablomartin4btc:
Updates:
- Fixed lint failure.
- Bring an already open dialog to the foreground as @flack [suggested](https://github.com/bitcoin-core/gui/pull/817#issuecomment-2051671919).

src/qt/transactionview.cpp
```diff
@@ -663,3 +665,14 @@ void TransactionView::closeOpenedDialogs()
     }
     m_opened_dialogs.clear();
 }
+
+bool TransactionView::detailsAlreadyShown(const QModelIndex &idx)
+{
+    for (TransactionDescDialog* dlg : m_opened_dialogs) {
+        if (dlg->getTransactionId() == idx.data(TransactionTableModel::TxHashRole).toString()) {
+            dlg->activateWindow();
```
alfonsoromanz, line 673: The dialog is not brought to the foreground in my case (I am using Mac). In order to make it work I had to call `raise()` instead. Thoughts?

```suggestion
            dlg->raise()
```

alfonsoromanz (commented):
Tested ACK b26d2e81ec6ac4b5bd97bc7f7ec0c44e502b6a18. Only one dialog is opened for a single tx.

shafi4117 (approved):

shafi4117 (approved):

src/qt/transactionview.cpp
pablomartin4btc: Thanks for reviewing and spotting that. I tried on Ubuntu to call **only** `dlg->raise()` but doesn't seem to work (also try other things like `dlg->setFocus()`, `dlg->setFocusPolicy(Qt::StrongFocus)`, `dlg->focusWidget()` and some combinations), until I found this in the QT documentation about [activateWindow()](https://doc.qt.io/qt-5/qwidget.html#activateWindow) (which I should have checked first):

_This function performs the same operation as clicking the mouse on the title bar of a top-level window. On **X11**, the result depends on the Window Manager. **If you want to ensure that the window is stacked on top as well you should also call [raise](https://doc.qt.io/qt-5/qwidget.html#raise)**. Note that the window must be visible, otherwise activateWindow() has no effect._

_On **Windows**, if you are calling this when the application is not currently the active one then it will not make it the active window. It will change the color of the taskbar entry to indicate that the window has changed in some way. This is because Microsoft does not allow an application to interrupt what the user is currently doing in another application._

So, since you tried it on Mac already, let me take a look on how it behaves on **Windows**.

alfonsoromanz: In Mac, it seems that `activateWindow()` only makes the window active, i.e: the title of the dialog gets highlighted and gets the keyboard input focus. The `raise()` function brings the dialog on top. So, at least in Mac, it makes sense to keep both IMO

pablomartin4btc: Yes, it needs both `activateWindow()` & `raise()`, let me try it on Windows and I'll incorporate your [suggestion](https://github.com/bitcoin-core/gui/pull/817#discussion_r1569483043), thanks!

BrandonOdiwuor (approved):
Tested ACK b26d2e81ec6ac4b5bd97bc7f7ec0c44e502b6a18

Tested successfully on Ubuntu 22.04.4 LTS. Only one dialogue is created per transaction and the clicked transaction is brought to the foreground

[Screencast from 24-04-2024 03:35:10 ALASIRI.webm](https://github.com/bitcoin-core/gui/assets/15610188/396507fc-bb8c-4773-8e74-68ac51c6b582)

luke-jr:
Why?

shafi4117 (approved):

pablomartin4btc:
> Why?

Discussed a bit offline... Please feel free to add more context here and any concerns you have.

pablomartin4btc:
Updates:
- [added the call](https://github.com/bitcoin-core/gui/blob/ba13f10f9ca5bad7de0e140b667fbae5e8b7b9a3/src/qt/transactionview.cpp#L675) to `raise()` as @alfonsoromanz [suggested](https://github.com/bitcoin-core/gui/pull/817#discussion_r1569483043) finding that the fix wasn't working properly on Mac without it, this is not causing any other effects on Linux or Windows which I both tested. 
- added includes of `QString` in both [`src/qt/transactiondescdialog.h`](https://github.com/bitcoin-core/gui/blob/ba13f10f9ca5bad7de0e140b667fbae5e8b7b9a3/src/qt/transactiondescdialog.h#L9) and [`src/qt/transactionview.cpp`](https://github.com/bitcoin-core/gui/blob/ba13f10f9ca5bad7de0e140b667fbae5e8b7b9a3/src/qt/transactionview.cpp#L38), without them `bitcoin-qt` was crashing on Windows on the [tx id hash string comparison](https://github.com/bitcoin-core/gui/blob/ba13f10f9ca5bad7de0e140b667fbae5e8b7b9a3/src/qt/transactionview.cpp#L673) (**_note_**: `bitcoin-qt` was cross-compiled for Windows with WSL Ubuntu 24.04 beta; it would be good to double check on native Windows compilation with VS 2022)

alfonsoromanz (approved):
Re ACK ba13f10f9ca5bad7de0e140b667fbae5e8b7b9a3. I tested in Mac and now the dialog is brought to the foreground

https://github.com/bitcoin-core/gui/assets/19962151/dd21afaa-b48d-4082-be4c-3c3bdc965cfa

hebasto:
> Currently a user can open unlimited tx details dialogs for the same tx id.

Concept ACK. Such a behavior can confuses the user.

hebasto:
> Tested successfully on Ubuntu 22.04.4 LTS. Only one dialogue is created per transaction and the clicked transaction is brought to the foreground

Tested on Ubuntu 24.04. It works with `QT_QPA_PLATFORM=xcb`, but fails with `QT_QPA_PLATFORM=wayland`.

pablomartin4btc:
> Tested on Ubuntu 24.04. It works with `QT_QPA_PLATFORM=xcb`, but fails with `QT_QPA_PLATFORM=wayland`.

I'll take a look, thanks!

pablomartin4btc:
@hebasto it seems there are some [issues](https://gitlab.freedesktop.org/wayland/wayland/-/issues/233) with Wayland on how these actions are being handled ([this one specific](https://bugs.kde.org/show_bug.cgi?id=462574) to KDE but found some old ones in [gnome](https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/730)).

As a workaround (tried other things as well), might be not elegant, but this works fine on both X11/Xorg and Wayland:

```diff --git a/src/qt/transactionview.cpp b/src/qt/transactionview.cpp
--- a/src/qt/transactionview.cpp
+++ b/src/qt/transactionview.cpp
@@ -671,8 +671,11 @@ bool TransactionView::detailsAlreadyShown(const QModelIndex &idx)
 {
     for (TransactionDescDialog* dlg : m_opened_dialogs) {
         if (dlg->getTransactionId() == idx.data(TransactionTableModel::TxHashRole).toString()) {
-            dlg->activateWindow();
-            dlg->raise();
+            auto eFlags = dlg->windowFlags();
+            dlg->setWindowFlags(eFlags|Qt::WindowStaysOnTopHint);
+            dlg->show();
+            dlg->setWindowFlags(eFlags);
+            dlg->show();
             return true;
         }
     }
```

I can update the code with that snippet above or we can do it on a follow-up.

hebasto:
> @hebasto it seems there are some [issues](https://gitlab.freedesktop.org/wayland/wayland/-/issues/233) with Wayland on how these actions are being handled ([this one specific](https://bugs.kde.org/show_bug.cgi?id=462574) to KDE but found some old ones in [gnome](https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/730)).
>
> As a workaround (tried other things as well), might be not elegant, but this works fine on both X11/Xorg and Wayland:

The approach you suggested works for me.

Here are other references to this workaround:
- https://stackoverflow.com/questions/6087887/bring-window-to-front-raise-show-activatewindow-don-t-work/10808934
- https://forum.qt.io/topic/6032/bring-window-to-front-raise-show-activatewindow-don-t-work-on-windows

> I can update the code with that snippet above or we can do it on a follow-up.

I think we should first fix our `GUIUtil::bringToFront()` first in a separate PR, which will also address similar issues in other cases.

Then we can apply the fixed `GUIUtil::bringToFront()` in this PR.

pablomartin4btc:
> I think we should first fix our `GUIUtil::bringToFront()` first in a separate PR, which will also address similar issues in other cases.
> 
> Then we can apply the fixed `GUIUtil::bringToFront()` in this PR.

Ok, I'll open a new PR for `GUIUtil::bringToFront()` and will put this one on draft in the meantime.

pablomartin4btc:
Updates:
- No functional changes:
  - Using GUIUtil::bringToFront() that was recently fixed (and merged) for `Wayland`;
  - Removed unnecessary extra line in (`src/qt/transactiondescdialog.h`).

pablomartin4btc:
Updates:
- Rebased (fixed previous CI failure - _previous releases, depends DEBUG..._ ).

w0xlt:
> would be nice if a click on a row where the dialog is already open would bring that dialog to the foreground

I think this is a better approach.
Simply limiting dialog windows can be confusing, as the user may not understand why more windows are not opening,

pablomartin4btc:
> > would be nice if a click on a row where the dialog is already open would bring that dialog to the foreground
> 
> I think this is a better approach. Simply limiting dialog windows can be confusing, as the user may not understand why more windows are not opening,

Yup, that's [done](https://github.com/bitcoin-core/gui/pull/817#issuecomment-2051968829). I'll update the PR's description to make it clearer.

pablomartin4btc:
Temporarily drafting it since I've found some issues on Qt 6 while reviewing #904. WIP.

diegoviola:
> from the current application

Application window/surface.

diegoviola:
@pablomartin4btc I've just tried your code from this branch and it works fine on top of the current master, take a look:

https://github.com/user-attachments/assets/a6dd0b43-15c0-42de-b379-598a581ab44a

PR #818, opened 2024-04-24
guiconstants: update ORG_DOMAIN to bitcoincore.org

MarnixCroes:
if desired, can also update `ORG_NAME` to _Bitcoin Core_

### DIFF
--- a/src/qt/guiconstants.h
+++ b/src/qt/guiconstants.h
@@ -47,7 +47,7 @@ static const int TOOLTIP_WRAP_THRESHOLD = 80;
 #define SPINNER_FRAMES 36
 
 #define QAPP_ORG_NAME "Bitcoin"
-#define QAPP_ORG_DOMAIN "bitcoin.org"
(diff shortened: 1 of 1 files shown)


laanwj:
> if desired, can also update ORG_NAME to Bitcoin Core

The reason that this was never done is because it changes where the Qt settings files are stored. For example this is `~/.config/<ORG_NAME>/<APP_NAME>`. 

i'm not sure what actually is the effect of changing the domain, but this shouldn't be done without checking.

hebasto:
> if desired, can also update `ORG_NAME` to _Bitcoin Core_

Are there any issues with the current code? If not, then this PR lacks motivation.

MarnixCroes:
ok, it came across in the qt-qml repo, I looked here and assumed it was forgotten to be updated.
nvm

laanwj:
Right. i think qt-qml does something substantially different with regard to settings, so it may be warranted there.

PR #819, opened 2024-04-29
Fix misleading signmessage error with segwit

willcl-ark:
As described in https://github.com/bitcoin/bitcoin/issues/10542 (and numerous other places), message signing in Bitcoin Core does not support "signing with a segwit address" and likely will not in the foreseeable future, or at least until a new message-signing standard is agreed upon.

Therefore update the possibly misleading error message presented to the user in the GUI to detail more specifically the reason their message cannot be signed, in the case that a non P2PKH address is entered.

This change takes the [suggested wording](https://github.com/bitcoin/bitcoin/issues/10542#issuecomment-1960313569) from @adiabat.

Perhaps with this we can close https://github.com/bitcoin/bitcoin/issues/10542 ?

### DIFF
--- a/src/qt/signverifymessagedialog.cpp
+++ b/src/qt/signverifymessagedialog.cpp
@@ -10,6 +10,7 @@
 #include <qt/platformstyle.h>
 #include <qt/walletmodel.h>
 
+#include <config/bitcoin-config.h> // IWYU pragma: keep
 #include <key_io.h>
 #include <util/message.h> // For MessageSign(), MessageVerify()
 #include <wallet/wallet.h>
@@ -123,7 +124,7 @@ void SignVerifyMessageDialog::on_signMessageButton_SM_clicked()
     if (!pkhash) {
         ui->addressIn_SM->setValid(false);
         ui->statusLabel_SM->setStyleSheet("QLabel { color: red; }");
-        ui->statusLabel_SM->setText(tr("The entered address does not refer to a key.") + QString(" ") + tr("Please check the address and try again."));
+        ui->statusLabel_SM->setText(tr("The entered address does not refer to a legacy (P2PKH) key. Message signing for SegWit and other non-P2PKH address types is not supported in this version of %1. Please check the address and try again.").arg(PACKAGE_NAME));
         return;
     }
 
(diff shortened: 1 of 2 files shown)


MarnixCroes (commented):
cack

src/qt/signverifymessagedialog.cpp
```diff
@@ -123,7 +123,10 @@ void SignVerifyMessageDialog::on_signMessageButton_SM_clicked()
     if (!pkhash) {
         ui->addressIn_SM->setValid(false);
         ui->statusLabel_SM->setStyleSheet("QLabel { color: red; }");
-        ui->statusLabel_SM->setText(tr("The entered address does not refer to a key.") + QString(" ") + tr("Please check the address and try again."));
+        ui->statusLabel_SM->setText(
+            tr("The entered address does not refer to a legacy (P2PKH) key.") + QString(" ") +
+            tr("Message signing for segwit addresses (addresses starting with bc1/tb1) is not supported in this version of Bitcoin Core.") + QString(" ") +
```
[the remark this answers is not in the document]
willcl-ark: Taken all suggestions, thanks.

MarnixCroes (approved):
lgtm 83def1c5a3741878aa63e6f28cc3def99f76c358
useful clarification

BrandonOdiwuor (approved):
Code Review ACK 83def1c5a3741878aa63e6f28cc3def99f76c358

hebasto (commented):
Approach ACK 83def1c5a3741878aa63e6f28cc3def99f76c358 on the content of the messages.

However, splitting lines makes translation work harder due to providing less context. Could you make every touched message a single translatable string?

Also, if you'd like, add translation comment will be very useful to avoid any potential issue similar to [ (for example, https://github.com/bitcoin/bitcoin/commit/](https://github.com/bitcoin/bitcoin/issues/29979).

hebasto (commented):
And "Bitcoin Core" phrase should be replaced with `PACKAGE_NAME` (see https://github.com/bitcoin/bitcoin/pull/18646, https://github.com/bitcoin/bitcoin/pull/19282 etc).

willcl-ark:
Hi @hebasto I took your suggestions save for the extra context comments -- I think the text as one blob like this is pretty self-explanatory?

willcl-ark:
Sorry, I've addressed the linter now.

src/qt/signverifymessagedialog.cpp
```diff
@@ -10,6 +10,7 @@
 #include <qt/platformstyle.h>
 #include <qt/walletmodel.h>
 
+#include <config/bitcoin-config.h>
```
hebasto, line 13: It is https://github.com/bitcoin/bitcoin/pull/29494 goal to include this headers unguarded :)

Effectively, this PR conflicts with it. However, I believe it will be merged soon.

Anyway, it seems reasonable in this PR to use:
```c++
#if defined(HAVE_CONFIG_H)
#include <config/bitcoin-config.h>
#endif
```
or
```c++
#include <config/bitcoin-config.h> // IWYU pragma: keep
```

willcl-ark: Thanks hebasto, I hadn't seen that in progress. Went for the latter suggestion in fb9f150759b22772dd48983a2be1ea397245e289 as it seems to be preferred by marco in the other PR.

hebasto (approved):
ACK fb9f150759b22772dd48983a2be1ea397245e289.

PR #820, opened 2024-05-11
GUI: Enable customisation of QR Code font

luke-jr:
May need the fixes in #506 to work cleanly in all cases, but should still be a net improvement on its own


hebasto:
Concept ACK on adding a new customization option for the QR code font.

Lightly tested b14c9d0572eeae0fa286e0d2b96474a6f587b18e. It seems to work as expected.

However, I think that QR image widget must consider the font size for adjusting its own size. Otherwise, the image can be unreadable and confusing for the user:
image

The second commit modifies the code introduced in the first one. Therefore, please squash them.

PR #824, opened 2024-06-10
Migrate legacy wallets that are not loaded

achow101:
Currently the Migrate Wallet menu item can only be used to migrate the currently loaded wallet. This is not suitable for the future when legacy wallets can no longer be loaded at all, but should still be able to be migrated. This PR changes that menu item into a menu list like Open Wallet and lets users migrate any legacy wallet in their wallet directory regardless of the wallets loaded.

One issue I ran into was dealing with encrypted wallets. Ideally, we would detect whether a wallet is encrypted, and prompt the user for their passphrase at that time. However, that's actually difficult to do in the GUI since migration will unload the wallet if it was already loaded, and reload it without connecting it to any signals or interfaces. Only then can it detect whether a wallet is encrypted, but then there is no `WalletModel` or even an `interfaces::Wallet` that the GUI could use to unlock it via a callback.

To deal with this, I've opted to just add a button to the migration dialog box that has the user enter their passphrase first, along with instructional text to use that button if their wallet was encrypted. If the user enters the wrong passphrase or clicked the other button that does not prompt for the passphrase, migration will fail with a message indicating that the passphrase was incorrect.

### DIFF
--- a/src/interfaces/wallet.h
+++ b/src/interfaces/wallet.h
@@ -343,7 +343,7 @@ public:
     virtual util::Result<WalletMigrationResult> migrateWallet(const std::string& name, const SecureString& passphrase) = 0;
 
     //! Return available wallets in wallet directory.
-    virtual std::vector<std::string> listWalletDir() = 0;
+    virtual std::vector<std::pair<std::string, std::string>> listWalletDir() = 0;
 
     //! Return interfaces for accessing wallets (if any).
     virtual std::vector<std::unique_ptr<Wallet>> getWallets() = 0;
--- a/src/interfaces/wallet.h
+++ b/src/interfaces/wallet.h
@@ -342,6 +342,9 @@ public:
     //! Migrate a wallet
     virtual util::Result<WalletMigrationResult> migrateWallet(const std::string& name, const SecureString& passphrase) = 0;
 
+    //! Returns true if wallet stores encryption keys
+    virtual bool isEncrypted(const std::string& wallet_name) = 0;
+
     //! Return available wallets in wallet directory.
     virtual std::vector<std::pair<std::string, std::string>> listWalletDir() = 0;
 
--- a/src/qt/askpassphrasedialog.h
+++ b/src/qt/askpassphrasedialog.h
@@ -26,6 +26,7 @@ public:
         Encrypt,    /**< Ask passphrase twice and encrypt */
         Unlock,     /**< Ask passphrase and unlock */
         ChangePass, /**< Ask old passphrase + new passphrase twice */
+        UnlockMigration, /**< Ask passphrase for unlocking during migration */
     };
 
     explicit AskPassphraseDialog(Mode mode, QWidget *parent, SecureString* passphrase_out = nullptr);
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -398,7 +398,7 @@ void BitcoinGUI::createActions()
             m_open_wallet_menu->clear();
             for (const auto& [path, info] : m_wallet_controller->listWalletDir()) {
                 const auto& [loaded, _] = info;
-                QString name = path.empty() ? QString("["+tr("default wallet")+"]") : QString::fromStdString(path);
+                QString name = GUIUtil::WalletDisplayName(path);
                 // An single ampersand in the menu item's text sets a shortcut for this item.
                 // Single & are shown when && is in the string. So replace & with &&.
                 name.replace(QChar('&'), QString("&&"));
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -396,15 +396,15 @@ void BitcoinGUI::createActions()
         connect(openAction, &QAction::triggered, this, &BitcoinGUI::openClicked);
         connect(m_open_wallet_menu, &QMenu::aboutToShow, [this] {
             m_open_wallet_menu->clear();
-            for (const std::pair<const std::string, bool>& i : m_wallet_controller->listWalletDir()) {
-                const std::string& path = i.first;
+            for (const auto& [path, info] : m_wallet_controller->listWalletDir()) {
+                const auto& [loaded, _] = info;
                 QString name = path.empty() ? QString("["+tr("default wallet")+"]") : QString::fromStdString(path);
                 // An single ampersand in the menu item's text sets a shortcut for this item.
                 // Single & are shown when && is in the string. So replace & with &&.
                 name.replace(QChar('&'), QString("&&"));
                 QAction* action = m_open_wallet_menu->addAction(name);
 
-                if (i.second) {
+                if (loaded) {
                     // This wallet is already loaded
                     action->setEnabled(false);
                     continue;
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -360,6 +360,7 @@ void BitcoinGUI::createActions()
     m_migrate_wallet_action = new QAction(tr("Migrate Wallet"), this);
     m_migrate_wallet_action->setEnabled(false);
     m_migrate_wallet_action->setStatusTip(tr("Migrate a wallet"));
+    m_migrate_wallet_menu = new QMenu(this);
 
     showHelpMessageAction = new QAction(tr("&Command-line options"), this);
     showHelpMessageAction->setMenuRole(QAction::NoRole);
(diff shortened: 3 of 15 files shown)


hebasto:
Concept ACK.

luke-jr:
>To deal with this, I've opted to just add a button to the migration dialog box that has the user enter their passphrase first, along with instructional text to use that button if their wallet was encrypted. 

If we're preemptively making UX changes just to avoid changing UX in the future, it seems like we shouldn't take shortcuts like this... :/

achow101:
> > To deal with this, I've opted to just add a button to the migration dialog box that has the user enter their passphrase first, along with instructional text to use that button if their wallet was encrypted.
> 
> If we're preemptively making UX changes just to avoid changing UX in the future, it seems like we shouldn't take shortcuts like this... :/

We already do this in the RPC too, for exactly the same reason.

src/qt/bitcoingui.cpp
```diff
@@ -456,10 +457,33 @@ void BitcoinGUI::createActions()
...
-            activity->migrate(walletFrame->currentWalletModel());
+        connect(m_migrate_wallet_menu, &QMenu::aboutToShow, [this] {
+            m_migrate_wallet_menu->clear();
+            for (const auto& [wallet_name, info] : m_wallet_controller->listWalletDir()) {
+                const auto& [loaded, format] = info;
+                QString name = GUIUtil::WalletDisplayName(wallet_name);
+                // Menu items remove single &. Single & are shown when && is in
+                // the string, but only the first occurrence. So replace only
+                // the first & with &&.
+                name.replace(name.indexOf(QChar('&')), 1, QString("&&"));
```
hebasto, line 468: https://github.com/bitcoin-core/gui/pull/828 can relevant to this code.

src/qt/askpassphrasedialog.h
```diff
@@ -26,6 +26,7 @@ class AskPassphraseDialog : public QDialog
         Encrypt,    /**< Ask passphrase twice and encrypt */
         Unlock,     /**< Ask passphrase and unlock */
         ChangePass, /**< Ask old passphrase + new passphrase twice */
+        UnlockMigration, // Ask passphrase for unlocking during migration
```
hebasto, line 29: Mind following the Doxygen comment formatting above?

src/interfaces/wallet.h
```diff
@@ -46,6 +46,7 @@ enum isminetype : unsigned int;
 struct CRecipient;
 struct WalletContext;
 using isminefilter = std::underlying_type<isminetype>::type;
+enum class DatabaseFormat;
```
hebasto, line 49: Why is this line needed?

hebasto:
> This PR changes that menu item into a menu list like Open Wallet and lets users migrate any legacy wallet in their wallet directory regardless of the wallets loaded.

Tested d45ac03a89cc500cd461f878f092cf8ec99e7760. The "Migrate Wallet" menu item is still disabled when no wallet is loaded.

hebasto:
Please rebase.

src/qt/bitcoingui.cpp
pablomartin4btc: Since #828 has been merged, please keep in mind this update...
```suggestion
                // Single & are shown when && is in the string. So replace & with &&.
                name.replace(QChar('&'), QString("&&"));
```

pablomartin4btc (commented):
tACK d45ac03a89cc500cd461f878f092cf8ec99e7760

<details>
<summary>Created a couple of legacy wallets (encrypted and not) to begin with.</summary>

 <ins>**edit**</ins>: running ./configure with: `--with-bdb`.

Using `bitcoind` (`./src/bitcoind -regtest -deprecatedrpc=create_bdb
`) and `bitcoin-cli` (`./src/bitcoin-cli -regtest createwallet legacyWallet false false "" false false
`) or directly on the `rpc-console` (`createwallet legacyWalletEnc false false "passphrase" false false`) in `bitcoin-qt` (`./src/qt/bitcoin-qt -regtest -deprecatedrpc=create_bdb`).

```
{
  "name": "legacyWalletEnc",
  "warnings": [
    "Wallet created successfully. The legacy wallet type is being deprecated and support for creating and opening legacy wallets will be removed in the future."
  ]
}
```

```
{
  "name": "legacyWallet",
  "warnings": [
    "Empty string given as passphrase, wallet will not be encrypted.",
    "Wallet created successfully. The legacy wallet type is being deprecated and support for creating and opening legacy wallets will be removed in the future."
  ]
}
```

</details>

<details>
<summary>The menu item "Migrate Wallet" is shown when there are already wallets loaded.</summary>

image

</details>

<details>
<summary>But "Migrate Wallet" is disabled when there are no wallets loaded as @hebasto mentioned <a href="https://github.com/bitcoin-core/gui/pull/824#issuecomment-2227310588">above</a>.</summary>

image

</details>

- I think that should be solved by removing the migration action from `BitcoinGUI::setWalletActionsEnabled` (the migration action has the same behaviour as the restore action which is not there).

<details>
<summary>"Migrate Wallet" is shown when there are no wallets available at all.</summary>

image

</details>


<details>
<summary>Tested migration of encrypted wallet with and without providing the passphrase, working both cases.</summary>

image

image

</details>

<details>
<summary>Every migration failure on a particular wallet still creates a backup file. Do we want this?</summary>

image
</details>

src/qt/bitcoingui.cpp
achow101: Fixed during rebase.

src/qt/askpassphrasedialog.h
achow101: Done

src/interfaces/wallet.h
achow101: Removed

achow101:
> Tested [d45ac03](https://github.com/bitcoin-core/gui/commit/d45ac03a89cc500cd461f878f092cf8ec99e7760). The "Migrate Wallet" menu item is still disabled when no wallet is loaded.

Should be fixed now.

pablomartin4btc:
> > "Migrate Wallet" is shown when there are no wallets available at all.
> 
> I think that's ok.

Yeah, it was just an observation (mainly to compare when the menu is disabled), thanks!

pablomartin4btc (approved):
> > Tested [d45ac03](https://github.com/bitcoin-core/gui/commit/d45ac03a89cc500cd461f878f092cf8ec99e7760). The "Migrate Wallet" menu item is still disabled when no wallet is loaded.
> 
> Should be fixed now.
> 

re-ACK bcf47b9f34e4f21e5ac75521ddd4f25ff3812412

src/qt/walletcontroller.cpp
```diff
@@ -436,47 +445,50 @@ void RestoreWalletActivity::finish()
...
+        AskPassphraseDialog dlg(AskPassphraseDialog::UnlockMigration, m_parent_widget, &passphrase);
+        if (dlg.exec() == QDialog::Rejected) return;
     }
 
     // GUI needs to remove the wallet so that it can actually be unloaded by migration
-    const std::string name = wallet_model->wallet().getWalletName();
-    m_wallet_controller->removeAndDeleteWallet(wallet_model);
+    if (WalletModel* wallet = m_wallet_controller->getWallet(name)) {
+        m_wallet_controller->removeAndDeleteWallet(wallet);
+    };
```
furszy, line 478: We don't actually need to do this. Could cherry-pick https://github.com/furszy/bitcoin-core/commit/f5c0c58977a8989fe17db361d179281d051b0806 and remove the introduced `getWallet`.

furszy (commented):
Few comments. Will continue reviewing.

src/qt/bitcoingui.cpp
```diff
@@ -455,10 +456,33 @@ void BitcoinGUI::createActions()
...
+                if (format != "bdb") {
+                    // This wallet is already migrated
+                    action->setEnabled(false);
+                    continue;
+                }
+
+                connect(action, &QAction::triggered, [this, wallet_name] {
+                    auto activity = new MigrateWalletActivity(m_wallet_controller, this);
+                    connect(activity, &MigrateWalletActivity::migrated, this, &BitcoinGUI::setCurrentWallet);
+                    activity->migrate(wallet_name);
```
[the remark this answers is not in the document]
achow101: What compiler? I'm not seeing this error, and it seems neither did CI.

src/qt/bitcoingui.cpp
```diff
@@ -455,10 +456,33 @@ void BitcoinGUI::createActions()
...
+                // the string, but only the first occurrence. So replace only
+                // the first & with &&.
+                name.replace(name.indexOf(QChar('&')), 1, QString("&&"));
+                QAction* action = m_migrate_wallet_menu->addAction(name);
+
+                if (format != "bdb") {
+                    // This wallet is already migrated
+                    action->setEnabled(false);
+                    continue;
+                }
```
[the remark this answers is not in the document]
achow101: Done

src/qt/walletcontroller.cpp
achow101: Cherry picked.

src/qt/bitcoingui.cpp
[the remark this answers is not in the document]
furszy: > What compiler? I'm not seeing this error, and it seems neither did CI.

clang 14.0.3 (clang-1403.0.22.14.1).

achow101:
> Could replace it directly for [furszy/bitcoin-core@e8d147e](https://github.com/furszy/bitcoin-core/commit/e8d147eeaacac7d434aad8dbbc5022aadf7d112a). It detects the wallet encryption status automatically regardless of its loaded/unloaded state.

Cherry picked.

Also fixed test failure.

furszy (commented):
ACK 0240454120a88915396b0eb68d9b5bee73e52fed

src/qt/bitcoingui.cpp
hebasto: > Fixed during rebase.

It still needs to be fixed.

hebasto (commented):
Almost ACK fd087d4667b94051d00617189dcecc4f88d849ce.

src/qt/bitcoingui.cpp
achow101: Ah, missed one, fixed.

hebasto (approved):
ACK 8f2522d242961ceb9e79672aa43e856863a1a6dd.

hebasto:
cc @furszy

furszy (commented):
ACK 8f2522d

src/qt/bitcoingui.cpp
```diff
@@ -455,10 +456,31 @@ void BitcoinGUI::createActions()
...
+                QAction* action = m_migrate_wallet_menu->addAction(name);
+
+                connect(action, &QAction::triggered, [this, wallet_name] {
+                    auto activity = new MigrateWalletActivity(m_wallet_controller, this);
+                    connect(activity, &MigrateWalletActivity::migrated, this, &BitcoinGUI::setCurrentWallet);
+                    activity->migrate(wallet_name);
+                });
+            }
+            if (m_migrate_wallet_menu->isEmpty()) {
+                QAction* action = m_migrate_wallet_menu->addAction(tr("No wallets available"));
```
MarnixCroes, line 481: nit: maybe a more descriptive text would be good?
something like `No wallets that need to be migrated`

for people who don't know about the migration and why/what it is.
seeing _No wallets available_ while they have wallet(s) might seem wrong/confusing.

sry for commenting post-merge

pablomartin4btc (commented):
post-merge tACK 8f2522d242961ceb9e79672aa43e856863a1a6dd

Checked changes proposed by @furszy.

<details>
<summary>Perhaps the PR's description needs to be updated (or a note added) since the previous extra button to enter the passphrase was removed (<code>isEncrypted</code> is in place). Nice solution, less prone to errors too.</summary>

image

</details>

PR #826, opened 2024-07-06
OptionsDialog: Allow Maximize of window

luke-jr:


hebasto:
The PR description is empty, so I have to ask: why is this needed?

luke-jr:
Perhaps not strictly needed, more of it's a bug that the user is prevented from doing so

hebasto (approved):
ACK 3dbd94b6610a58460b52a6cc02892d881c091ccd.

PR #828, opened 2024-07-13
Rendering an amp characters in the wallet name for QMenu

knst:
In the current implementation Qt uses '&' as a signal to underscore letter and use it as a hot-key, which is not expected for case of wallet name.

The [comment in the code](https://github.com/bitcoin/bitcoin/pull/30446/files#diff-2ecf8cbf369cf3d2f3d2b1cf5cfe4c1a647d63e11e2885d2fd0ac11fb5f7a804L402-L404) regarding the use of an "&" on a menu item is misleading.
If a wallet name has an "&" in it, it is not supposed to be interpreted as a hot-key, but it should be shown as it is without replacing it to an underscore.

See screenshots before & after:
Screenshot_20240713_122454
Screenshot_20240713_131304

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -399,10 +399,9 @@ void BitcoinGUI::createActions()
             for (const std::pair<const std::string, bool>& i : m_wallet_controller->listWalletDir()) {
                 const std::string& path = i.first;
                 QString name = path.empty() ? QString("["+tr("default wallet")+"]") : QString::fromStdString(path);
-                // Menu items remove single &. Single & are shown when && is in
-                // the string, but only the first occurrence. So replace only
-                // the first & with &&.
-                name.replace(name.indexOf(QChar('&')), 1, QString("&&"));
+                // An single ampersand in the menu item's text sets a shortcut for this item.
+                // Single & are shown when && is in the string. So replace & with &&.
+                name.replace(QChar('&'), QString("&&"));
                 QAction* action = m_open_wallet_menu->addAction(name);
 
                 if (i.second) {


hebasto:
> In the current implementation Qt uses '&' as a signal to underscore letter and use it as a hot-key, which is not expected for case of wallet name.

Does this hold for all [supported](https://github.com/bitcoin-core/gui/blob/master/doc/dependencies.md) Qt versions?

> 
> The [comment in the code](https://github.com/bitcoin/bitcoin/pull/30446/files#diff-2ecf8cbf369cf3d2f3d2b1cf5cfe4c1a647d63e11e2885d2fd0ac11fb5f7a804L402-L404) is misleading, if you replace one & to double &&, the next & will be a "single one". All of them in the string should be replaced, not only the first occurrence.

I'm not certain about that, but the mentioned comment might have been correct for some earlier version of Qt.

knst:
> I'm not certain about that, but the mentioned comment might have been correct for some earlier version of Qt.
I tested with 2 version: Qt 5.15.14 and Qt 5.15.11.

Doc for [qt 6](https://doc.qt.io/qt-6/qmenubar.html) says:
> The ampersand in the menu item's text sets Alt+F as a shortcut for this menu. (You can use "&&" to get a real ampersand in the menu bar.)

Doc for [qt 5](https://doc.qt.io/qt-5/qmenubar.html) says the same:
> The ampersand in the menu item's text sets Alt+F as a shortcut for this menu. (You can use "&&" to get a real ampersand in the menu bar.)

And even doc for [qt 4](https://het.as.utexas.edu/HET/Software/html/qmenubar.html) says the same:
> The ampersand in the menu item's text sets Alt+F as a shortcut for this menu. (You can use "&&" to get a real ampersand in the menu bar.)

I believe that in original PR introduced a ["partial fix"](https://github.com/bitcoin/bitcoin/pull/16826) which works for the first ampersand no one just tested a wallet with several ampersands.

hebasto:
> I believe that in original PR introduced a ["partial fix"](https://github.com/bitcoin/bitcoin/pull/16826) which works for the first ampersand no one just tested a wallet with several ampersands.

It might be a platform-specific issue though. See https://bugreports.qt.io/browse/QTBUG-63361.

src/qt/bitcoingui.cpp
```diff
@@ -399,10 +399,7 @@ void BitcoinGUI::createActions()
             for (const std::pair<const std::string, bool>& i : m_wallet_controller->listWalletDir()) {
                 const std::string& path = i.first;
                 QString name = path.empty() ? QString("["+tr("default wallet")+"]") : QString::fromStdString(path);
-                // Menu items remove single &. Single & are shown when && is in
-                // the string, but only the first occurrence. So replace only
-                // the first & with &&.
```
hebasto, line 404: I think that the comment should be adjusted rather than being deleted entirely.

hebasto (commented):
Tested 948520b72d7fe8e7a3aa0d774159807058188627 on Ubuntu 24.04 with Qt 5.15.3. It works as expected.

Going to test other platforms.

hebasto:
cc @achow101 @jarolrod

src/qt/bitcoingui.cpp
knst: Could you suggest wording? English is not my native language

knst:
> It might be a platform-specific issue though. See https://bugreports.qt.io/browse/QTBUG-63361.

could someone test on Mac?
I will make probably platform's depend implementation of char replacement

src/qt/bitcoingui.cpp
hebasto: Maybe:
```C++
                // Menu items remove single &. Single & are shown when && is in
                // the string. So replace & with &&.
```
?

hebasto (approved):
ACK 948520b72d7fe8e7a3aa0d774159807058188627, module https://github.com/bitcoin-core/gui/pull/828/files#r1677136023.

Tested on Ubuntu 24.04, Windows 11 Pro and macOS 12.7.5 Monterey (x86).

src/qt/bitcoingui.cpp
```diff
@@ -399,10 +399,9 @@ void BitcoinGUI::createActions()
...
-                name.replace(name.indexOf(QChar('&')), 1, QString("&&"));
+                // The single ampersand in the menu item's text sets a shortcut for this menu.
```
hebasto, line 402: Maybe
```suggestion
                // The single ampersand in the menu item's text sets a shortcut for this item.
```
?

hebasto (approved):
re-ACK 7581817826e000c4f0a505e7f2785a611cac8f55

pablomartin4btc (commented):
Concept ACK - I'll test it soon.

There's a typo in the commit body ("missleading"), since you are there, perhaps a better prefix for its title could be "`gui:`" and a <ins>nit</ins> on the text could be:

```
The comment in the code regarding the use of an "&"
on a menu item is misleading. If a wallet name has an "&" in it,
it is not supposed to be interpreted as a hot-key, but it should be
shown as it is without replacing it to an underscore.
```

hebasto (approved):
re-ACK 8233ee41ab9648cd0c3bd78bc2a8d692a54d9ea0.

CI failures are unrelated.

1alexbc1 (approved):

src/qt/bitcoingui.cpp
pablomartin4btc: I agree with @hebasto. Small nit:
```
                // Menu items remove single &. Single & is shown when && are in
                // the string. So replace & with &&.
```

src/qt/bitcoingui.cpp
```diff
@@ -399,10 +399,9 @@ void BitcoinGUI::createActions()
             for (const std::pair<const std::string, bool>& i : m_wallet_controller->listWalletDir()) {
                 const std::string& path = i.first;
                 QString name = path.empty() ? QString("["+tr("default wallet")+"]") : QString::fromStdString(path);
-                // Menu items remove single &. Single & are shown when && is in
-                // the string, but only the first occurrence. So replace only
-                // the first & with &&.
-                name.replace(name.indexOf(QChar('&')), 1, QString("&&"));
+                // An single ampersand in the menu item's text sets a shortcut for this item.
```
pablomartin4btc, line 402: In case the above suggestion is not taken, typo here:
```suggestion
                // A single ampersand in the menu item's text sets a shortcut for this item.
```

pablomartin4btc (approved):
tACK 8233ee41ab9648cd0c3bd78bc2a8d692a54d9ea0

<details>
<summary>It corrects the bug in master.</summary>

- before
image


- after
image


</details>

BrandonOdiwuor (approved):
ACK 8233ee41ab9648cd0c3bd78bc2a8d692a54d9ea0. Tested on Ubuntu 22.04 using Qt version 5.15.3

PR #831, opened 2024-07-30
GUIUtil::brintToFront workaround for Wayland

pablomartin4btc:
There are known issues around handling windows focus in `Wayland` ([this one specific](https://bugs.kde.org/show_bug.cgi?id=462574) in KDE but also in [gnome](https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/730)).

The idea is that the workaround will be executed if `bitcoin-qt` is running using `Wayland` platform (e.g.: `QT_QPA_PLATFORM=wayland ./src/qt/bitcoin-qt -regtest`), since the workaround behaviour looks like re-opening the window again (which I tried to fix by moving the window to the original position and/ or re-setting the original geometry without success) while in `X11` (not sure in Mac) the current `GUIUtil::brintToFront` actually sets the focus to the desired window, keeping its original position as expected, and I didn't want to change that (`X11` behaviour).

The solution was [initially discussed](https://github.com/bitcoin-core/gui/pull/817#issuecomment-2256158902) with hebasto in #817.

### DIFF
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -406,19 +406,26 @@ bool isObscured(QWidget *w)
 
 void bringToFront(QWidget* w)
 {
-#ifdef Q_OS_MACOS
-    ForceActivation();
-#endif
-
     if (w) {
-        // activateWindow() (sometimes) helps with keyboard focus on Windows
-        if (w->isMinimized()) {
-            w->showNormal();
-        } else {
+        if (QGuiApplication::platformName() == "wayland") {
+            auto flags = w->windowFlags();
+            w->setWindowFlags(flags|Qt::WindowStaysOnTopHint);
             w->show();
+            w->setWindowFlags(flags);
+            w->show();
+        } else {
+#ifdef Q_OS_MACOS
+            ForceActivation();
+#endif
+            // activateWindow() (sometimes) helps with keyboard focus on Windows
+            if (w->isMinimized()) {
+                w->showNormal();
+            } else {
+                w->show();
+            }
+            w->activateWindow();
+            w->raise();
         }
-        w->activateWindow();
-        w->raise();
     }
 }
 


hebasto:
Concept ACK.

src/qt/guiutil.cpp
```diff
@@ -1007,4 +1009,15 @@ void ShowModalDialogAsynchronously(QDialog* dialog)
     dialog->show();
 }
 
+void bringToFrontWorkaroundForWayland(QWidget* w)
```
hebasto, line 1012: Any specific reason to not inline this code?

src/qt/guiutil.cpp
```diff
@@ -1007,4 +1009,15 @@ void ShowModalDialogAsynchronously(QDialog* dialog)
...
+{
+    if (QGuiApplication::platformName() == "wayland") {
+        auto eFlags = w->windowFlags();
```
hebasto, line 1015: The variable name looks confusing to me. Why not just "flags"?

pablomartin4btc: Yeah, I'll change it.

src/qt/guiutil.cpp
pablomartin4btc: Just to have it separately, also in case we need to touch it again.

hebasto: This makes diff bigger, you need to touch the header. Also, on Wayland, the `activateWindow()` and `raise()` can be unneeded.

pablomartin4btc: Ok, I'll add a conditional there then.

src/qt/guiutil.cpp
```diff
@@ -411,14 +411,22 @@ void bringToFront(QWidget* w)
...
-            w->showNormal();
-        } else {
+        if (QGuiApplication::platformName() == "wayland") {
+            auto flags = w->windowFlags();
+            w->setWindowFlags(flags|Qt::WindowStaysOnTopHint);
+            w->show();
+            w->setWindowFlags(flags);
             w->show();
+        } else {
+            // activateWindow() (sometimes) helps with keyboard focus on Windows
```
hebasto, line 421: Shouldn't macOS-specific `ForceActivation();` go in this branch as well?

pablomartin4btc: Yeah, it makes sense.

hebasto (approved):
ACK 15aa7d023688700a47997b92108de95f2d864f5a.

Tested on Ubuntu 24.04. The behavior with `QT_QPA_PLATFORM=wayland`  has become more consistent with the behavior with `QT_QPA_PLATFORM=xcb`.

Tested on macOS 12.7.6 Monterey. No behavior changes observed.

PR #832, opened 2024-07-30
gui: improve user dialog when signing multisig psbts

mjdietzx:
Prior to this commit, when a psbt requiring multiple signers is signed via gui the dialogue is vague/confusing. Whether the signer successfully partially signs the psbt or not the same warning is displayed "Could not sign any more inputs." After signing the tx, the only way for the user to know if their action had any effect is to inspect the psbt (unless it was the last signature that completed the psbt, in which case there is a success dialoge).

Now we indicate when a psbt has been partially signed, and show how many partial signatures the psbt has in the transaction description.

Here is what a "daisy chain" signing flow for a 4-of-4 multisig looks like via gui. Same as in bitcoin core's functional test [wallet_multisig_descriptor_psbt](https://github.com/bitcoin/bitcoin/blob/d367a4e36f7357c4ebd018e8e1c9c5071db2e1c2/test/functional/wallet_multisig_descriptor_psbt.py#L140-L147)

First signer: 
image

Second signer: 
image

Third signer: 
image

Fourth and final signer (psbt is complete and can be broadcasted): 
image

Before this commit, the first-third signers would just see this after signing. It is at best confusing because it is the same warning a signer would see if they were not in the multisig and no signature was added to the psbt:
image


hebasto:
cc @achow101 @furszy

src/qt/psbtoperationsdialog.cpp
```diff
@@ -217,9 +234,17 @@ QString PSBTOperationsDialog::renderTransaction(const PartiallySignedTransaction
...
+    const size_t num_partial_sigs{psbtx.inputs.at(0).partial_sigs.size()};
+    const bool all_inputs_n_partial_sigs_equal{std::all_of(psbtx.inputs.begin(), psbtx.inputs.end(),
+        [&](const PSBTInput& input) { return num_partial_sigs == input.partial_sigs.size(); })};
+
     if (num_unsigned > 0) {
         tx_description.append("<br><br>");
         tx_description.append(tr("Transaction has %1 unsigned inputs.").arg(QString::number(num_unsigned)));
+        if (all_inputs_n_partial_sigs_equal && num_partial_sigs > 0) {
+            tx_description.append(" ");
+            tx_description.append(tr("Each input has %1 partial signatures.").arg(QString::number(num_partial_sigs)));
```
pablomartin4btc, line 246: optional nit (IMO makes it more clear that includes the just added partial signature):
```suggestion
            tx_description.append(tr("Each input now has %1 partial signatures.").arg(QString::number(num_partial_sigs)));
```

pablomartin4btc (commented):
Concept ACK

pablomartin4btc:
It would be interesting to see a PSBT window's output sample when the user signed an input but couldn't sign another input which is also part of the transaction? (Is that possible?)

src/qt/psbtoperationsdialog.cpp
mjdietzx: Before you sign, if there are already partial sigs on a psbt, this text will show that. For example, 3rd signer in a multisig:
- before signing `tx_description` tells them there are already 2 partial sigs
- they sign and see they have successfully added a partial signature. `tx_description` now tells them there are 3 partial sigs
- if they were to try to sign again, then they'd see the warning "Could not sign any more inputs." at the top of dialog, but `tx_description` would still show there are 3 psbts

Therefore I think `now` should not be added

mjdietzx:
> It would be interesting to see a PSBT window's output sample when the user signed an input but couldn't sign another input which is also part of the transaction? (Is that possible?)

It's possible but I considered it outside the scope of this PR -  in such cases behavior is not changed by this PR. Because i want to keep this simple (seems trying to handle cases like this and communicate it to the user in a straightforward manner would be complicated).

In this PR I just improve try to improve the UX for "normal" multisigs like [wallet_multisig_descriptor_psbt.py](https://github.com/bitcoin/bitcoin/blob/master/test/functional/wallet_multisig_descriptor_psbt.py) and https://github.com/bitcoin/bitcoin/pull/29156 - I don't think it can ever be the case that one input is signed but another is not, and I'd expect this to be true for vast majority of wallets where users are using bitcoin core's gui to sign psbts

All that said, interested to see what reviewers with more experience than me think

w0xlt:
Concept ACK.
Clear improvement in UX.

PR #837, opened 2024-09-18
qt6: Fix linking when configured with `-DENABLE_WALLET=OFF`

hebasto:
Split from https://github.com/bitcoin/bitcoin/pull/30997.

When building with Qt 6 in my dev branch, I encountered a linker error when configured with `-DENABLE_WALLET=OFF`:
```
$ cmake -B build -DENABLE_WALLET=OFF -DBUILD_GUI=ON
$ cmake --build build -t bitcoin-qt
<snip>
[100%] Linking CXX executable bitcoin-qt
/usr/bin/ld: libbitcoinqt.a(rpcconsole.cpp.o): in function `QtPrivate::MetaObjectForType<WalletModel const*, void>::metaObjectFunction(QtPrivate::QMetaTypeInterface const*)':
/usr/include/x86_64-linux-gnu/qt6/QtCore/qmetatype.h:903:(.text._ZN9QtPrivate17MetaObjectForTypeIPK11WalletModelvE18metaObjectFunctionEPKNS_18QMetaTypeInterfaceE[_ZN9QtPrivate17MetaObjectForTypeIPK11WalletModelvE18metaObjectFunctionEPKNS_18QMetaTypeInterfaceE]+0x2b): undefined reference to `WalletModel::staticMetaObject'
/usr/bin/ld: libbitcoinqt.a(rpcconsole.cpp.o): in function `QMetaTypeIdQObject<WalletModel const*, 8>::qt_metatype_id()':
/usr/include/x86_64-linux-gnu/qt6/QtCore/qmetatype.h:1313:(.text._ZZN9QtPrivate16QMetaTypeForTypeIPK11WalletModelE17getLegacyRegisterEvENUlvE_4_FUNEv[_ZZN9QtPrivate16QMetaTypeForTypeIPK11WalletModelE17getLegacyRegisterEvENUlvE_4_FUNEv]+0x53): undefined reference to `WalletModel::staticMetaObject'
collect2: error: ld returned 1 exit status
gmake[3]: *** [src/qt/CMakeFiles/bitcoin-qt.dir/build.make:154: src/qt/bitcoin-qt] Error 1
gmake[2]: *** [CMakeFiles/Makefile2:2107: src/qt/CMakeFiles/bitcoin-qt.dir/all] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:2114: src/qt/CMakeFiles/bitcoin-qt.dir/rule] Error 2
gmake: *** [Makefile:998: bitcoin-qt] Error 2
```

This PR resolves the issue.

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -16,7 +16,9 @@
 #include <qt/guiutil.h>
 #include <qt/peertablesortproxy.h>
 #include <qt/platformstyle.h>
+#ifdef ENABLE_WALLET
 #include <qt/walletmodel.h>
+#endif // ENABLE_WALLET
 #include <rpc/client.h>
 #include <rpc/server.h>
 #include <util/strencodings.h>


fanquake:
What is the error? Why does it only happen with Qt6?

hebasto:
> What is the error?

I put it to the PR description.

> Why does it only happen with Qt6?

Differences in Qt implementation between Qt 5 and Qt 6? (did not dig into the exact code change though)

src/qt/rpcconsole.cpp
```diff
@@ -16,7 +16,9 @@
 #include <qt/guiutil.h>
 #include <qt/peertablesortproxy.h>
 #include <qt/platformstyle.h>
+#ifdef ENABLE_WALLET
 #include <qt/walletmodel.h>
+#endif // ENABLE_WALLET
```
furszy, line 21: How the `RPCExecutor::request` method compiles when this line doesn't exist?

pablomartin4btc (approved):
tACK 5be34bacf6d07fb73a0cedfb63a384968d538f3e

It builds correctly on Qt5 and qt runs fine, with both `-DENABLE_WALLET=OFF` & `-DENABLE_WALLET=ON`. When `wallet support` is `off` also tested the rpc console and several rpc commands.

src/qt/rpcconsole.cpp
pablomartin4btc: I think because the compiler only needs to know WalletModel, it doesn't need the full definition unless you dereference the pointer (eg as in `RPCConsole::RPCParseCommandLine` -> `wallet_model->getWalletName()` - which is guarded).

pablomartin4btc (commented):
utACK 33657e1c958146312e4c68765a92871920401396

Guarded all `WalletModel` references (and overriding functions without it) in `rpcconsole`.

promag (commented):
I'll try a different approach.

promag:
ACK 5be34bacf6d07fb73a0cedfb63a384968d538f3e, all other changes in 33657e1c958146312e4c68765a92871920401396 are not required, makes the code harder to read.

src/qt/rpcconsole.cpp
hebasto: > Reworked.

Reverted back.

> How the `RPCExecutor::request` method compiles when this line doesn't exist?

Due to https://github.com/bitcoin-core/gui/blob/5be34bacf6d07fb73a0cedfb63a384968d538f3e/src/qt/rpcconsole.h#L24

furszy:
Arrived a bit late, but see #841. It drops `WalletModel` linkage and removes an extra `#ifdef ENABLE_WALLET` in the RPC executor class.

PR #840, opened 2024-10-04
qt6: Handle different signatures of `QANEF::nativeEventFilter`

hebasto:
Split from https://github.com/bitcoin/bitcoin/pull/30997.

This PR ensures compatibility across all supported Qt versions.

For more details, please refer to https://github.com/qt/qtbase/commit/3b38c73c7ffa71c00c172cf0e05742835a304300.

No behaviour change.

### DIFF
--- a/src/qt/winshutdownmonitor.cpp
+++ b/src/qt/winshutdownmonitor.cpp
@@ -12,7 +12,11 @@
 
 // If we don't want a message to be processed by Qt, return true and set result to
 // the value that the window procedure should return. Otherwise return false.
+#if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))
+bool WinShutdownMonitor::nativeEventFilter(const QByteArray &eventType, void *pMessage, qintptr *pnResult)
+#else
 bool WinShutdownMonitor::nativeEventFilter(const QByteArray &eventType, void *pMessage, long *pnResult)
+#endif
 {
        Q_UNUSED(eventType);
 
--- a/src/qt/winshutdownmonitor.h
+++ b/src/qt/winshutdownmonitor.h
@@ -20,7 +20,11 @@ public:
     WinShutdownMonitor(std::function<void()> shutdown_fn) : m_shutdown_fn{std::move(shutdown_fn)} {}
 
     /** Implements QAbstractNativeEventFilter interface for processing Windows messages */
+#if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))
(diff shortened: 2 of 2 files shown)


fanquake:
If Qt 6.7.0 is going to be required for Windows after the switchover; why change this code now just to immediately drop it after migrating to Qt 6?

hebasto:
> If Qt 6.7.0 is going to be required for Windows after the switchover; why change this code now just to immediately drop it after migrating to Qt 6?

To make https://github.com/bitcoin/bitcoin/pull/30997 focused on the build system changes only.

We already have:
```
$ git grep 'QT_VERSION_CHECK(6, 0, 0)'
src/qt/addressbookpage.cpp:#if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))
src/qt/addressbookpage.cpp:#if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))
src/qt/bitcoin.cpp:#if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
src/qt/bitcoin.cpp:#if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
src/qt/bitcoin.cpp:#if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
```

fanquake:
Not sure we should be having to open and review multiple additional PRs (one now, one later to remove the code again), and be making pointless code changes, just to keep a different PR "focussed". Especially when the changes you're trying to extract are completely relevant.

promag (commented):
I think small changes to make code compatible with both versions are fine until support for qt5 is dropped.

Code review ACK 80761afced12c774a1768fb27a3975d456981ae0. 

Alternative is to use type alias
```cpp
#if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))
using result_t = qintptr*;
#else
using result_t = long*;
#endif

fanquake:
> I think small changes to make code compatible with both versions are fine

I don't think that is the case here? Qt 6 isn't supported for windows until the later PR, at which point Qt 5 isn't supported?

promag:
@fanquake oh I see your point, we won't need to support the old signature of qt5 so the diff could be just the arg type change.

> at which point Qt 5 isn't supported?

Unless this is not the case? @hebasto?

I'm keeping the ACK, but what @fanquake suggests is fine too, with less maintenance efforts.

fanquake:
@promag sed this commit in the switchover PR: https://github.com/bitcoin/bitcoin/commit/812f3e7727211a60989e613bd5edf0a4ec3f146f. Qt 6.7.0 or later will be required for Windows (although as far as I'm aware, Qt 5 won't be supported generally in any case).

PR #841, opened 2024-10-04
Decouple WalletModel from RPCExecutor

furszy:
A more comprehensive fix for the issue described in #837.

Since the `WalletModel` class is unavailable when compiling without wallet support
`(-DENABLE_WALLET=0)`, the RPC executor class should not be coupled to it.
This decoupling ensures GUI compatibility with builds that omit wallet support.

This also drops an extra `#ifdef ENABLE_WALLET` block which is always good.

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -92,7 +92,7 @@ public:
     explicit RPCExecutor(interfaces::Node& node) : m_node(node) {}
 
 public Q_SLOTS:
-    void request(const QString &command, const WalletModel* wallet_model);
+    void request(const QString &command, const QString& wallet_name);
 
 Q_SIGNALS:
     void reply(int category, const QString &command);
@@ -169,7 +169,7 @@ public:
  * @param[out]   pstrFilteredOut  Command line, filtered to remove any sensitive data
  */
 
-bool RPCConsole::RPCParseCommandLine(interfaces::Node* node, std::string &strResult, const std::string &strCommand, const bool fExecute, std::string * const pstrFilteredOut, const WalletModel* wallet_model)
+bool RPCConsole::RPCParseCommandLine(interfaces::Node* node, std::string &strResult, const std::string &strCommand, const bool fExecute, std::string * const pstrFilteredOut, const QString& wallet_name)
 {
     std::vector< std::vector<std::string> > stack;
     stack.emplace_back();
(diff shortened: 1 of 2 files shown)


promag (commented):
Concept meh, `WalletModel` is coupled to `RPCConsole`, why bother with `RPCExecutor`?

hebasto:
@furszy 3f34c04628f1cb0f59afb5176dae3a0e8d414d7d "gui: bugfix, decouple WalletModel from RPCExecutor"

Maybe remove "bugfix" from the commit message, as there is no reproducible bug in the master branch?

pablomartin4btc (commented):
tACK 002b792b9a85100d89e47706c29cf1fd355d9727

Tested the rpc console with and without `-DENABLE_WALLET`. I think the refactor makes sense as the rpc just needs the wallet name for the related commands.

Out of the scope of this PR, I noticed that the order of the wallets in the combo-boxes corresponds to the `settings.json` file and if the user changes the wallet in the main window, when the rpc console is open, the wallet in the console combo doesn't correspond with the one in the main window, which is ok if you want to compare data but I think when the console opens it should have the same. Also perhaps we should show the last used wallet (in `./config/Bitcoin-Qt-*.conf`?) when qt starts (if still in the loaded wallets in `settings.json`) instead of showing the first of the list.

furszy:
> Out of the scope of this PR, I noticed that the order of the wallets in the combo-boxes corresponds to the settings.json file and if the user changes the wallet in the main window, when the rpc console is open, the wallet in the console combo doesn't correspond with the one in the main window, which is ok if you want to compare data but I think when the console opens it should have the same. Also perhaps we should show the last used wallet (in ./config/Bitcoin-Qt-*.conf?) when qt starts (if still in the loaded wallets in settings.json) instead of showing the first of the list.

This surely comes from the fact that the RPC console view is created during initialization. The presentation of the dialog in screen is merely a `show()` call. So there is no "open for the first time" concept right now.
If the goal is still to move to a new GUI, I wouldn't start changing this now here. The current behavior doesn't seem to be "that" bad anyway.

pablomartin4btc:
> If the goal is still to move to a new GUI, I wouldn't start changing the views behavior now.

Yeah, neither do I, just mentioning it here since I just passed thru while testing.

jarolrod (commented):
strong concept ack

BrandonOdiwuor (approved):
tACK 002b792b9a85100d89e47706c29cf1fd355d9727

Was able to build successfully on Ubuntu 24.04 with Qt version 5.15.13 with `-DENABLE_WALLET=OFF` and `-BUILD_GUI=ON` flags and was able to successfully try a couple of RPCs on the GUI RPC console

Screenshot from 2024-11-19 10-10-48

w0xlt:
Code Review ACK https://github.com/bitcoin-core/gui/pull/841/commits/002b792b9a85100d89e47706c29cf1fd355d9727

Passing the wallet name instead of the WalletModel object makes more sense in this case and solves the described issue.

hebasto (approved):
ACK 002b792b9a85100d89e47706c29cf1fd355d9727, I have reviewed the code and it looks OK.

PR #850, opened 2025-01-06
psbt: Use SIGHASH_DEFAULT when signing PSBTs

achow101:
SIGHASH_DEFAULT should be used to indicate SIGHASH_DEFAULT for taproot inputs, and SIGHASH_ALL for all other input types. This avoids adding an unnecessary byte to the end of all Taproot signatures added to PSBTs signed in the GUI.

See also bitcoin/bitcoin#22514

### DIFF
--- a/src/qt/psbtoperationsdialog.cpp
+++ b/src/qt/psbtoperationsdialog.cpp
@@ -83,7 +83,7 @@ void PSBTOperationsDialog::signTransaction()
 
     WalletModel::UnlockContext ctx(m_wallet_model->requestUnlock());
 
-    const auto err{m_wallet_model->wallet().fillPSBT(SIGHASH_ALL, /*sign=*/true, /*bip32derivs=*/true, &n_signed, m_transaction_data, complete)};
+    const auto err{m_wallet_model->wallet().fillPSBT(SIGHASH_DEFAULT, /*sign=*/true, /*bip32derivs=*/true, &n_signed, m_transaction_data, complete)};
 
     if (err) {
         showStatus(tr("Failed to sign transaction: %1")


luke-jr:
>This avoids adding an unnecessary byte to the end of all Taproot signatures added to PSBTs signed in the GUI.

Two bytes, isn't it? For the PSBT_IN_SIGHASH keytype

achow101:
> > This avoids adding an unnecessary byte to the end of all Taproot signatures added to PSBTs signed in the GUI.
> 
> Two bytes, isn't it? For the PSBT_IN_SIGHASH keytype

No, that field is not added. This is referring to the actual signature that will show up in the scriptWitness, and as appears in `PSBT_IN_TAPROOT_*_SIG` fields beforehand.

Sjors:
utACK 3e97ff9c5eaa3160426ba112930b047404c54c9e

It would be useful for the `FillPSBT` doc in `wallet.h` to explain that passing `SIGHASH_DEFAULT` will use `SIGHASH_ALL` for pre-taproot inputs.

You have to crawl all the way to `MutableTransactionSignatureCreator::CreateSig` to find this out:

```cpp
// BASE/WITNESS_V0 signatures don't support explicit SIGHASH_DEFAULT, use SIGHASH_ALL instead.
const int hashtype = nHashType == SIGHASH_DEFAULT ? SIGHASH_ALL : nHashType;
```

pablomartin4btc (commented):
utACK 3e97ff9c5eaa3160426ba112930b047404c54c9e

I agree with @Sjors but tha'ts not covered by doc at the `/** Signature hash types/flags */` `enum` definition (`src/script/interpreter.h`)?

> https://github.com/bitcoin/bitcoin/commit/e94bb4e5fc6dea67844a8495b948b679649b7b65 makes this problem go away though.

is that part of a PR?

Sjors:
@pablomartin4btc it's part of #31622.

The description in `interpreter.h` is ambiguous. It says "Taproot only". It also says "equivalent to SIGHASH_ALL", but doesn't say that it gets converted (and obviously a constant can't guarantee that actually happens).

Theschorpioen (approved):

hebasto (approved):
ACK 3e97ff9c5eaa3160426ba112930b047404c54c9e, I have reviewed the code and it looks OK.

PR #864, opened 2025-04-07
Crash fix, disconnect numBlocksChanged() signal during shutdown

### LINKED ISSUE #862: Segfault during shutdown in SendCoinsDialog::updateCoinControlState
During IBD, while shutting down the gui, it crashed with `Thread 1 "bitcoin-qt" received signal SIGSEGV, Segmentation fault.` Version: v29.99.0-af3dee0b8d45

The `bt` is:

```
(gdb) bt
#0  0x00005555558ba7b2 in SendCoinsDialog::updateCoinControlState (
    this=this@entry=0x55555833f710) at ./qt/sendcoinsdialog.cpp:850
#1  0x00005555558ba88e in SendCoinsDialog::updateSmartFeeLabel (
    this=0x55555833f710) at ./qt/sendcoinsdialog.cpp:863
#2  0x0000555556a99c3e in QObject::event(QEvent*) ()
#3  0x0000555555c4e603 in QApplicationPrivate::notify_helper(QObject*, QEvent*)
    ()
#4  0x0000555556a6829a in QCoreApplication::notifyInternal2(QObject*, QEvent*)
    ()
#5  0x0000555556a6c2bb in QCoreApplicationPrivate::sendPostedEvents(QObject*, int, QThreadData*) ()
#6  0x0000555556abfd45 in QEventDispatcherUNIX::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) ()
#7  0x000055555636ae12 in QXcbUnixEventDispatcher::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) ()
#8  0x0000555556a66da6 in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) ()
#9  0x0000555556a70128 in QCoreApplication::exec() ()
#10 0x00005555557bacda in GuiMain (argc=3, argv=0x7fffffffdff8)
    at ./qt/bitcoin.cpp:691
#11 0x00007ffff7867d90 in __libc_start_call_main (
    main=main@entry=0x55555573a580 <main(int, char**)>, argc=argc@entry=3,
    argv=argv@entry=0x7fffffffdff8)
    at ../sysdeps/nptl/libc_start_call_main.h:58
#12 0x00007ffff7867e40 in __libc_start_main_impl (
    main=0x55555573a580 <main(int, char**)>, argc=3, argv=0x7fffffffdff8,
    init=<optimised out>, fini=<optimised out>, rtld_fini=<optimised out>,
    stack_end=0x7fffffffdfe8) at ../csu/libc-start.c:392
#13 0x00005555557b29f5 in _start ()
```

Thus, the code seems to be:

https://github.com/bitcoin-core/gui/blob/af3dee0b8d45/src/qt/sendcoinsdialog.cpp#L850

furszy:
Aiming to fix bitcoin-core/gui#862.

The crash stems from the order of the shutdown procedure:
We first unset the client model, then destroy the wallet controller—but we leave
the internal wallet models (`m_wallets`) untouched for a brief period. As a result,
there’s a point in time where views still have connected signals and access to
wallet models that are not connected to any wallet controller.
Now.. since the `clientModel` is only replaced with nullptr locally and not destroyed
yet, signals like `numBlocksChanged` can still emit. Thus, when wallet views receive
them, they see a non-null wallet model ptr, and proceed to call backend functions
from a model that is being torn down.

As the shutdown procedure begins by unsetting `clientModel` from all views. It’s safe
to ignore events when `clientModel` is nullptr.

### DIFF
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -855,6 +855,10 @@ void SendCoinsDialog::updateCoinControlState()
 }
 
 void SendCoinsDialog::updateNumberOfBlocks(int count, const QDateTime& blockDate, double nVerificationProgress, SyncType synctype, SynchronizationState sync_state) {
+    // During shutdown, clientModel will be nullptr. Attempting to update views at this point may cause a crash
+    // due to accessing backend models that might no longer exist.
+    if (!clientModel) return;
+    // Process event
     if (sync_state == SynchronizationState::POST_INIT) {
         updateSmartFeeLabel();
     }


yeyeypro (approved):

luke-jr:
I think Qt is supposed to automatically disconnect signals when objects are destroyed? Seems like we should just ensure the model gets destroyed before the underlying interfaces...?

luke-jr:
OTOH, it seems like we should unconditionally disconnect the old clientmodel before setting a new one, so this makes sense up at the top of the method.

furszy:
> OTOH, it seems like we should unconditionally disconnect the old clientmodel before setting a new one, so this makes sense up at the top of the method.

Yeah. We could do what the rest of the code is doing too.. just check for the local `clientModel` before consuming the event:
```diff
diff --git a/src/qt/sendcoinsdialog.cpp b/src/qt/sendcoinsdialog.cpp
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -860,7 +860,7 @@
 }
 
 void SendCoinsDialog::updateNumberOfBlocks(int count, const QDateTime& blockDate, double nVerificationProgress, SyncType synctype, SynchronizationState sync_state) {
-    if (sync_state == SynchronizationState::POST_INIT) {
+    if (sync_state == SynchronizationState::POST_INIT && clientModel) {
         updateSmartFeeLabel();
     }
 }
```

Just to be clear about this issue, it happens because we first unset the client model, then destroy the wallet controller but we do not destroy the internal wallet models (m_wallets). So, there is a point in time just after these two calls where our views still have the signals connected (because the `clientModel` has not been destroyed yet, just replaced by a nullptr locally) and the view still has access to the wallet model. So they receive the `clientModel` new block signal, check if the wallet model is there and try to call some backend functions.
So, the simplest solution would be to check for a local nullptr client model when handling the signal.

A more general solution would be to close all wallets before destroying the wallet controller, but… I’m not sure I want to go down that rabbit hole. Better to fix this crash first, then change shutdown behavior.

pablomartin4btc (commented):
tACK c6f4b0d79606b525064ad702de06bcde11582dd1

laanwj (approved):
Code review ACK 0ee5376fb302ebb21fb6192ed708800ee8035eb5
Nice solution, also handles the (hypothethical) case where a client model is replaced with a new one.

laanwj:
That's really strange. Is there some thread race maybe?

furszy:
> That's really strange. Is there some thread race maybe?
> Does the old solution (to check clientModel in the callback) work better?

Hmm, I should check, but I have the hunch that events already queued are still delivered after disconnection. Which, if that is the case, the previous `clientModel` check would work better.

laanwj:
> Hmm, I should check, but I have the hunch that events already queued are still delivered after disconnection. Which, if that is the case, the previous clientModel check would work better.

The signals are queued, but if no handler is connected, they sould just be dropped.

furszy:
> > Hmm, I should check, but I have the hunch that events already queued are still delivered after disconnection. Which, if that is the case, the previous clientModel check would work better.
> 
> It's worth checking but imo it's unlikely it works that way. The signals are queued, but if no handler is connected, they sould just be dropped (next time the event loop runs).

It seems to be the case: https://stackoverflow.com/questions/2532341/problem-with-qtqueuedconnection-signal-delivered-after-disconnect

So.. it seems that reverting to the null clientModel check is the simplest fix path.
We could discuss the general solution of closing all wallets before destroying the wallet controller in another PR too (if/when we ever want to do that).

laanwj:
Okay, that's really awful (doubt this is the only such case), but yes, better to revert to previous solution then.

furszy:
Updated. Back into the `clientModel` nullptr check approach.

> (doubt this is the only such case)

For sure. We could create a global connection function that checks if the app is tearing down before processing the event. Still, I think I'd rather rework the shutdown process than go that route. Yet, not something for this PR.

pablomartin4btc (approved):
re-ACK 71656bdfaa6bfe08ce9651246a3ef606f923351b

hebasto (approved):
ACK 71656bdfaa6bfe08ce9651246a3ef606f923351b, I have reviewed the code and it looks OK.

fanquake:
Backported to 29.x in #32292.

PR #866, opened 2025-04-09
Enhanced Traffic Graph Widget with Multi-timeframe Support and Data Persistence

rebroad:
This PR improves the network traffic graph widget in the Debug window to provide:
1. Multiple time range options (from 5 minutes to 28 days)
2. Data persistence between sessions
3. Interactive visualization features including tooltips and logarithmic scaling

## Motivation
The existing network traffic graph has limited utility with its fixed time range and lack of historical data preservation. This enhancement allows users to:
- Analyze network traffic patterns over different timeframes
- Preserve historical traffic data between session restarts
- Interact with the graph to get specific data points
- Better visualize varying traffic volumes with logarithmic scaling

These improvements are valuable for:
- Network debugging and troubleshooting
- Understanding Bitcoin Core's network behaviour
- Monitoring traffic patterns for optimization
- Research purposes

## Implementation
The implementation preserves all existing functionality while adding new features:
- Added a pre-configured timeframe selection (13 different timeframes)
- Implemented traffic data serialization and deserialization
- Enhanced the visualization with interactive features
- Improved tooltip information with precise timestamps and traffic rates

Supporting changes:
- Added formatBytesps function for a prettier display of traffic rates
- Added FormatISO8601Time for better time display

## Testing
Tested on Linux with various network conditions. The new functionality can be exercised by:
1. Using the slider (or arrow keys) to select different time ranges
2. Restarting the application to verify data persistence
3. Clicking on the graph to toggle between linear and logarithmic scales
4. Hovering over data points to view detailed information

## Documentation
The changes are mostly self-documenting through the UI and are constrained to the Qt interface without affecting core functionality.

## Compatibility
This PR maintains compatibility with existing functionality. The data persistence file uses proper serialization versioning to allow for future format changes if needed.

### DIFF
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -665,19 +665,28 @@
            <item>
             <widget class="QSlider" name="sldGraphRange">
              <property name="minimum">
-              <number>1</number>
+              <number>0</number>
              </property>
              <property name="maximum">
-              <number>288</number>
+              <number>2400</number>
+             </property>
+             <property name="singleStep">
+              <number>200</number>
              </property>
              <property name="pageStep">
-              <number>12</number>
+              <number>200</number>
              </property>
              <property name="value">
-              <number>6</number>
+              <number>0</number>
              </property>
              <property name="orientation">
               <enum>Qt::Horizontal</enum>
+             </property>
+             <property name="tickPosition">
+              <enum>QSlider::TicksBelow</enum>
+             </property>
+             <property name="tickInterval">
+              <number>200</number>
              </property>
             </widget>
            </item>
@@ -691,16 +700,6 @@
              </property>
              <property name="alignment">
               <set>Qt::AlignCenter</set>
-             </property>
-            </widget>
-           </item>
-           <item>
-            <widget class="QPushButton" name="btnClearTrafficGraph">
-             <property name="text">
-              <string>&amp;Reset</string>
-             </property>
-             <property name="autoDefault">
-              <bool>false</bool>
              </property>
             </widget>
            </item>
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -54,7 +54,6 @@
 using util::Join;
 
 const int CONSOLE_HISTORY = 50;
-const int INITIAL_TRAFFIC_GRAPH_MINS = 30;
 const QSize FONT_RANGE(4, 40);
 const char fontSizeSettingsKey[] = "consoleFontSize";
 
@@ -537,13 +536,12 @@ RPCConsole::RPCConsole(interfaces::Node& node, const PlatformStyle *_platformSty
     connect(ui->clearButton, &QAbstractButton::clicked, [this] { clear(); });
     connect(ui->fontBiggerButton, &QAbstractButton::clicked, this, &RPCConsole::fontBigger);
     connect(ui->fontSmallerButton, &QAbstractButton::clicked, this, &RPCConsole::fontSmaller);
-    connect(ui->btnClearTrafficGraph, &QPushButton::clicked, ui->trafficGraph, &TrafficGraphWidget::clear);
 
     // disable the wallet selector by default
     ui->WalletSelector->setVisible(false);
     ui->WalletSelectorLabel->setVisible(false);
 
-    setTrafficGraphRange(INITIAL_TRAFFIC_GRAPH_MINS);
+    setTrafficGraphRange(1); // 1 is the lowest setting (0 bumps up)
     updateDetailWidget();
 
     consoleFontSize = settings.value(fontSizeSettingsKey, QFont().pointSize()).toInt();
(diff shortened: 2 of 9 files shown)


src/qt/trafficgraphwidget.cpp
```diff
@@ -20,152 +22,603 @@
...
+        m_client_model = model;
+        if (m_samples_in[0].empty() && m_samples_out[0].empty()) {
+            // Load saved traffic data if available and the arrays are empty
+            loadData();
+
+            // Restore the saved total bytes counts to the node if they exist
+            if (m_total_bytes_recv > 0 || m_total_bytes_sent > 0) {
+                model->node().setTotalBytesRecv(m_total_bytes_recv);
+                model->node().setTotalBytesSent(m_total_bytes_sent);
+            }
```
luke-jr, line 46: The GUI should not be changing low-level internals.

If for some reason the clientModel is changed, its data should be taken as authoritative, even if it contradicts the history so far (if that's an issue, wipe the history).

Alternatively, maybe the history should be stored with the node, and loaded into the GUI (or RPC, in another hypothetical PR).

src/qt/forms/debugwindow.ui
```diff
@@ -694,16 +703,6 @@
              </property>
             </widget>
            </item>
-           <item>
-            <widget class="QPushButton" name="btnClearTrafficGraph">
```
luke-jr, line 698: If we're deleting this, maybe there should be a way for the user to insert a reference line?

For now, I'd move removal of anything to a separate PR.

src/qt/rpcconsole.cpp
```diff
@@ -578,7 +576,7 @@ RPCConsole::RPCConsole(interfaces::Node& node, const PlatformStyle *_platformSty
     // based timer interface
     m_node.rpcSetTimerInterfaceIfUnset(rpcTimerInterface);
 
-    setTrafficGraphRange(INITIAL_TRAFFIC_GRAPH_MINS);
```
luke-jr, line 581: Why are we losing this default constant?

src/util/time.cpp
```diff
@@ -75,6 +75,13 @@ void MockableSteadyClock::ClearMockTime()
 
 int64_t GetTime() { return GetTime<std::chrono::seconds>().count(); }
 
+int64_t GetTimeMillis()
```
luke-jr, line 78: I doubt we'll be restoring this. It was removed intentionally.

luke-jr (changes requested):
Merging it all together likes this makes review very annoying. It was better as separate PRs.

src/util/time.cpp
rebroad: yes, I'm going to remove this, will push an update shortly

src/qt/rpcconsole.cpp
rebroad: because it isn't needed - on startup it "defaults" to the first non-full range

src/qt/forms/debugwindow.ui
rebroad: The user can effectively remove the data by deleting the .dat file - is there any basis for being able to do this while the client is running?

src/qt/trafficgraphwidget.cpp
rebroad: The idea is that the data is not persistent - but yes, I see your point - it should probably not be something that only happens when the GUI is running, and should perhaps also happen in bitcoind also.

rebroad:
> Merging it all together likes this makes review very annoying. It was better as separate PRs.

I vaguely remember people complaining that I was raising several PRs that could be combined when I first raised the PRs years ago.... I guess one cannot please everyone all of the time.

rebroad:
There is a bug in the shutdown sequence (null model), so this is not ready to merge yet.

src/qt/trafficgraphwidget.cpp
rebroad: > The GUI should not be changing low-level internals.
> 
> If for some reason the clientModel is changed, its data should be taken as authoritative, even if it contradicts the history so far (if that's an issue, wipe the history).
> 
> Alternatively, maybe the history should be stored with the node, and loaded into the GUI (or RPC, in another hypothetical PR).

@luke-jr I've removed the setTotalBytes functions as I agree with your reasoning. The GUI keeps a local track of the totals from the last session and simply adds them to the totals provided by the node now. I did think about storing the data on the node, but I am not sure it makes much sense given the data is currently stored on the GUI and so it makes more sense to keep the totals local with the traffic history for now. Perhaps someone thinks it's worth storing this on the node in the future, but it doesn't seem like an urgent need for now.

rebroad:
I've removed the code that changed the totals on the node - the GUI just keeps track now without writing values to the node. Also, added some checks for corrupt save files so that it doesn't load bad data.

rebroad:
I realise this PR is starting to look a bit messy with all the updates. I also realise I ought to separate various functionality changes into separate commits. Is it better to raise a new PR once this is done, force push to this one?

rebroad:
Have just added some more GUI improvements - the text on the graph now has a black outline around it making it easier to read (and is printed AFTER the graph rather than before) .Also, the grey line at the bottom of the graph is now printed after the graph also, causing the graph not to spill onto that line.

Also, the bright green and red lines no longer surround the sides of the graph but only draw an outline along actual graph values. This makes it easier to distinguish about the graph going off the edge of the display areas, vs a sudden drop in traffic (which previously looked the same). The now graph also fills the area fully to the edges at the side now (which it did not before due to rounding issues).

It would be useful to know whether I should split all these changes into individual pull requests, or individual commits, or better all together like this.

RandyMcMillan:
concept ACK

th30neee (approved):

th30neee (approved):

PR #868, opened 2025-04-24
Replace stray tfm::format to cerr with qWarning

laanwj:
GUI warnings should go to the log, not to the console (which may not be connected at all).


hebasto (approved):
ACK edd46566bd66cea7d7f4116429fe1c11d2187ba2, I have reviewed the code and it looks OK.

PR #871, opened 2025-05-06
qt, docs: Unify term "clipboard"

hebasto:
A translator on Transifex noticed:
> The term "system clipboard" appears twice. The term "clipboard" appears 10 times. Perhaps we could standardize on just saying "clipboard"?

This PR addresses this issue.

### DIFF
--- a/src/qt/forms/addressbookpage.ui
+++ b/src/qt/forms/addressbookpage.ui
@@ -78,7 +78,7 @@
      <item>
       <widget class="QPushButton" name="copyAddress">
        <property name="toolTip">
-        <string>Copy the currently selected address to the system clipboard</string>
+        <string>Copy the currently selected address to the clipboard</string>
        </property>
        <property name="text">
         <string>&amp;Copy</string>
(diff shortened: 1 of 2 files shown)


hebasto:
> A translator on Transifex noticed:
> 
> > The term "system clipboard" appears twice. The term "clipboard" appears 10 times. Perhaps we could standardize on just saying "clipboard"?

Reported by @jesterhodl.

jesterhodl:
Would be good to do the same for all remaining *.ts files in ```src/qt/locale/``` as well as these two:
`
src/qt/forms/signverifymessagedialog.ui:            <string>Copy the current signature to the system clipboard</string>
src/qt/forms/addressbookpage.ui:        <string>Copy the currently selected address to the system clipboard</string>
`

I used ```grep -ri "system clipboard" *``` to find them

pablomartin4btc (commented):
ACK 3bbdbc0a5e1b409969cedaf249d1d01dea9bcf73

PR #876, opened 2025-05-26
Including exception what() in Runaway dialog box

Krellan:
If an exception happens, including e.what() in the text passed to the dialog box, so that the user can see it.  This is in addition to any text from getWarnings() that might already be present, separated by newlines if so.

This will make it easier for users to find what went wrong and get the help they need, instead of having to tell the user to go back and dig through the debug.log file.

### DIFF
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -46,7 +46,9 @@
 
 #include <boost/signals2/connection.hpp>
 #include <chrono>
+#include <exception>
 #include <memory>
+#include <string>
 
 #include <QApplication>
 #include <QDebug>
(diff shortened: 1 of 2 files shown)


furszy (commented):
> If an exception happens, including e.what() in the text passed to the dialog
box, so that the user can see it.  This is in addition to any text from
getWarnings() that might already be present, separated by newlines if so.

Have you verified that we won't output the same error message twice?

Krellan:
> Have you verified that we won't output the same error message twice?

I looked at the `Warning` class, and looked at its usages, and the messages that are passed in differ from the text that would arise if an exception were to be thrown (they come from different areas of the code).

Ari4ka (commented):
****

PR #877, opened 2025-05-26
gui: Add a menu action to restore then migrate a legacy wallet

achow101:
Some users will have a backup of their legacy wallet. These cannot be restored since the "Restore Wallet" action expects to be able to load the wallet after restoring, and this fails for legacy wallets now that they are deleted. Furthermore, the "Migrate Wallet" action only allows users to migrate wallets that are in the wallets directory, so such backups cannot be migrated from the GUI.

This PR resolves this issue by adding a menu item in the "Migrate Wallet" menu which allows users to select their backup file so that it will first be restored but not loaded, and then migrated.

Depends on https://github.com/bitcoin/bitcoin/pull/32620

### DIFF
--- a/src/interfaces/wallet.h
+++ b/src/interfaces/wallet.h
@@ -328,7 +328,7 @@ public:
     virtual std::string getWalletDir() = 0;
 
     //! Restore backup wallet
-    virtual util::Result<std::unique_ptr<Wallet>> restoreWallet(const fs::path& backup_file, const std::string& wallet_name, std::vector<bilingual_str>& warnings) = 0;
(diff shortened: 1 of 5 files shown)


pablomartin4btc (commented):
tACK 56407ca69e34778cfd0d054fbaa94ab0b28ffb6d

I'll try to review the code later (first commit is not longer needed, perhaps rebase?).

Regarding the approach IMHO:

> integrate this functionality to automatically detect if it is a legacy wallet and offer the migration option in the same flow.

I thought the same at first glance as an additional to the current PR logic, if the user knows that has the backup to restore maybe goes directly there first.

I'd update the menu item title to "Wallet file...", since you could restore a .dat file or migrate a backup file directly - checked with `CLI`.

Perhaps @GBKS could add his view here...

image

achow101:
Rebased and ready for review

PR #878, opened 2025-06-26
Fix broken navigation link to files documentation in README_doxygen.md

VolodymyrBg:
Replaced the outdated link to files.html with the correct files.md in the navigation section of README_doxygen.md, ensuring users are directed to the current documentation format.


hebasto:
It works fine as is in https://doxygen.bitcoincore.org/.

In the future, please open non-GUI-specific PRs in the main [repo](https://github.com/bitcoin/bitcoin).

PR #879, opened 2025-07-10
New SVG, Icons, PNGs and X PixMaps

ghost:
This PR clean up the old SVG, Icons, PNGs and X PixMaps by providing optimized much smaller file size versions while at the same time keeping resolutions untouched. Shadows are removed from the files.


hebasto:
> Shadows are removed from the files.

Why do you think this change is necessary?

Also, the size of the logo has been increased. Could you please mention this in the PR description as well, along with a justification for the change?

jonasschnelli:
I briefly discussed the issue of copyright and authorship on the SVG with the PR author in #bitcoin (IRC).

It sounded to me that the PR author was worried about the copyright and MIT licensing as well as who was the original designer. Unclear what the intention of this PR is as this also changes the copyright in the file.

I have no opinion on visual style.

PR #881, opened 2025-07-26
Move `FreespaceChecker` class into its own module

hebasto:
For some reason, the MOC compiler in older versions of Qt 6 fails to parse `qt/intro.cpp`, as noted in [this comment](https://github.com/bitcoin/bitcoin/pull/32998#issuecomment-3082011233).

This PR proposes a move-only refactoring to simplify the source structure by eliminating the need for the inline `#include <qt/intro.moc>`, thereby effectively working around the issue.

Required for https://github.com/bitcoin/bitcoin/pull/32998.

### DIFF
--- a/src/qt/CMakeLists.txt
+++ b/src/qt/CMakeLists.txt
@@ -75,6 +75,8 @@ add_library(bitcoinqt STATIC EXCLUDE_FROM_ALL
   clientmodel.h
   csvmodelwriter.cpp
   csvmodelwriter.h
+  freespacechecker.cpp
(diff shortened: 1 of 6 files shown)


hebasto:
@ajtowns 

Thank you for the review! Your feedback has been addressed.

ajtowns:
ACK 4542412ae063903654fc1cd95f77889bae76d61c

nit: qt/intro.h forward decl of `class FreespaceChecker` is redundant since the header is included now.

PR #882, opened 2025-07-30
Add console commands for clearing output and history

waketraindev:
Adds new console commands for clearing the screen and managing command history in the GUI RPC console.

Introduces four new commands:

- `clear` - clears the console output area.
- `clear-all` - clears both output and command history.
- `history` - lists stored command history.
- `history-clear` - clears command history only.

Provides explicit, privacy-orientd ways to clear console history and output without altering default behavior.
Useful in shared or long-running sessions.

### DIFF
--- a/doc/release-notes-gui-882.md
+++ b/doc/release-notes-gui-882.md
@@ -1,0 +1,12 @@
+GUI changes
+-----------
+
+- Added new console commands to the `Debug window -> Console` for improved privacy:
+
+    - `clear` -- clears the console output area
+    - `clear-all` -- clears both output and command history.
+    - `history` -- lists stored command history.
+    - `history-clear` -- clears command history only.
+
+  These additions improve privacy and convenience in shared or long-lived
+  sessions by allowing users to remove sensitive commands and output from view.
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -93,6 +93,8 @@ public Q_SLOTS:
 
 Q_SIGNALS:
     void reply(int category, const QString &command);
+    void clearConsole(bool clear_screen, bool clear_history);
+    void printHistory();
 
 private:
     interfaces::Node& m_node;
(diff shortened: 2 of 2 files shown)


pablomartin4btc (commented):
I don't think this change is practical. Clearing the console should remove the visible output, but it shouldn't also clear the command history. In most terminal environments (e.g., bash, zsh, cmd, PowerShell, etc.), using clear or Ctrl+L clears the screen while preserving command history, which remains accessible via the arrow keys. Users often rely on this history to repeat or modify previous commands, and removing it could unintentionally disrupt that workflow. I suggest keeping the command history intact when clearing the console.

---

Also, just a heads-up: this GUI repository is currently in maintenance mode and primarily focused on bugfixes — new feature additions are unlikely to be considered or prioritized unless they're closely tied to functionality in Bitcoin Core. If you're interested in contributing new features or improvements, I encourage you to take a look at [bitcoin-core/gui-qml](https://github.com/bitcoin-core/gui-qml), which is the new GUI project under active development. Your contributions would be very welcome there.

luke-jr:
Concept NACK, agree we should match expected behaviour here (not clearing command history). Maybe some way can be added, but I'm not sure there's an obvious right way.

pablomartin4btc (commented):
Thanks for the update, but I don’t think relying on a hidden keyboard combination (Shift modifier and Ctrl+Shift+L) is a good long-term solution. Most users will never discover it, and without documentation, it risks becoming an obscure feature that only a handful of people know about. It also deviates from common terminal behavior, where clearing the output and clearing history are separate, explicit actions. On top of that, shortcut bindings can be difficult to maintain long-term — they may conflict with other shortcuts, differ across platforms, or get broken by `Qt` or `OS` changes.

I think that a more user-friendly and discoverable approach would be to support terminal-style commands in the console, so users can interact in a way they already understand from bash, zsh, cmd, PowerShell, etc. For example:

```
history               # list cached commands
help history          # show usage info for the history command
history -c            # clear command history
clear                 # clear visible console output only
help clear            # explain clear options (e.g. 'clear history')
clear history         # clear command history
```

This way, clearing history becomes an intentional action the user explicitly types, and it can be documented alongside other commands in the console’s help output. It’s discoverable, consistent with terminal norms but mainly with the current rpc console behaviour, and extensible if more sub-commands are needed in the future.

pablomartin4btc (commented):
Sorry I haven't had the chance to check the PR earlier.

For the shortcut perhaps you can document it in the welcome message of the console:

image


Replacing this line:
`Use up and down arrows to navigate history, and Ctrl+L to clear screen.`
By:
`Use up and down arrows to navigate history, Ctrl+L to clear screen and Ctrl+L to clear screen and Ctrl+Shift+L to clear the command history.`

For the new console commands...
>    - clear -> clears output
>    - history-clear -> clears output and history
>    - history -> prints command history

I find them useful but I'd split the clear screen from the clear history (`history-clear`) and if you want to clear both perhaps a "`clear-all`".

Same for the button, I'd add a different button for clearing the history.

---

Please keep in mind that new features like this [may not be merged](https://github.com/bitcoin-core/gui/pull/753#issuecomment-2629332355) and help would be also much appreciated in the new [QML GUI](https://github.com/bitcoin-core/gui-qml/).

waketraindev:
I'd appreciate any suggestions on discoverability of these console commands.

PR #898, opened 2025-10-05
gui: add coins (UTXOs) tab and makes it view-only

apogio:
GUI: Add coins tab and sets view-only mode

Adds Coins (UTXOs) tab at the top and:

- Removes selection checkboxes from the dialog
- Disables coin selection functionality
- Maintains UTXO viewing capabilities
- Removes all the information from the top area of the dialog.

The following pictures depict how the GUI would look like:

coins_1

coins_3


hebasto:
cc @achow101 @Sjors

hebasto:
> I'm not sure what the comment about motivation means...

It means that the following only describes the feature itself, without providing the goals, use cases, or rationale behind it:
> GUI: Add coins tab and sets view-only mode
> 
> Adds Coins (UTXOs) tab at the top and:
> 
>    * Removes selection checkboxes from the dialog
> 
>    * Disables coin selection functionality
> 
>    * Maintains UTXO viewing capabilities
> 
>    * Removes all the information from the top area of the dialog.

PR #901, opened 2025-10-10
Add createwallet, createwalletdescriptor, and migratewallet to history filter

waketraindev:
Added `createwallet`, `createwalletdescriptor` and `migratewallet` RPC commands to the Qt console history filter since they may include passphrases or other sensitive data that should not be stored in command history.


fanquake:
Backported to `30.x` in https://github.com/bitcoin/bitcoin/pull/33609.

PR #908, opened 2025-10-30
Remove HD seed reference from blank wallet tooltip

john-moffett:
Blank descriptor wallets currently do not have HD seeds and none can be added (or 'set') by the user, so remove the reference in the tooltip.

As I understand it, descriptor wallets don't have a global HD seed and don't store the HD seeds for keys they generate. Currently, no new HD seeds can be added by the user (even for old wallets since `sethdseed` was removed), though it may be possible in the future, eg -  https://github.com/bitcoin/bitcoin/pull/33043


hebasto:
cc @achow101

PR #909, opened 2025-11-06
Prevent re-execution of sensitive commands from console history

waketraindev:
Sensitive RPC commands such as `walletpassphrase` or `createwallet`
have their arguments redacted when stored in the console history.
Even though their parameters are hidden, these commands could still
be recalled and executed again, which might lead to unintended or
harmful actions.

This change extends the existing blocking filter that prevents
re-execution of commands considered sensitive or risky when recalled
from history. Such entries are prefixed with a leading character (`!`),
marking them as non-executable. When the user attempts to run them
again from history, the console blocks the action and displays an
informational message. Commands entered manually remain unaffected.

In addition to wallet and key-related RPCs, this filter now also
covers transaction-related commands such as `send`, `sendall`,
`sendmany`, and `sendtoaddress`, which may cause unwanted effects
if repeated from history.

Test coverage has been expanded to verify redaction and blocking
behavior, ensuring that sensitive commands are correctly identified
and prefixed. The console help text has been updated to describe
this functionality.

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -83,6 +83,14 @@ const QStringList historyFilter = QStringList()
 
 }
 
+// List of commands that may cause unintended effects if accidentally re-executed from history
+const QStringList sensitiveFilter = QStringList()
+    << "send"
+    << "sendall"
+    << "sendmany"
+    << "sendtoaddress"
+;
+
 /* Object for executing console RPC commands in a separate thread.
 */
 class RPCExecutor : public QObject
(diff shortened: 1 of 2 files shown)


src/qt/rpcconsole.cpp
```diff
@@ -1094,6 +1104,12 @@ void RPCConsole::startExecutor()
         m_is_executing = false;
     });
 
+    connect(m_executor, &RPCExecutor::noop, this, [this]() {
+        ui->messagesWidget->undo();
+        scrollToEnd();
+        m_is_executing = false;
+    });
```
luke-jr: Why does this need a signal/slot?

luke-jr (changes requested):

waketraindev:
* Blocking character was changed from '#' to '!' in order to reserve '#' for printing comments such as like bash
* Removed noop slot and signal
* Added alert window when a command starting with ! is entered
* Commands starting with ! don't print, and don't go to history
* Updated PR title and description to reflect the changes

hebasto:
Please rebase.

PR #910, opened 2025-11-06
Added test coverage for qt gui#901 console history filter

waketraindev:
Add test coverage for the QT rpc console updated filtered commands in gui#901


src/qt/test/rpcnestedtests.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2021 The Bitcoin Core developers
+// Copyright (c) 2016-2025 The Bitcoin Core developers
```
pablomartin4btc: so we don't have to update it every time...
```suggestion
// Copyright (c) 2016-present The Bitcoin Core developers
```

PR #911, opened 2025-11-19
Adds non-mempool wallet balance to overview

ajtowns:
The wallet can contain transactions that are not accepted into the node's mempool (eg due to containing a too large OP_RETURN output, due to too low a feerate, or due to too many unconfirmed ancestors). In the event you end up in this situation, it can appear as if funds have gone missing from your wallet due to the non-mempool balance not being reported. Correct this by reporting the non-mempool balance.

See bitcoin/bitcoin#33671 for further context.

### DIFF
--- a/src/qt/bitcoinunits.cpp
+++ b/src/qt/bitcoinunits.cpp
@@ -142,7 +142,6 @@ QString BitcoinUnits::formatHtmlWithUnit(Unit unit, const CAmount& amount, bool 
 
 QString BitcoinUnits::formatWithPrivacy(Unit unit, const CAmount& amount, SeparatorStyle separators, bool privacy)
 {
-    assert(amount >= 0);
     QString value;
     if (privacy) {
         value = format(unit, 0, false, separators, true).replace('0', '#');
--- a/src/qt/forms/overviewpage.ui
+++ b/src/qt/forms/overviewpage.ui
@@ -130,14 +130,14 @@
               </property>
              </widget>
             </item>
-            <item row="3" column="0" colspan="2">
+            <item row="4" column="0" colspan="2">
              <widget class="Line" name="line">
               <property name="orientation">
                <enum>Qt::Horizontal</enum>
               </property>
              </widget>
             </item>
-            <item row="4" column="0">
+            <item row="5" column="0">
              <widget class="QLabel" name="labelTotalText">
               <property name="text">
                <string>Total:</string>
@@ -183,7 +183,33 @@
               </property>
              </widget>
             </item>
-            <item row="4" column="1">
+            <item row="3" column="1">
+             <widget class="QLabel" name="labelNonMempool">
+              <property name="cursor">
+               <cursorShape>IBeamCursor</cursorShape>
+              </property>
+              <property name="toolTip">
+               <string>Balance for wallet transactions not in the mempool</string>
+              </property>
+              <property name="text">
+               <string notr="true">0.00000000 BTC</string>
+              </property>
+              <property name="alignment">
+               <set>Qt::AlignRight|Qt::AlignTrailing|Qt::AlignVCenter</set>
+              </property>
+              <property name="textInteractionFlags">
+               <set>Qt::LinksAccessibleByMouse|Qt::TextSelectableByKeyboard|Qt::TextSelectableByMouse</set>
+              </property>
+             </widget>
+            </item>
+            <item row="3" column="0">
+             <widget class="QLabel" name="labelNonMempoolText">
+              <property name="text">
+               <string>Non-mempool:</string>
+              </property>
+             </widget>
+            </item>
+            <item row="5" column="1">
              <widget class="QLabel" name="labelTotal">
               <property name="cursor">
                <cursorShape>IBeamCursor</cursorShape>
--- a/src/qt/overviewpage.cpp
+++ b/src/qt/overviewpage.cpp
@@ -190,13 +190,20 @@ void OverviewPage::setBalance(const interfaces::WalletBalances& balances)
     ui->labelBalance->setText(BitcoinUnits::formatWithPrivacy(unit, balances.balance, BitcoinUnits::SeparatorStyle::ALWAYS, m_privacy));
     ui->labelUnconfirmed->setText(BitcoinUnits::formatWithPrivacy(unit, balances.unconfirmed_balance, BitcoinUnits::SeparatorStyle::ALWAYS, m_privacy));
     ui->labelImmature->setText(BitcoinUnits::formatWithPrivacy(unit, balances.immature_balance, BitcoinUnits::SeparatorStyle::ALWAYS, m_privacy));
-    ui->labelTotal->setText(BitcoinUnits::formatWithPrivacy(unit, balances.balance + balances.unconfirmed_balance + balances.immature_balance, BitcoinUnits::SeparatorStyle::ALWAYS, m_privacy));
+    ui->labelNonMempool->setText(BitcoinUnits::formatWithPrivacy(unit, balances.nonmempool_balance, BitcoinUnits::SeparatorStyle::ALWAYS, m_privacy));
+    ui->labelTotal->setText(BitcoinUnits::formatWithPrivacy(unit, balances.balance + balances.unconfirmed_balance + balances.immature_balance + balances.nonmempool_balance, BitcoinUnits::SeparatorStyle::ALWAYS, m_privacy));
     // only show immature (newly mined) balance if it's non-zero, so as not to complicate things
     // for the non-mining users
     bool showImmature = balances.immature_balance != 0;
 
     ui->labelImmature->setVisible(showImmature);
     ui->labelImmatureText->setVisible(showImmature);
+
+    // likewise for non-mempool balances
+    bool showNonMempool = balances.nonmempool_balance != 0;
+
+    ui->labelNonMempool->setVisible(showNonMempool);
+    ui->labelNonMempoolText->setVisible(showNonMempool);
 }
 
 void OverviewPage::setClientModel(ClientModel *model)
@@ -296,5 +303,6 @@ void OverviewPage::setMonospacedFont(const QFont& f)
     ui->labelBalance->setFont(f);
     ui->labelUnconfirmed->setFont(f);
     ui->labelImmature->setFont(f);
+    ui->labelNonMempool->setFont(f);
     ui->labelTotal->setFont(f);
 }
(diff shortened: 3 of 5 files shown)


ajtowns:
Rebased after merge of parent PR, should be ready to review. Happy to close if some someone wants to take over.

polespinasa:
Maybe I'm not understanding the issue, but does this matter?
If that is a transaction we created we will not broadcast it.
If that is a transaction we are receiving we will not be able to spend the outputs until it is confirmed as the child will be rejected by our mempool.

So I am not really getting the issue, we just need to wait for it to be confirmed.

IIUC this only makes sense for private-broadcast.

ajtowns:
Worst case, I guess: if you make a transaction that doesn't hit the mempool, it won't be visible in your balance, which may lead you to try to make the same transaction again. That might happen if the first transaction was spending some unconfirmed outputs from large transactions, so your new transaction would have exceeded the cluster size limits. Once those large transactions confirm, your nonmempool transaction will be broadcast and able to be confirmed.

pablomartin4btc (commented):
Concept ACK

>The negative sign for the non-mempool stuff isn't really very clear, perhaps.

I think for negative amounts we use `COLOR_NEGATIVE`  - defined in `guiconstants.h` (at least for transactions).

Thinking out loud... shouldn't be the non-mempool balance shown under the pending one like a subitem...

polespinasa:
> Worst case, I guess: if you make a transaction that doesn't hit the mempool, it won't be visible in your balance, which may lead you to try to make the same transaction again. That might happen if the first transaction was spending some unconfirmed outputs from large transactions, so your new transaction would have exceeded the cluster size limits. Once those large transactions confirm, your nonmempool transaction will be broadcast and able to be confirmed.

I see, didn't check the code when asking, had the same concerns as here https://github.com/bitcoin/bitcoin/pull/33671#issuecomment-3554858459, but I saw the approach taken there was pretty nice tbh.

Concept ACK

polespinasa:
> The negative sign for the non-memool stuff isn't really very clear, perhaps.

It is not :)

I don't think we show in the balance overview the "negative" (outgoing) balance, we just reduce the available amount.
imagen

src/qt/forms/overviewpage.ui
```diff
@@ -183,7 +183,33 @@
...
+              </property>
+              <property name="textInteractionFlags">
+               <set>Qt::LinksAccessibleByMouse|Qt::TextSelectableByKeyboard|Qt::TextSelectableByMouse</set>
+              </property>
+             </widget>
+            </item>
+            <item row="3" column="0">
+             <widget class="QLabel" name="labelNonMempoolText">
+              <property name="text">
+               <string>Non-mempool:</string>
```
polespinasa, line 208: just a silly feel-free-to-ignore nit: in the rpc we use `nonmempool` no `non_mempool`. Maybe we don't want to use the `-` for consistency?

Also maybe mempool can be tooo techincal for some users? What about `Pending (unbroadcast)` or something similar?

src/qt/bitcoinunits.cpp
```diff
@@ -142,7 +142,6 @@ QString BitcoinUnits::formatHtmlWithUnit(Unit unit, const CAmount& amount, bool
 
 QString BitcoinUnits::formatWithPrivacy(Unit unit, const CAmount& amount, SeparatorStyle separators, bool privacy)
 {
-    assert(amount >= 0);
```
polespinasa, line 145: nit: Maybe this change motivation should be documented in the commit message. It is not clear why is needed.
Just a simple line saying that `m_mine_nonmempool` is always negative so this assert is no longer true should be enough.

src/qt/overviewpage.cpp
```diff
@@ -190,13 +190,20 @@ void OverviewPage::setBalance(const interfaces::WalletBalances& balances)
...
+    ui->labelTotal->setText(BitcoinUnits::formatWithPrivacy(unit, balances.balance + balances.unconfirmed_balance + balances.immature_balance + balances.nonmempool_balance, BitcoinUnits::SeparatorStyle::ALWAYS, m_privacy));
     // only show immature (newly mined) balance if it's non-zero, so as not to complicate things
     // for the non-mining users
     bool showImmature = balances.immature_balance != 0;
 
     ui->labelImmature->setVisible(showImmature);
     ui->labelImmatureText->setVisible(showImmature);
+
+    // likewise for non-mempool balances
+    bool showNonMempool = balances.nonmempool_balance != 0;
```
polespinasa, line 203: maybe `< 0` ?

src/qt/test/wallettests.cpp
polespinasa, line 278: Maybe we can use this PR to update this to also use:
`walletModel.wallet().getBalances().balance`

Like you did below?

src/qt/forms/overviewpage.ui
```diff
@@ -183,7 +183,33 @@
...
              </widget>
             </item>
-            <item row="4" column="1">
+            <item row="3" column="1">
+             <widget class="QLabel" name="labelNonMempool">
+              <property name="cursor">
+               <cursorShape>IBeamCursor</cursorShape>
+              </property>
+              <property name="toolTip">
+               <string>Balance for wallet transactions not in the mempool</string>
```
pablomartin4btc, line 192: nit: I think the tooltip text doesn't tell the user what that means in practice, or why it's negative...

```suggestion
               <string>Balance about to be spent by wallet transactions not currently in the mempool. May never confirm.</string>
```

src/qt/forms/overviewpage.ui
pablomartin4btc: I was thinking the same and have expressed it in my previous [comment](https://github.com/bitcoin-core/gui/pull/911#pullrequestreview-4538226876) but I've checked in the code and it seems that unbroadcast has a different meaning I think (transactions that ARE in the mempool but haven't been relayed to peers yet - `m_unbroadcast_txids` in `txmempool.h`), and for the pending section, there's a key difference: Pending = mempool → almost certain to confirm soon; Non-mempool = might never confirm,  so it would be  also misleading.

src/qt/overviewpage.cpp
```diff
@@ -190,13 +190,20 @@ void OverviewPage::setBalance(const interfaces::WalletBalances& balances)
     ui->labelBalance->setText(BitcoinUnits::formatWithPrivacy(unit, balances.balance, BitcoinUnits::SeparatorStyle::ALWAYS, m_privacy));
     ui->labelUnconfirmed->setText(BitcoinUnits::formatWithPrivacy(unit, balances.unconfirmed_balance, BitcoinUnits::SeparatorStyle::ALWAYS, m_privacy));
     ui->labelImmature->setText(BitcoinUnits::formatWithPrivacy(unit, balances.immature_balance, BitcoinUnits::SeparatorStyle::ALWAYS, m_privacy));
-    ui->labelTotal->setText(BitcoinUnits::formatWithPrivacy(unit, balances.balance + balances.unconfirmed_balance + balances.immature_balance, BitcoinUnits::SeparatorStyle::ALWAYS, m_privacy));
+    ui->labelNonMempool->setText(BitcoinUnits::formatWithPrivacy(unit, balances.nonmempool_balance, BitcoinUnits::SeparatorStyle::ALWAYS, m_privacy));
+    ui->labelTotal->setText(BitcoinUnits::formatWithPrivacy(unit, balances.balance + balances.unconfirmed_balance + balances.immature_balance + balances.nonmempool_balance, BitcoinUnits::SeparatorStyle::ALWAYS, m_privacy));
```
pablomartin4btc, line 194: nit: the non-mempool balance is currently a plain text with no visual emphasis, perhaps since it's always negative, it should be red (like negative amounts in the transaction list)... (haven't tested it)
```suggestion
    ui->labelNonMempool->setStyleSheet("QLabel { color: " + COLOR_NEGATIVE.name() + "; }");
    ui->labelTotal->setText(BitcoinUnits::formatWithPrivacy(unit, balances.balance + balances.unconfirmed_balance + balances.immature_balance + balances.nonmempool_balance, BitcoinUnits::SeparatorStyle::ALWAYS, m_privacy));
```

pablomartin4btc (commented):
@ajtowns, I've added a test that perhaps you may want to include here, it checks the label visibility and its value, please check e746d9ef6225b1c4281ce514fb4e21fc98b4c56a (based on top of your changes).

I think this PR should include a `doc/release-notes-911.md` file. Suggested text:

```
GUI Changes
-----------
  - The Overview page now displays a "Non-mempool:" balance row showing funds that would be spent if wallet transactions currently outside the mempool were to confirm. This row is only shown when non-mempool wallet transactions are present.
```

src/qt/forms/overviewpage.ui
polespinasa: Maybe `non-standard` and then add a `?` icon next to it, if putting the mouse on top it shows a small explanation? IIRC we used that `?` in the past for some rbf comments.

pablomartin4btc: The label has a tooltip already, I left a suggestion [there](https://github.com/bitcoin-core/gui/pull/911#discussion_r3471624578).

src/qt/overviewpage.cpp
ajtowns: I think it's better to show the figure if it somehow ends up being positive (due to a bug or some future change) -- hiding surprising values is bad for auditability.

src/qt/forms/overviewpage.ui
ajtowns: I think "about to be spent" and "may never confirm" are mutually contradictory.

"Not in the mempool" describes what's going on, I think "violates local node policy" would accurately describe the "why" (conflicting with txs in the  mempool should show up differently iirc), but I'm not sure that's much more helpful -- if you want to act on the information you need a lot more context/details: it could just be that other txs need to confirm first (cluster/ancestor limits), it could be that you've deliberately set datacarriersize limits and could change them to let the tx into your mempool, it could be that you've somehow gotten a tx in your wallet that violates unconfigurable policies (perhaps via a different bitcoind version?).

Saying something simple and google-able like "not in the mempool" and letting users get more information from either docs or an AI agent or similar seems like the best approach to me, fwiw.

ajtowns:
Reviews seem to be on UI issues now which is fair enough for the gui, but not really my wheelhouse, so closing as up-for-grabs.

polespinasa:
> Reviews seem to be on UI issues now which is fair enough for the gui, but not really my wheelhouse, so closing as up-for-grabs.

Went ahead with it in #952
Added you as a co-author.

src/qt/overviewpage.cpp
polespinasa: Taken in #952 if you want to review

PR #915, opened 2025-11-22
Defer transaction signing until user clicks Send

### LINKED ISSUE #30070: Load PSBT error: Unable to decode PSBT

### Current behaviour

"Load PSBT from ..." from menu, error shows:
`Unable to decode PSBT 
Unsigned tx does not have empty scriptSigs and scriptWitnesses.: unspecified iostream_category error`

decodepsbt and analyzepsbt show the same error too:
`TX decode failed Unsigned tx does not have empty scriptSigs and scriptWitnesses.: unspecified iostream_category error (code -22)`

### Expected behaviour

Load PSBT success.

### Steps to reproduce

I test PSBT in my wallet today, create a transaction in Bitcoin Core, click "send" and "create unsigned", then "Load PSBT from ..." from menu

I'm using prune mode for block chain. This address I made tx was obtained by importprivkey, and then covert to Descriptor Wallets by migratewallet.

### Relevant log output

no relevant log.

Pre-built binaries

Bitcoin Core version v27.0.0

MacOS 14.4.1

### Machine specifications

_No response_

151henry151:
Fixes [#30070](https://github.com/bitcoin/bitcoin/issues/30070)

When creating an unsigned PSBT from the GUI, the transaction was already signed during preparation, causing legacy inputs to have non-empty scriptSig fields. The PSBT parser then rejects them.

This defers signing until the user clicks "Send" instead of signing during preparation. Fee calculation still works since transactions can be created without signing.

Follows the approach suggested by @achow101 in the issue comments.

### DIFF
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -281,6 +281,8 @@ bool SendCoinsDialog::PrepareSendText(QString& question_string, QString& informa
     }
 
     // prepare transaction for getting txFee earlier
+    // Create unsigned transaction to support creating unsigned PSBTs.
+    // Signing is deferred until the user clicks "Send".
     m_current_transaction = std::make_unique<WalletModelTransaction>(recipients);
     WalletModel::SendCoinsReturn prepareStatus;
 
@@ -357,7 +359,7 @@ bool SendCoinsDialog::PrepareSendText(QString& question_string, QString& informa
 
         // append transaction size
         //: When reviewing a newly created PSBT (via Send flow), the transaction fee is shown, with "virtual size" of the transaction displayed for context
-        question_string.append(" (" + tr("%1 kvB", "PSBT transaction creation").arg((double)m_current_transaction->getTransactionSize() / 1000, 0, 'g', 3) + "): ");
+        question_string.append(" (" + tr("%1 kvB (unsigned)", "PSBT transaction creation").arg((double)m_current_transaction->getTransactionSize() / 1000, 0, 'g', 3) + "): ");
 
         // append transaction fee value
         question_string.append("<span style='color:#aa0000; font-weight:bold;'>");
@@ -515,6 +517,12 @@ void SendCoinsDialog::sendButtonClicked([[maybe_unused]] bool checked)
         presentPSBT(psbtx);
     } else {
         // "Send" clicked
+        WalletModel::UnlockContext ctx(model->requestUnlock());
+        if (!ctx.isValid()) {
+            fNewRecipientAllowed = true;
+            return;
+        }
+
         assert(!model->wallet().privateKeysDisabled() || model->wallet().hasExternalSigner());
         bool broadcast = true;
         if (model->wallet().hasExternalSigner()) {
@@ -539,6 +547,24 @@ void SendCoinsDialog::sendButtonClicked([[maybe_unused]] bool checked)
                 } else {
                     presentPSBT(psbtx);
                 }
+            }
+        } else {
+            // Sign the transaction now that the user has confirmed they want to send.
+            CMutableTransaction mtx = CMutableTransaction{*(m_current_transaction->getWtx())};
+            PartiallySignedTransaction psbtx(mtx);
+            bool complete = false;
+            // Fill and sign the PSBT
+            const auto err{model->wallet().fillPSBT(std::nullopt, /*sign=*/true, /*bip32derivs=*/false, /*n_signed=*/nullptr, psbtx, complete)};
+            if (err || !complete) {
+                Q_EMIT message(tr("Send Coins"), tr("Failed to sign transaction."),
+                    CClientUIInterface::MSG_ERROR);
+                send_failure = true;
+                broadcast = false;
+            } else {
+                // Extract the signed transaction
+                CHECK_NONFATAL(FinalizeAndExtractPSBT(psbtx, mtx));
+                const CTransactionRef tx = MakeTransactionRef(mtx);
+                m_current_transaction->setWtx(tx);
             }
         }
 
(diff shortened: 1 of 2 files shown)


151henry151:
It looks like the one failing check is a CI issue with disk space too low. I'm not sure how to address that. I think the code of the PR is correct.

151henry151:
@achow101 would you like to check this out and let me know if it looks OK?

151henry151:
Should I push an empty commit to trigger a re-run of the CI here, or is there some other action needed?

src/qt/walletmodel.cpp
```diff
@@ -147,7 +147,7 @@ bool WalletModel::validateAddress(const QString& address) const
     return IsValidDestinationString(address.toStdString());
 }
 
-WalletModel::SendCoinsReturn WalletModel::prepareTransaction(WalletModelTransaction &transaction, const CCoinControl& coinControl)
+WalletModel::SendCoinsReturn WalletModel::prepareTransaction(WalletModelTransaction &transaction, const CCoinControl& coinControl, bool sign)
```
achow101: `prepareTransaction` has only a single caller that already sets `sign = false`. I don't think this needs to take a parameter, it can just always not sign.

achow101 (commented):
Concept ACK

src/qt/walletmodel.cpp
151henry151: suggestion implemented

src/qt/sendcoinsdialog.cpp
```diff
@@ -540,6 +542,24 @@ void SendCoinsDialog::sendButtonClicked([[maybe_unused]] bool checked)
                     presentPSBT(psbtx);
                 }
             }
+        } else {
+            // Sign the transaction now that the user has confirmed they want to send.
+            CMutableTransaction mtx = CMutableTransaction{*(m_current_transaction->getWtx())};
+            PartiallySignedTransaction psbtx(mtx);
+            bool complete = false;
+            // Fill and sign the PSBT
+            const auto err{model->wallet().fillPSBT(std::nullopt, /*sign=*/true, /*bip32derivs=*/false, /*n_signed=*/nullptr, psbtx, complete)};
```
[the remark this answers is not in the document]
151henry151: refactored as suggested

johnny9 (commented):
Double check this works with locked encrypted wallets

johnny9 (changes requested):
Need to move the UnlockContext when you sign the transaction otherwise it fails to send now.

https://github.com/user-attachments/assets/6b7e6894-b803-4753-9f20-66541af63633

src/qt/sendcoinsdialog.cpp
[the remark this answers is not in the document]
151henry151: @johnny9 I think I've addressed this correctly with [39a5fed](https://github.com/bitcoin-core/gui/commit/39a5fed760273d796d6b7e4cb298fc36d4d27022) -- it passed my local tests; do you want to check it out and see if it looks good to you?

johnny9: This is a little different now that the current transaction isn't signed yet so the kvB is different now than before.  Can either change the dialog to make this clear or adjust this size to be the same as the estimated size the fee calculation uses.

johnny9 (commented):
Moving the UnlockContext in the "Send" click flow resolved the previous issue. I think the previous UnlockContext is no longer needed.

I also see that the send confirmation dialog isn't as accurate as it was previously so should adjust what the dialog says or expose the correct fee estimated size through the model.

src/qt/sendcoinsdialog.cpp
```diff
@@ -269,23 +272,18 @@ bool SendCoinsDialog::PrepareSendText(QString& question_string, QString& informa
     }
 
     fNewRecipientAllowed = false;
-    WalletModel::UnlockContext ctx(model->requestUnlock());
-    if(!ctx.isValid())
-    {
-        // Unlock wallet was cancelled
```
johnny9: I think I was wrong about removing this. I think you will need unlock here too to be sure you can get a change address.

johnny9 (commented):
I actually think I was wrong and we need both UnlockContexts. Sorry to flp-flop on that.  Also take a look at achow's comment again. Seems like you didnt remove the parameter he mentioned. Other than those things I think this looks good now

Ari4ka (approved):

151henry151:
@johnny9 @achow101 I think this is ready for review, if anybody wants to check it out?

Ari4ka (approved):

151henry151:
Just checking in; anything further I can do here to get this ready for merge?

Ari4ka (approved):

maflcko:
According to https://btctranscripts.com/bitcoin-core-dev-tech/2026-05/qml-planning and https://btctranscripts.com/bitcoin-core-dev-tech/2026-05/qml-update, there seems to be momentum toward QML, so review in the legacy gui is reduced for now?

151henry151:
Rebased onto current master (4b91316643). Updated for the current `fillPSBT`/`PSBTFillOptions` API and post–gui#807 `createTransaction` flow.

`prepareTransaction` now always calls `createTransaction` with `sign=false`; internal wallets sign via `fillPSBT` only after the user clicks Send, with `UnlockContext` at prepare and send (encrypted wallets). The confirmation dialog labels tx size as unsigned. This should also subsume the overlapping `walletmodel.cpp` hunk in bitcoin/bitcoin#33112.

Tested with `cmake --build build --target bitcoin-qt test_bitcoin-qt` and `./build/bin/test_bitcoin-qt --run_all` — passed.

Re QML: agreed there's momentum toward a QML replacement, but this is a small fix for the GUI we still ship today (#30070); I don't think that direction makes this redundant in the meantime.

151henry151:
CI on the latest push is mostly failing at "Restore caches" with warpbuilds/cache/restore@v1 not allowed on this repo — could that be from the rebased master CI workflow rather than the patch itself, or did I miss something I should have done?

hebasto:
> CI on the latest push is mostly failing at "Restore caches" with warpbuilds/cache/restore@v1 not allowed on this repo — could that be from the rebased master CI workflow rather than the patch itself, or did I miss something I should have done?

Should work now.

Ari4ka (approved):

PR #922, opened 2026-01-09
Fix transactions disable problem

beati:
- In bitcoin-qt, there is under Settings the "Mask values" option
- When a wallet is open and the "Mask values" checkbox is checked, the "Transactions" tab is disabled
- When no wallet was open and the "Mask values" checkbox was checked and a wallet was opened, the "Transactions" tab was enabled
- This fix changes this, so that there is always the same behavior
- The "Transactions" tab is now always disabled when the "Mask values" checkbox is checked
- This fix is manually tested on the fork


hebasto:
Does this PR solve the same issue as https://github.com/bitcoin-core/gui/pull/815?

PR #924, opened 2026-01-12
Show an error message if the restored wallet name is empty

achow101:
The Restore Wallet dialog rejects wallet names that are empty, but was doing so silently. This is confusing, we should be presenting an error message to the user.


polespinasa (approved):
Code review and tested ack dd904298c13b14ef518e24fa63c6d0962f4a2de0

Tested the current (master) version and it's pretty annoying that the retire wallet screen just closes if the name is empty. Now the message is pretty clear.

Image

hebasto (approved):
ACK dd904298c13b14ef518e24fa63c6d0962f4a2de0. Tested on Fedora 43.

PR #925, opened 2026-02-01
Add "Hide conflicted" checkbox to Transactions tab

w0xlt:
Conflicted transactions can accumulate in the wallet over time. These transactions clutter the transaction list and cannot be filtered out.

While the Overview page already hides conflicted transactions using `TransactionFilterProxy::setShowInactive(false)`, the Transactions tab has no equivalent option, forcing users to scroll through potentially many obsolete entries.

This PR adds a "Hide conflicted" checkbox to the Transactions tab filter row, allowing users to toggle visibility of conflicted transactions. The implementation reuses the existing `setShowInactive()` function.

**Before:** 
Users must manually scroll past conflicted transactions with no filtering option. 

**After:** 
A checkbox in the filter row lets users hide conflicted transactions with one click.

image

### DIFF
--- a/src/qt/transactionview.cpp
+++ b/src/qt/transactionview.cpp
@@ -23,6 +23,7 @@
 #include <optional>
 
 #include <QApplication>
+#include <QCheckBox>
 #include <QComboBox>
 #include <QDateTimeEdit>
 #include <QDesktopServices>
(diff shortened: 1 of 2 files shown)


pablomartin4btc (commented):
Concept ACK

One thought: perhaps the Hide button should only appear when there are actually conflicted transactions present.

Alternatively, if Luke’s approach is preferred, it might be worth adding some visual indication (e.g. a label or status message) that conflicted transactions exist, or enabling a button/action to view them in a separate window.

Finally, if feasible, it would be great to include brief test instructions in the PR description, outlining how to verify the behavior before and after the change.

RandyMcMillan:
Concept ACK

> Approach NACK. "Conflicted" should be reduced to an implementation detail IMO. Instead, I suggest hiding conflicted transactions by default after N blocks, and have an advanced setting (in the settings dialog?) to show them.

@luke-jr suggestion ACK

making it a user option in settings will avoid ui clutter.

---

Show/Hide (default Hide)
Blocks N  (default disabled N=144)

---

PR #928, opened 2026-02-14
Replace three dots with ellipsis

hebasto:
From the translator's [comment](https://app.transifex.com/bitcoin/bitcoin/translate/#pl/qt-translation-031x/611215465/) on Transifex:
> Source text has 3 dots, instead of 3-dot character. Most commonly used form is the 3-dot character: …
You could consider updating it.


hebasto:
Reported by @jesterhodl.

PR #935, opened 2026-03-28
bugfix: truncate header sync percentage

### LINKED ISSUE #65: GUI misses some header updates when in IBD
Moved from https://github.com/bitcoin/bitcoin/issues/9464. Low prio, could probably just be closed as wont-fix.

OSX 10.12.2, Qt 5.7.1, master at 2a524b8e8fe69ef487fd8ea1b4f7a03f473ed201.

Running with `bitcoin-qt -reindex-chainstate`, "Number of blocks left" in the modal overlay never updates, and remains as "Unknown. Syncing Headers (446424)...". Even though we've synced > 200'000 blocks.
modal

arejula27:
During IBD, if `LoadingBlocks()` is true (e.g. during a chainstate rebuild after a crash), the node rejects incoming headers from peers. The GUI has headers loaded from the block index on disk, but `bestHeaderDate` becomes stale over time.

After ~4 hours (`HEADER_HEIGHT_DELTA_SYNC * nPowTargetSpacing`), the time-based heuristic in `modaloverlay.cpp` considers headers as "not synced" and falls into the `else` branch, displaying "Unknown" for both blocks remaining and estimated time. Meanwhile, `UpdateHeaderSyncLabel()` computes a percentage that rounds up to 100.0%, creating a contradictory display: `"Unknown. Syncing Headers (942188, 100.0%)"`. Please read issue #65 for more context.

 This patch truncates the percentage to one decimal place instead of rounding, so it only shows 100.0% when headers are truly up to date. The same fix is applied to
  `UpdateHeaderPresyncLabel()` for consistency.

Fixes https://github.com/bitcoin-core/gui/issues/65 .

### DIFF
--- a/src/qt/modaloverlay.cpp
+++ b/src/qt/modaloverlay.cpp
@@ -163,12 +163,14 @@ void ModalOverlay::tipUpdate(int count, const QDateTime& blockDate, double nVeri
 
 void ModalOverlay::UpdateHeaderSyncLabel() {
     int est_headers_left = bestHeaderDate.secsTo(QDateTime::currentDateTime()) / Params().GetConsensus().nPowTargetSpacing;
-    ui->numberOfBlocksLeft->setText(tr("Unknown. Syncing Headers (%1, %2%)…").arg(bestHeaderHeight).arg(QString::number(100.0 / (bestHeaderHeight + est_headers_left) * bestHeaderHeight, 'f', 1)));
+    const int pct = bestHeaderHeight ? static_cast<int>(1000LL * bestHeaderHeight / (bestHeaderHeight + est_headers_left)) : 0;
+    ui->numberOfBlocksLeft->setText(tr("Unknown. Syncing Headers (%1, %2%)…").arg(bestHeaderHeight).arg(QStringLiteral("%1.%2").arg(pct / 10).arg(pct % 10)));
 }
 
 void ModalOverlay::UpdateHeaderPresyncLabel(int height, const QDateTime& blockDate) {
     int est_headers_left = blockDate.secsTo(QDateTime::currentDateTime()) / Params().GetConsensus().nPowTargetSpacing;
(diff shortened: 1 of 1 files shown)


src/qt/modaloverlay.cpp
```diff
@@ -163,12 +165,18 @@ void ModalOverlay::tipUpdate(int count, const QDateTime& blockDate, double nVeri
 
 void ModalOverlay::UpdateHeaderSyncLabel() {
     int est_headers_left = bestHeaderDate.secsTo(QDateTime::currentDateTime()) / Params().GetConsensus().nPowTargetSpacing;
-    ui->numberOfBlocksLeft->setText(tr("Unknown. Syncing Headers (%1, %2%)…").arg(bestHeaderHeight).arg(QString::number(100.0 / (bestHeaderHeight + est_headers_left) * bestHeaderHeight, 'f', 1)));
+    double pct = 100.0 * bestHeaderHeight / (bestHeaderHeight + est_headers_left);
+    // Truncate to one decimal place to avoid displaying 100.0% when headers are still missing
+    pct = std::floor(pct * 10.0) / 10.0;
+    ui->numberOfBlocksLeft->setText(tr("Unknown. Syncing Headers (%1, %2%)…").arg(bestHeaderHeight).arg(QString::number(pct, 'f', 1)));
```
luke-jr:
```suggestion
    const int pct = bestHeaderHeight ? static_cast<int>(1000LL * bestHeaderHeight / (bestHeaderHeight + est_headers_left)) : 0;
    ui->numberOfBlocksLeft->setText(tr("Unknown. Syncing Headers (%1, %2%)…").arg(bestHeaderHeight).arg(QStringLiteral("%1.%2").arg(pct / 10).arg(pct % 10)));
```

arejula27: Good one, i guess it is more efficient as we do not work with floats, right?

Sjors:
@hebasto can you add a bug label to this PR and other bug fixes? That makes it easier to track them (e.g. with RSS feed).

arejula27:
@polespinasa

PR #936, opened 2026-05-05
Remove opt-in RBF

polespinasa:
With the whole network mostly running with Full Replace By Fee, it does not make sense to let the user signal or not to signal for RBF on their transactions.

With this PR the transactions created using the GUI will fallback to the wallet configuration in order to signal or not to signal BIP 125. (True by default).

### DIFF
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -179,13 +179,6 @@ void SendCoinsDialog::setModel(WalletModel *_model)
         connect(ui->groupFee, &QButtonGroup::idClicked, this, &SendCoinsDialog::coinControlUpdateLabels);
 
         connect(ui->customFee, &BitcoinAmountField::valueChanged, this, &SendCoinsDialog::coinControlUpdateLabels);
-#if (QT_VERSION >= QT_VERSION_CHECK(6, 7, 0))
-        connect(ui->optInRBF, &QCheckBox::checkStateChanged, this, &SendCoinsDialog::updateSmartFeeLabel);
-        connect(ui->optInRBF, &QCheckBox::checkStateChanged, this, &SendCoinsDialog::coinControlUpdateLabels);
-#else
-        connect(ui->optInRBF, &QCheckBox::stateChanged, this, &SendCoinsDialog::updateSmartFeeLabel);
-        connect(ui->optInRBF, &QCheckBox::stateChanged, this, &SendCoinsDialog::coinControlUpdateLabels);
-#endif
         CAmount requiredFee = model->wallet().getRequiredFee(1000);
         ui->customFee->SetMinValue(requiredFee);
         if (ui->customFee->value() < requiredFee) {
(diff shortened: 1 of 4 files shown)


src/qt/sendcoinsdialog.cpp
```diff
@@ -360,15 +350,6 @@ bool SendCoinsDialog::PrepareSendText(QString& question_string, QString& informa
         question_string.append("<span style='color:#aa0000; font-weight:bold;'>");
         question_string.append(BitcoinUnits::formatHtmlWithUnit(model->getOptionsModel()->getDisplayUnit(), txFee));
         question_string.append("</span><br />");
-
-        // append RBF message according to transaction's signalling
-        question_string.append("<span style='font-size:10pt; font-weight:normal;'>");
-        if (ui->optInRBF->isChecked()) {
-            question_string.append(tr("You can increase the fee later (signals Replace-By-Fee, BIP-125)."));
```
sedited, line 367: Might be good to keep the "You can increase the fee later" string?

polespinasa: Yes, makes sens, added back.

sedited (approved):
ACK 90eda67bb8f9c3b82a3eccc787446269c83b2558

polespinasa:
Failing test seems unrelated

src/qt/sendcoinsdialog.cpp
```diff
@@ -360,17 +350,12 @@ bool SendCoinsDialog::PrepareSendText(QString& question_string, QString& informa
...
-        } else {
-            question_string.append(tr("Not signalling Replace-By-Fee, BIP-125."));
-        }
-        question_string.append("</span>");
     }
 
+    // append RBF message
+    question_string.append("<span style='font-size:10pt; font-weight:normal;'>");
+    question_string.append(tr("You can increase the fee later."));
```
hebasto, line 357: What if Bitcoin Core is run with `-walletrbf=0`?

hebasto (commented):
Is the "RPC only" note still correct here:https://github.com/bitcoin-core/gui/blob/cad5f5604582eac95bcfbf08cca8af1c186c36eb/src/wallet/init.cpp#L77?

src/qt/sendcoinsdialog.cpp
polespinasa: Have to test to ensure there's no random check that blocks the transaction from being broadcasted but it is redundant.

Anyway there is work ongoing towards deprecating and removing it.
https://github.com/bitcoin/bitcoin/pull/34917


Edit: Just checked and it doesn't matter, the wallet still broadcast it, we are not enforcing RBF even if `-walletrbf=0`. The only difference is the sequence number value, but has no effect.

polespinasa:
> Is the "RPC only" note still correct here

Not at all. If this gets merged I can open a follow-up PR to change that line, maybe can just ask to get it modified here https://github.com/bitcoin/bitcoin/pull/34917.
Happens when repos are split 🤷‍♂️

achow101:
ACK 90eda67bb8f9c3b82a3eccc787446269c83b2558

PR #938, opened 2026-06-03
qt: Preserve headers presync progress

djh58:
## Summary

During headers pre-sync, the pre-sync height and the accepted-header height can describe different phases of initial sync. The GUI previously allowed both `HEADER_PRESYNC` and normal header/block sync updates to repaint the same status and modal overlay labels.

As a result, an active `Pre-syncing Headers (...)` display could be replaced by a lower `Syncing Headers (...)` accepted-header progress value while headers pre-sync was still active. The modal overlay could also continue showing normal block verification progress during headers pre-sync.

This PR keeps the headers pre-sync display active until accepted headers catch up, then restores normal header/block sync display behavior. It also labels modal progress as headers pre-sync progress during that phase instead of showing stale block verification progress.

This is a GUI display-only change. It does not change header validation, block validation, peer behavior, block download, or sync state.

## Testing

- `cmake -B build-qt -DCMAKE_BUILD_TYPE=Debug -DBUILD_GUI=ON -DBUILD_TESTS=ON -DBUILD_GUI_TESTS=ON -DBUILD_BENCH=OFF -DBUILD_FUZZ_BINARY=OFF`
- `cmake --build build-qt --target test_bitcoin-qt -j"$(sysctl -n hw.ncpu)"`
- `ulimit -n 10240 && ctest --test-dir build-qt -R test_bitcoin-qt --output-on-failure`

The new regression coverage is `ModalOverlayTests::headersPresyncProgressStaysVisible`.

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -1161,9 +1161,21 @@ void BitcoinGUI::setNumBlocks(int count, const QDateTime& blockDate, double nVer
     switch (blockSource) {
         case BlockSource::NETWORK:
             if (synctype == SyncType::HEADER_PRESYNC) {
+                m_headers_presync_active = true;
+                m_headers_presync_height = count;
+                m_headers_presync_block_date = blockDate;
                 updateHeadersPresyncProgressLabel(count, blockDate);
                 return;
-            } else if (synctype == SyncType::HEADER_SYNC) {
+            }
+            if (m_headers_presync_active) {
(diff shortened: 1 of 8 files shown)


pinheadmz:
Smells like AI. Do you, the human, understand the changes? Before we run CI can you post some screenshots of what you have done here and explain why the change is needed?

djh58:
I observed this in a local build while setting up a node from a fresh datadir. When it was syncing, the progress tracker would alternate b/t "Pre-syncing Headers" and "Syncing Headers", which are two separate phases of header sync to my understanding. I'll try to reproduce again and get screenshots or a recording.

PR #939, opened 2026-06-15
gui: show wallet name in progress messages

### LINKED ISSUE #259: When opening or autoloading wallets there should be clear messages about rescanning in progress and wallets' names.

ShauryaaSharma:
## What

Include the wallet display name in the `initMessage` calls made during
wallet loading so the splash screen clearly identifies which wallet is
being processed:

- `LoadWalletInternal`: `"Loading wallet…"` → `"[wallet_name] Loading wallet…"`
- `CreateWallet`: `"Creating wallet…"` → `"[wallet_name] Creating wallet…"`
- `CWallet::AttachChain`: `"Rescanning…"` → `"[wallet_name] Rescanning…"` *(primary fix)*

For unnamed (default) wallets, `DisplayName()` already returns the
translated `"default wallet"` string, handling that case correctly.

A comment is added to `LoadWalletsActivity::load()` explaining why no
specific wallet name is shown there — it loads multiple wallets
simultaneously from the node's wallet list.

## Why

When a wallet that has not been synchronized recently is opened — either
via `settings.json` at startup or via File > Open Wallet — the splash
screen shows a bare `"Rescanning…"` with no indication of which wallet
is being scanned. Users with multiple wallets have no way to know which
one is being processed without checking the debug log.

The `ShowProgress` calls inside `ScanForWalletTransactions` already use
`"[wallet_name] Rescanning…"` (lines 1888/1903/2008); this change makes
the preceding `initMessage` consistent with them.

Fixes #259

### DIFF
--- a/src/qt/walletcontroller.cpp
+++ b/src/qt/walletcontroller.cpp
@@ -375,6 +375,8 @@ LoadWalletsActivity::LoadWalletsActivity(WalletController* wallet_controller, QW
 
 void LoadWalletsActivity::load(bool show_loading_minimized)
 {
+    // No specific wallet name is shown here because multiple wallets may be
+    // loaded simultaneously when syncing the GUI with the node's wallet list.
     showProgressDialog(
         //: Title of progress window which is displayed when wallets are being loaded.
         tr("Load Wallets"),
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -283,7 +283,7 @@ std::shared_ptr<CWallet> LoadWalletInternal(WalletContext& context, const std::s
             return nullptr;
         }
 
-        context.chain->initMessage(_("Loading wallet…"));
+        context.chain->initMessage(strprintf("[%s] %s", name.empty() ? std::string{_("default wallet")} : name, _("Loading wallet…")));
         std::shared_ptr<CWallet> wallet = CWallet::LoadExisting(context, name, std::move(database), error, warnings);
         if (!wallet) {
             error = Untranslated("Wallet loading failed.") + Untranslated(" ") + error;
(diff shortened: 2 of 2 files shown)


ShauryaaSharma:

Small UX consistency fix, no logic changes, only `initMessage` call-site text updates across two files.

**Problem:** when a wallet that hasn't synced recently is opened (via `settings.json` at startup or File > Open Wallet), the splash screen shows a bare `"Rescanning…"` with no indication of which wallet is being scanned. Users with multiple wallets have to check the debug log to find out.

**Fix:** include `DisplayName()` in three `initMessage` calls:

-   `LoadWalletInternal`: `"Loading wallet…"` → `"[wallet_name] Loading wallet…"`
-   `CreateWallet`: `"Creating wallet…"` → `"[wallet_name] Creating wallet…"`
-   `CWallet::AttachChain`: `"Rescanning…"` → `"[wallet_name] Rescanning…"` ← primary fix

`DisplayName()` already returns the translated `"default wallet"` string for unnamed wallets, so that case is handled correctly without extra logic.

**Consistency note:** `ShowProgress` calls inside `ScanForWalletTransactions` (lines 1888/1903/2008) already use `"[wallet_name] Rescanning…"` , this change makes the preceding `initMessage` consistent with them.

**Re conflicts with #30343 and #25722:** both are drafts and neither has been merged. Happy to rebase once either lands if that helps move things forward.

Fixes #259.

PR #942, opened 2026-06-20
qt: prevent console hang on large RPC responses

ShauryaaSharma:
## Problem

Fixes #932.

Running `getrawtransaction` on a very large non-standard transaction in the bitcoin-qt console causes the GUI to become completely unresponsive.

Reproduction:

getrawtransaction 30794005e228089a56cca58b6ad85e8cc0667cdc06b0ee6f9fe72aa19f50acd2

The transaction is ~3.98 MB and sits in block 938523. Entering the command above leaves the node window frozen until the process is killed. Reported in #932.

## Root cause

The hang is not in the node or the RPC execution path. It occurs in the Qt UI thread after the result is delivered to the console widget.

Two operations in `RPCConsole::message()` scale badly with payload size:

1. `GUIUtil::HtmlEscape()` processes the result string character-by-character. A 3.98 MB transaction encodes to ~7.96 MB of  hex, all of which this function walks before returning.

2. `QTextEdit::append()` is then called with an HTML string of the same magnitude. Qt re-lays-out the entire text document on each call; for a payload this size the layout pass does not complete in any practical amount of time, stalling the event loop indefinitely.

## Fix

Add a 1 MiB size guard in `RPCExecutor::request()`, immediately after `RPCExecuteCommandLine()` returns and before `QString::fromStdString()` is called. When the result exceeds the limit the widget receives a short notice instead of the raw output:

Response too large to display (7962624 bytes). Use bitcoin-cli to retrieve the full result.

Because the guard fires before `QString` conversion, the oversized string is never converted to UTF-16, never HTML-escaped, and never handed to the text widget.

The 1 MiB threshold:
- reuses the `_MiB` literal already present in this file (`ui->lineEdit->setMaxLength(16_MiB)`)
- is well above any realistic interactive debugging output
- is well below the ~7.96 MiB that triggers the hang

## Tests

A mock RPC command `rpcLargeOutput` is registered in `src/qt/test/rpcnestedtests.cpp`. It returns a string of `1_MiB + 1` bytes. The test asserts that `RPCExecuteCommandLine` returns the complete result unmodified, confirming the size guard in
`RPCExecutor::request()` does not affect the underlying RPC call, only what is rendered in the console widget.

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -414,6 +414,18 @@ void RPCExecutor::request(const QString &command, const QString& wallet_name)
             return;
         }
 
+        // Avoid passing a very large string to the Qt text widget. Appending
+        // many megabytes of HTML to QTextEdit causes the UI thread to hang
+        // while the document is re-laid-out. Direct the user to bitcoin-cli
+        // for commands whose output exceeds the display limit.
+        if (result.size() > 1_MiB) {
+            Q_EMIT reply(RPCConsole::CMD_REPLY,
+                QString("Response too large to display (%1 bytes). "
+                        "Use bitcoin-cli to retrieve the full result.")
+                    .arg(result.size()));
+            return;
+        }
+
         Q_EMIT reply(RPCConsole::CMD_REPLY, QString::fromStdString(result));
     }
     catch (UniValue& objError)
--- a/src/qt/test/rpcnestedtests.cpp
+++ b/src/qt/test/rpcnestedtests.cpp
@@ -9,6 +9,7 @@
 #include <qt/rpcconsole.h>
 #include <rpc/server.h>
 #include <test/util/setup_common.h>
+#include <util/byte_units.h>
 #include <univalue.h>
 
 #include <QTest>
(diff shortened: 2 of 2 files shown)


ShauryaaSharma:

This PR fixes a hard hang in the bitcoin-qt console when running RPC commands that return large payloads (e.g. `getrawtransaction` on a ~3.98 MB transaction).

**Root cause:** two operations in `RPCConsole::message()` scale badly with payload size:

-   `GUIUtil::HtmlEscape()` walks the entire result string character-by-character
-   `QTextEdit::append()` re-lays-out the full text document on each call, for a ~7.96 MB hex payload this stalls the Qt event loop indefinitely

**Fix:** a 1 MiB size guard in `RPCExecutor::request()` fires before `QString::fromStdString()` is called, so the oversized string is never converted, never escaped, and never handed to the widget. The user sees a short notice instead:

Response too large to display (7962624 bytes). Use bitcoin-cli to retrieve the full result.


**Test coverage:** a mock RPC command `rpcLargeOutput` is registered in `src/qt/test/rpcnestedtests.cpp`, returning `1_MiB + 1` bytes, asserting the guard does not affect the underlying RPC result, only the console rendering.

No behaviour change for normal-sized responses. Happy to adjust the threshold or message wording if needed.

Fixes #932.

RandyMcMillan:
Concept ACK

PR #943, opened 2026-06-20
qt: warn about rescan in wallet migration dialogs

### LINKED ISSUE #930: Wallet migration includes rescan
Users may want to migrate legacy wallets they haven't used in years. The migration itself is typically very quick, but it's followed by a rescan (if the wallet is behind).

Depending on whether the user has `-blockfilterindex` enabled, this rescan takes somewhere between long and really really long.

There's no visual indication, only the logs show what's going on:

Image

Ideally the migration should finish and then prompt the user that a rescan is about to happen. That way the user can abort the rescan and resume it some other time if it takes too long.

Alternatively, the migrate wallet dialog could at least warn that a rescan is (or may be) happening and could take a while.

Fun aside: block filters don't work on legacy wallets, so it's much faster to first migrate and then rescan.

ShauryaaSharma:
## Problem

Fixes #930.

Migrating a legacy wallet triggers an automatic blockchain rescan
immediately after the conversion completes. Depending on chain size
and whether `-blockfilterindex` is enabled, this rescan can take
anywhere from minutes to hours. During this time the progress dialog
shows "Migrating Wallet…" with no indication that a rescan is
happening — only the debug log reveals what is going on.

## Fix

Three changes, all in `MigrateWalletActivity`:

**1 & 2 — Confirmation dialogs** (`migrate()` and
`restore_and_migrate()`): append a rescan warning to the informative
text so users know what to expect before clicking Confirm:

> After migration, the wallet will be rescanned for transactions.
> Depending on the size of the blockchain and whether block filter
> indexes are available, this rescan may take a significant amount of
> time.

**3 — Progress dialog** (`do_migrate()`): update the label from the
bare "Migrating Wallet…" to also mention the rescan, so users who are
already in the waiting phase have context:

> Migrating Wallet **name**…
>
> A blockchain rescan will follow and may take a significant amount of
> time.

## Notes

- All changes are confined to `src/qt/walletcontroller.cpp`.
- No behaviour or logic is changed — this is purely informational UI
  text.
- The rescan itself is performed inside `migrateWallet()` as a single
  blocking call; separating migration from rescan would require wallet
  backend changes and is out of scope here.

### DIFF
--- a/src/qt/walletcontroller.cpp
+++ b/src/qt/walletcontroller.cpp
@@ -448,7 +448,10 @@ void MigrateWalletActivity::do_migrate(const std::string& name)
         if (dlg.exec() == QDialog::Rejected) return;
     }
 
-    showProgressDialog(tr("Migrate Wallet"), tr("Migrating Wallet <b>%1</b>…").arg(GUIUtil::HtmlEscape(name)));
+    showProgressDialog(tr("Migrate Wallet"),
+        tr("Migrating Wallet <b>%1</b>…<br><br>"
+           "A blockchain rescan will follow and may take a significant amount of time.")
+            .arg(GUIUtil::HtmlEscape(name)));
 
     QTimer::singleShot(0, worker(), [this, name, passphrase] {
         auto res{node().walletLoader().migrateWallet(name, passphrase)};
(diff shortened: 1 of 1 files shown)


ShauryaaSharma:

This is a small UX-only fix for wallet migration, no logic or behaviour changes, purely informational text updates confined to `src/qt/walletcontroller.cpp`.

**Problem:** migrating a legacy wallet silently triggers a blockchain rescan that can take minutes to hours. The progress dialog just shows `"Migrating Wallet…"` with no indication a rescan is coming or is in progress, users have no way to know without watching the debug log.

**Three text-only changes in `MigrateWalletActivity`:**

-   `migrate()` confirmation dialog, appends rescan warning before the user clicks Confirm
-   `restore_and_migrate()` confirmation dialog, same warning appended
-   `do_migrate()` progress label, updated from bare `"Migrating Wallet…"` to also mention the rescan:

    ```
    Migrating Wallet name…

    A blockchain rescan will follow and may take a significant amount of time.
    ```

**Out of scope (noted for clarity):** separating the migration step from the rescan would require wallet backend changes, this PR intentionally doesn't touch that.

Happy to adjust the wording if it doesn't match the tone used elsewhere in the Qt dialogs.

Fixes #930.

RandyMcMillan:
Concept ACK

RandyMcMillan:
an additional PR may include the multichain error handling and alerts. I noticed those need some attention.

PR #944, opened 2026-07-03
Fix out-of-bounds read in RPCParseCommandLine on empty command

nabhan06:
When a console line has no command name (it starts with `)`, or is `()`, `(`, or `,`), RPCParseCommandLine reaches the command-execution branch while the current argument frame is still empty, so `stack.back()[0]` reads out of bounds and the argument list built from `stack.back().begin() + 1` to `end()` is an invalid iterator range (throws std::length_error in practice, UBSan flags the null-pointer reference otherwise).

The `(` branch already guards the frame with `stack.back().size() > 0`, so I add the same check to the `)`/newline branch and the empty frame is skipped. To be clear, `(` alone isn't safe on master either: it fails via the `\n` branch, not via the `(` branch itself (the state there isn't `STATE_ARGUMENT`), and `,` alone fails the same way.

Since there's no command to run in any of these cases, the parser now returns `false` so the console reports an invalid command line, consistent with other fully-invalid input like a bare `'` or `"`, rather than silently ignoring it.

Regression cases added to rpcNestedTests for `)`, `()`, `(` and `,` (all abort on master without the guard), plus `getblockchaininfo)` which stays tolerated.

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -159,6 +159,7 @@ bool RPCConsole::RPCParseCommandLine(interfaces::Node* node, std::string &strRes
     } state = STATE_EATING_SPACES;
     std::string curarg;
     UniValue lastResult;
+    bool command_parsed = false;
     unsigned nDepthInsideSensitive = 0;
     size_t filter_begin_pos = 0, chpos;
     std::vector<std::pair<size_t, size_t>> filter_ranges;
@@ -290,7 +291,7 @@ bool RPCConsole::RPCParseCommandLine(interfaces::Node* node, std::string &strRes
                         curarg.clear();
                         state = STATE_EATING_SPACES_IN_BRACKETS;
                     }
-                    if ((ch == ')' || ch == '\n') && stack.size() > 0)
+                    if ((ch == ')' || ch == '\n') && stack.size() > 0 && stack.back().size() > 0)
                     {
                         if (fExecute) {
                             // Convert argument list to JSON objects in method-dependent way,
@@ -306,6 +307,7 @@ bool RPCConsole::RPCParseCommandLine(interfaces::Node* node, std::string &strRes
                             lastResult = node->executeRpc(method, params, uri);
                         }
 
+                        command_parsed = true;
                         state = STATE_COMMAND_EXECUTED;
                         curarg.clear();
                     }
@@ -372,8 +374,11 @@ bool RPCConsole::RPCParseCommandLine(interfaces::Node* node, std::string &strRes
                 strResult = lastResult.write(2);
             [[fallthrough]];
         case STATE_ARGUMENT:
-        case STATE_EATING_SPACES:
             return true;
+        case STATE_EATING_SPACES:
+            // Reaching this state without ever parsing a command means the line
+            // held no command name (e.g. ")", "()", "(", ","); treat it as invalid.
+            return command_parsed;
         default: // ERROR to end in one of the other states
             return false;
     }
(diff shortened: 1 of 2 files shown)


RandyMcMillan:
Concept ACK

src/qt/rpcconsole.cpp
```diff
@@ -290,7 +290,7 @@ bool RPCConsole::RPCParseCommandLine(interfaces::Node* node, std::string &strRes
                         curarg.clear();
                         state = STATE_EATING_SPACES_IN_BRACKETS;
                     }
-                    if ((ch == ')' || ch == '\n') && stack.size() > 0)
+                    if ((ch == ')' || ch == '\n') && stack.size() > 0 && stack.back().size() > 0)
```
pablomartin4btc, line 294: The guard is the right approach and the scope is correct: both `)` and `\n` need it since `)` alone / `()` fail at the `)` check before the appended `\n` is ever reached, while `(` alone and `,` alone fail only via the `\n` path.

src/qt/test/rpcnestedtests.cpp
```diff
@@ -134,6 +134,8 @@ void RPCNestedTests::rpcNestedTests()
     QVERIFY_EXCEPTION_THROWN(RPCConsole::RPCExecuteCommandLine(m_node, result, "getblockchaininfo() getblockchaininfo()"), std::runtime_error); //invalid syntax
     RPCConsole::RPCExecuteCommandLine(m_node, result, "getblockchaininfo("); //tolerate non closing brackets if we have no arguments
     RPCConsole::RPCExecuteCommandLine(m_node, result, "getblockchaininfo()()()"); //tolerate non command brackets
+    RPCConsole::RPCExecuteCommandLine(m_node, result, ")"); //tolerate a closing bracket with no command (empty argument stack)
+    RPCConsole::RPCExecuteCommandLine(m_node, result, "()"); //tolerate empty brackets with no command
```
pablomartin4btc: Missing some regression cases. Two additional inputs fail on master with the same OOB but aren't covered by the PR's two new tests:
- `RPCConsole::RPCExecuteCommandLine(m_node, result, "(");`  // fails on master via `\n` branch, fixed by this PR
- `RPCConsole::RPCExecuteCommandLine(m_node, result, ",");`  // same

This test doesn't fail but it's also missing:
- `RPCConsole::RPCExecuteCommandLine(m_node, result, "getblockchaininfo)”); //tolerate a closing bracket with no argument stack`

pablomartin4btc (commented):
Concept ACK on the fix — `Error: vector` is a raw internal exception message with no meaning to the user.

I think the PR description is slightly misleading. It says "The `(` branch already guards this with `stack.back().size() > 0`", implying `(` alone is safe on master. It isn't — `(` alone does fail on master via the `\n` branch, just not via the `(` branch itself (state != `STATE_ARGUMENT`). Worth clarifying.

Tested at fb98a780355419aa5942b09f3cc4bc5342d714bd.

**_Behaviour suggestion_** — rather than silently ignoring these inputs as proposed, I'd prefer returning `false` so the UI shows "Invalid command line" popup, consistent with other fully-invalid input (e.g. bare `'` or `"`) as there is no command name to speak of in these cases either. Silent ignore feels inconsistent with existing error handling.

Left a couple of comments.

src/qt/test/rpcnestedtests.cpp
nabhan06: Added all three: `(` and `,` now have their own cases (both abort on master via the `\n` branch), and `getblockchaininfo)` as a tolerated case. Since these no-command inputs now return false, the `)`/`()`/`(`/`,` cases assert that with QVERIFY.

nabhan06:
Good points, both addressed.

Switched the no-command inputs (`)`, `()`, `(`, `,`) to return false instead of silently ignoring them, so the console shows an invalid command line like it does for a bare `'` or `"`. A `command_parsed` flag tracks whether the execution branch ever ran, and the final `STATE_EATING_SPACES` returns it. The size guard stays since it's what keeps the OOB read from happening in the first place.

Also fixed the description: `(` alone does fail on master, via the `\n` branch rather than the `(` branch itself. Rewrote that paragraph and added the extra regression cases. rpcNestedTests is green.

nabhan06:
any update?

hebasto (approved):
ACK fef99e6563ae284811904b980069166621c4fa22, tested on Ubuntu 26.04.

PR #945, opened 2026-07-09
qt: Fix sign message address book filtering

### LINKED ISSUE #62: Sign Message UI allows selection of keyless addresses
Moved from: https://github.com/bitcoin/bitcoin/issues/17969/

## What Happens
When you click File|Sign Message, the UI asks you to choose an address with which to sign.  You can enter one by hand, paste from clipboard, or "Choose a previously used address."  The third option provides a list of addresses that includes addresses for which there is no private key (P2SH addresses).  This is intimately related to #10542 but the fix _here_ is straightforward.

## Expected Behavior
I expect addresses for which no private key is known to be excluded from the list presented.  I realize that the same list is used to select an address to which to receive a payment, but that code could be adjusted to accept a filter, or the private-key-less addresses could be filtered out during display.

## To reproduce
Ensure that you have at least one P2SH address in your Bitcoin core wallet.
Select File | Sign Message
Click the address book.
Note that there are P2SH addresses offered for selection.

This is in Bitcoin Core 0.18.

Bushstar:
Two bugs here found when signing some pretty ancient addresses! Fixed in one PR as they share the same root cause and code path.

This PR fixes sign message address selection so it only shows signable legacy addresses and no longer mutates the wallet’s shared address book model.

The problem was that the sign message picker used `WalletModel::refresh(bool pk_hash_only)` as a filter, it was was too loose a filter and it replaced the wallet's shared address book model leaving other views stale.

The fix is one change, stop mutating the shared wallet model and give the signing picker its own local filter.

This is related it seems and could be closed: https://github.com/bitcoin-core/gui/issues/62

#### Reproduce address picker showing invalid addresses

image

1. In console: `createwallet "watchonly" true true`
2. Select watchonly wallet in console
3. In console: `importdescriptors '[{"desc":"addr(mipcBbFg9gMiCh81Kj8tqqdgoZub1ZJRfn)#eccnnjdy","timestamp":"now","label":"keyless legacy"}]'`
4. In main window select the watchonly wallet
5. Open File > Sign message....
6. Click the address book button next to the signing address field.
7. Actual current behavior: keyless legacy appears as a selectable signing address.
8. Choose it, enter a message, click Sign Message.

#### Reproduce Receiving Addresses window stops updating

1. In console: `createwallet "non-updating"`
2. In the main window select "non-updating" wallet
3. Open File > Sign message..., click the signing address-book button, then close the picker.
4. In console: `getnewaddress "after picker" bech32`
5. Open Window > Receiving addresses, the new address is not shown

### DIFF
--- a/src/qt/addressbookpage.h
+++ b/src/qt/addressbookpage.h
@@ -38,7 +38,12 @@ public:
         ForEditing  /**< Open address book for editing */
     };
 
-    explicit AddressBookPage(const PlatformStyle *platformStyle, Mode mode, Tabs tab, QWidget *parent = nullptr);
+    enum class AddressFilter {
+        ShowAll,
+        Signable, /**< Show only addresses that can sign messages */
+    };
+
+    explicit AddressBookPage(const PlatformStyle *platformStyle, Mode mode, Tabs tab, QWidget *parent = nullptr, AddressFilter address_filter = AddressFilter::ShowAll);
     ~AddressBookPage();
 
     void setModel(AddressTableModel *model);
@@ -52,6 +57,7 @@ private:
     AddressTableModel* model{nullptr};
     Mode mode;
     Tabs tab;
+    AddressFilter address_filter;
     QString returnValue;
     AddressBookSortFilterProxyModel *proxyModel;
     QMenu *contextMenu;
--- a/src/qt/walletmodel.cpp
+++ b/src/qt/walletmodel.cpp
@@ -583,11 +583,6 @@ bool WalletModel::isMultiwallet() const
     return m_node.walletLoader().getWallets().size() > 1;
 }
 
-void WalletModel::refresh(bool pk_hash_only)
-{
-    addressTableModel = new AddressTableModel(this, pk_hash_only);
-}
-
 uint256 WalletModel::getLastBlockProcessed() const
 {
     return m_client_model ? m_client_model->getBestBlockHash() : uint256{};
--- a/src/qt/addressbookpage.cpp
+++ b/src/qt/addressbookpage.cpp
@@ -20,11 +20,13 @@
 class AddressBookSortFilterProxyModel final : public QSortFilterProxyModel
 {
     const QString m_type;
+    const AddressBookPage::AddressFilter m_address_filter;
 
 public:
-    AddressBookSortFilterProxyModel(const QString& type, QObject* parent)
+    AddressBookSortFilterProxyModel(const QString& type, AddressBookPage::AddressFilter address_filter, QObject* parent)
         : QSortFilterProxyModel(parent)
         , m_type(type)
+        , m_address_filter(address_filter)
     {
         setDynamicSortFilter(true);
         setFilterCaseSensitivity(Qt::CaseInsensitive);
@@ -42,6 +44,10 @@ protected:
         }
 
         auto address = model->index(row, AddressTableModel::Address, parent);
+        if (m_address_filter == AddressBookPage::AddressFilter::Signable &&
+            !model->data(address, AddressTableModel::CanSignMessageRole).toBool()) {
+            return false;
+        }
 
         const auto pattern = filterRegularExpression();
         return (model->data(address).toString().contains(pattern) ||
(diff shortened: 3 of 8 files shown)


src/qt/walletmodel.cpp
```diff
@@ -583,11 +583,6 @@ bool WalletModel::isMultiwallet() const
     return m_node.walletLoader().getWallets().size() > 1;
 }
 
-void WalletModel::refresh(bool pk_hash_only)
```
pablomartin4btc, line 586: nit (only if you have to retouch and you agree ofc):
I'd create another commit for the removal of `WalletModel::refresh()` as a separate cleanup commit, so a reviewer clearly sees: here's the new mechanism, here's the removal of the old broken one.

src/qt/addressbookpage.h
```diff
@@ -38,7 +38,12 @@ class AddressBookPage : public QDialog
         ForEditing  /**< Open address book for editing */
     };
 
-    explicit AddressBookPage(const PlatformStyle *platformStyle, Mode mode, Tabs tab, QWidget *parent = nullptr);
+    enum class Filter {
```
pablomartin4btc: nit: I think `Filter` is too generic — even it sits alongside `Mode` and `Tabs` in the same class; could be `AddressFilter` or just be more descriptive (other options to consider `RowFilter` or `EntryFilter`). Since it's enum class it's scoped (externally `AddressBookPage::Filter`) which helps, but within the class itself `Filter` somehow clashes conceptually with `QSortFilterProxyModel` filtering already happening.

src/qt/addressbookpage.h
```diff
@@ -38,7 +38,12 @@ class AddressBookPage : public QDialog
...
+        None,
+        Signable,
```
pablomartin4btc: nit: `Mode` already uses that doxygen `/**<` style inline on each value...
```suggestion
        Signable, /**< Show only addresses that can sign messages */
```

src/qt/addressbookpage.h
```diff
@@ -38,7 +38,12 @@ class AddressBookPage : public QDialog
         ForEditing  /**< Open address book for editing */
     };
 
-    explicit AddressBookPage(const PlatformStyle *platformStyle, Mode mode, Tabs tab, QWidget *parent = nullptr);
+    enum class Filter {
+        None,
```
pablomartin4btc: nit: minor, but `None` as the default (meaning "no filter applied, show all") perhaps is slightly counterintuitive. `ShowAll` would read more naturally as the default I think (or `NoFilter`).
```suggestion
        ShowAll,
```

pablomartin4btc (commented):
tACK 1238afb8e5190d6a9929e10b1a64aa1f28d6bcc9

Reviewed both bugs carefully, reproduced both on master, confirmed fixed on this branch.

<details>
<summary>No Qt tests existed for this flow. Worked on two regression tests in <code>addressbooktests.cpp</code> on top if this PR if you want to include them...</summary>
<br>

- `TestAddressTableModelStability`: spins up `SignVerifyMessageDialog`, invokes `on_addressBookButton_SM_clicked()` via `QMetaObject::invokeMethod`, closes the modal via `QTimer::singleShot`, verifies the shared `addressTableModel` pointer is unchanged after the interaction; (d318af73a9ce35a4a59d1a4023343ebd998b16f1)

- `TestSignableAddressFilter`: sets up three addresses (watch-only `PKHash`, `bech32`, spendable `PKHash` via `pkh()` descriptor); verifies only the spendable `PKHash` reports `CanSignMessageRole=true` and that `Filter::Signable` shows exactly one row. (c671e8782e7e64ca798bab4710d87ae13769c1fe)

Both are skipped on macOS, but run successfully on Linux.

</details>

Left a couple of suggestions (none blocker).

Bushstar:
@pablomartin4btc thanks for your review and tests. I've updated the PR as per your suggestions and updated the filter var as well to be consistent with the renamed AddressFilter enum for better clarity. Few minor updates to the tests for the rename changes and to satisfy the LLM linter.

pablomartin4btc (commented):
ACK 891068acf31db586b274f7dbf0be463098e7c788

PR #947, opened 2026-07-14
gui: Update RPC console wallet when the active wallet changes

kwsantiago:
When multiple wallets are loaded, switching the active wallet via the main window's wallet selector doesn't update the wallet used by the RPC console. Console commands keep running against the previously selected wallet, so a wallet RPC can be sent to a different wallet than the one shown as active.

The open/create/restore/migrate paths already call `RPCConsole::setCurrentWallet()`, but the main window selector was missed. Add the call to the central `setCurrentWallet()`.

To reproduce: load two wallets, switch the active wallet in the toolbar, then run a wallet command such as `getwalletinfo` in the console.

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -826,6 +826,7 @@ void BitcoinGUI::setCurrentWallet(WalletModel* wallet_model)
 {
     if (!walletFrame || !m_wallet_controller) return;
     walletFrame->setCurrentWallet(wallet_model);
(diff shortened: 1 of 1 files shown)


polespinasa (commented):
ACK 9ee30341032cdd414d6b74c9ebe97fb8daf1b320

Before this fix:

Image


After the fix:

Image


Worth noticing that the console has it's own wallet selector and that does not update the wallet frame current wallet only the console one. Which imho that is ok as it allows you execute commands from different wallets.

PR #948, opened 2026-07-16
GUI/Intro: pass -blocksdir to getPathToCheck if present

### LINKED ISSUE #946: Consider blocksdir when estimating available disk space to determine default pruning
When running bitcoin-qt for the first time, the intro window is shown where the user is prompted for where they want to store the datadir, the default is `~/.bitcoin`.

If there's not a lot of space in `~/.bitcoin`, the prune option is enabled by default. However, a common use case is to point `-blocksdir` to a larger partition/drive.

The current behavior as of Bitcoin Core v31.1 only [checks for the available space of the datadir](https://github.com/bitcoin-core/gui/blob/master/src/qt/freespacechecker.cpp) which results in prune getting enabled by default even when `-blocksdir` points to a larger drive.

It would be nice to have bitcoin-qt check whether blocksdir is on a separate path/partition/drive before enabling the prune option. Additionally, this sounds similar to #709 although I'm not entirely sure it's the same issue.

diegoviola:
Fixes: #946

### DIFF
--- a/src/qt/intro.cpp
+++ b/src/qt/intro.cpp
@@ -275,7 +275,13 @@ void Intro::startThread()
 void Intro::checkPath(const QString &dataDir)
 {
     mutex.lock();
-    pathToCheck = dataDir;
+
+    if (gArgs.IsArgSet("-blocksdir")) {
+        pathToCheck = GUIUtil::PathToQString(gArgs.GetBlocksDirPath());
+    } else {
+        pathToCheck = dataDir;
+    }
+
     if(!signalled)
     {
         signalled = true;


pablomartin4btc (commented):
Concept ACK

The actual problem in #946 is the intro dialog's prune suggestion (only shown the first time `bitcoin-qt` runs) — `FreespaceChecker` checks datadir's free space but the prune trigger uses the full blockchain + chainstate threshold (~880 GB), incorrectly suggesting prune when `-blocksdir` is set since only ~14 GB of chainstate goes to datadir. Skipping the dialog avoids the wrong suggestion.

hebasto (commented):
I'm not sure this is the correct approach.

For example, it breaks the `-choosedatadir` option, which is now ignored when combined with `-blocksdir`:
```console
$ ./build_dev_mode/bin/bitcoin-qt -choosedatadir -blocksdir=/tmp/test_blocksdir
# no intro dialog is shown
```

diegoviola:
> I'm not sure this is the correct approach.
> 
> For example, it breaks the `-choosedatadir` option, which is now ignored when combined with `-blocksdir`:
> 
> ```
> $ ./build_dev_mode/bin/bitcoin-qt -choosedatadir -blocksdir=/tmp/test_blocksdir
> # no intro dialog is shown
> ```

Can you please try this new approach? The dialog is now being shown for me and prune is disabled when running `./bin/bitcoin-qt -choosedatadir -blocksdir=/path/to/bitcoin-blocks/` which _I think_ is the correct behavior.

PR #950, opened 2026-07-20
qt, test: Run GUI tests on macOS with `minimal` QPA plugin

hebasto:
On macOS, `test_bitcoin-qt` has skipped `AppTests`, `WalletTests`, and `AddressBookTests` under the `minimal` QPA plugin since https://github.com/bitcoin/bitcoin/commit/a3197c5294df4711bab0ff6dcdd061ceab115c7d (2018, Qt 5.11), pointing to [QTBUG-49686](https://qt-project.atlassian.net/browse/QTBUG-49686).

The recent re-evaluation of the Bitcoin Core codebase showed that the Qt bug report is no longer relevant. The remaining causes were in our own code, and they are fixed in the first commit.

### DIFF
--- a/src/qt/macdockiconhandler.mm
+++ b/src/qt/macdockiconhandler.mm
@@ -20,7 +20,8 @@ bool dockClickHandler(id self, SEL _cmd, ...) {
 }
 
 void setupDockClickHandler() {
-    Class delClass = (Class)[[[NSApplication sharedApplication] delegate] class];
+    if (NSApp == nil) return;
+    Class delClass = (Class)[[NSApp delegate] class];
     SEL shouldHandle = sel_registerName("applicationShouldHandleReopen:hasVisibleWindows:");
     class_replaceMethod(delClass, shouldHandle, (IMP)dockClickHandler, "B@:");
 }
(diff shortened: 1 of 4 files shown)


maflcko:
lgtm ACK fd59d68c2629440e544ffa6c9a96b4252ebf45c3

pablomartin4btc (commented):
tACK fd59d68c2629440e544ffa6c9a96b4252ebf45c3

I recently ran into [this](https://github.com/bitcoin-core/gui/pull/945#pullrequestreview-4678017739) when writing tests for `addressbooktests.cpp` in #945 — they had to be run on Linux because they were skipped on macOS.

Tested locally — `AppTests`, `WalletTests`, and `AddressBookTests` now run with `QT_QPA_PLATFORM=minimal` on macOS.

From the macOS native CI logs, `test_bitcoin-qt` runs and passes but it's unclear whether those three test classes were previously being skipped or actually executing.

hebasto:
> From the macOS native CI logs, `test_bitcoin-qt` runs and passes but it's unclear whether those three test classes were previously being skipped or actually executing.

For the tests run by the `ctest` driver, you can use the `--verbose` option:
```console
ctest --test-dir build_dev_mode -R bitcoin-qt --verbose
```

fanquake:
After this change the tests segfault if you build using depends: https://github.com/bitcoin/bitcoin/issues/35771.

maflcko:
Fixed in https://github.com/bitcoin-core/gui/pull/951

PR #953, opened 2026-07-27
Adds option to not load the wallet after migration

polespinasa:
Following https://github.com/bitcoin/bitcoin/pull/35266 this PR adds the option to not load the wallet after migrating to the GUI.

It is only added for the `migrate` option, not for the `restore_and_migrate`. I guess if we are restoring the wallet we always want to load it.
In any case, it's pretty straightforward to implement it there too.

Inside the original migration pop-up box it appears a checkbox that allows the user to choose if want to load the wallet or not, it is checked by default:

imagen

If yes, the wallet is loaded and shown, if not the wallet gets migrated and the GUI returns to it's state.

The checkbox has a tooltip that informs when not loading a migrated wallet can be useful:
imagen

### DIFF
--- a/src/qt/walletcontroller.cpp
+++ b/src/qt/walletcontroller.cpp
@@ -23,6 +23,7 @@
 #include <chrono>
 
 #include <QApplication>
+#include <QCheckBox>
 #include <QMessageBox>
 #include <QMetaObject>
 #include <QMutexLocker>
@@ -439,7 +440,7 @@ void RestoreWalletActivity::finish()
     Q_EMIT finished();
 }
 
-void MigrateWalletActivity::do_migrate(const std::string& name)
+void MigrateWalletActivity::do_migrate(const std::string& name, bool load_wallet)
 {
     SecureString passphrase;
     if (node().walletLoader().isEncrypted(name)) {
@@ -450,8 +451,8 @@ void MigrateWalletActivity::do_migrate(const std::string& name)
 
     showProgressDialog(tr("Migrate Wallet"), tr("Migrating Wallet <b>%1</b>…").arg(GUIUtil::HtmlEscape(name)));
 
-    QTimer::singleShot(0, worker(), [this, name, passphrase] {
-        auto res{node().walletLoader().migrateWallet(name, passphrase)};
+    QTimer::singleShot(0, worker(), [this, name, passphrase, load_wallet] {
+        auto res{node().walletLoader().migrateWallet(name, passphrase, load_wallet)};
 
         if (res) {
             m_success_message = tr("The wallet '%1' was migrated successfully.").arg(GUIUtil::HtmlEscape(GUIUtil::WalletDisplayName(name)));
@@ -461,7 +462,12 @@ void MigrateWalletActivity::do_migrate(const std::string& name)
             if (res->solvables_wallet_name) {
                 m_success_message += QChar(' ') + tr("Solvable but not watched scripts have been migrated to a new wallet named '%1'.").arg(GUIUtil::HtmlEscape(GUIUtil::WalletDisplayName(res->solvables_wallet_name.value())));
             }
-            m_wallet_model = m_wallet_controller->getOrCreateWallet(std::move(res->wallet));
+            if (load_wallet) {
+                assert(res->wallet);
+                m_wallet_model = m_wallet_controller->getOrCreateWallet(std::move(res->wallet));
+            } else {
+                m_success_message += QChar(' ') + tr("The wallet was not loaded after migration. You can open it from the \"File > Open wallet\" menu.");
+            }
         } else {
             m_error_message = util::ErrorString(res);
         }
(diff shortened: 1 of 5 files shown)


src/qt/walletcontroller.h
```diff
@@ -196,7 +196,7 @@ class MigrateWalletActivity : public WalletControllerActivity
 private:
     QString m_success_message;
 
-    void do_migrate(const std::string& name);
+    void do_migrate(const std::string& name, const bool load_wallet);
```
pablomartin4btc: nit: I think `const` on a by-value parameter is redundant here, only protects the local copy and doesn't affect callers. Same applies to the other occurrences in `walletcontroller.cpp`, `interfaces/wallet.h` and `interfaces.cpp`.

```suggestion
    void do_migrate(const std::string& name, bool load_wallet);
```

src/qt/walletcontroller.cpp
```diff
@@ -486,7 +491,16 @@ void MigrateWalletActivity::migrate(const std::string& name)
     box.setDefaultButton(QMessageBox::Yes);
     if (box.exec() != QMessageBox::Yes) return;
 
-    do_migrate(name);
+    // Warn the user about loading the wallet
+    QMessageBox box_loadwallet(m_parent_widget);
+    box_loadwallet.setWindowTitle(tr("Migrate wallet"));
+    box_loadwallet.setText(tr("Do you want to load the wallet <i>%1</i> after migrating?").arg(GUIUtil::HtmlEscape(GUIUtil::WalletDisplayName(name))));
+    box_loadwallet.setInformativeText(tr("If the node is pruned, loading the wallet after migrating can fail the migration if the wallet was created before the pruned height."));
```
pablomartin4btc: nit (perhaps cleaner...):
```suggestion
    box_loadwallet.setInformativeText(tr("If the node is pruned and the wallet was created before the pruned height, the migration process may fail trying to load the migrated wallet."));
```

pablomartin4btc (commented):
Concept ACK

It'd be nice to have the screenshots in the description (perhaps collapsed with markup `<details>`) so reviewers that don't want to test it can also give their feedback.

Left a couple of nits.

src/qt/walletcontroller.cpp
polespinasa: done

src/qt/walletcontroller.h
polespinasa: done

achow101:
Concept ACK

In terms of UX, I think it would be better to have this be a checkbox option in the initial dialog rather than a second dialog that the user has to click through.

polespinasa:
> In terms of UX, I think it would be better to have this be a checkbox option in the initial dialog rather than a second dialog that the user has to click through.

Done, modified the description to show the current UI.

src/qt/walletcontroller.cpp
```diff
@@ -482,11 +488,15 @@ void MigrateWalletActivity::migrate(const std::string& name)
                 "The migration process will create a backup of the wallet before migrating. This backup file will be named "
                 "<wallet name>-<timestamp>.legacy.bak and can be found in the directory for this wallet. In the event of "
                 "an incorrect migration, the backup can be restored with the \"Restore Wallet\" functionality."));
+    auto* load_wallet_checkbox = new QCheckBox(tr("Load wallet after migration"), &box);
```
pablomartin4btc, line 491: usually we include the widget into the .ui form, but this is ok I think.

src/wallet/interfaces.cpp
```diff
@@ -601,9 +601,9 @@ class WalletLoaderImpl : public WalletLoader
         }
         return wallet;
     }
-    util::Result<WalletMigrationResult> migrateWallet(const std::string& name, const SecureString& passphrase) override
+    util::Result<WalletMigrationResult> migrateWallet(const std::string& name, const SecureString& passphrase, const bool load_wallet) override
```
pablomartin4btc: nit: I think this one was missed from my previous [suggestion](https://github.com/bitcoin-core/gui/pull/953#discussion_r3659487901)...
```suggestion
    util::Result<WalletMigrationResult> migrateWallet(const std::string& name, const SecureString& passphrase, bool load_wallet) override
```

pablomartin4btc (commented):
ACK 2b4f57cad66ad2e58c51ec0f351a5c112d59b534

w0xlt:
Concept ACK

src/wallet/interfaces.cpp
polespinasa: wopps I missed that one, thanks!

src/qt/walletcontroller.cpp
polespinasa: I will resolve it for now, unless someone requires that too, I also think this is ok as is, easier to modify and remove if ever have to

pablomartin4btc (commented):
ACK 4cea59573c8310fd182d7b4f8ecf04fd0bbc3897

hebasto (approved):
ACK 4cea59573c8310fd182d7b4f8ecf04fd0bbc3897, I have reviewed the code and it looks OK.

achow101:
ACK 4cea59573c8310fd182d7b4f8ecf04fd0bbc3897

PR #957, opened 2026-08-05
fix: add .dat file extension automatically when exporting watchonly

### LINKED ISSUE #956: Export Watch only missing .dat file extension
When exporting a watchonly wallet using the GUI the .dat extension is not automatically added to the file.
The GUI then is not able to restore a wallet without the explicit .dat extension.

~Note: I did not check if this also happens using RPC, will update the issue when tested.~
Edit: The RPC is unaffected by this, it does allow non `.dat` files.

With .dat works:

Image

Without .dat does not see it:

Image

polespinasa:
fixes https://github.com/bitcoin-core/gui/issues/956

Unlike `backup wallet`, `export watch-only wallet` was not automatically adding the file extension to the exported file, making restoring difficult if the user doesn't manually add the file extension after exporting.

Allows also to restore a wallet from a non specified `.dat` file extension. This is achieved by removing the filter in the select file screen, matching the RPC behavior.

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -444,7 +444,7 @@ void BitcoinGUI::createActions()
             //: The title for Restore Wallet File Windows
             QString title_windows = tr("Load Wallet Backup");
 
-            QString backup_file = GUIUtil::getOpenFileName(this, title_windows, QString(), name_data_file + QLatin1String(" (*.dat)"), nullptr);
+            QString backup_file = GUIUtil::getOpenFileName(this, title_windows, QString(), name_data_file +  QLatin1String(" (*.dat *.wallet);;") + tr("All Files") + QLatin1String(" (*)"), nullptr);
             if (backup_file.isEmpty()) return;
 
             bool wallet_name_ok;
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -529,7 +529,11 @@ void BitcoinGUI::createActions()
         connect(m_mask_values_action, &QAction::toggled, this, &BitcoinGUI::setPrivacy);
         connect(m_mask_values_action, &QAction::toggled, this, &BitcoinGUI::enableHistoryAction);
         GUIUtil::ExceptionSafeConnect(m_export_watchonly_action, &QAction::triggered, [this](bool) {
-            QString destination = GUIUtil::getSaveFileName(this, tr("Save Watch-only Wallet Export"), QString(), QString(), nullptr);
+            QString destination = GUIUtil::getSaveFileName(this,
+                tr("Save Watch-only Wallet Export"), QString(),
+                //: Name of the wallet data file format.
+                tr("Wallet Data") + QLatin1String(" (*.dat)"), nullptr);
+
             if (destination.isEmpty()) return;
             WalletModel* model = walletFrame->currentWalletModel();
             if (!Assume(model)) return;


hebasto:
> Unlike `backup wallet`, `export watch-only wallet` was not automatically adding the file extension to the exported file, making restoring difficult if the user doesn't manually add the file extension after exporting.

As I noted [here](https://github.com/bitcoin-core/gui/issues/956#issuecomment-5226394767):
> The GUI behaviour is consistent with the RPC.

This change does not prevent using `exportwatchonlywallet no_dat_extension` RPC and subsequently opening the file in the GUI.

polespinasa:
> This change does not prevent using `exportwatchonlywallet no_dat_extension` RPC and subsequently opening the file in the GUI.

Right, probably more correct to say partially-fixes, this PRs tries to mimic what backup wallet does. Which does add the file extension by default when using the gui.

I think the normal non-technical user (will never use the RPC) workflow is just using the GUI options, and that should not imply adding some unknown data file extensions manually. A user might not know what a `.dat` is. Adding it by default seems the correct way to me.

polespinasa:
> This change does not prevent using `exportwatchonlywallet no_dat_extension` RPC and subsequently opening the file in the GUI.

Now it does :)

Second commit allows to load a non `.dat` file. Force file type filtering might cause the issue reported in https://github.com/bitcoin-core/gui/issues/956, thus breaking UX.

Still I think the first commit is good as adding the file extension is normally good expected behavior and is what the GUI already does with backup wallet.

achow101:
ACK 10ea03ee5164b584a9155ca260d6ecb85058b6fe

src/qt/bitcoingui.cpp
jmechanical13-cloud, line 559: File=Qmenu

jmechanical13-cloud (changes requested):

src/qt/bitcoingui.cpp
```diff
@@ -444,7 +444,7 @@ void BitcoinGUI::createActions()
             //: The title for Restore Wallet File Windows
             QString title_windows = tr("Load Wallet Backup");
 
-            QString backup_file = GUIUtil::getOpenFileName(this, title_windows, QString(), name_data_file + QLatin1String(" (*.dat)"), nullptr);
+            QString backup_file = GUIUtil::getOpenFileName(this, title_windows, QString(), "Wallet Files (*.dat *.wallet);;All Files (*)", nullptr);
```
hebasto: 10ea03ee5164b584a9155ca260d6ecb85058b6fe

1. The `name_data_file` local variable seems to be unused now.

2. The "Wallet Files" and "All Files" substrings are not translated.

polespinasa: Done, thanks :)

src/qt/bitcoingui.cpp
```diff
@@ -444,7 +444,7 @@ void BitcoinGUI::createActions()
             //: The title for Restore Wallet File Windows
             QString title_windows = tr("Load Wallet Backup");
 
-            QString backup_file = GUIUtil::getOpenFileName(this, title_windows, QString(), name_data_file + QLatin1String(" (*.dat)"), nullptr);
+            QString backup_file = GUIUtil::getOpenFileName(this, title_windows, QString(), name_data_file +  QLatin1String(" (*.dat *.wallet);;") + tr("All Files") + QLatin1String(" (*)"), nullptr);
```
hebasto, line 447: nit: A stray space?
```suggestion
            QString backup_file = GUIUtil::getOpenFileName(this, title_windows, QString(), name_data_file + QLatin1String(" (*.dat *.wallet);;") + tr("All Files") + QLatin1String(" (*)"), nullptr);
```

hebasto (approved):
ACK 75a4e6c67881a791d6cbc81008c0d70d12c26710.

A new translatable string "All Files" should be added to the translation source file before the "Translation string freeze" [step](https://github.com/bitcoin/bitcoin/issues/35122).

polespinasa:
@hebasto sorry I never did the translation process.
I followed the guide and I have a commit for it, should I open a PR for it? Or in the future there will be a PR with more strings to be translated?

hebasto:
> I followed the guide and I have a commit for it, should I open a PR for it?

It was simply a note for maintainers (including myself), as the v32.0 release process has already passed the "Soft translation string freeze" step.

> Or in the future there will be a PR with more strings to be translated?

Yes.

polespinasa:
> It was simply a note for maintainers (including myself), as the v32.0 release process has already passed the "Soft translation string freeze" step.

Oh okay, thanks :)

hebasto:
> A new translatable string "All Files" should be added to the translation source file before the "Translation string freeze" [step](https://github.com/bitcoin/bitcoin/issues/35122).

Done in https://github.com/bitcoin/bitcoin/pull/36034.

PR #962, opened 2026-08-30
Improve GUI appearance on macOS 26

Portles:
On macOS 26, Bitcoin Core’s existing Qt widgets retain the appearance used on earlier macOS releases.
This makes text fields, spin boxes, the main toolbar, and the Node window tabs look liquid glass design with the updated system interface.
This change applies macOS 26-specific liquid glass styling to these controls. It rounds text and spin boxes, unifies the main title bar and toolbar, and styles the Node window tabs using qt palette colors.
The liquid glass styling is enabled only at runtime on macOS 26 or later and is not activated on earlier releases.
Spin-box arrows are provided as 1x and 2x png resources to avoid introducing a QtSvg runtime dependency.

Cool note: Not manually tested on macOS versions earlier than 26.


hebasto (commented):
@Portles 

Thank you for your contribution!

However, we usually rely solely on the themes provided by Qt and do not attempt to adjust the GUI appearance to match the native style of any particular OS. Otherwise, that would impose an extra burden on code review, testing, and maintenance.

Portles:
@hebasto

Thank you for your reply

I understand the maintenance concern.
To clarify, this change is not intended to make the application look more like native macOS. It only ensures that the existing Qt components continue to behave correctly on newer macOS versions, without changing the intended design. These behaviors are being enforced by newer macOS SDKs and toolchains, so this PR proactively prevents the UI issues that would otherwise appear in future builds.

And here is the proof: https://developer.apple.com/documentation/bundleresources/information-property-list/uidesignrequirescompatibility

